PR #41, opened 2026-08-11
Fix notification session reload loop

kaishi00:
## Summary

Fix notification taps so they reopen the intended Hermes conversation without repeatedly reloading a failed session.

## Root cause

Hermes notifications carry the live runtime session ID, while `session.resume` requires the durable stored session ID. The app treated the runtime ID as the resume key, received a session-not-found failure, and retained the pending target. Its delayed retry task then incremented the navigation attempt indefinitely, causing the constant reload loop seen in TestFlight.

Hermes Desktop documents the same runtime-to-stored identity distinction in its notification/session routing: [session-ids.ts](https://raw.githubusercontent.com/NousResearch/hermes-agent/main/apps/desktop/src/lib/session-ids.ts).

## Changes

- Preserve `storedSessionId` separately on `SessionSummary` while retaining runtime IDs as alternates.
- Resolve notification runtime IDs to stored IDs before calling `session.resume`.
- Bound a failed notification route to one delayed retry for the same target, then clear it if the retry also fails.
- Inject retry timing and test the complete retry-fire → second-failure → clear sequence on an isolated service instance.
- Prefer explicit stored-ID fields; do not treat a lone ambiguous `id` field as a verified stored ID.
- Remove the redundant `discardPendingTarget` alias and normalize resolver fallback IDs.
- Add regression coverage for normalization, runtime-to-stored resolution, bounded retries, and AppState failure cleanup.

## Verification

- Focused notification/session tests: 31 passed.
- Full iOS simulator suite on iPhone 17 Pro: 434 passed, 0 failures.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved notification navigation by reliably matching sessions using their durable identities and alternate IDs.
  * Failed notification opens now retry in a controlled manner and clear the pending target after repeated failures.
  * Improved handling of session identity differences when resuming conversations.
* **Tests**
  * Added coverage for notification routing, session identity resolution, retry behavior, and clearing failed navigation state.

### DIFF
--- a/Conduit/ConduitApp.swift
+++ b/Conduit/ConduitApp.swift
@@ -74,8 +74,8 @@ struct ConduitApp: App {
                     guard appState.isConnected, let target = notifications.pendingTarget else { return }
                     if await appState.openNotificationTarget(target) {
                         notifications.clearPendingTarget(target)
-                    } else if !notifications.retryPendingTarget(target) {
-                        notifications.clearPendingTarget(target)
+                    } else {
+                        notifications.handleFailedNotificationRoute(target)
                     }
                 }
                 .task(id: voiceIntentRouteKey) {
--- a/Conduit/ConduitApp.swift
+++ b/Conduit/ConduitApp.swift
@@ -74,8 +74,8 @@ struct ConduitApp: App {
                     guard appState.isConnected, let target = notifications.pendingTarget else { return }
                     if await appState.openNotificationTarget(target) {
                         notifications.clearPendingTarget(target)
-                    } else {
-                        notifications.discardPendingTarget(target)
+                    } else if !notifications.retryPendingTarget(target) {
+                        notifications.clearPendingTarget(target)
                     }
                 }
                 .task(id: voiceIntentRouteKey) {
--- a/Conduit/ConduitApp.swift
+++ b/Conduit/ConduitApp.swift
@@ -75,7 +75,7 @@ struct ConduitApp: App {
                     if await appState.openNotificationTarget(target) {
                         notifications.clearPendingTarget(target)
                     } else {
-                        notifications.retryPendingTarget()
+                        notifications.discardPendingTarget(target)
                     }
                 }
                 .task(id: voiceIntentRouteKey) {
--- a/Conduit/Models/Models.swift
+++ b/Conduit/Models/Models.swift
@@ -274,6 +274,9 @@ struct DelegateAgentActivity: Identifiable, Equatable {
 
 struct SessionSummary: Identifiable, Equatable {
     let id: String
+    /// The durable database identity used by `session.resume`. Hermes stream
+    /// events and notifications may carry the live runtime identity instead.
+    var storedSessionId: String? = nil
     var alternateIds: [String]
     var title: String
     var model: String
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -3511,11 +3511,12 @@ final class AppState: ObservableObject {
             return false
         }
         let requestedID = target.sessionId
-        let matchingSession = (sessions + cronSessions).first { session in
-            session.id == requestedID || session.alternateIds.contains(requestedID)
-        }
+        let resumableID = NotificationSessionResolver.resumableSessionID(
+            for: requestedID,
+            in: sessions + cronSessions
+        )
         let opened = await openSession(
-            matchingSession?.id ?? requestedID,
+            resumableID,
             reusing: transitionGeneration
         )
         guard notificationOpenAttemptIsCurrent(
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -1156,6 +1156,9 @@ enum MessageNormalizer {
         return records.enumerated().map { index, item in
             let obj = item.objectValue ?? [:]
             let id = obj["session_id"]?.stringValue ?? obj["id"]?.stringValue ?? String(index)
+            let storedSessionId = ["stored_session_id", "storedSessionId", "id"]
+                .compactMap { obj[$0]?.stringValue?.trimmingCharacters(in: .whitespacesAndNewlines) }
+                .first { !$0.isEmpty }
 
             // Collect alternate IDs
             let altKeys = ["session_id", "id", "stored_session_id", "runtime_session_id", "session_key"]
@@ -1178,6 +1181,7 @@ enum MessageNormalizer {
 
             return SessionSummary(
                 id: id,
+                storedSessionId: storedSessionId,
                 alternateIds: Array(altIds),
                 title: obj["title"]?.stringValue ?? obj["preview"]?.stringValue ?? "Untitled conversation",
                 model: obj["model"]?.stringValue ?? "Hermes",
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -1155,10 +1155,18 @@ enum MessageNormalizer {
 
         return records.enumerated().map { index, item in
             let obj = item.objectValue ?? [:]
-            let id = obj["session_id"]?.stringValue ?? obj["id"]?.stringValue ?? String(index)
-            let storedSessionId = ["stored_session_id", "storedSessionId", "id"]
+            let runtimeSessionId = obj["session_id"]?.stringValue
+            let id = runtimeSessionId ?? obj["id"]?.stringValue ?? String(index)
+            let explicitStoredSessionId = ["stored_session_id", "storedSessionId"]
                 .compactMap { obj[$0]?.stringValue?.trimmingCharacters(in: .whitespacesAndNewlines) }
                 .first { !$0.isEmpty }
+            let storedSessionId = explicitStoredSessionId ?? {
+                // The legacy catalog shape exposes the durable ID as `id`
+                // alongside a separate runtime `session_id`. A lone `id` is
+                // ambiguous, so do not label it as a verified stored ID.
+                guard runtimeSessionId != nil else { return nil }
+                return obj["id"]?.stringValue?.trimmingCharacters(in: .whitespacesAndNewlines)
+            }()
 
             // Collect alternate IDs
             let altKeys = ["session_id", "id", "stored_session_id", "runtime_session_id", "session_key"]
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -9,6 +9,26 @@ struct ConduitNotificationTarget: Equatable, Identifiable {
     var id: String { "\(profile ?? "default"):\(sessionId):\(type ?? "")" }
 }
 
+enum NotificationSessionResolver {
+    /// Hermes notifications identify a live runtime session, while
+    /// `session.resume` is keyed by the durable stored session. Catalog rows
+    /// retain both identities so a notification can be routed without asking
+    /// the gateway to resume a runtime-only key.
+    static func resumableSessionID(
+        for notificationSessionID: String,
+        in sessions: [SessionSummary]
+    ) -> String {
+        let normalizedID = notificationSessionID.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !normalizedID.isEmpty else { return notificationSessionID }
+        guard let session = sessions.first(where: { session in
+            session.id == normalizedID || session.alternateIds.contains(normalizedID)
+        }) else {
+            return notificationSessionID
+        }
+        return session.storedSessionId ?? session.id
+    }
+}
+
 struct ConduitNotificationPreferences: Codable, Equatable {
     var enabled = true
     var approvalNeeded = true
@@ -55,7 +75,6 @@ final class PushNotificationService: ObservableObject {
     private var registration: StoredRegistration?
     private var deviceToken: String?
     private var tokenContinuation: CheckedContinuation<String, Error>?
-    private var navigationRetryTask: Task<Void, Never>?
 
     var isEnabled: Bool { registration != nil && preferences.enabled }
     var statusText: String {
@@ -167,26 +186,17 @@ final class PushNotificationService: ObservableObject {
 
     func receiveNotificationPayload(_ userInfo: [AnyHashable: Any]) {
         guard let target = notificationTarget(from: userInfo) else { return }
-        navigationRetryTask?.cancel()
         pendingTarget = target
         navigationAttempt += 1
     }
 
     func clearPendingTarget(_ target: ConduitNotificationTarget) {
         guard pendingTarget == target else { return }
-        navigationRetryTask?.cancel()
-        navigationRetryTask = nil
         pendingTarget = nil
     }
 
-    func retryPendingTarget() {
-        guard pendingTarget != nil else { return }
-        navigationRetryTask?.cancel()
-        navigationRetryTask = Task { [weak self] in
-            try? await Task.sleep(for: .seconds(1.5))
-            guard !Task.isCancelled else { return }
-            self?.navigationAttempt += 1
-        }
+    func discardPendingTarget(_ target: ConduitNotificationTarget) {
+        clearPendingTarget(target)
     }
 
     private func notificationTarget(from userInfo: [AnyHashable: Any]) -> ConduitNotificationTarget? {
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -19,11 +19,11 @@ enum NotificationSessionResolver {
         in sessions: [SessionSummary]
     ) -> String {
         let normalizedID = notificationSessionID.trimmingCharacters(in: .whitespacesAndNewlines)
-        guard !normalizedID.isEmpty else { return notificationSessionID }
+        guard !normalizedID.isEmpty else { return normalizedID }
         guard let session = sessions.first(where: { session in
             session.id == normalizedID || session.alternateIds.contains(normalizedID)
         }) else {
-            return notificationSessionID
+            return normalizedID
         }
         return session.storedSessionId ?? session.id
     }
@@ -75,6 +75,9 @@ final class PushNotificationService: ObservableObject {
     private var registration: StoredRegistration?
     private var deviceToken: String?
     private var tokenContinuation: CheckedContinuation<String, Error>?
+    private var navigationRetryTask: Task<Void, Never>?
+    private var pendingRetryCount = 0
+    private let maxNotificationRetriesPerTarget = 1
 
     var isEnabled: Bool { registration != nil && preferences.enabled }
     var statusText: String {
@@ -186,17 +189,32 @@ final class PushNotificationService: ObservableObject {
 
     func receiveNotificationPayload(_ userInfo: [AnyHashable: Any]) {
         guard let target = notificationTarget(from: userInfo) else { return }
+        navigationRetryTask?.cancel()
         pendingTarget = target
+        pendingRetryCount = 0
         navigationAttempt += 1
     }
 
     func clearPendingTarget(_ target: ConduitNotificationTarget) {
         guard pendingTarget == target else { return }
+        navigationRetryTask?.cancel()
+        navigationRetryTask = nil
         pendingTarget = nil
+        pendingRetryCount = 0
     }
 
-    func discardPendingTarget(_ target: ConduitNotificationTarget) {
-        clearPendingTarget(target)
+    @discardableResult
+    func retryPendingTarget(_ target: ConduitNotificationTarget) -> Bool {
+        guard pendingTarget == target,
+              pendingRetryCount < maxNotificationRetriesPerTarget else { return false }
+        pendingRetryCount += 1
+        navigationRetryTask?.cancel()
+        navigationRetryTask = Task { [weak self] in
+            try? await Task.sleep(for: .seconds(1.5))
+            guard !Task.isCancelled else { return }
+            self?.navigationAttempt += 1
+        }
+        return true
     }
 
     private func notificationTarget(from userInfo: [AnyHashable: Any]) -> ConduitNotificationTarget? {
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -78,6 +78,7 @@ final class PushNotificationService: ObservableObject {
     private var navigationRetryTask: Task<Void, Never>?
     private var pendingRetryCount = 0
     private let maxNotificationRetriesPerTarget = 1
+    private let retryDelay: Duration
 
     var isEnabled: Bool { registration != nil && preferences.enabled }
     var statusText: String {
@@ -87,7 +88,8 @@ final class PushNotificationService: ObservableObject {
         return "Off"
     }
 
-    private init() {
+    init(retryDelay: Duration = .seconds(1.5)) {
+        self.retryDelay = retryDelay
         if let data = KeychainHelper.loadPushRegistration(),
            let saved = try? JSONDecoder().decode(StoredRegistration.self, from: data) {
             registration = saved
@@ -190,6 +192,7 @@ final class PushNotificationService: ObservableObject {
     func receiveNotificationPayload(_ userInfo: [AnyHashable: Any]) {
         guard let target = notificationTarget(from: userInfo) else { return }
         navigationRetryTask?.cancel()
+        navigationRetryTask = nil
         pendingTarget = target
         pendingRetryCount = 0
         navigationAttempt += 1
@@ -204,15 +207,21 @@ final class PushNotificationService: ObservableObject {
     }
 
     @discardableResult
-    func retryPendingTarget(_ target: ConduitNotificationTarget) -> Bool {
+    func handleFailedNotificationRoute(_ target: ConduitNotificationTarget) -> Bool {
         guard pendingTarget == target,
-              pendingRetryCount < maxNotificationRetriesPerTarget else { return false }
+              pendingRetryCount < maxNotificationRetriesPerTarget else {
+            clearPendingTarget(target)
+            return false
+        }
         pendingRetryCount += 1
         navigationRetryTask?.cancel()
+        let retryDelay = self.retryDelay
         navigationRetryTask = Task { [weak self] in
-            try? await Task.sleep(for: .seconds(1.5))
+            try? await Task.sleep(for: retryDelay)
             guard !Task.isCancelled else { return }
-            self?.navigationAttempt += 1
+            guard let self else { return }
+            self.navigationRetryTask = nil
+            self.navigationAttempt += 1
         }
         return true
     }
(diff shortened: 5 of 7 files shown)


kaishi00:
## Summary
Fixes the notification-tap reload loop by preserving the durable `storedSessionId` on `SessionSummary`, resolving notification runtime session IDs to stored IDs before `session.resume`, and making a failed notification route terminal (discarding the pending target and removing the unbounded retry task).

## Findings

### [minor] Failed notification taps are now permanently dropped on any transient failure
**File:** Conduit/ConduitApp.swift:78, Conduit/Services/PushNotificationService.swift:198
**Issue:** The retry mechanism is removed entirely; any failure inside `openNotificationTarget` (forced catalog reload hiccup, gateway momentarily unavailable during `session.resume`, mid-profile-switch) now permanently swallows the tap with no recovery path. A single transient failure means the user taps a notification and nothing ever happens. The previous unbounded retry was indeed the loop bug, but the fix goes to the opposite extreme.
**Suggestion:** Replace the unbounded retry with a *bounded* one (e.g., one retry capped via `navigationAttempt` threshold, or keep the target pending and re-attempt on the next successful catalog sync / connectivity change). At minimum, document the drop behavior in-app (e.g., surface the failure rather than silently no-op).

### [minor] `discardPendingTarget` is a redundant alias
**File:** Conduit/Services/PushNotificationService.swift:198-200
**Issue:** `discardPendingTarget(_:)` is just `clearPendingTarget(_:)` under a new name; it adds a second public mutation path with identical behavior and no added semantics.
**Suggestion:** Drop the alias and call `clearPendingTarget(target)` from ConduitApp.swift:78, or have `discardPendingTarget` remain but remove `clearPendingTarget` if `discard` better reflects intent.

### [minor] Resolver returns untrimmed ID on the miss path
**File:** Conduit/Services/PushNotificationService.swift:21-22,26
**Issue:** Matching is done against `normalizedID` (whitespace-trimmed), but on the miss path the resolver returns the original untrimmed `notificationSessionID`, which is then passed to `openSession` → `session.resume`. A payload with surrounding whitespace would match nothing and be passed through untrimmed.
**Suggestion:** Return `normalizedID` in both fallbacks for consistency.

### [minor] Regression test doesn't exercise the actual failure wiring
**File:** ConduitTests/MessageNormalizerTests.swift:127-145
**Issue:** `testFailedNotificationRouteDiscardsPendingTarget` directly calls `discardPendingTarget`, so it only verifies the trivial alias exists. The actual fix — ConduitApp's `.task(id:)` calling `discardPendingTarget` when `openNotificationTarget` returns `false` (instead of the old `retryPendingTarget`) — has no coverage.
**Suggestion:** Add an `AppState`-level test that drives `openNotificationTarget` to failure and asserts the pending target is cleared (the existing `AppStateChatResumeTests` harness already supports gated/forced catalog failures), or a test asserting no retry task is scheduled after a failure.

### [nit] Test mutates the shared singleton without isolation
**File:** ConduitTests/MessageNormalizerTests.swift:129
**Issue:** The test writes to `PushNotificationService.shared.pendingTarget`; if other tests in the suite touch the same singleton, ordering could matter. It does reset state, so risk is low.
**Suggestion:** Fine as-is, or inject/construct a fresh instance for isolation.

### [nit] `storedSessionId` heuristic depends on unverified server key semantics
**File:** Conduit/Services/HermesClient.swift:1158-1161
**Issue:** The `id` fallback for `storedSessionId` assumes that when `stored_session_id`/`storedSessionId` are absent, the payload's `id` key holds the durable identity. If the server ever emits `id` as the runtime ID (mirroring `session_id`), the resolver would route `session.resume` to the runtime ID again and the loop would resurface. Tests pin the current convention, but there is no guard against `storedSessionId == session.id == runtime` ambiguity.
**Suggestion:** Consider treating a `storedSessionId` that equals `session.id` as unresolved (fall back to the catalog primary), or validate against a known stored-ID pattern from Hermes.

No retain cycles, force unwraps, concurrency violations, or SwiftUI lifecycle issues were introduced. The removal of the weak-self retry task is clean, and all `SessionSummary` construction sites are covered by the defaulted parameter.

## Verdict
APPROVE

<a href="https://opencode.ai/s/6ytC6np7">New%20session%20-%202026-08-11T17%3A02%3A23.052Z</a>
[opencode session](https://opencode.ai/s/6ytC6np7)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31515426639)

kaishi00:
## Summary
Fixes the notification-tap reload loop by separating the durable `storedSessionId` from runtime IDs on `SessionSummary`, resolving notification runtime IDs to stored IDs before `session.resume`, and bounding a failed notification route to a single delayed retry that is then terminal. The previous review feedback (redundant alias, untrimmed resolver miss-path) was addressed in the follow-up commit.

## Findings

### [minor] The loop-fix wiring itself has no direct regression test
**File:** Conduit/ConduitApp.swift:73-80
**Issue:** The tests cover the leaf units — `retryPendingTarget`'s boolean logic (returns false after count exhausted), `clearPendingTarget`, and `openNotificationTarget`'s busy-state cleanup — but nothing exercises the actual glue that makes a failure terminal: a failed open → `retryPendingTarget` returns true → the 1.5s task bumps `navigationAttempt` → the `.task(id:)` re-runs → second failure → `retryPendingTarget` returns false → `clearPendingTarget`. The very regression this PR fixes (the reload loop) is not asserted anywhere; the 1.5s delayed-fire mechanism that drives the retry is also never awaited in tests.
**Suggestion:** Add a test that drives the full sequence (e.g., feed `receiveNotificationPayload`, fail `openNotificationTarget`, call `retryPendingTarget`, simulate the retry firing by bumping `navigationAttempt`, fail again, and assert `pendingTarget == nil` and `navigationRetryTask == nil`). If wiring a fake `Task.sleep` is needed, inject the delay.

### [minor] Tests spawn a live 1.5s retry task against the shared singleton
**File:** ConduitTests/MessageNormalizerTests.swift:160-174
**Issue:** The first `retryPendingTarget(target)` call in `testNotificationRetryIsBoundedPerPendingTarget` returns `true` and schedules a real `Task` that sleeps 1.5s and then mutates `navigationAttempt` on `PushNotificationService.shared`. The deferred `clearPendingTarget` cancels it only if the test completes before the timer fires, making the shared instance's `navigationAttempt` timing-dependent across the suite — which can spuriously re-trigger ConduitApp's `.task(id:)` in unrelated UI tests.
**Suggestion:** Inject a short sleep (or a closure producing the delay) so the retry task resolves within the test, or construct a fresh `PushNotificationService` instance for the test rather than mutating the shared singleton.

### [minor] `storedSessionId` fallback to `id` can silently reintroduce the runtime-as-resume-key bug
**File:** Conduit/Services/HermesClient.swift:1158-1161
**Issue:** When `stored_session_id`/`storedSessionId` are absent, the fallback is the payload's `id` key — but the base `id` derivation also falls back to `id` (as the runtime identity) when `session_id` is absent. If a server emits only `id` and that `id` is the runtime session, then `storedSessionId == session.id == runtime`, and `NotificationSessionResolver` returns the runtime key, so `session.resume` fails again and the loop could resurface. The tests pin the current convention, but there is no guard for the ambiguity.
**Suggestion:** Treat a `storedSessionId` that equals the derived runtime `id` (i.e., `storedSessionId == id` when it came from the `id` fallback) as unresolved, and have the resolver prefer an explicitly-present `stored_session_id`/`storedSessionId` key over the `id` fallback.

### [nit] Inconsistent whitespace trimming across identity fields
**File:** Conduit/Services/HermesClient.swift:1159-1165
**Issue:** `storedSessionId` is trimmed and the resolver's input is trimmed, but the base `id` (line 1158) and the `alternateIds` (line 1164-1165) are inserted untrimmed. A payload carrying whitespace in `stored_session_id` would be trimmed into `storedSessionId` but kept raw in `alternateIds`, and a whitespace-padded notification ID could fail to match either field.
**Suggestion:** Apply the same trimming to `id` and to the `altKeys` values for consistency.

### [nit] Silent permanent drop of the tap after the single retry
**File:** Conduit/ConduitApp.swift:77-78
**Issue:** After one 1.5s retry fails, `clearPendingTarget` permanently swallows the tap with no user-visible feedback and no re-attempt on the next connectivity change. A transient failure (mid-profile-switch, gateway blip during the retry's `session.resume`) means the user taps a notification and nothing happens.
**Suggestion:** Acceptable trade-off to kill the loop, but surface the failure (e.g., set `errorMessage`) or keep the target pending and re-attempt on the next successful catalog sync / connectivity change rather than dropping silently.

## Verdict
APPROVE

The bounded-retry logic is correct — after the single delayed retry fires, `retryPendingTarget` returns `false` and `clearPendingTarget` terminates the route, so the infinite reload loop is genuinely closed. No retain cycles, force unwraps, or concurrency violations introduced; the retry task's `[weak self]` is correct and the class is `@MainActor`. The remaining issues are coverage and robustness polish.

<a href="https://opencode.ai/s/hbV60owo">New%20session%20-%202026-08-11T18%3A47%3A49.868Z</a>
[opencode session](https://opencode.ai/s/hbV60owo)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31524480586)

kaishi00:
## Summary
This PR fixes an infinite reload loop when tapping a Hermes notification by separating the durable `storedSessionId` from runtime IDs on `SessionSummary`, resolving notification runtime IDs to stored IDs before `session.resume`, and bounding a failed notification route to a single delayed retry that is then terminal.

## Findings

### [minor] Retry budget can be consumed by unrelated `.task(id:)` re-runs before the 1.5s retry fires
**File:** Conduit/Services/PushNotificationService.swift:210-227, Conduit/ConduitApp.swift:73-79
**Issue:** `navigationRetryTask` is only cancelled by `receiveNotificationPayload`/`clearPendingTarget`, but the `.task(id: notificationRouteKey)` also re-runs when `appState.isConnected` or `navigationAttempt` changes for other reasons (e.g., a connectivity flap, another notification attempt bump). Each such re-run that fails calls `handleFailedNotificationRoute`, sees `pendingRetryCount == 1`, and clears the target — consuming the single retry without it ever firing. A transient flap between the first failure and the scheduled retry permanently drops the tap.
**Suggestion:** Track the pending target on the retry task itself (e.g., capture `target` in the task and have it verify `pendingTarget == target` before bumping `navigationAttempt`), or don't count a failure as consuming the budget when the retry task for this target hasn't fired yet. At minimum, document this trade-off.

### [minor] No direct regression test for the loop-fix glue
**File:** ConduitTests/MessageNormalizerTests.swift:150-196
**Issue:** The tests cover the leaf units — `handleFailedNotificationRoute`'s boolean logic, `clearPendingTarget`, and the busy-state cleanup in `AppStateChatResumeTests` — but nothing exercises the actual wiring that makes a failure terminal: `.task(id:)` → `openNotificationTarget` fails → `handleFailedNotificationRoute` returns true → retry fires → `navigationAttempt` bump → re-run → second failure → clear. The exact regression this PR fixes is not asserted end-to-end.
**Suggestion:** Add a test driving the full sequence against a fresh `PushNotificationService` plus an injected `AppState` failure, asserting `pendingTarget == nil` after the second failure and that no further `navigationAttempt` change occurs.

### [minor] Test relies on a 10ms sleep for a `.zero`-delay task to fire
**File:** ConduitTests/MessageNormalizerTests.swift:186-189
**Issue:** `testFailedNotificationRouteRetriesOnceThenClearsTarget` depends on the `.zero`-delay retry task having executed within the 10ms sleep window. `Task.sleep(for: .zero)` plus a `navigationAttempt` bump on the MainActor is timing-sensitive; under CI load this could be flaky.
**Suggestion:** Inject a `Continuation`/signal, or poll with an expectation on `navigationAttempt` rather than a fixed sleep.

### [minor] `storedSessionId` heuristic can still equal the runtime ID
**File:** Conduit/Services/HermesClient.swift:1158-1169
**Issue:** When a `session_id` (runtime) and `id` are both present, `storedSessionId` is set to `id`. If the server emits `id` equal to the runtime ID (a plausible convention), then `storedSessionId == session.id == runtime`, the resolver returns the runtime key, and `session.resume` fails again — bounded now (no loop), but the tap is dropped after one retry. There is no guard for `storedSessionId == runtimeSessionId`.
**Suggestion:** Treat a `storedSessionId` that equals the derived runtime ID as unresolved (fall through to the catalog primary), and prefer an explicitly-present `stored_session_id`/`storedSessionId` key over the `id` fallback.

### [nit] Trimming inconsistency across identity fields
**File:** Conduit/Services/HermesClient.swift:1159,1173; Conduit/Services/PushNotificationService.swift:21
**Issue:** `storedSessionId` and the resolver input are whitespace-trimmed, but `id` and `alternateIds` are inserted raw. A whitespace-padded catalog `id`/`stored_session_id` would be trimmed into `storedSessionId` yet kept raw in `alternateIds`, so a padded notification ID may fail to match either field.
**Suggestion:** Apply the same trimming to `id` and the `altKeys` values for consistency.

### [nit] Test mutates the shared singleton and permanently bumps `navigationAttempt`
**File:** ConduitTests/MessageNormalizerTests.swift:150-171
**Issue:** `testFailedNotificationRouteClearsTarget` uses `PushNotificationService.shared`, which increments the shared `navigationAttempt` and sets `pendingTarget` on the global instance. `clearPendingTarget` resets `pendingTarget`/`pendingRetryCount` but not `navigationAttempt`, so the singleton's `navigationAttempt` is left incremented for the rest of the suite. The sibling test correctly uses a fresh instance.
**Suggestion:** Construct a fresh `PushNotificationService` for this test (or restore `navigationAttempt`) for isolation.

### [nit] Silent drop of the tap after retries are exhausted
**File:** Conduit/ConduitApp.swift:77-78
**Issue:** After the single retry fails, the target is cleared with no user-visible feedback and no re-attempt on the next connectivity change. A transient mid-profile-switch or gateway failure means the user taps a notification and nothing happens.
**Suggestion:** Acceptable trade-off to kill the loop, but surface the failure (e.g., set `errorMessage`) or keep the target pending and re-attempt on the next successful catalog sync / connectivity change.

## Verdict
APPROVE

The bounded-retry logic genuinely closes the reload loop — after the single retry fires, `handleFailedNotificationRoute` returns `false` and `clearPendingTarget` makes the route terminal, and stale-attempt/`pendingTarget == target` guards prevent cross-target retry pollution. No retain cycles, force unwraps, concurrency violations, or SwiftUI lifecycle issues were introduced; the retry task's `[weak self]` is correct, the class is `@MainActor`, and `retryDelay` is safely captured by value. Remaining issues are robustness polish and test-coverage gaps for the loop-fix glue.

<a href="https://opencode.ai/s/Wv0ZsIsf">New%20session%20-%202026-08-11T19%3A04%3A28.847Z</a>
[opencode session](https://opencode.ai/s/Wv0ZsIsf)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31525901334)

kaishi00:
## Summary
Fixes the notification-tap reload loop by separating the durable `storedSessionId` from runtime IDs on `SessionSummary`, resolving notification runtime IDs to stored IDs before `session.resume`, and bounding a failed notification route to one delayed retry that is then terminal.

## Findings

### [minor] Retry budget can be consumed before the scheduled retry fires, dropping the tap after a single failed attempt
**File:** Conduit/ConduitApp.swift:73-79, Conduit/Services/PushNotificationService.swift:210-227
**Issue:** `handleFailedNotificationRoute` increments `pendingRetryCount` and schedules the 1.5s retry on the *first* failure. If `.task(id:)` re-runs for an unrelated reason before that retry fires (an `isConnected` flap, another `navigationAttempt` bump) and fails again, the second call sees `pendingRetryCount == 1` and clears the target — cancelling the still-pending retry task. The scheduled retry never fires and the tap is permanently dropped after one attempt, which is exactly the transient-failure case the retry was meant to cover. The scheduled task also bumps `navigationAttempt` without verifying it still owns the target.
**Suggestion:** Only count a failure against the budget when it originates from the retry task firing (e.g., have the retry task itself perform the final `handleFailedNotificationRoute`-style check, or track `hasScheduledRetryForTarget` separately from the count so an intermediate re-run doesn't consume the budget).

### [minor] Loop-fix glue is not asserted end-to-end; the retry test drives the service in isolation
**File:** ConduitTests/MessageNormalizerTests.swift:173-196
**Issue:** `testFailedNotificationRouteRetriesOnceThenClearsTarget` calls `handleFailedNotificationRoute` directly, so it proves the leaf logic (retry-fire → second-failure → clear) but never exercises the actual wiring that made the loop: `openNotificationTarget` fails → `handleFailedNotificationRoute` → `navigationAttempt` bump → `.task(id:)` re-runs → second failure → terminal. `testFailedNotificationOpenReleasesNotificationBusyState` (AppStateChatResumeTests.swift:263-278) only asserts busy-state cleanup, not that the target is cleared. The exact regression this PR fixes has no direct regression test.
**Suggestion:** Add a test that feeds `receiveNotificationPayload`, drives `openNotificationTarget` to failure, lets the retry task fire (injected delay), fails again, and asserts `pendingTarget == nil` and no further `navigationAttempt` change — through the app-side path rather than calling the service API directly.

### [minor] Fabricated `storedSessionId` can equal the runtime ID, so the original failure mode still occurs (now bounded)
**File:** Conduit/Services/HermesClient.swift:1163-1169
**Issue:** When `session_id` and `id` are both present, `storedSessionId = id` with no guard against `id == session_id`. If the server emits `id` mirroring the runtime value, then `storedSessionId == session.id == runtime`, `NotificationSessionResolver` returns the runtime key, and `session.resume` fails again — no longer an infinite loop (bounded to one retry), but the tap is dropped rather than routed. There is no check that the fabricated stored ID differs from the runtime ID.
**Suggestion:** Treat a `storedSessionId` that equals `runtimeSessionId` as unresolved (return `nil`), and prefer an explicitly-present `stored_session_id`/`storedSessionId` key over the `id` fallback. If no distinct durable ID exists, surface a message rather than silently failing `session.resume`.

### [minor] Whitespace trimming is inconsistent across identity fields
**File:** Conduit/Services/HermesClient.swift:1159,1160-1162,1172-1173
**Issue:** `storedSessionId` and the resolver input are whitespace-trimmed, but the base `id` (line 1159) and `alternateIds` (lines 1172-1173) are inserted raw. A whitespace-padded catalog `id`/`session_id` is trimmed into `storedSessionId` yet kept padded in `alternateIds`, so a whitespace-padded notification ID (which the resolver trims) can fail to match any catalog row and fall through to an unresolvable `session.resume`.
**Suggestion:** Apply the same `trimmingCharacters(in: .whitespacesAndNewlines)` to `id` and the `altKeys` values for consistency.

### [nit] Test mutates the shared singleton and leaves `navigationAttempt` permanently bumped
**File:** ConduitTests/MessageNormalizerTests.swift:150-171
**Issue:** `testFailedNotificationRouteClearsPendingTarget` uses `PushNotificationService.shared`; `receiveNotificationPayload` increments the shared `navigationAttempt`, which `clearPendingTarget` never restores. The sibling retry test correctly constructs a fresh instance with an injected delay, so the two tests are inconsistent in isolation strategy. Impact is low (unit tests don't run ConduitApp's `.task(id:)`), but any future test asserting on the singleton's `navigationAttempt` would be order-dependent.
**Suggestion:** Construct a fresh `PushNotificationService(retryDelay: .zero)` here too, matching the sibling test.

### [nit] Silent permanent drop after the retry is exhausted
**File:** Conduit/ConduitApp.swift:77-78
**Issue:** After the single retry fails, `clearPendingTarget` swallows the tap with no user-visible feedback and no re-attempt on the next connectivity change. A mid-profile-switch or gateway blip means the user taps a notification and nothing ever happens.
**Suggestion:** Acceptable trade-off for killing the loop, but surface the failure (e.g., set `appState.errorMessage`) or re-attempt on the next successful catalog sync instead of dropping silently.

### [nit] Retry test relies on a polling busy-wait around a `.zero`-delay task
**File:** ConduitTests/MessageNormalizerTests.swift:186-196
**Issue:** `testFailedNotificationRouteRetriesOnceThenClearsTarget` spins with `await Task.yield()` inside a `ContinuousClock` deadline to observe the `.zero`-delay retry task fire. This is much better than a fixed sleep, but it is still timing-dependent; under MainActor saturation in CI the 1s window is generous, yet the pattern can be made deterministic.
**Suggestion:** Signal the retry fired via a `CheckedContinuation`/expectation, or poll on `navigationAttempt` with an XCTest expectation rather than a wall-clock deadline.

No retain cycles (retry task uses `[weak self]` and self only holds the `Task` weakly), force unwraps, concurrency violations (class is `@MainActor`, Swift 5.9 without strict concurrency), or SwiftUI lifecycle issues were introduced. `storedSessionId` is added with a defaulted parameter so all existing construction sites remain valid. The bounded-retry logic genuinely terminates: after the retry fires and the second failure occurs, `clearPendingTarget` cancels the task and clears the target, so the reload loop is closed.

## Verdict
APPROVE

<a href="https://opencode.ai/s/dyavktoD">New%20session%20-%202026-08-12T01%3A39%3A25.149Z</a>
[opencode session](https://opencode.ai/s/dyavktoD)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31554257461)

kaishi00:
## Summary
Fixes the notification-tap reload loop by keeping the durable `storedSessionId` separate from runtime IDs on `SessionSummary`, resolving notification runtime IDs to stored IDs before `session.resume`, and making a failed notification route terminal after a single delayed retry.

## Findings

### [minor] Retry budget can be consumed by an unrelated `.task(id:)` re-run before the scheduled retry fires
**File:** Conduit/Services/PushNotificationService.swift:210-227, Conduit/ConduitApp.swift:73-79
**Issue:** `notificationRouteKey` includes `appState.isConnected` and `navigationAttempt`, so `.task(id:)` re-runs on any connectivity flap or attempt bump — not just the retry task's bump. If the route fails again before the 1.5s retry fires (e.g., a gateway flap that re-runs the task immediately), `handleFailedNotificationRoute` sees `pendingRetryCount == 1` and clears the target, cancelling the still-pending retry. The one retry meant to absorb transient failures is consumed by an unrelated re-run, so a tap can be permanently dropped after a single immediate failure. The retry task also bumps `navigationAttempt` without verifying it still owns the target.
**Suggestion:** Don't count a failure against the budget when the scheduled retry for that target hasn't fired yet (e.g., track `hasScheduledRetryFor(target)` separately from the count), or have the retry task perform the final failure check itself so only a retry-originated failure clears the target.

### [minor] No end-to-end regression test for the loop-fix glue
**File:** ConduitTests/MessageNormalizerTests.swift:173-196, ConduitTests/AppStateChatResumeTests.swift:263-278
**Issue:** Tests cover the leaf units — `handleFailedNotificationRoute`'s boolean logic, `clearPendingTarget`, and busy-state cleanup — but nothing drives the actual wiring that made the loop: `.task(id:)` → `openNotificationTarget` fails → `handleFailedNotificationRoute` schedules retry → retry bumps `navigationAttempt` → `.task(id:)` re-runs → second failure → `clearPendingTarget`. The exact regression this PR fixes (the reload loop) has no direct assertion.
**Suggestion:** Add a test that feeds `receiveNotificationPayload`, fails `openNotificationTarget` through an injected `AppState`, lets the retry fire, fails again, and asserts `pendingTarget == nil` and no further `navigationAttempt` change, exercising the app-side path rather than calling the service API directly.

### [minor] `storedSessionId` fallback can silently equal the runtime ID
**File:** Conduit/Services/HermesClient.swift:1163-1169
**Issue:** When both `session_id` and `id` are present, `storedSessionId = id` with no guard against `id == session_id`. If the server emits `id` mirroring the runtime value, then `storedSessionId == session.id == runtime`, the resolver returns the runtime key, and `session.resume` fails again — no longer an infinite loop (bounded to one retry), but the tap is dropped rather than routed.
**Suggestion:** Treat a `storedSessionId` equal to the derived runtime ID as unresolved, and prefer an explicitly-present `stored_session_id`/`storedSessionId` key over the `id` fallback. If no distinct durable ID exists, surface a message instead of silently failing `session.resume`.

### [minor] Whitespace trimming is inconsistent across identity fields
**File:** Conduit/Services/HermesClient.swift:1158-1173, Conduit/Services/PushNotificationService.swift:21
**Issue:** `storedSessionId` and the resolver input are whitespace-trimmed, but the base `id` (line 1158) and `alternateIds` (lines 1172-1173) are inserted raw. A whitespace-padded catalog `session_id`/`id` is trimmed into `storedSessionId` yet kept padded in `id`/`alternateIds`, so a trimmed notification ID can fail to match any catalog row and fall through to an unresolvable `session.resume`.
**Suggestion:** Apply the same `trimmingCharacters(in: .whitespacesAndNewlines)` to `id` and the `altKeys` values for consistency.

### [nit] Test mutates the shared singleton and permanently bumps `navigationAttempt`
**File:** ConduitTests/MessageNormalizerTests.swift:150-171
**Issue:** `testFailedNotificationRouteClearsPendingTarget` uses `PushNotificationService.shared`; `receiveNotificationPayload` increments the shared `navigationAttempt`, which `clearPendingTarget` never restores. The sibling retry test correctly uses a fresh instance, so isolation strategy is inconsistent.
**Suggestion:** Construct a fresh `PushNotificationService(retryDelay: .zero)` here too, matching the sibling test.

### [nit] Retry test relies on wall-clock polling around a `.zero`-delay task
**File:** ConduitTests/MessageNormalizerTests.swift:185-200
**Issue:** The test spins `await Task.yield()` against a `ContinuousClock` deadline to observe the retry task fire. Better than a fixed sleep, but still timing-dependent; under MainActor saturation the 1s window is generous, yet the pattern could be made deterministic.
**Suggestion:** Poll on `navigationAttempt` with an XCTest expectation or signal the retry via a continuation rather than a wall-clock deadline.

### [nit] Silent permanent drop of the tap after retries are exhausted
**File:** Conduit/ConduitApp.swift:77-78
**Issue:** After the single retry fails, `clearPendingTarget` swallows the tap with no user-visible feedback and no re-attempt on the next connectivity change. A mid-profile-switch or gateway blip means the user taps a notification and nothing happens.
**Suggestion:** Acceptable trade-off to kill the loop, but surface the failure (e.g., set `appState.errorMessage`) or re-attempt on the next successful catalog sync instead of dropping silently.

### [nit] `storedSessionId` declared `var` while sibling fields are `let`
**File:** Conduit/Models/Models.swift:277
**Issue:** `storedSessionId` is `var` though `SessionSummary`'s other identity fields are `let`; nothing mutates it after construction.
**Suggestion:** Declare it `let` (a `let` with a default value still keeps existing call sites compiling).

No retain cycles (retry task uses `[weak self]`, and `self.navigationRetryTask = nil` inside the task is safe — the runtime retains the running task), no force unwraps, and no concurrency violations introduced (class is `@MainActor`, `retryDelay` captured by value). The bounded-retry logic genuinely terminates: after the retry fires and the second failure occurs, `clearPendingTarget` cancels the task and clears the target, so the reload loop is closed. Remaining issues are robustness polish and a coverage gap for the glue that this PR fixes.

## Verdict
APPROVE

<a href="https://opencode.ai/s/hxMFaFzs">New%20session%20-%202026-08-12T01%3A56%3A38.061Z</a>
[opencode session](https://opencode.ai/s/hxMFaFzs)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31555200344)

PR #42, opened 2026-08-11
Fix send-freeze (regex backtracking), hung RPCs on socket death, and WebKit process leak

angel12:
While testing Conduit we hit a hard app freeze (force-close required) when sending a message during a streaming turn. A code review traced it to three separate defects; this PR fixes them, one commit each. Full write-ups with line-pinned references live on our fork: [freeze](https://github.com/angel12/hermes-conduit/issues/1), [RPC hang](https://github.com/angel12/hermes-conduit/issues/4), [WebKit leak](https://github.com/angel12/hermes-conduit/issues/5).

## Fix catastrophic regex backtracking in SyntaxHighlighter
`SyntaxHighlighter.highlight`'s string-literal alternations are ambiguous — `\\.` and `[^"]` both match a backslash — so an **unterminated** string literal containing escape sequences backtracks exponentially (~2^k paths). An unterminated escaped string is the normal transient state while a code block streams, and the highlighter runs synchronously in SwiftUI `body` on every reveal tick, so streaming tool output like JSON-encoded shell commands (`{"command":"python3 - <<'PY'\nimport os…`) wedges the main thread permanently. The send-message state burst makes it trigger reliably right as the keyboard dismisses: composer frozen mid-screen over black space, force-close required.

Fix: exclude the backslash from the negated character classes (`(?:\\.|[^"\\])*`), making the branches disjoint. Measured with NSRegularExpression: 16 escapes 5 ms → 20 escapes 80 ms → 24 escapes 1.2 s → ~30+ effectively infinite; after the fix, sub-millisecond at 2,000 escapes with identical tokenization of terminated strings, escapes, and comments.

Verified end-to-end in the simulator against a live Hermes instance: streamed a 300-escape unterminated string literal, sent a message mid-stream (the exact force-close repro), interacted with the UI throughout — no freeze.

## Fail in-flight RPC continuations when the WebSocket dies
`receiveLoop`'s catch sets `isConnected = false` and fires `onDisconnected`, but never touches the `pending` continuation map — cleanup only happens in `disconnect()`, which AppState calls only when it replaces the client after minting a new ticket. Until then, anything awaiting an RPC on the dead socket (send, clarify/approval responses, session open) hangs silently for the full per-request timeout (180 s for `prompt.submit`). Now the receive-loop failure path fails all pending requests with `connectionClosed`, matching `disconnect()`.

## Break WKScriptMessageHandler retain cycle in DashboardTicketBridge
`configuration.userContentController.add(self, name:)` retains the bridge strongly, forming a cycle (bridge → webView → configuration → userContentController → bridge). The `deinit` that would remove the handler can never run, so every bridge replacement (base-URL change, Cloudflare credential change, sign-out/sign-in) leaks the bridge and keeps its WebKit content process alive for the app's lifetime. The bridge now registers a weak proxy `WKScriptMessageHandler`; the cycle never forms and `deinit` removes the proxy registration.

## Testing
- Rebased onto current `main` (`ea4d95f`); no conflicts, including with the #36 cookie-clearing changes in `DashboardTicketBridge.swift`.
- Full unit suite passes on the iPhone 17 Pro simulator (`xcodebuild test`, 0 failures).
- Freeze repro exercised live in the simulator as described above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dashboard message handling to prevent communication issues during web content interactions.
  * Pending requests now fail promptly when a connection closes unexpectedly or the bridge is unavailable.
  * Improved syntax highlighting performance and stability for incomplete streamed code strings.

### DIFF
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -133,6 +133,13 @@ enum DashboardTicketBridgeError: LocalizedError {
 /// configuration → userContentController → bridge) that `deinit` can never
 /// break — every replaced bridge would keep a WebKit content process alive
 /// for the app's lifetime. The proxy holds the bridge weakly instead.
+///
+/// WebKit delivers `userContentController(_:didReceive:)` on the main thread,
+/// and the proxied bridge is `@MainActor`; marking the proxy `@MainActor` too
+/// makes that isolation contract explicit so the forward into the bridge needs
+/// no implicit cross-actor hop and stays correct under Swift 6 strict
+/// concurrency.
+@MainActor
 private final class WeakScriptMessageHandler: NSObject, WKScriptMessageHandler {
     private weak var delegate: WKScriptMessageHandler?
 
@@ -182,6 +189,17 @@ final class DashboardTicketBridge: NSObject {
     }
 
     deinit {
+        // The retain-cycle fix is what makes deallocation reachable at all, so
+        // `deinit` can now run while requests are still pending. Removing the
+        // message handler alone leaves their continuations awaiting a JS reply
+        // that can never arrive: the proxy's weak delegate is already nil and
+        // silently drops the posted message. Resume them now so callers see
+        // `.notReady` instead of hanging for a response that will never come.
+        // (The dictionary is torn down with the bridge, so only the resumptions
+        // matter here — nothing else can resume them once the bridge is gone.)
+        for continuation in pendingRequests.values {
+            continuation.resume(throwing: DashboardTicketBridgeError.notReady)
+        }
         webView.configuration.userContentController.removeScriptMessageHandler(forName: "dashboard-response")
     }
 
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -128,6 +128,23 @@ enum DashboardTicketBridgeError: LocalizedError {
     }
 }
 
+/// WKUserContentController retains its script message handlers strongly, so
+/// registering the bridge directly forms a cycle (bridge → webView →
+/// configuration → userContentController → bridge) that `deinit` can never
+/// break — every replaced bridge would keep a WebKit content process alive
+/// for the app's lifetime. The proxy holds the bridge weakly instead.
+private final class WeakScriptMessageHandler: NSObject, WKScriptMessageHandler {
+    private weak var delegate: WKScriptMessageHandler?
+
+    init(_ delegate: WKScriptMessageHandler) {
+        self.delegate = delegate
+    }
+
+    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
+        delegate?.userContentController(userContentController, didReceive: message)
+    }
+}
+
 @MainActor
 final class DashboardTicketBridge: NSObject {
     let baseURL: String
@@ -151,7 +168,7 @@ final class DashboardTicketBridge: NSObject {
         }
         self.webView = WKWebView(frame: .zero, configuration: configuration)
         super.init()
-        configuration.userContentController.add(self, name: "dashboard-response")
+        configuration.userContentController.add(WeakScriptMessageHandler(self), name: "dashboard-response")
         webView.navigationDelegate = self
         Task { [weak self] in
             guard let self else { return }
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -453,6 +453,10 @@ final class HermesClient: ObservableObject {
                 // Socket closed or errored
                 logger.error("WebSocket receive failed: \(error.localizedDescription, privacy: .public)")
                 isConnected = false
+                // No response can ever arrive on a dead socket; failing the
+                // in-flight requests here keeps awaiting UI from hanging for
+                // the remainder of each request's timeout.
+                failAllPendingRequests(with: HermesError.connectionClosed)
                 if !closedIntentionally {
                     onDisconnected?()
                 }
@@ -509,10 +513,13 @@ final class HermesClient: ObservableObject {
         session?.invalidateAndCancel()
         socketDelegate = nil
         isConnected = false
-        // Fail all pending requests
-        for (_, pending) in pending {
-            pending.timer?.invalidate()
-            pending.continuation.resume(throwing: HermesError.connectionClosed)
+        failAllPendingRequests(with: HermesError.connectionClosed)
+    }
+
+    private func failAllPendingRequests(with error: Error) {
+        for (_, request) in pending {
+            request.timer?.invalidate()
+            request.continuation.resume(throwing: error)
         }
         pending.removeAll()
     }
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -1279,7 +1279,11 @@ private enum MarkdownLanguage {
 private enum SyntaxHighlighter {
     static func highlight(_ source: String, language: String) -> AttributedString {
         let keywords: Set<String> = ["as", "async", "await", "break", "case", "catch", "class", "const", "continue", "def", "else", "enum", "false", "final", "for", "func", "guard", "if", "import", "in", "init", "let", "nil", "null", "private", "public", "return", "self", "static", "struct", "switch", "throw", "true", "try", "var", "while"]
-        let pattern = #"(//[^\n]*|#[^\n]*|/\*[\s\S]*?\*/|\"(?:\\.|[^\"])*\"|'(?:\\.|[^'])*'|\b\d+(?:\.\d+)?\b|\b[A-Za-z_][A-Za-z0-9_]*\b)"#
+        // String-literal alternatives must be disjoint (`\\.` vs `[^"\\]`):
+        // if both branches can match a backslash, an unterminated literal with
+        // many escapes — the normal transient state while a code block streams —
+        // backtracks exponentially and wedges the main thread (~2^k paths).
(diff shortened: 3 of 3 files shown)


kaishi00:

[github run](/kaishi00/hermes-conduit/actions/runs/31531095422)

kaishi00:
Failed to check permissions for user : HttpError: Not Found - https://docs.github.com/rest

[github run](/kaishi00/hermes-conduit/actions/runs/31554513029)

kaishi00:
## Summary
This PR fixes three independent defects: catastrophic regex backtracking in the syntax highlighter (main-thread freeze), hung in-flight RPCs when the WebSocket dies, and a WKScriptMessageHandler retain cycle that leaked the dashboard bridge and its WebKit content process. All three changes are technically correct and well-documented; findings below are minor robustness gaps in the WebKit fix.

## Findings

### [minor] In-flight dashboard requests can hang forever once the bridge deallocates
**File:** Conduit/Services/DashboardTicketBridge.swift:184-186
**Issue:** The retain-cycle fix makes bridge deallocation possible, which is the point — but `deinit` only removes the message handler; it never fails `pendingRequests`. `requestJSON` (line 252) has no Swift-side timeout; the only timeout lives in the injected JS (`setTimeout` → `controller.abort()`), and its error path posts a response to `window.webkit.messageHandlers['dashboard-response']`. If the bridge deallocates mid-request (base-URL change in `prepareDashboardBridge`, Cloudflare credential change, or `requireSignIn` setting the bridge to `nil`), that posted message arrives at the `WeakScriptMessageHandler` with a nil delegate and is silently dropped, and the awaiting continuation is never resumed. This path was unreachable before the PR (the cycle kept the bridge alive and its handler registered), so the fix makes it newly reachable.
**Suggestion:** In `deinit`, reject any still-pending requests, e.g. `rejectPending(with: DashboardTicketBridgeError.notReady)` before `removeScriptMessageHandler`. Continuations can be resumed from `deinit`; this guarantees no caller hangs indefinitely.

### [minor] Weak proxy relies on unannotated main-thread delivery / deinit actor assumptions
**File:** Conduit/Services/DashboardTicketBridge.swift:136-186
**Issue:** `WeakScriptMessageHandler` is a plain (nonisolated) `NSObject`, and `DashboardTicketBridge` is `@MainActor`. Correctness currently depends on two implicit guarantees: (1) WebKit delivers `userContentController(_:didReceive:)` on the main thread so the weakly-held bridge's actor-isolated method runs on its actor, and (2) the last strong reference to the bridge (and thus its `deinit`, which touches the main-thread-only `WKWebView`) is dropped on the main actor. Both hold today (Swift 5.9 mode, AppState replaces bridges on the main actor), but under Swift 6 strict concurrency the nonisolated proxy calling into a `@MainActor` bridge and the nonisolated `deinit` touching `WKWebView` are fragile.
**Suggestion:** Annotate `WeakScriptMessageHandler` with `@MainActor` (and `@preconcurrency WKScriptMessageHandler` conformance if needed) to make the isolation contract explicit and future-proof for strict concurrency.

### [nit] None found in the regex and RPC fixes
**File:** Conduit/Views/Components/MarkdownText.swift:1286, Conduit/Services/HermesClient.swift:456-459, 519-525
**Issue:** No issues found. The `(?:\\.|[^"\\])*` change correctly makes the string-literal branches disjoint, eliminating the exponential backtracking while preserving tokenization of terminated strings, escapes, and comments; the remaining worst case on an unterminated literal is a linear scan, which is fine. The `failAllPendingRequests` refactor exactly preserves `disconnect()` semantics and the new receive-loop failure path is safe against double-resume (all resume paths remove from `pending` first), correct on the `@MainActor` client, and a no-op in the intentional-disconnect ordering. Note `SyntaxHighlighter.highlight` does run synchronously in `body` (MarkdownText.swift:893), confirming the freeze diagnosis.
**Suggestion:** No change needed.

## Verdict
APPROVE — all three defects are correctly and cleanly fixed. The minor findings (fail pending requests on bridge dealloc; annotate the proxy/deinit for isolation) are recommended follow-ups rather than blockers.

<a href="https://opencode.ai/s/Beemdx3A">New%20session%20-%202026-08-12T01%3A47%3A54.367Z</a>
[opencode session](https://opencode.ai/s/Beemdx3A)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31554724636)

kaishi00:
@angel12 — I pushed a follow-up commit (`0020227`) onto your branch addressing the two **[minor]** findings from the review. You'd said you weren't picky about me touching it, so I just took care of them. The regex and RPC fixes needed no changes (the **[nit]** finding confirmed they're correct).

**1. Fail in-flight dashboard requests on bridge dealloc** — `DashboardTicketBridge.swift`
`deinit` previously only removed the message handler. Once your retain-cycle fix made deallocation reachable mid-request, any still-pending `requestJSON` continuation would hang forever: the proxy's weak delegate is already nil by then, so the JS reply is dropped and nothing resumes the continuation. `deinit` now resumes all pending continuations with `.notReady` before unregistering the handler. Safe against double-resume — nothing else can reach them once the bridge is gone.

**2. Annotate the weak proxy `@MainActor`** — `DashboardTicketBridge.swift`
`WeakScriptMessageHandler` is now `@MainActor`. WebKit already delivers the callback on the main thread and the bridge is `@MainActor`, so this makes the isolation contract explicit (no implicit cross-actor hop) and keeps the conformance clean under Swift 6 strict concurrency.

Project is Swift 5 language mode, so no `@preconcurrency` was needed. CI is re-running on the new commit.

kaishi00:
## Summary
This PR fixes three defects: catastrophic regex backtracking in `SyntaxHighlighter` (main-thread freeze on unterminated streamed strings), hung in-flight RPCs when the WebSocket dies (failing `pending` continuations in the receive-loop catch), and a `WKScriptMessageHandler` retain cycle in `DashboardTicketBridge` that leaked the bridge and its WebKit content process. All three fixes are correct and the RPC/WebKit changes match the upstream gateway protocol (JSON-RPC 2.0 responses correlated by `id`).

## Findings

### [minor] `DashboardTicketBridge.deinit` pending-request loop is unreachable dead code
**File:** Conduit/Services/DashboardTicketBridge.swift:200-203
**Issue:** The `deinit` resumes all `pendingRequests` continuations, and the commit message claims deallocation is now reachable "mid-request" with continuations left hanging. That premise is wrong: every pending continuation corresponds to a coroutine suspended inside `requestJSON` (`withCheckedThrowingContinuation`), and a suspended async method's frame retains `self` strongly. So `pendingRequests` non-empty ⟹ the bridge's refcount is ≥ 1 from the suspended task ⟹ `deinit` can never fire while a request is pending. The loop can never execute. This also means the hang scenario it was added to guard against (JS reply dropped by nil-delegate proxy) cannot occur through this path — the bridge stays alive until the JS reply or the JS-side `AbortController` timeout resolves the request. The code is harmless, but it does not do what the comment/commit claims.
**Suggestion:** Either remove the loop and correct the comment, or keep it purely as defensive documentation but fix the reasoning. If you want the resumption guarantee spelled out, note that the suspended `requestJSON` coroutine is what keeps the bridge alive, not AppState.

### [minor] `failAllPendingRequests` resumes while iterating `pending`, clearing only afterward
**File:** Conduit/Services/HermesClient.swift:519-525
**Issue:** The loop resumes each continuation before `pending.removeAll()`. Resuming a `CheckedContinuation` can run the awaiting code synchronously on the same (main) actor; if that error-handling code re-enters the client (issues a new RPC, or calls `disconnect()`, which also calls `failAllPendingRequests`), `pending` is mutated during enumeration and Swift traps with "Collection mutated while being enumerated". This pattern pre-existed in `disconnect()`, but the new `receiveLoop` call site (HermesClient.swift:459) fires automatically on every socket death, widening exposure. Likelihood is low (AppState's `handleDisconnect` defers reconnect work), and all current paths are double-resume-safe.
**Suggestion:** Clear before resuming, mirroring the snapshot pattern already used in `DashboardTicketBridge.rejectPending` (line 369-373): `let pending = pending; pendingRequests.removeAll(); pending.values.forEach { $0.continuation.resume(throwing: error) }`.

### [minor] Nonisolated `deinit` touches `@MainActor` state and main-thread-bound WebKit objects
**File:** Conduit/Services/DashboardTicketBridge.swift:191-204
**Issue:** `deinit` reads `pendingRequests` and calls `removeScriptMessageHandler` on the `WKWebView`'s configuration. `DashboardTicketBridge` is `@MainActor`, but `deinit` is nonisolated; correctness depends on the implicit guarantee that the last strong reference is dropped on the main actor (true today, since AppState replaces bridges on the main actor). Under Swift 6 strict concurrency this is an error, and nothing enforces the main-thread-drop contract.
**Suggestion:** Annotate with `isolated deinit` once the project moves to Swift 6, or extract the handler removal to a `@MainActor` helper invoked from an explicit teardown path in addition to `deinit`.

### [nit] None in the regex and RPC fixes
**File:** Conduit/Views/Components/MarkdownText.swift:1286, Conduit/Services/HermesClient.swift:456-459
**Issue:** No issues. The `(?:\\.|[^"\\])*` change makes the string-literal branches disjoint, eliminating the exponential backtracking (linear worst case on an unterminated literal) while preserving tokenization of terminated strings/escapes/comments. The receive-loop failure path correctly fails all pending RPCs with `connectionClosed` before `onDisconnected`, is a no-op in the intentional-disconnect ordering, and is double-resume-safe (every resume path removes from `pending` first; timers are invalidated).

## Verdict
APPROVE — all three defects are correctly fixed and the failure-path handling matches the upstream Hermes gateway protocol. The findings are robustness/cleanup items, not blockers.

<a href="https://opencode.ai/s/6KrUU8et">New%20session%20-%202026-08-12T13%3A42%3A05.503Z</a>
[opencode session](https://opencode.ai/s/6KrUU8et)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31602669049)

PR #43, opened 2026-08-11
Deduplicate buffered stream deltas against the resume inflight snapshot

angel12:
## Problem
Foreground reconciliation buffers stream events while `session.list` + `session.resume` run, seeds the live bubble from the resume snapshot's **cumulative** `inflight` projection, then replays the buffered events on top. The inflight text already contains the deltas that arrived during the buffer window, so foregrounding the app mid-turn visibly repeated a chunk of the streaming reply.

## Fix
Before replay, collapse the buffered `.messageDelta` events to the suffix the snapshot has not covered. The delta stream is contiguous, so the snapshot's tail and the buffered concatenation meet at exactly one offset — computed as the longest overlap between the inflight's suffix and the buffer's prefix. Fully covered deltas are dropped; a partial overlap replays only the remainder (as a single delta, in the last delta's position); zero overlap (non-cumulative gateways, or a turn that started after the snapshot) leaves the events untouched. Non-delta events always pass through in order.

Applied only when `snapshot.hasLiveProjection` — the case where the bubble was seeded from inflight.

## Testing
- 11 new unit tests (`BufferedEventDeduplicationTests`) covering full/partial/zero overlap, longest-match selection, non-delta pass-through, empty inflight, and session-id preservation.
- Full suite passes on the iPhone 17 Pro simulator.

Full write-up: [angel12/hermes-conduit#19](https://github.com/angel12/hermes-conduit/issues/19)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery when resuming interrupted streaming sessions.
  * Prevented already-displayed text from appearing again after a resume, including partial and whitespace variations.
  * Preserved the correct session context while replaying buffered updates.
  * Maintained event order and retained updates that cannot be safely matched, improving reliability in edge cases.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2324,6 +2324,11 @@ final class AppState: ObservableObject {
                     acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(
                         knownSessionIDs(for: boundarySessionID)
                     )
+                    if !acceptedSessionIDs.contains(result.sessionId) {
+                        sessionCatalogLog.debug(
+                            "Skipping buffered delta dedup because resumed session \(result.sessionId, privacy: .public) is not a catalog-confirmed alias of the reconciliation boundary"
+                        )
+                    }
                 } else {
                     acceptedSessionIDs = [result.sessionId]
                 }
@@ -2852,18 +2857,17 @@ final class AppState: ObservableObject {
             )
         } else if normalizedCoveredText.hasPrefix(normalizedBufferedDeltaText) {
             coveredRawCharacters = bufferedDeltaText.count
-        } else if normalizedCoveredText.hasSuffix(normalizedBufferedDeltaText) {
-            // A reconnect can leave the local boundary behind a persisted
-            // portion of the turn. In that case the buffered window may begin
-            // in the middle of the covered span; consume it only when the
-            // entire buffered text is the covered span's suffix.
-            coveredRawCharacters = bufferedDeltaText.count
         } else {
-            let overlap = Self.longestSuffixPrefixOverlapLength(
+            let overlaps = Self.suffixPrefixOverlapLengths(
                 covered: normalizedCoveredText,
                 buffered: normalizedBufferedDeltaText
             )
-            guard overlap > 0 else { return events }
+            guard overlaps.count == 1, let overlap = overlaps.first else {
+                // Repeated content can produce multiple valid alignments. A
+                // content-only guess could consume genuinely new text, so
+                // preserve the events when the offset is ambiguous.
+                return events
+            }
             let leadingWhitespaceCount = bufferedDeltaText.prefix { $0.isWhitespace }.count
             coveredRawCharacters = leadingWhitespaceCount + overlap
         }
@@ -2895,15 +2899,16 @@ final class AppState: ObservableObject {
         return deduplicated
     }
 
-    /// Returns the longest prefix of `buffered` that is also a suffix of
+    /// Returns every non-empty prefix of `buffered` that is also a suffix of
     /// `covered`. The prefix-function scan stays linear in the cumulative
-    /// projection size while preserving the boundary-derived alignment rule.
-    nonisolated static func longestSuffixPrefixOverlapLength(
+    /// projection size; callers can reject repeated-content ambiguity when
+    /// more than one alignment is possible.
+    nonisolated static func suffixPrefixOverlapLengths(
         covered: String,
         buffered: String
-    ) -> Int {
+    ) -> [Int] {
         let pattern = Array(buffered)
-        guard !pattern.isEmpty, !covered.isEmpty else { return 0 }
+        guard !pattern.isEmpty, !covered.isEmpty else { return [] }
 
         var prefixLengths = Array(repeating: 0, count: pattern.count)
         var prefixLength = 0
@@ -2917,8 +2922,10 @@ final class AppState: ObservableObject {
             prefixLengths[index] = prefixLength
         }
 
+        let text = Array(covered)
         var matched = 0
-        for character in covered {
+        var overlaps: [Int] = []
+        for (index, character) in text.enumerated() {
             while matched > 0, pattern[matched] != character {
                 matched = prefixLengths[matched - 1]
             }
@@ -2926,10 +2933,18 @@ final class AppState: ObservableObject {
                 matched += 1
             }
             if matched == pattern.count {
+                if index == text.count - 1 {
+                    overlaps.append(pattern.count)
+                }
                 matched = prefixLengths[matched - 1]
             }
         }
-        return matched
+
+        while matched > 0 {
+            overlaps.append(matched)
+            matched = prefixLengths[matched - 1]
+        }
+        return overlaps
     }
 
     /// `session.resume.inflight` is a cumulative projection on some gateways.
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2311,16 +2311,37 @@ final class AppState: ObservableObject {
                 : []
             if result.snapshot.hasLiveProjection {
                 let resumedInflightText = streamingBuffer
+                let boundary = reconciliation?.token == token ? reconciliation : nil
+                let acceptedSessionIDs: Set<String>
+                if let boundary,
+                   let boundarySessionID = boundary.streamSessionIDAtBoundary {
+                    acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(
+                        knownSessionIDs(for: boundarySessionID)
+                    )
+                } else {
+                    acceptedSessionIDs = [result.sessionId]
+                }
                 let knownPrefix: String?
-                if let boundary = reconciliation, boundary.token == token {
+                let coveredText: String?
+                if let boundary {
                     knownPrefix = Self.normalizedReconciliationBoundaryPrefix(
                         boundaryText: boundary.streamTextAtBoundary,
                         boundarySessionID: boundary.streamSessionIDAtBoundary,
                         resumedSessionID: result.sessionId,
+                        acceptedSessionIDs: acceptedSessionIDs,
+                        after: messages
+                    )
+                    coveredText = Self.reconciliationBoundaryCoverageText(
+                        boundaryText: boundary.streamTextAtBoundary,
+                        boundarySessionID: boundary.streamSessionIDAtBoundary,
+                        resumedSessionID: result.sessionId,
+                        acceptedSessionIDs: acceptedSessionIDs,
+                        snapshotInflightText: result.snapshot.inflightAssistantText,
                         after: messages
                     )
                 } else {
                     knownPrefix = nil
+                    coveredText = nil
                 }
 
                 // The live bubble was just seeded from the cumulative inflight
@@ -2331,7 +2352,9 @@ final class AppState: ObservableObject {
                     bufferedEvents,
                     againstInflight: resumedInflightText,
                     knownPrefix: knownPrefix,
-                    sessionID: result.sessionId
+                    sessionID: result.sessionId,
+                    acceptedSessionIDs: acceptedSessionIDs,
+                    coveredText: coveredText
                 )
             }
             bufferedEvents.forEach(applyStreamEvent)
@@ -2698,34 +2721,85 @@ final class AppState: ObservableObject {
         boundaryText: String?,
         boundarySessionID: String?,
         resumedSessionID: String,
+        acceptedSessionIDs: Set<String>,
         after messages: [ChatMessage]
     ) -> String? {
         guard let boundaryText,
               let boundarySessionID,
               !boundarySessionID.isEmpty,
-              boundarySessionID == resumedSessionID else {
+              acceptedSessionIDs.contains(boundarySessionID),
+              acceptedSessionIDs.contains(resumedSessionID) else {
             return nil
         }
         return unpersistedInflightAssistantText(boundaryText, after: messages)
+    }
+
+    nonisolated static func reconciliationBoundaryCoverageText(
+        boundaryText: String?,
+        boundarySessionID: String?,
+        resumedSessionID: String,
+        acceptedSessionIDs: Set<String>,
+        snapshotInflightText: String,
+        after messages: [ChatMessage]
+    ) -> String? {
+        guard let boundaryText,
+              let boundarySessionID,
+              !boundarySessionID.isEmpty,
+              acceptedSessionIDs.contains(boundarySessionID),
+              acceptedSessionIDs.contains(resumedSessionID) else {
+            return nil
+        }
+
+        let normalizedBoundaryText = boundaryText.trimmingCharacters(in: .whitespacesAndNewlines)
+        let normalizedSnapshotText = snapshotInflightText.trimmingCharacters(in: .whitespacesAndNewlines)
+        if normalizedSnapshotText.hasPrefix(normalizedBoundaryText) {
+            return String(normalizedSnapshotText.dropFirst(normalizedBoundaryText.count))
+        }
+
+        let normalizedUnpersistedBoundary = unpersistedInflightAssistantText(
+            boundaryText,
+            after: messages
+        )
+        let normalizedUnpersistedSnapshot = unpersistedInflightAssistantText(
+            snapshotInflightText,
+            after: messages
+        )
+        guard normalizedUnpersistedSnapshot.hasPrefix(normalizedUnpersistedBoundary) else {
+            return nil
+        }
+        return String(
+            normalizedUnpersistedSnapshot.dropFirst(normalizedUnpersistedBoundary.count)
+        )
     }
 
     nonisolated static func deduplicatingBufferedEvents(
         _ events: [StreamEvent],
         againstInflight inflight: String,
         knownPrefix: String?,
-        sessionID: String
+        sessionID: String,
+        acceptedSessionIDs: Set<String> = [],
+        coveredText explicitCoveredText: String? = nil
     ) -> [StreamEvent] {
-        let normalizedInflight = inflight.trimmingCharacters(in: .whitespacesAndNewlines)
-        let normalizedKnownPrefix = knownPrefix?.trimmingCharacters(in: .whitespacesAndNewlines)
-        guard let normalizedKnownPrefix,
-              normalizedInflight.hasPrefix(normalizedKnownPrefix) else {
-            return events
+        let coveredText: String
+        if let explicitCoveredText {
+            coveredText = explicitCoveredText
+        } else {
+            guard let knownPrefix else { return events }
+            if inflight.hasPrefix(knownPrefix) {
+                coveredText = String(inflight.dropFirst(knownPrefix.count))
+            } else {
+                let normalizedInflight = inflight.trimmingCharacters(in: .whitespacesAndNewlines)
+                let normalizedKnownPrefix = knownPrefix.trimmingCharacters(in: .whitespacesAndNewlines)
+                guard normalizedInflight.hasPrefix(normalizedKnownPrefix) else {
+                    return events
+                }
+                coveredText = String(
+                    normalizedInflight.dropFirst(normalizedKnownPrefix.count)
+                )
+            }
         }
-
-        let coveredText = String(
-            normalizedInflight.dropFirst(normalizedKnownPrefix.count)
-        ).trimmingCharacters(in: .whitespacesAndNewlines)
-        guard !coveredText.isEmpty else { return events }
+        let normalizedCoveredText = coveredText.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !normalizedCoveredText.isEmpty else { return events }
 
         let bufferedDeltaTexts = events.compactMap { event in
             if case .messageDelta(_, let text) = event {
@@ -2741,8 +2815,12 @@ final class AppState: ObservableObject {
             }
             return nil
         })
+        let allowedSessionIDs = acceptedSessionIDs.isEmpty
+            ? Set([sessionID])
+            : acceptedSessionIDs
         guard deltaSessionIDs.count == 1,
-              deltaSessionIDs.first == sessionID else {
+              let deltaSessionID = deltaSessionIDs.first,
+              allowedSessionIDs.contains(deltaSessionID) else {
             return events
         }
 
@@ -2751,12 +2829,20 @@ final class AppState: ObservableObject {
         guard !normalizedBufferedDeltaText.isEmpty else { return events }
 
         let coveredRawCharacters: Int
-        if normalizedBufferedDeltaText == coveredText {
+        if normalizedBufferedDeltaText == normalizedCoveredText {
             coveredRawCharacters = bufferedDeltaText.count
-        } else if normalizedBufferedDeltaText.hasPrefix(coveredText) {
+        } else if normalizedBufferedDeltaText.hasPrefix(normalizedCoveredText) {
             let leadingWhitespaceCount = bufferedDeltaText.prefix { $0.isWhitespace }.count
-            coveredRawCharacters = leadingWhitespaceCount + coveredText.count
-        } else if coveredText.hasPrefix(normalizedBufferedDeltaText) {
+            let coveredEnd = leadingWhitespaceCount + normalizedCoveredText.count
+            let coveredTrailingWhitespaceCount = coveredText.reversed().prefix { $0.isWhitespace }.count
+            let bufferedTrailingWhitespaceCount = bufferedDeltaText.dropFirst(coveredEnd)
+                .prefix { $0.isWhitespace }
+                .count
+            coveredRawCharacters = coveredEnd + min(
+                coveredTrailingWhitespaceCount,
+                bufferedTrailingWhitespaceCount
+            )
+        } else if normalizedCoveredText.hasPrefix(normalizedBufferedDeltaText) {
             coveredRawCharacters = bufferedDeltaText.count
         } else {
             return events
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2795,6 +2795,25 @@ final class AppState: ObservableObject {
     ) -> [StreamEvent] {
         let coveredText: String
         if let explicitCoveredText {
+            guard let knownPrefix else { return events }
+            let normalizedInflight = inflight.trimmingCharacters(in: .whitespacesAndNewlines)
+            let normalizedKnownPrefix = knownPrefix.trimmingCharacters(in: .whitespacesAndNewlines)
+            guard normalizedInflight.hasPrefix(normalizedKnownPrefix) else {
+                return events
+            }
+            let expectedCoveredText = String(
+                normalizedInflight.dropFirst(normalizedKnownPrefix.count)
+            )
+            let normalizedExplicitCoveredText = explicitCoveredText
+                .trimmingCharacters(in: .whitespacesAndNewlines)
+            let explicitCoverageOverlapsSeededText = Self.suffixPrefixOverlapLengths(
+                covered: normalizedExplicitCoveredText,
+                buffered: expectedCoveredText
+            )
+            guard !expectedCoveredText.isEmpty,
+                  explicitCoverageOverlapsSeededText.count == 1 else {
+                return events
+            }
             coveredText = explicitCoveredText
         } else {
             guard let knownPrefix else { return events }
@@ -2835,6 +2854,26 @@ final class AppState: ObservableObject {
               let deltaSessionID = deltaSessionIDs.first,
               allowedSessionIDs.contains(deltaSessionID) else {
             return events
+        }
+
+        if let newTurnIndex = events.firstIndex(where: { event in
+            guard case .messageStart(let startSessionID) = event else { return false }
+            return allowedSessionIDs.contains(startSessionID)
+        }) {
+            // A message start observed after the reconciliation boundary is
+            // an explicit new-turn marker. Deduplicate any older buffered
+            // events, but preserve the marker and everything after it because
+            // the same text can be fresh content from the new turn.
+            let eventsBeforeNewTurn = Array(events[..<newTurnIndex])
+            let eventsAfterNewTurn = Array(events[newTurnIndex...])
+            return deduplicatingBufferedEvents(
+                eventsBeforeNewTurn,
+                againstInflight: inflight,
+                knownPrefix: knownPrefix,
+                sessionID: sessionID,
+                acceptedSessionIDs: acceptedSessionIDs,
+                coveredText: explicitCoveredText
+            ) + eventsAfterNewTurn
         }
 
         let bufferedDeltaText = bufferedDeltaTexts.joined()
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2315,6 +2315,12 @@ final class AppState: ObservableObject {
                 let acceptedSessionIDs: Set<String>
                 if let boundary,
                    let boundarySessionID = boundary.streamSessionIDAtBoundary {
+                    // Do not infer that a newly returned runtime ID belongs
+                    // to this boundary solely because the resume RPC returned
+                    // it. Keep only IDs already accepted for the boundary and
+                    // known as aliases of that catalog session; an empty
+                    // result intentionally disables deduplication rather than
+                    // risking text from a different session.
                     acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(
                         knownSessionIDs(for: boundarySessionID)
                     )
@@ -2713,10 +2719,12 @@ final class AppState: ObservableObject {
     /// in the resume snapshot's cumulative `inflight` projection — replaying
     /// them on top of the seeded live bubble repeats that text. When the exact
     /// stream text at the matching session boundary is known, consume only the
-    /// corresponding prefix of the buffered deltas. Edge whitespace is ignored
+    /// corresponding span of the buffered deltas. Edge whitespace is ignored
     /// consistently with resume seeding, and the raw covered count preserves
-    /// event order when deltas carry that whitespace. Without a matching
-    /// boundary or session, repeated text is ambiguous, so leave events intact.
+    /// event order when deltas carry that whitespace. Interior whitespace is
+    /// not rewritten: a mismatch may be real content, so leave it intact.
+    /// Without a matching boundary or session, repeated text is ambiguous, so
+    /// leave events intact.
     nonisolated static func normalizedReconciliationBoundaryPrefix(
         boundaryText: String?,
         boundarySessionID: String?,
@@ -2843,6 +2851,12 @@ final class AppState: ObservableObject {
                 bufferedTrailingWhitespaceCount
             )
         } else if normalizedCoveredText.hasPrefix(normalizedBufferedDeltaText) {
+            coveredRawCharacters = bufferedDeltaText.count
+        } else if normalizedCoveredText.hasSuffix(normalizedBufferedDeltaText) {
+            // A reconnect can leave the local boundary behind a persisted
+            // portion of the turn. In that case the buffered window may begin
+            // in the middle of the covered span; consume it only when the
+            // entire buffered text is the covered span's suffix.
             coveredRawCharacters = bufferedDeltaText.count
         } else {
             return events
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2859,7 +2859,13 @@ final class AppState: ObservableObject {
             // entire buffered text is the covered span's suffix.
             coveredRawCharacters = bufferedDeltaText.count
         } else {
-            return events
+            let overlap = Self.longestSuffixPrefixOverlapLength(
+                covered: normalizedCoveredText,
+                buffered: normalizedBufferedDeltaText
+            )
+            guard overlap > 0 else { return events }
+            let leadingWhitespaceCount = bufferedDeltaText.prefix { $0.isWhitespace }.count
+            coveredRawCharacters = leadingWhitespaceCount + overlap
         }
 
         var remainingCoverage = coveredRawCharacters
@@ -2887,6 +2893,43 @@ final class AppState: ObservableObject {
             }
         }
         return deduplicated
+    }
+
+    /// Returns the longest prefix of `buffered` that is also a suffix of
+    /// `covered`. The prefix-function scan stays linear in the cumulative
+    /// projection size while preserving the boundary-derived alignment rule.
+    nonisolated static func longestSuffixPrefixOverlapLength(
+        covered: String,
+        buffered: String
+    ) -> Int {
+        let pattern = Array(buffered)
+        guard !pattern.isEmpty, !covered.isEmpty else { return 0 }
+
+        var prefixLengths = Array(repeating: 0, count: pattern.count)
+        var prefixLength = 0
+        for index in 1..<pattern.count {
+            while prefixLength > 0, pattern[index] != pattern[prefixLength] {
+                prefixLength = prefixLengths[prefixLength - 1]
+            }
+            if pattern[index] == pattern[prefixLength] {
+                prefixLength += 1
+            }
+            prefixLengths[index] = prefixLength
+        }
+
+        var matched = 0
+        for character in covered {
+            while matched > 0, pattern[matched] != character {
+                matched = prefixLengths[matched - 1]
+            }
+            if pattern[matched] == character {
+                matched += 1
+            }
+            if matched == pattern.count {
+                matched = prefixLengths[matched - 1]
+            }
+        }
+        return matched
     }
 
     /// `session.resume.inflight` is a cumulative projection on some gateways.
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2286,11 +2286,21 @@ final class AppState: ObservableObject {
                 return false
             }
 
-            let bufferedEvents = reconciliation?.token == token
+            var bufferedEvents = reconciliation?.token == token
                 ? (reconciliation?.bufferedEvents ?? []).filter {
                     reconciliation?.accepts(sessionID(for: $0)) == true
                 }
                 : []
+            if result.snapshot.hasLiveProjection {
+                // The live bubble was just seeded from the cumulative inflight
+                // projection, which already includes deltas emitted while this
+                // reconciliation was in flight. Replaying them verbatim repeats
+                // that text in the bubble.
+                bufferedEvents = Self.deduplicatingBufferedEvents(
+                    bufferedEvents,
+                    againstInflight: result.snapshot.inflightAssistantText
+                )
+            }
             bufferedEvents.forEach(applyStreamEvent)
             return settleReconciliationAndPublish(
                 token,
@@ -2641,6 +2651,61 @@ final class AppState: ObservableObject {
                 messages[index].approval = approval
             }
         }
+    }
+
+    /// Deltas buffered while reconciliation ran are usually already contained
+    /// in the resume snapshot's cumulative `inflight` projection — replaying
+    /// them on top of the seeded live bubble repeats that text. Collapse the
+    /// buffered deltas to the suffix the snapshot has not covered yet (their
+    /// concatenation overlaps the snapshot's tail), emitted as a single delta
+    /// in place of the last buffered one. Non-delta events pass through.
+    nonisolated static func deduplicatingBufferedEvents(
+        _ events: [StreamEvent],
+        againstInflight inflight: String
+    ) -> [StreamEvent] {
+        guard !inflight.isEmpty else { return events }
+        let deltaTexts: [String] = events.compactMap {
+            if case .messageDelta(_, let text) = $0 { return text }
+            return nil
+        }
+        guard !deltaTexts.isEmpty else { return events }
+
+        let combined = deltaTexts.joined()
+        let covered = overlapLength(betweenSuffixOf: inflight, andPrefixOf: combined)
+        let remainder = String(combined.dropFirst(covered))
+        guard remainder.count < combined.count else { return events }
+
+        let lastDeltaIndex = events.lastIndex {
+            if case .messageDelta = $0 { return true }
+            return false
+        }
+        var deduplicated: [StreamEvent] = []
+        for (index, event) in events.enumerated() {
+            guard case .messageDelta(let sessionId, _) = event else {
+                deduplicated.append(event)
+                continue
+            }
+            if index == lastDeltaIndex, !remainder.isEmpty {
+                deduplicated.append(.messageDelta(sessionId: sessionId, text: remainder))
+            }
+        }
+        return deduplicated
+    }
+
+    /// Longest `k` where the last `k` characters of `text` equal the first
+    /// `k` characters of `candidate`. The buffered stream is contiguous, so
+    /// the snapshot's tail and the buffer's head meet at exactly one offset.
+    nonisolated static func overlapLength(betweenSuffixOf text: String, andPrefixOf candidate: String) -> Int {
+        let textChars = Array(text)
+        let candidateChars = Array(candidate)
+        var k = min(textChars.count, candidateChars.count)
+        while k > 0 {
+            if textChars.suffix(k).elementsEqual(candidateChars.prefix(k)) {
+                return k
+            }
+            k -= 1
+        }
+        return 0
     }
 
     /// `session.resume.inflight` is a cumulative projection on some gateways.
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2707,6 +2707,14 @@ final class AppState: ObservableObject {
         let covered = inflight.count - knownPrefix.count
         guard covered > 0 else { return events }
 
+        let bufferedDeltaTexts = events.compactMap { event in
+            if case .messageDelta(_, let text) = event {
+                return text
+            }
+            return nil
+        }
+        guard !bufferedDeltaTexts.isEmpty else { return events }
+
         let deltaSessionIDs = Set(events.compactMap { event in
             if case .messageDelta(let sessionId, _) = event {
                 return sessionId
@@ -2714,6 +2722,13 @@ final class AppState: ObservableObject {
             return nil
         })
         guard deltaSessionIDs.count == 1 else {
+            return events
+        }
+
+        let bufferedDeltaText = bufferedDeltaTexts.joined()
+        let inflightSuffix = String(inflight.dropFirst(knownPrefix.count))
+        guard bufferedDeltaText.hasPrefix(inflightSuffix)
+                || inflightSuffix.hasPrefix(bufferedDeltaText) else {
             return events
         }
 
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2289,9 +2289,6 @@ final class AppState: ObservableObject {
                 settleReconciliation(token, automaticSyncOperationID: automaticSyncOperationID)
                 return false
             }
-            let resumedInflightText = result.snapshot.hasLiveProjection
-                ? streamingBuffer
-                : nil
             await refreshChatResumeContext(sessionId: result.sessionId, using: client)
 
             guard automaticChatResumeWorkIsCurrent(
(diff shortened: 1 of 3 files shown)


kaishi00:

[github run](/kaishi00/hermes-conduit/actions/runs/31533029559)

kaishi00:
## Summary
This PR deduplicates stream events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection so the live bubble isn't re-fed text it already contains, and bundles three bug fixes (WebKit handler retain cycle, in-flight RPC hang on socket death, regex catastrophic backtracking) plus a new test file.

## Findings

### [major] Deduped delta remainder is replayed out of order across interleaved non-delta events
**File:** Conduit/Services/AppState.swift:2673-2692
**Issue:** All buffered delta text is concatenated, the covered prefix is stripped, and the remaining text is emitted as a single delta at the position of the *last* delta. Any uncovered text that originally appeared *before* an intervening non-delta event (toolStart/toolComplete, sessionBusy, etc.) is moved to after it. Example: buffer `[.messageDelta("abcd"), .toolStart(...), .messageDelta("ef")]` with an inflight tail of `"abc"` should replay `.messageDelta("d")`, `.toolStart`, `.messageDelta("ef")`, but the current code replays `.toolStart`, `.messageDelta("def")`. Agent turns routinely interleave tool events between text deltas, so this misorders the streaming bubble relative to tool cards after foregrounding.
**Suggestion:** Consume the covered count sequentially across each `.messageDelta` in original event order and emit each non-empty uncovered suffix at its own event position (dropping fully-covered deltas), e.g.:

```swift
guard covered > 0 else { return events }
var remainingCoverage = covered
return events.compactMap { event in
    guard case .messageDelta(let sessionId, let text) = event else { return event }
    let coveredInEvent = min(remainingCoverage, text.count)
    remainingCoverage -= coveredInEvent
    let suffix = String(text.dropFirst(coveredInEvent))
    return suffix.isEmpty ? nil : .messageDelta(sessionId: sessionId, text: suffix)
}
```

Add a regression test with an interleaved tool event (the new tests only cover `toolStart` *before* any deltas, which cannot expose this).

### [minor] `overlapLength` is O(n²) with repeated array allocation on the main actor
**File:** Conduit/Services/AppState.swift:2698-2709
**Issue:** `textChars.suffix(k).elementsEqual(candidateChars.prefix(k))` builds two new `Array`s each iteration and scans every k from `min(...)` down to 0. For the no-overlap path (non-cumulative gateways or a turn started after the snapshot — the very cases where dedup is skipped anyway) this is quadratic; a long inflight text (thousands of chars) can cause a noticeable main-thread hitch since this runs synchronously inside the `@MainActor` reconcile flow.
**Suggestion:** Return `0` early when either input is empty (already handled at the caller), and use a linear-time longest-border algorithm (KMP prefix function or a Z-array) over the concatenated string, or compare via `String.Index` slices without materializing arrays.

### [minor] Longest-match overlap is ambiguous for self-repeating content
**File:** Conduit/Services/AppState.swift:2674
**Issue:** The assumption that "the snapshot's tail and the buffer's head meet at exactly one offset" only holds for non-repetitive text. For streams with repeated runs (e.g. inflight `"aaaa"`, buffered `"aaa"`), the longest suffix/prefix match over-covers and silently drops genuinely new text; the opposite can also under-dedup. The doc comment states the streams "meet at exactly one offset," but content-based matching cannot guarantee this.
**Suggestion:** Document the heuristic limitation, and consider bounding the search (e.g. only consider overlaps ≥ some minimum) or matching against the exact byte/event position when available. Minor spacing corruption can also occur because the bubble is seeded with the whitespace-trimmed `unpersistedInflightAssistantText` while the overlap is computed against the untrimmed raw inflight.

### [minor] Deltas from multiple sessions are concatenated and re-attributed
**File:** Conduit/Services/AppState.swift:2667-2692
**Issue:** `deltaTexts` concatenates all `.messageDelta` texts regardless of session, and the emitted remainder is tagged with the *last* delta's `sessionId`. With `reconciliation.acceptsAnySession == true` (or multiple accepted session IDs), interleaved deltas from different sessions would be merged and mis-attributed to one session.
**Suggestion:** Group the dedup per session (or bail out when more than one distinct session id is present in the buffered deltas), so text is never moved across a session boundary.

### [nit] `WeakScriptMessageHandler` is not actor-annotated (strict-concurrency forward-compat)
**File:** Conduit/Services/DashboardTicketBridge.swift:136-146
**Issue:** The proxy is a non-isolated `WKScriptMessageHandler` whose `delegate` is the `@MainActor` `DashboardTicketBridge` conformance. This compiles today, but under Swift 6 strict concurrency the `@MainActor`-isolated conformance to a non-isolated protocol plus the cross-isolation call from the proxy would surface warnings/errors. WebKit delivers script messages on the main thread, so the proxy can be safely annotated.
**Suggestion:** Mark `WeakScriptMessageHandler` `@MainActor` (matching where the callbacks actually arrive) so the delegate hop is explicitly main-actor isolated. The cycle fix itself is correct.

## Verdict
REQUEST_CHANGES

The dedup approach is sound and well-tested, but the major event-ordering defect (uncovered delta text repositioned across interleaved tool events) contradicts the stated "non-delta events always pass through in order" invariant and has a straightforward fix.

<a href="https://opencode.ai/s/2JDV3OwE">New%20session%20-%202026-08-12T02%3A09%3A55.546Z</a>
[opencode session](https://opencode.ai/s/2JDV3OwE)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31555926985)

kaishi00:
## Summary
This PR deduplicates stream events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection so the live bubble isn't re-fed text it already contains: it captures the stream text/session at the reconciliation boundary, computes the overlap between the snapshot's tail and the buffered delta concatenation, and consumes only the covered prefix (sequential per-event, preserving interleaved non-delta order). Prior review comments (event ordering, suffix validation) are addressed in the current HEAD.

## Findings

### [major] Boundary prefix domain mismatch: dedup silently no-ops when the current turn has an already-persisted prefix
**File:** Conduit/Services/AppState.swift:2316-2337
**Issue:** `knownPrefix` is the raw `streamingBuffer` captured at `beginReconciliation` (AppState.swift:2025), which holds the *full* current-turn streamed text including any prefix already persisted to the transcript. But the value dedup is compared against, `resumedInflightText` (line 2292-2294), is `streamingBuffer` *after* `applyChatResume` seeded it with `unpersistedInflightAssistantText` (line 2622-2628) — a whitespace-trimmed, persisted-prefix-stripped suffix. The guard `inflight.hasPrefix(knownPrefix)` (line 2703) then fails and the function returns `events` unchanged. This is precisely the scenario `unpersistedInflightAssistantText`'s own docstring describes ("its already-persisted prefix is also present in the recovered transcript"), so the duplicate-text bug this PR sets out to fix persists for cumulative gateways whose current-turn prefix was already persisted. The mismatch is only detectable at integration level — the unit tests hand-construct aligned inputs.
**Suggestion:** Derive `knownPrefix` through the same normalization the seeding uses (e.g. strip any persisted assistant message that is a prefix of the boundary text), or drop the boundary requirement and compute coverage directly as the longest overlap between the *seeded* `resumedInflightText` and the concatenated buffered delta text (the ordering + suffix-validation logic already makes this safe). Add an integration-style test where the persisted message is a prefix of the buffered deltas.

### [minor] Boundary session check can apply one session's text length to another session's deltas
**File:** Conduit/Services/AppState.swift:2317-2327
**Issue:** `knownPrefix` is used whenever `boundarySessionID` equals the resumed session *or* the requested session *or* is merely in `acceptedSessionIDs`. With `acceptsAnySession == true` (from `beginReconciliation`, line 2032) and the session-change path that carries the prior boundary forward (line 2021-2027), a boundary text captured for an older session can be applied against buffered deltas of a different resumed session. The `deltaSessionIDs.count == 1` guard binds the buffered deltas to a single session but not to the resumed one. If the old session's stream text happens to be a prefix of the new session's seeded inflight (e.g. both start with "Okay"), genuinely new content is dropped.
**Suggestion:** Require `boundarySessionID == result.sessionId` for applying `knownPrefix`, or store the boundary text keyed by session and only reuse it when it belongs to the resumed session. Add a test where the boundary session differs from the delta session.

### [minor] Trailing-whitespace trimming causes slight over-inclusion in replay
**File:** Conduit/Services/AppState.swift:2707, 2728-2733
**Issue:** `covered = inflight.count - knownPrefix.count` is computed against the trimmed seed, while the buffered delta text retains its raw trailing whitespace. When the final buffered delta ends in whitespace that the seeding trimmed off, `covered` underestimates the true overlap, so a few trailing whitespace characters are replayed on top of the seeded bubble. Conversely, if the boundary text ends in whitespace, `inflight.hasPrefix(knownPrefix)` can fail and skip dedup entirely. Cosmetic but visible as stray spacing / a repeated blank line after foregrounding.
**Suggestion:** Trim both sides consistently (or match on the untrimmed cumulative inflight rather than the trimmed seed), and document that the overlap heuristic is whitespace-sensitive.

### [nit] Dead `?? ""` fallback
**File:** Conduit/Services/AppState.swift:2335
**Issue:** `resumedInflightText` is only non-`nil` when `result.snapshot.hasLiveProjection` is true (line 2292), and the surrounding `if result.snapshot.hasLiveProjection` block guarantees it is non-`nil` here, so `?? ""` never fires. The empty-inflight case is instead exercised by the `covered > 0` guard.
**Suggestion:** Pass `resumedInflightText` directly (the guard handles empty) or restructure to make the invariant explicit.

### [nit] No tests cover the boundary-capture → seeding integration path
**File:** ConduitTests/BufferedEventDeduplicationTests.swift:28-213
**Issue:** All 11 tests drive the pure function with hand-aligned inputs. None exercise the wiring at AppState.swift:2016-2037/2284-2338, so the domain mismatch in the major finding above is invisible to the suite (as is the whitespace trimming interplay).
**Suggestion:** Add at least one test for the full resume path with a partially-persisted current turn, asserting the replayed bubble equals the seeded text plus only the truly-uncovered suffix.

## Verdict
REQUEST_CHANGES

The dedup algorithm itself is now correct (sequential per-event consumption preserves interleaved ordering, the suffix validation prevents length-only over-consumption, and it fails safe — it never drops text when inputs are inconsistent). The concern is the boundary-vs-seed domain mismatch: the core fix silently degrades to a no-op in a documented common scenario (already-persisted prefix in a cumulative inflight), and the integration path is untested.

<a href="https://opencode.ai/s/eWYFT0Xr">New%20session%20-%202026-08-12T17%3A07%3A02.527Z</a>
[opencode session](https://opencode.ai/s/eWYFT0Xr)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31620864770)

kaishi00:
## Summary
This PR deduplicates stream events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection: it captures the stream text/session at the reconciliation boundary, strips any persisted assistant prefix, and consumes only the covered prefix of the buffered `.messageDelta` events before replay. The earlier ordering and suffix-validation defects are fixed in HEAD (sequential per-event consumption, exact-prefix validation), and the algorithm is fail-safe in the data-loss direction (it never drops text when inputs are inconsistent).

## Findings

### [major] Prefix-only overlap silently no-ops (duplicate persists) when the local boundary lags the persisted transcript
**File:** Conduit/Services/AppState.swift:2718-2763
**Issue:** `coveredText` is computed as `unpersistedInflight − unpersistedBoundary` and dedup only fires when that text is an exact *prefix* of the concatenated buffered deltas. When the local `streamingBuffer` at the boundary is a proper prefix of the last persisted assistant message (e.g., deltas lost while backgrounded / socket death, then a cumulative gateway re-sends the gap during the window), the buffered deltas are `" brown fox"` while the seeded bubble is the unpersisted suffix `"fox"` and `coveredText` is `"fox"`. `"fox"` is a *suffix*, not a prefix, of `" brown fox"` → all three branches fail → `return events` unchanged → `" brown fox"` replays on top of the seeded `"fox"`, producing `"fox brown fox"`. This is precisely the cumulative-gateway/persisted-prefix scenario the PR targets (and the `unpersistedInflightAssistantText` docstring), so the duplicate-text bug the PR sets out to fix persists in this edge.
**Suggestion:** Document the prefix-contiguity assumption explicitly, and detect the gap case — e.g., if the seeded bubble's unpersisted inflight is non-empty and is *not* a suffix of the boundary-captured cumulative text, skip dedup deliberately (or compute coverage against the raw boundary/inflight domain rather than the unpersisted domain). Add a regression test with boundary `"The quick"`, persisted `"The quick brown"`, inflight `"The quick brown fox"`, and buffered delta `" brown fox"`.

### [minor] Buffered deltas carrying a requested/alternate session ID bypass dedup
**File:** Conduit/Services/AppState.swift:2738-2747
**Issue:** The filter at line 2307 accepts events via `reconciliation.accepts(...)`, which admits `requestedSessionId`, `acceptedSessionIDs`, and runtime aliases. But the dedup guard requires `deltaSessionIDs.first == sessionID` where `sessionID` is the resolved `result.sessionId` (line 2744). If the gateway emits window deltas tagged with the requested/alternate ID rather than the resolved runtime ID, dedup silently returns `events` unchanged and the already-covered text replays — the same visible duplicate the PR fixes. This is plausible because `applyStreamEvent` (via `eventBelongsToActiveSession`, line 5843) explicitly supports alias IDs.
**Suggestion:** Normalize buffered delta IDs against the reconciliation's accepted ID set before comparing (map each delta ID to the resolved session), or relax the guard to accept when all delta IDs are in `acceptedSessionIDs ∪ {result.sessionId}`.

### [nit] No integration coverage for partial overlap or the persisted-prefix gap
**File:** ConduitTests/AppStateChatResumeTests.swift:123-227
**Issue:** The two integration tests cover only full coverage (`" I found it."` dropped) and session mismatch. The partial-overlap path — the interleaved-tool ordering case and the gap/lag case above — is exercised only through the pure-function unit tests with hand-aligned inputs, so the boundary-capture → seeding wiring (AppState.swift:2016-2027 / 2284-2336) is never validated end-to-end with partial coverage.
**Suggestion:** Add an integration test where the buffered window contains an interleaved `.toolStart` plus a partially covered delta, asserting the replayed bubble equals the seeded text plus only the uncovered suffix.

### [nit] PR description no longer matches the implementation
**File:** PR description
**Issue:** The body describes the algorithm as "the longest overlap between the inflight's suffix and the buffer's prefix"; the current code instead does deterministic exact-prefix matching derived from the boundary (`inflight.dropFirst(knownPrefix)` with equality/prefix validation). The "longest-match selection" tests likewise reflect exact matching, not a longest-overlap search. Keeping these in sync matters because the actual (stronger) invariant is prefix-contiguity, not longest overlap.
**Suggestion:** Update the PR description to state that coverage is computed deterministically from the captured boundary and applied only when the buffered delta text begins with exactly that covered prefix.

## Verdict
COMMENT

The dedup mechanism is sound, well-tested at the unit level, and fails safe (it never over-consumes text; the ordering and suffix-validation majors from earlier reviews are genuinely addressed). The remaining gap — prefix-only matching no-ops when the local boundary is misaligned with the persisted prefix — is real and in-scope but narrow and non-destructive; it is worth a follow-up commit plus a regression test rather than blocking this merge.

<a href="https://opencode.ai/s/zXDAJs2W">New%20session%20-%202026-08-12T18%3A12%3A38.093Z</a>
[opencode session](https://opencode.ai/s/zXDAJs2W)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31625124678)

kaishi00:
## Summary
This PR deduplicates stream events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection: it captures the stream text/session at the reconciliation boundary, derives the covered span, and consumes only that prefix of the buffered `.messageDelta` events before replay. Earlier review majors (per-event ordering, inflight-suffix validation, whitespace handling, session aliasing) are addressed and covered by tests in the current HEAD.

## Findings

### [major] Dedup silently no-ops when the buffered window is a suffix (not prefix) of the covered span, re-enabling the duplicate
**File:** Conduit/Services/AppState.swift:2831-2849
**Issue:** `deduplicatingBufferedEvents` only fires when the normalized buffered delta text is equal to, a prefix of, or an extension of the covered text. When the buffered window is a *suffix* of the covered span, all three branches fail and the events replay unchanged. This is reachable in the PR's own target scenario. `reconciliationBoundaryCoverageText` (line 2753-2757) computes coverage from the *raw* snapshot minus the raw boundary, and the consumption must then align with the *seeded* bubble (persisted-prefix-stripped, per `applyChatResume` line 2644-2649). Concrete reproduction: boundary `"A"` (local `streamingBuffer` lags), transcript persists `"AB"` mid-window, snapshot `"ABC"`, buffered delta `"C"`. Covered text is `"BC"`; `"C"` is neither equal nor a prefix of `"BC"`, so nothing is dropped — `"C"` replays onto the seeded bubble `"C"`, producing `"CC"`. The local boundary is captured once and carried forward across reconnects (AppState.swift:2021-2027, 2203-2204), so a mid-window gap (socket death / lost deltas) leaves the buffer suffix-aligned to the coverage and reproduces the exact duplicate this PR exists to remove. The code fails safe in the data-loss direction only; the duplicate-text direction is unhandled.
**Suggestion:** When the prefix checks fail but `normalizedCoveredText.hasSuffix(normalizedBufferedDeltaText)` holds, consume the buffered events in full (the snapshot cumulative projection covers them). When they only partially overlap, restrict consumption to the correctly-aligned intersection and keep the rest. Add a regression test for boundary `"A"`, snapshot `"ABC"`, persisted `"AB"`, buffered `"C"` (asserting `"C"` is not replayed), plus one for a mid-window gap (buffered `"CD"`, covered `"BCD"`).

### [minor] `acceptedSessionIDs` silently collapses to empty when reconcile runs without catalog-known IDs
**File:** Conduit/Services/AppState.swift:2315-2323
**Issue:** `acceptedSessionIDs` is `boundary.acceptedSessionIDs.intersection(knownSessionIDs(for: boundarySessionID))`. Every guard downstream (`normalizedReconciliationBoundaryPrefix` and `reconciliationBoundaryCoverageText`, lines 2730-2731/2748-2749) requires `acceptedSessionIDs.contains(boundarySessionID)` and `.contains(resumedSessionID)`. Today this works because `refreshActiveSession` passes `knownSessionIDs(for: sessionId)` into `reconcile` (line 3821), but any caller invoking `reconcile`/`beginReconciliation` without accepted IDs (or with an `activeSessionId` whose runtime alias isn't yet in the catalog) produces an empty set and silently disables dedup — no error, no notification. It is also brittle: an alias that was registered in `acceptedSessionIDs` but not present at intersection time (e.g. `knownSessionIDs` falls back to `[boundarySessionID]` at line 3568) can exclude the resolved `result.sessionId`.
**Suggestion:** Default the intersection to `[boundarySessionID, result.sessionId]` (or union the session's aliases from `applyChatResume`'s `sessionIDs` at line 2562) and add a guard/comment so a failed intersection cannot silently skip dedup in the primary reconcile path.

### [minor] Coverage arithmetic is whitespace-sensitive in ways the seeded bubble is not
**File:** Conduit/Services/AppState.swift:2753-2757, 2834-2844
**Issue:** `coveredText` is computed on whitespace-*trimmed* strings while consumption is applied to raw buffered text via the `leadingWhitespaceCount`/`coveredEnd` arithmetic. This works for the tested shapes, but any divergence in interior or repeated whitespace (e.g. buffered `"foo  bar"` vs covered `"foo bar"`) fails the `hasPrefix` check and leaves the duplicate in place; conversely, when the seeded bubble was trimmed and the covered text ends in whitespace, the trailing-whitespace `min(...)` (lines 2841-2844) can under- or over-consume slightly, leaving a stray separator or consuming a genuine single-space prefix of new content.
**Suggestion:** Compute coverage in the same domain the seeding uses — derive `coveredText` from `unpersistedInflightAssistantText` on both boundary and snapshot — and document the trim sensitivity (the reconciliation comment at 2716-2718 claims it is handled consistently, but the raw-vs-normalized mixing above is not).

### [nit] Coverage helpers ignore the seeded-bubble domain; two parallel normalization paths can disagree
**File:** Conduit/Services/AppState.swift:2737-2773 vs 2783-2801
**Issue:** The integration path passes an explicit `coveredText` computed from the raw snapshot/boundary, while the fallback inside `deduplicatingBufferedEvents` computes coverage from the *seeded* `streamingBuffer` minus `knownPrefix`. When `reconciliationBoundaryCoverageText` first branch fails and the fallback runs, `knownPrefix` (`unpersistedInflightAssistantText(boundary)`) is compared against the trimmed seeded buffer — a different domain. If the persisted transcript is a proper prefix of the boundary (unpersisted boundary is non-empty but shorter than the raw boundary), the prefix check can fail and skip dedup even though coverage genuinely exists.
**Suggestion:** Unify on one source of truth: compute `coveredText` from the persisted-normalized values in both paths, or drop the `knownPrefix` fallback entirely and always require the explicit coverage (the seeded-buffer prefix is already validated by the equality/prefix guards).

### [nit] No test covers the suffix-aligned (gap) case at unit or integration level
**File:** ConduitTests/BufferedEventDeduplicationTests.swift:28-213, ConduitTests/AppStateChatResumeTests.swift:123-254
**Issue:** All 11 unit tests and 3 integration tests align `coveredText` with a buffered prefix (equal/prefix/extension). None feed the suffix-aligned inputs that trigger the major finding above, so the suite cannot detect that the PR's duplicate bug survives in the gap path.
**Suggestion:** Add the unit regression (inflight/snapshot `"ABC"`, boundary `"A"`, buffered `"C"`) and an integration test where a mid-window reconnect carries the stale boundary forward and the replay stays free of duplicates.

## Verdict
REQUEST_CHANGES

The algorithm is now sound for contiguous windows — ordering, suffix validation, whitespace, and session aliasing are handled and fail safe against data loss — but the core fix still lets the duplicate resurface when the buffered window is suffix-aligned to the covered span (mid-window gap or a persisted-prefix inside the coverage), which is precisely the reconnect-driven foreground scenario in scope. No retain cycles, force unwraps, or concurrency issues were introduced; the new statics are `nonisolated` pure functions and the replay loop is linear.

<a href="https://opencode.ai/s/MS81o0kK">New%20session%20-%202026-08-12T18%3A40%3A01.171Z</a>
[opencode session](https://opencode.ai/s/MS81o0kK)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31628694597)

kaishi00:
## Summary
This PR deduplicates `.messageDelta` events buffered during foreground reconciliation against the resume snapshot's cumulative inflight projection: it captures stream text/session at the reconciliation boundary, derives the covered span (persisted-prefix normalized), validates the alignment before consuming, and sequentially trims the covered prefix off each buffered delta while preserving interleaved non-delta event order. The current HEAD addresses the earlier ordering, suffix-validation, whitespace, and session-alias majors and fails safe against data loss.

## Findings

### [major] Partial overlap where the buffered window straddles the covered span is never deduplicated
**File:** Conduit/Services/AppState.swift:2839-2863
**Issue:** The four alignment branches (equal, buffered.hasPrefix(covered), covered.hasPrefix(buffered), covered.hasSuffix(buffered)) miss the straddle case: buffered text begins inside the covered span and extends into genuinely new text. Example: boundary `"A"`, snapshot `"ABCD"` (covered `"BCD"`), window captured `"CDE"` — `"CD"` is covered, `"E"` is new. None of the four branches match (`"CDE"` is neither equal/prefix/suffix of `"BCD"`), so `return events` fires and the full `"CDE"` replays on top of the seeded bubble `"CD"`, producing `"CDCDE"`. This is the exact "local boundary lagged a persisted/gap portion" scenario the new suffix branch's own comment documents (line 2856-2859), but the suffix branch only consumes when the *entire* buffered text is covered; the boundary-crossing continuation is the common reconnect/lossy-window variant and reproduces the duplicate this PR exists to remove.
**Suggestion:** When no pure prefix/suffix/equality relationship holds, compute the longest overlap between `coveredText`'s suffix and the buffered concatenation's prefix (the original `overlapLength` approach) and consume only that intersection, keeping the trailing new text. Add a regression for covered `"BCD"` with buffered `"CDE"` (and the interleaved-event variant) at both the unit and integration level.

### [minor] Resume returning a runtime session ID unknown to the catalog silently disables dedup in the primary path
**File:** Conduit/Services/AppState.swift:2315-2351
**Issue:** `acceptedSessionIDs` is `boundary.acceptedSessionIDs ∩ knownSessionIDs(boundarySessionID)`, and `normalizedReconciliationBoundaryPrefix`/`reconciliationBoundaryCoverageText` both require `acceptedSessionIDs.contains(resumedSessionID)`. When `result.sessionId` (inserted into `reconciliation.acceptedSessionIDs` at line 2245) is not yet a known catalog alias, the intersection drops it, both helpers return `nil`, and dedup returns events unchanged — the visible duplicate persists on the very first resume that produces a fresh runtime ID, with no diagnostic. The only test passing with an alternate ID pre-registers it (`session("stored-a", alternateIDs: ["runtime-a"])` in AppStateChatResumeTests.swift), so a catalog-lagging runtime ID is untested.
**Suggestion:** Union the resolved runtime ID (or the reconcile `sessionIDs` from applyChatResume:2568) into the accepted set before intersecting, or log/no-op explicitly when dedup is disabled for this reason. Add a test where the buffered delta is tagged with the runtime ID before it appears in the catalog.

### [minor] Equality branch consumes edge whitespace that may belong to genuinely new content
**File:** Conduit/Services/AppState.swift:2840-2841
**Issue:** When `normalizedBufferedDeltaText == normalizedCoveredText`, `coveredRawCharacters = bufferedDeltaText.count` consumes the entire raw buffered string, including leading/trailing whitespace that sits *outside* the covered span (e.g. covered `"foo bar"`, buffered `" foo bar "`). If that trailing/leading whitespace is the separator before the next (uncovered) token, it is silently eaten. Cosmetic today, but in the straddle case this compounds the ordering impact. The hasPrefix branch handles this with the `min(...)` trailing-whitespace cap; the equality branch does not.
**Suggestion:** Apply the same whitespace-bounded consumption in the equality branch (compute covered span from the raw strings with the leading/trailing-whitespace arithmetic instead of consuming the whole raw buffer), or trim both sides before comparing and consume only the true raw overlap.

### [nit] No test covers the straddle/partial-overlap-with-extension case
**File:** ConduitTests/BufferedEventDeduplicationTests.swift:196-226
**Issue:** The suffix-aligned tests (`testSuffixAlignedCoveredDeltaIsNotReplayed`, `testMidWindowGapDropsSuffixAlignedBufferedDeltas`) only feed buffered text that is fully contained in the covered span. The boundary-crossing continuation (covered `"BCD"`, buffered `"CDE"` → expect `"E"`) is exercised nowhere, so the major finding above is invisible to the suite.
**Suggestion:** Add `coveredText: "BCD"` with buffered `["CDE"]` expecting `["E"]`, plus an integration variant with an interleaved tool event.

### [nit] PR description no longer matches the implementation
**File:** PR description
**Issue:** The body describes "the longest overlap between the inflight's suffix and the buffer's prefix" and "a partial overlap replays only the remainder (as a single delta, in the last delta's position)". The current code uses deterministic boundary-derived exact matching with sequential per-event consumption, and the straddle case (which longest-overlap *would* have handled) is not covered. The mismatch makes the remaining gap harder to spot.
**Suggestion:** Update the description to state coverage is derived deterministically from the captured boundary and applied only under exact prefix/suffix/equality alignment, and explicitly note the boundary-crossing continuation as a known limitation (or fix it).

## Verdict
REQUEST_CHANGES

The core algorithm is sound and fail-safe for contiguous windows: per-event sequential consumption preserves interleaved ordering, mismatched suffixes are validated before consumption, and nothing non-whitespace is dropped unless it is provably within the snapshot. However, the suffix-aligned handling added in the final commit only covers the "buffered ⊆ covered" case; the boundary-straddling continuation (buffered starts inside the covered span and extends into new text) matches no branch and replays the duplicate the PR is meant to remove, and it is the same reconnect/lag scenario the code itself documents. No retain cycles, force unwraps, or concurrency issues were introduced — the new helpers are `nonisolated` pure functions, and the replay loop is linear.

<a href="https://opencode.ai/s/W5LeAO7k">New%20session%20-%202026-08-12T18%3A58%3A44.388Z</a>
[opencode session](https://opencode.ai/s/W5LeAO7k)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31630258042)

kaishi00:
## Summary
This PR deduplicates `.messageDelta` events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection: it captures the stream text/session at the reconciliation boundary, derives the covered span (with persisted-prefix normalization), validates alignment, and sequentially trims the covered prefix off each buffered delta while preserving interleaved non-delta event order. Earlier majors (per-event ordering, suffix validation, whitespace handling, session aliasing, the mid-window gap) are genuinely addressed in HEAD, and the algorithm is otherwise fail-safe against data loss for contiguous, non-repeating streams.

## Findings

### [major] Straddle/suffix branches silently drop genuinely new text on self-repeating content
**File:** Conduit/Services/AppState.swift:2855-2868
**Issue:** The `hasSuffix` branch (2855-2860) and the KMP `longestSuffixPrefixOverlapLength` fallback (2861-2868) consume buffered deltas based purely on the longest content suffix/overlap between the covered span and the buffered concatenation. On self-repeating text this over-consumes and drops new content. Concrete repro: boundary `""`, snapshot `"abcabc"` (covered `"abcabc"`), buffered window `"bcabc"` (a reconnect left the window starting at position 4 of the cumulative stream, which then continued to `"abcabcabc"`). The true overlap is `"bc"` (positions 4-5); the new continuation is `"abc"` and must be replayed. Because `"bcabc"` is itself a suffix of `"abcabc"`, the `hasSuffix` branch fires and consumes all 5 characters, replaying nothing — the seeded bubble stays `"abcabc"` and the real continuation `"abc"` is lost. This is the one data-loss direction the other branches deliberately fail safe on, and it is reachable whenever the stream contains a repeated run and the window straddles the covered span (the same reconnect/lag scenario the straddle branches exist for).
**Suggestion:** Don't consume on a content suffix/overlap alone. Bound the straddle consumption to the alignment that is provable — e.g. prefer the *shortest* plausible overlap, or verify the overlap position against the boundary-derived coverage (the buffered window's start position relative to the covered span is known at buffering time and could be recorded on `Reconciliation`), or skip dedup entirely when more than one content overlap is plausible. Add a regression test for covered `"abcabc"` with buffered `["bcabc"]` expecting `["abc"]`, plus the interleaved-tool variant.

### [minor] `acceptedSessionIDs` intersection can silently disable dedup for a fresh runtime ID
**File:** Conduit/Services/AppState.swift:2315-2326
**Issue:** `acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(knownSessionIDs(for: boundarySessionID))` drops `result.sessionId` (inserted at line 2245) whenever the resolved runtime ID is not yet a catalog-known alias. All downstream guards then return `nil`/skip and the buffered deltas replay unchanged — the visible duplicate this PR removes persists on the first resume that produces a fresh runtime ID, with no diagnostic. The comment documents this as intentional, but the primary reconcile path can hit it; the only existing test (`testResumeDedupAcceptsAlternateSessionIDForBufferedDelta`) pre-registers the alias in the catalog.
**Suggestion:** Union the resolved runtime ID (or the reconcile `sessionIDs` from `applyChatResume`) into the accepted set before intersecting, or emit a log when dedup is skipped for this reason. Add a test where the buffered delta is tagged with a runtime ID that is not yet in the catalog.

### [minor] Coverage is computed in the trimmed domain but consumed against raw buffered text
**File:** Conduit/Services/AppState.swift:2761-2765, 2842-2852
**Issue:** `reconciliationBoundaryCoverageText` computes the covered span from whitespace-trimmed boundary/snapshot strings, while the `hasPrefix` branch maps it back to raw characters using the buffered text's own `leadingWhitespaceCount` and a `min(...)` cap on trailing whitespace. This is consistent only when the boundary/snapshot and buffered window carry matching edge whitespace. A boundary ending in whitespace (e.g. `"The "` → trimmed `"The"`) shifts the covered span by that whitespace, and the seeded bubble (which trims on both sides) plus the replayed remainder can then join words without a separator (e.g. `"bugis"`), or stray whitespace can be replayed.
**Suggestion:** Derive coverage in the same domain the seeding uses (normalize both boundary and snapshot with `unpersistedInflightAssistantText`, as the second branch already does) and document that the overlap is whitespace-sensitive, or drop the raw-domain first branch.

### [nit] The `knownPrefix` fallback is dead in the integration path and unvalidated in the explicit path
**File:** Conduit/Services/AppState.swift:2794-2807
**Issue:** The integration path always passes an explicit `coveredText`, so the fallback (which performs the `hasPrefix(knownPrefix)` sanity check) never runs in production — it is only exercised by unit tests. Conversely, the explicit path never cross-checks that `knownPrefix + coveredText` actually equals the seeded `inflight`, so the suffix-validation safeguard added for the fallback does not protect the real path.
**Suggestion:** Either unify on one source of truth (derive `coveredText` from the seeded buffer in both paths) or add the equivalent consistency check to the explicit path.

### [nit] Integration tests flush the published projection via a `showSidebar` toggle
**File:** ConduitTests/AppStateChatResumeTests.swift:180-181, 246-247, 302-303, 356-357
**Issue:** The `showSidebar = true; showSidebar = false` pair is an undocumented side-effect-driven publish flush; it silently couples each test's correctness to `showSidebar` triggering a publish.
**Suggestion:** Use a deterministic publish hook (or await the streaming publish task) instead of toggling unrelated state.

### [nit] No coverage for the repetition-ambiguity straddle or catalog-lagging runtime ID
**File:** ConduitTests/BufferedEventDeduplicationTests.swift:196-273
**Issue:** The suffix-aligned and straddle tests all use non-repeating content, so none exercise the over-consumption path above; the alternate-session test pre-registers the runtime alias, so the fresh-ID no-op path is also untested.
**Suggestion:** Add the covered `"abcabc"`/buffered `"bcabc"` → `["abc"]` regression and a runtime-ID-not-in-catalog case.

## Verdict
REQUEST_CHANGES

The core mechanism is sound and the earlier ordering/suffix/whitespace/alias defects are resolved, with strong unit and integration coverage. The remaining blocker is the straddle handling: the suffix-aligned and KMP-overlap branches introduced in the last two commits consume on content matching alone and can silently drop genuinely new continuation text on self-repeating streams — the one data-loss case the rest of the algorithm deliberately avoids. No retain cycles, force unwraps, or concurrency issues were introduced; the new helpers are `nonisolated` pure functions and the replay loop is linear.

<a href="https://opencode.ai/s/YVHd603Z">New%20session%20-%202026-08-12T19%3A11%3A16.947Z</a>
[opencode session](https://opencode.ai/s/YVHd603Z)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31631298518)

kaishi00:
## Summary
This PR deduplicates `.messageDelta` events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection: it captures the stream text/session at the reconciliation boundary, derives the covered span, validates alignment, and sequentially trims the covered prefix off each buffered delta while preserving interleaved non-delta event order.

## Findings

### [major] Content-only consumption can drop genuinely new text when the buffered window is not a true continuation of the covered span
**File:** Conduit/Services/AppState.swift:2844-2873
**Issue:** Branches 1-2 (and the KMP fallback) consume buffered text purely on a content prefix/equality match, with the ambiguity guard only applied to the *multi-overlap* KMP case. A *single* coincidental overlap still consumes. This is the residual data-loss direction the other branches deliberately fail safe on, and it is reachable in the PR's own target scenario (mid-window gap / turn boundary / non-cumulative gateway). Repro: boundary `""`, snapshot `"xyz"` (covered `"xyz"`), a new turn (or post-gap delta) streams `"xyz more"`. `normalizedBufferedDeltaText.hasPrefix(normalizedCoveredText)` fires, `coveredRawCharacters = 3`, and `"xyz"` of the genuinely new text is consumed — the seeded bubble shows `"xyz more"` instead of `"xyzxyz more"`. The PR description's stated invariant — "a turn that started after the snapshot ... leaves the events untouched" — only holds when the content overlap is *zero*; a coincidental single overlap violates it.
**Suggestion:** Validate the overlap positionally before consuming. The boundary's exact stream offset is known (the buffered window starts immediately after the boundary text), so reject consumption when the union `(knownPrefix + coveredText + bufferedRemainder)` is not consistent with `(boundaryText + bufferedText)`; failing that, treat a content match with no positional proof like the ambiguous case and leave events intact. Add a regression for covered `"xyz"` with buffered `["xyz more"]` from a distinct turn (assert no consumption).

### [minor] `knownPrefix` sanity check is dead in the production path
**File:** Conduit/Services/AppState.swift:2337-2356, 2796-2813
**Issue:** Production always passes `coveredText` explicitly, so the `knownPrefix` parameter is never consulted — the `inflight.hasPrefix(knownPrefix)` consistency check and the fallback coverage derivation run only in unit tests. The explicit path trusts `reconciliationBoundaryCoverageText`'s content prefix-match with no independent cross-check against the seeded bubble, which is exactly the check that would have caught the major finding above (the unit tests exercise it, the integration path does not).
**Suggestion:** Unify on one source of truth — derive `coveredText` from the seeded buffer in both paths (or add the equivalent `inflight.hasPrefix(knownPrefix)` + seeded-bubble consistency check to the explicit path).

### [minor] `acceptedSessionIDs` intersection silently disables dedup on a fresh runtime ID
**File:** Conduit/Services/AppState.swift:2315-2334
**Issue:** `acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(knownSessionIDs(for: boundarySessionID))`. When `result.sessionId` is a runtime ID not yet registered as a catalog alias, the intersection drops it, both helpers return `nil`, and the already-covered buffered deltas replay in full — the exact duplicate this PR removes — on the first resume that produces a fresh runtime ID. It is logged, but the only alias test (`testResumeDedupAcceptsAlternateSessionIDForBufferedDelta`) pre-registers the alias, so the catalog-lagging path is untested.
**Suggestion:** Union the resolved runtime ID (or the reconcile `sessionIDs` from `applyChatResume`) into the accepted set before intersecting, and add a test where the buffered delta is tagged with a runtime ID that is not yet in the catalog.

### [nit] Integration tests force a publish via a `showSidebar` toggle
**File:** ConduitTests/AppStateChatResumeTests.swift:180-181, 246-247, 302-303, 356-357, 420-421, 464-465
**Issue:** The `showSidebar = true; showSidebar = false` pair is an undocumented side-effect-driven flush of the throttled streaming publish. Each test's assertions silently depend on `showSidebar` triggering a publish; if the publish mechanism changes, the dedup tests break for the wrong reason.
**Suggestion:** Use a deterministic publish hook (or await the streaming publish task) instead of toggling unrelated state.

### [nit] PR description no longer matches the implementation
**File:** PR description
**Issue:** The body still describes "the longest overlap between the inflight's suffix and the buffer's prefix"; the code now does deterministic boundary-derived matching with per-event consumption and an ambiguity fail-safe. Keeping the description accurate matters because the actual invariant is contiguity + fail-safe, not longest overlap.
**Suggestion:** Update the description to state that coverage is derived from the captured boundary and consumed only under validated prefix/suffix/equality alignment, and note the coincidental-overlap case as a known limitation (or fix it per the major finding).

## Verdict
REQUEST_CHANGES

The mechanism is well-engineered: sequential per-event consumption preserves interleaved ordering, whitespace arithmetic holds up under scrutiny, the ambiguity guard protects multi-overlap repeated content, and the replay loop is linear. No retain cycles, force unwraps, or concurrency issues were introduced — the new helpers are `nonisolated` pure functions. The remaining blocker is the coincidental single-overlap path (branch 2 / KMP), which can silently consume genuinely new text in the gap/new-turn/non-cumulative scenarios the PR explicitly targets, contradicting the documented "turn started after the snapshot is left untouched" invariant.

<a href="https://opencode.ai/s/8ct7uX0d">New%20session%20-%202026-08-12T19%3A32%3A33.839Z</a>
[opencode session](https://opencode.ai/s/8ct7uX0d)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31633066494)

kaishi00:
## Summary
This PR deduplicates `.messageDelta` events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection: it captures stream text/session at the reconciliation boundary, derives the covered span (persisted-prefix normalized), validates alignment against the seeded bubble, and sequentially trims the covered prefix off each buffered delta while preserving interleaved non-delta event order.

## Findings

### [major] Content-only consumption can drop genuinely new text when no `.messageStart` is buffered
**File:** Conduit/Services/AppState.swift:2883-2912
**Issue:** Branches 1–3 consume on content equality/prefix matching alone, with no positional proof that the buffered window is the continuation of the boundary stream. The explicit-path cross-check (AppState.swift:2797-2817) only validates `explicitCoveredText` against the *seeded* bubble (both are coverage candidates), never against the buffered text that actually gets consumed. So the documented invariant — "a turn that started after the snapshot ... leaves the events untouched" — holds only for *zero* content overlap. Repro: covered `"xyz"`, buffered `[.messageDelta("xyz more")]` with no `.messageStart` in the window (marker predates the buffer, or a gateway without turn markers / an agent continuation): branch 2 fires, `coveredRawCharacters = 3`, and the genuinely new `"xyz"` is consumed — the seeded bubble renders `"xyz more"` instead of `"xyzxyz more"`. Branch 1 (exact equality, e.g. covered `"foo"` + buffered `"foo"` from a new turn) drops the whole delta. The `.messageStart` split (AppState.swift:2859-2877) is the only positional guard and is not reachable in these cases.
**Suggestion:** Validate alignment positionally before consuming — e.g. record the buffered window's start offset relative to the boundary at buffering time and require the union `(knownPrefix + coveredText + bufferedRemainder)` to be consistent with `(boundaryText + bufferedText)`; otherwise treat a content match with no positional proof like the ambiguous case and leave events intact. Add a unit test for covered `"xyz"` + buffered `["xyz more"]` (no marker) asserting no consumption.

### [minor] Runtime-ID not in the catalog silently disables dedup in the primary foreground path
**File:** Conduit/Services/AppState.swift:2316-2334
**Issue:** `acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(knownSessionIDs(for: boundarySessionID))` drops `result.sessionId` whenever the resolved runtime ID is not yet a catalog-confirmed alias (e.g. a server-created session or a freshly minted runtime alias not persisted into `SessionSummary.alternateIds`). Both `normalizedReconciliationBoundaryPrefix` and `reconciliationBoundaryCoverageText` then return `nil`, `deduplicatingBufferedEvents` returns events unchanged, and the exact duplicate this PR removes replays. It is logged, but the only alias test (AppStateChatResumeTests.swift:231-293) pre-registers the alias, so the catalog-lagging path is untested and a user-visible regression occurs with no diagnostic.
**Suggestion:** Union the resolved runtime ID (already inserted into `boundary.acceptedSessionIDs` at AppState.swift:2245) or the reconcile `sessionIDs` from `applyChatResume` into the accepted set before intersecting; add a test where the buffered delta is tagged with a runtime ID that is not in the catalog.

### [minor] Whitespace-domain mixing silently skips dedup on interior-whitespace divergence
**File:** Conduit/Services/AppState.swift:2766-2785, 2886-2896
**Issue:** `coveredText` is derived from whitespace-trimmed boundary/snapshot strings, then mapped onto raw buffered text via `leadingWhitespaceCount` + a `min(...)` trailing-whitespace cap. This is only consistent when boundary/snapshot and buffered text carry matching edge whitespace. Any interior or repeated-whitespace divergence (e.g. covered `"foo  bar"` vs buffered `"foo bar"`, or a boundary ending in whitespace) fails the `hasPrefix` branches and leaves the duplicate in place — the exact bug this PR fixes — with no signal; conversely the trimmed `coveredEnd` arithmetic can over/under-consume a single space at the seam.
**Suggestion:** Derive coverage in the same domain the seeding uses (normalize boundary and snapshot with `unpersistedInflightAssistantText` on both sides, as the second branch already does) and document that the overlap is whitespace-sensitive, or drop the raw-domain first branch.

### [nit] Integration tests flush the published projection via a `showSidebar` toggle
**File:** ConduitTests/AppStateChatResumeTests.swift:180-181, 246-247, 302-303, 356-357, 420-421, 464-465
**Issue:** The `showSidebar = true; showSidebar = false` pair is an undocumented side-effect-driven flush of the throttled streaming publish; each test's assertions silently depend on `showSidebar` triggering a publish. If the publish mechanism changes, the dedup tests break for the wrong reason.
**Suggestion:** Use a deterministic publish hook (or await the streaming publish task) instead of toggling unrelated state.

### [nit] No test covers the no-marker coincidental-overlap path
**File:** ConduitTests/BufferedEventDeduplicationTests.swift:291-328
**Issue:** `testNewTurnMarkerPreservesCoincidentallyRepeatedText` and `testNewTurnMarkerOnlyPreservesEventsAfterTheMarker` both rely on a buffered `.messageStart`; there is no test where covered `"xyz"` is paired with buffered `["xyz more"]` and no marker, so the major finding above is invisible to the suite (the straddle/ambiguity tests at 228-289 use non-repeating or fully-ambiguous content only).
**Suggestion:** Add `coveredText: "xyz"` with buffered `["xyz more"]` (no marker) asserting no consumption, plus the branch-1 equality variant.

## Verdict
REQUEST_CHANGES

The mechanism is well-engineered: sequential per-event consumption preserves interleaved ordering, the ambiguity guard protects multi-overlap repeated content, the KMP scan is linear, and the new helpers are `nonisolated` pure functions with no retain cycles, force unwraps, or concurrency issues introduced. The remaining blocker is the coincidental single-overlap path (branches 1–3), which can silently consume genuinely new continuation text whenever a `.messageStart` is not present in the buffered window — a data-loss direction the rest of the algorithm deliberately fails safe on, and one that contradicts the PR's documented "turn started after the snapshot is left untouched" invariant.

<a href="https://opencode.ai/s/RVVvX559">New%20session%20-%202026-08-12T19%3A55%3A29.224Z</a>
[opencode session](https://opencode.ai/s/RVVvX559)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31634957639)

kaishi00:
## Summary
This PR deduplicates `.messageDelta` events buffered during foreground reconciliation against the resume snapshot's cumulative `inflight` projection: it captures the stream text/session at the reconciliation boundary, derives the covered span (persisted-prefix normalized), validates alignment against the seeded bubble, and sequentially trims the covered span off buffered deltas while preserving interleaved non-delta event order. It bundles a linear KMP overlap helper and 11 unit + 7 integration tests.

## Findings

### [major] Cross-turn single overlap still consumes genuinely new text; documented invariant is overstated
**File:** Conduit/Services/AppState.swift:2884-2925
**Issue:** Consumption is gated only by `hasBoundaryAnchor` (a non-empty boundary text) and an explicit `.messageStart` marker. A boundary anchor only proves a stream was live at the boundary — it does not prove the buffered window is the *same turn's* continuation. When a new turn starts inside the post-snapshot slice of the reconciliation window and the gateway emits no `.messageStart` (or it predates the buffer), a *single, coincidental* content overlap passes every guard and consumes genuinely new text. Concrete repro: boundary `"The quick brown"`, snapshot `"The quick brown fox"` (covered `"fox"`, seed `"fox"`), buffered `[.messageDelta("fox jumped over")]` from a new turn with no marker → branch 2 (2899) computes `coveredRawCharacters = 3`, drops `"fox"`, replays `" jumped over"` → bubble renders `"fox jumped over"` instead of `"fox"` + `"fox jumped over"`. The PR body claims "a turn that started after the snapshot ... leaves the events untouched," but that invariant only holds for *zero* content overlap; the `hasBoundaryAnchor`/multi-overlap/marker guards are the only protections and all pass here.
**Suggestion:** Validate alignment positionally, not just by content. The cumulative stream length at the boundary is known when the window opens (`streamingBuffer.count` at `beginReconciliation`, AppState.swift:2025) and the snapshot length is known, so the covered span's raw extent is deterministic. Reject consumption unless `boundaryText + bufferedText` is consistent with `knownPrefix + coveredText + remainder` (i.e., the buffered window is a true continuation of the boundary stream). Failing that, treat a single content overlap without positional proof like the ambiguous case and leave events intact. Add a regression with covered `"fox"` / buffered `["fox jumped over"]` / no marker / anchor present asserting no consumption.

### [minor] Fresh runtime session ID that is not yet a catalog alias silently disables dedup
**File:** Conduit/Services/AppState.swift:2316-2334
**Issue:** `acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(knownSessionIDs(for: boundarySessionID))` drops `result.sessionId` whenever the resolved runtime ID is not yet registered as an alias of the catalog session (e.g., a server-created session or a freshly minted runtime alias not yet persisted into `SessionSummary.alternateIds`). Both `normalizedReconciliationBoundaryPrefix` and `reconciliationBoundaryCoverageText` then return `nil` (they require `acceptedSessionIDs.contains(resumedSessionID)`), and `deduplicatingBufferedEvents` returns events unchanged — the exact duplicate this PR removes replays on the first resume that produces a fresh runtime ID. It is logged at debug, but the only alias test (`AppStateChatResumeTests.swift:406`) pre-registers the alias, so the catalog-lagging path is untested and the regression is user-visible with no diagnostic.
**Suggestion:** Union the resolved runtime ID (already admitted via `reconciliation.accepts`) or the `sessionIDs` from `applyChatResume` (AppState.swift:2576) into the accepted set before intersecting, or promote the skip to a more visible diagnostic. Add a test where the buffered delta is tagged with a runtime ID that is not yet in the catalog.

### [minor] Interior-whitespace divergence silently skips dedup (duplicate persists)
**File:** Conduit/Services/AppState.swift:2769-2788, 2899-2909
**Issue:** `reconciliationBoundaryCoverageText` derives the covered span from whitespace-trimmed boundary/snapshot strings, then the `hasPrefix` branch maps it back onto raw buffered text via `leadingWhitespaceCount` plus a `min(...)` trailing-whitespace cap. This is consistent only when edge whitespace lines up. Any interior or repeated-whitespace divergence between the covered span and the buffered window (e.g., covered `"foo  bar"` vs buffered `"foo bar"`, or a boundary ending in whitespace) fails the `hasPrefix`/`hasSuffix`/equality branches and the KMP overlap, so dedup no-ops and the covered text replays on top of the seeded bubble — with no signal. Safe in the data-loss direction but silently defeats the fix.
**Suggestion:** Compute coverage in the same domain the seeding uses (`unpersistedInflightAssistantText` on both boundary and snapshot, as the second branch at 2775-2788 already does) and document that the overlap is whitespace-sensitive, or drop the raw-domain first branch.

### [nit] Integration tests flush the published projection via a `showSidebar` toggle
**File:** ConduitTests/AppStateChatResumeTests.swift:180-181, 221-222, 287-288, 343-344, 397-398, 461-462, 506-507
**Issue:** The `showSidebar = true; showSidebar = false` pair is an undocumented side-effect-driven flush of the throttled streaming publish. Each test's assertions silently depend on `showSidebar` invalidating the publish guard (`scheduleStreamingPublish` at AppState.swift:6871); if the publish mechanism changes, the dedup tests break for the wrong reason.
**Suggestion:** Use a deterministic publish hook (or await the streaming publish task) instead of toggling unrelated state.

### [nit] PR description no longer matches the implementation
**File:** PR description
**Issue:** The body describes "the longest overlap between the inflight's suffix and the buffer's prefix" and "a partial overlap replays only the remainder (as a single delta, in the last delta's position)." The current code does deterministic boundary-derived matching with sequential per-event consumption, and longest-overlap only survives as the KMP ambiguity-fail-safe. The mismatch hides the real invariant (contiguity + fail-safe) and obscures the residual cross-turn gap above.
**Suggestion:** Update the description to state coverage is derived deterministically from the captured boundary and consumed only under validated prefix/suffix/equality alignment, and explicitly note the cross-turn-without-marker case as a known limitation (or fix it per the major finding).

No retain cycles, force unwraps, or concurrency issues were introduced: the new helpers are `nonisolated` pure functions on the `@MainActor` class, `deltaSessionIDs.first`/`overlaps.first` are guarded by `count == 1`, the KMP scan is linear (the earlier O(n²) concern is resolved), and the replay loop is linear.

## Verdict
COMMENT

The mechanism is well-engineered: sequential per-event consumption preserves interleaved ordering, the whitespace arithmetic holds under the tests, the multi-overlap ambiguity guard protects repeated content, and it fails safe against data loss in every ambiguous case except one. The residual blocker — a new turn arriving in the post-snapshot window without a `.messageStart` whose text coincides with the covered span — is narrow but is the one data-loss direction the rest of the algorithm deliberately avoids, and it contradicts the PR's own documented invariant. That warrants a positional guard (the boundary stream length is known at buffering time) plus the missing regression test, and the description should be updated to match the deterministic matching actually implemented.

<a href="https://opencode.ai/s/4fl3ASh5">New%20session%20-%202026-08-12T20%3A13%3A46.906Z</a>
[opencode session](https://opencode.ai/s/4fl3ASh5)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31636487335)

PR #44, opened 2026-08-11
Cache markdown parsing, syntax highlighting, and attachment previews

angel12:
## Problem
Every visible `MarkdownText` body re-evaluates ~30×/s while a reply streams (each delta publishes an AppState change), and each evaluation re-parsed its source and rebuilt the attributed string from scratch — O(visible transcript) main-thread work per frame, growing with message length. Measured on the simulator: a single streaming code block pegged the app at ~93% CPU, with scroll hitching and delayed taps on long transcripts.

## Fix
Three `NSCache` layers, all evicted under memory pressure:

- **`MarkdownRenderCache`** — parsed blocks + prebuilt selectable `NSAttributedString`, keyed by (source, style, gateway-media flag, Dynamic Type size). Settled messages render from cache; only genuinely new content pays for parsing. Streaming-tail fade opacities are applied to a copy of the cached base each frame instead of forcing a full rebuild.
- **`SyntaxHighlighter`** — tokenized code memoized by (language, source). Settled code blocks across the transcript stop re-tokenizing at frame rate; only the still-growing streaming block misses.
- **`UserImageAttachmentPreview`** — decoded local attachment previews cached by path instead of re-reading the file from disk on every render.

## Measurements (iPhone 17 Pro simulator, debug build)
- Worst case — fresh conversation, giant single-line code block streaming: **93% → ~75% CPU**. The residual is the still-growing block itself, whose source changes every reveal tick; making that incremental is future work.
- The O(transcript)-per-frame term is eliminated: settled messages behind the stream render from cache, and the app idles at 2–3% once the turn completes.

## Testing
Full suite passes (427 tests, including the existing `ChatTextSelectionTests` that exercise the parser/formatter directly). Streaming, fades, code blocks, and attachments visually verified in the simulator against a live Hermes instance.

Full write-up: [angel12/hermes-conduit#2](https://github.com/angel12/hermes-conduit/issues/2)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling so pending requests are resolved promptly when a connection closes.
  * Prevented potential memory retention issues in dashboard messaging.

  * Faster reuse of local image attachments.
  * Improved Markdown rendering performance, including streamed content and syntax highlighting.
  * Reduced unnecessary processing during repeated content updates.

### DIFF
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -128,6 +128,23 @@ enum DashboardTicketBridgeError: LocalizedError {
     }
 }
 
+/// WKUserContentController retains its script message handlers strongly, so
+/// registering the bridge directly forms a cycle (bridge → webView →
+/// configuration → userContentController → bridge) that `deinit` can never
+/// break — every replaced bridge would keep a WebKit content process alive
+/// for the app's lifetime. The proxy holds the bridge weakly instead.
+private final class WeakScriptMessageHandler: NSObject, WKScriptMessageHandler {
+    private weak var delegate: WKScriptMessageHandler?
+
+    init(_ delegate: WKScriptMessageHandler) {
+        self.delegate = delegate
+    }
+
+    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
+        delegate?.userContentController(userContentController, didReceive: message)
+    }
+}
+
 @MainActor
 final class DashboardTicketBridge: NSObject {
     let baseURL: String
@@ -151,7 +168,7 @@ final class DashboardTicketBridge: NSObject {
         }
         self.webView = WKWebView(frame: .zero, configuration: configuration)
         super.init()
-        configuration.userContentController.add(self, name: "dashboard-response")
+        configuration.userContentController.add(WeakScriptMessageHandler(self), name: "dashboard-response")
         webView.navigationDelegate = self
         Task { [weak self] in
             guard let self else { return }
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -453,6 +453,10 @@ final class HermesClient: ObservableObject {
                 // Socket closed or errored
                 logger.error("WebSocket receive failed: \(error.localizedDescription, privacy: .public)")
                 isConnected = false
+                // No response can ever arrive on a dead socket; failing the
+                // in-flight requests here keeps awaiting UI from hanging for
+                // the remainder of each request's timeout.
+                failAllPendingRequests(with: HermesError.connectionClosed)
                 if !closedIntentionally {
                     onDisconnected?()
                 }
@@ -509,10 +513,13 @@ final class HermesClient: ObservableObject {
         session?.invalidateAndCancel()
         socketDelegate = nil
         isConnected = false
-        // Fail all pending requests
-        for (_, pending) in pending {
-            pending.timer?.invalidate()
-            pending.continuation.resume(throwing: HermesError.connectionClosed)
+        failAllPendingRequests(with: HermesError.connectionClosed)
+    }
+
+    private func failAllPendingRequests(with error: Error) {
+        for (_, request) in pending {
+            request.timer?.invalidate()
+            request.continuation.resume(throwing: error)
         }
         pending.removeAll()
     }
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -777,9 +777,22 @@ private struct UserImageAttachmentPreview: View {
     @State private var gatewayImage: UIImage?
     @State private var gatewayLoadFailed = false
 
+    /// Body re-evaluates at streaming frame rate, and re-reading the file
+    /// from disk each time hitches scrolling. Decoded previews are shared
+    /// across rows; NSCache evicts them under memory pressure.
+    private static let localImageCache: NSCache<NSString, UIImage> = {
+        let cache = NSCache<NSString, UIImage>()
+        cache.countLimit = 64
+        return cache
+    }()
+
     private var localImage: UIImage? {
         guard let url = localFileURL else { return nil }
-        return UIImage(contentsOfFile: url.path)
+        let key = url.path as NSString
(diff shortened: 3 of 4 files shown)


kaishi00:

[github run](/kaishi00/hermes-conduit/actions/runs/31533031286)

kaishi00:
## Summary
This PR adds three `NSCache` layers (Markdown parsing/render, syntax highlighting, and local attachment previews) to eliminate the per-frame O(visible transcript) re-parse/rebuild work while a reply streams, plus a WebKit weak message-handler proxy and centralized pending-RPC failure handling.

## Findings

### [severity: major] Dead WebSocket is not cleared on receive failure and the failure transition is not socket-identity-guarded
**File:** Conduit/Services/HermesClient.swift:455-459
**Issue:** When `socket.receive()` throws, the loop sets `isConnected = false` and fails pending requests, but never clears/cancels `self.socket`. `rpc`'s guard only checks `socket.closeCode == .invalid` (line 530), which is typically `.invalid` after a receive error (it means "no close frame received", not "healthy"), so a subsequent RPC can be issued against a stopped socket. More importantly, the catch block runs unconditionally on the MainActor: if a reconnect (`connect()`, line 360) already replaced `self.socket` and spawned a new `receiveLoop` (the old `receiveTask` is overwritten without cancellation), a late-resuming old loop will set `isConnected = false` on the healthy new connection and fail *its* in-flight pending RPCs via `failAllPendingRequests`. This PR makes that stale-loop clobbering more damaging than before by adding the pending-failure call.
**Suggestion:** In the catch block, guard the whole transition with the same identity check used elsewhere (e.g., capture the loop's socket at the top and only proceed if `self.socket === socket`), set `self.socket = nil` (and optionally `socket.cancel()`), and have `rpc`'s guard also require `isConnected`.

### [severity: minor] Streaming transient snapshots pollute both render caches
**File:** Conduit/Views/Components/MarkdownText.swift:113-125 and 1369-1374
**Issue:** Each streaming delta produces a new `source`, so every frame inserts a new entry into `MarkdownRenderCache.cache` (a full parsed-blocks + `NSAttributedString` snapshot) and `SyntaxHighlighter.cache` (a tokenized snapshot). These intermediate snapshots are never reused once the stream advances — they only churn the 256/128-entry budgets and force eviction of entries that were actually reusable. NSCache's LRU behavior keeps settled visible messages hot (they re-hit every frame), so the damage is bounded, but at peak a single growing message can dominate both caches with dead entries and each insertion copies the growing source string into the key.
**Suggestion:** Skip `setObject` for in-progress streaming content, or use a single replaceable key for the active stream and promote only the final settled snapshot into the content-addressed cache.

### [severity: minor] Local attachment image cache holds full-resolution decoded bitmaps
**File:** Conduit/Views/ChatView.swift:783-795
**Issue:** `localImageCache` uses `countLimit = 64` with no `totalCostLimit`, and caches `UIImage(contentsOfFile:)` decoded at full resolution even though the preview is rendered at ~224pt (line 857). Photos from a user's library (10–20 MP) decode to tens of MB each; 64 such entries can reach hundreds of MB before memory pressure evicts, causing warnings/jetsam risk. `localImage` is also a computed property evaluated in `body`, so repeated misses still touch disk until the image is cached.
**Suggestion:** Set `totalCostLimit` and pass an estimated pixel-memory cost via `setObject(_:forKey:cost:)`; downsample to the ~224pt preview size before caching.

### [severity: minor] `String(describing: foregroundStyle)` is not a stable cache-key identity
**File:** Conduit/Views/Components/MarkdownText.swift:105-111
**Issue:** The render-cache key stringifies a SwiftUI `Color`. `Color` wraps an opaque provider whose `description` is not documented as a stable identifier — for dynamic/adaptive colors (e.g., `conduitAdaptiveAccent`, Theme.swift:31) it can include provider identity/pointer info. If the description varies between calls for the same logical color, every body evaluation misses and the cache is useless (perf regression); if two distinct colors stringify alike, a stale-color rendering could be served.
**Suggestion:** Key on an explicit discriminant (e.g., the `usesAccentSurface` flag plus a small enum/style token the caller supplies) rather than on `String(describing:)` of the `Color` value.

### [severity: nit] Weak handler forwards WKScriptMessage on a non-main thread
**File:** Conduit/Services/DashboardTicketBridge.swift:143-145
**Issue:** WebKit invokes `userContentController(_:didReceive:)` on a background thread; the proxy forwards directly to the `@MainActor` bridge's handler (line 414), which then touches `pendingRequests`/`baseURL` off the main actor. This is pre-existing behavior (WebKit called the bridge directly before), so the proxy doesn't regress it — but it's the natural place to hop to the MainActor.
**Suggestion:** Wrap the forward in `Task { @MainActor in ... }` to close the data-race on the bridge's isolated state.

## Verdict
REQUEST_CHANGES

The caching design is sound and the regex-disjointness fix for `SyntaxHighlighter` correctly eliminates the exponential backtracking. The blocking issue is the HermesClient socket-failure path: leaving a dead socket installed (which passes the `closeCode == .invalid` guard) and letting an unguarded stale receive loop fail a replacement connection's pending RPCs. The transient-stream-snapshot caching and unbounded-decoded-image cache should also be addressed before merge.

<a href="https://opencode.ai/s/4YspWfnz">New%20session%20-%202026-08-12T01%3A58%3A15.196Z</a>
[opencode session](https://opencode.ai/s/4YspWfnz)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31555293068)

kaishi00:
Thanks @angel12 — the caching work here is a real win and we want to land it. Quick note on what's happening with this PR:

This branch bundled the three fixes that already landed on `main` (regex backtracking, RPC-fail-on-socket-death, WebKit weak-handler proxy) alongside the genuinely new caching commit, which left it conflicting with `main`. Rather than block on a rebase, we cherry-picked **only** the new caching commit onto current `main` (your authorship preserved) and folded in the review findings (stable cache key, skip transient streaming snapshots, cost-bounded + downsampled image cache) in **#50**.

The `HermesClient` socket-lifecycle note from the reviews is real but lives in already-merged code, so it'll be handled in a separate PR.

Closing this in favor of #50 — appreciate the contribution!

PR #45, opened 2026-08-11
Guard the image fallback link against unparseable URLs

angel12:
## Scope

This PR preserves the original image-fallback crash fix and adds the dashboard bridge teardown fix that was identified during review.

## Changes

- Avoid force-unwrapping model-authored image URLs. The fallback tries the raw URL, then a percent-encoded variant, and renders a non-link “Image unavailable” label when neither is valid.
- Reject in-flight dashboard WebKit requests before `AppState` disconnects, signs out, or replaces the bridge, so pending continuations do not remain suspended.
- Keep the current `main` implementations from PRs #43, #50, and #51 as the source of truth; this branch no longer duplicates their older bundled changes.

## Validation

- Full iOS simulator suite: 480 tests, 0 failures.
- Added regressions for bridge pending-request invalidation and invalid-image fallback copy.

Original issue: [angel12/hermes-conduit#3](https://github.com/angel12/hermes-conduit/issues/3)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling when dashboard connections are disconnected, replaced, or require sign-in.
  * Pending dashboard requests now complete safely instead of remaining unresolved.
  * Prevented outdated dashboard updates after a connection is invalidated.
  * Fixed image fallback rendering for invalid or unavailable links, avoiding crashes and showing appropriate text.
  * Improved handling of encoded and malformed image URLs, including safer HTTP(S) link validation and accessible fallback descriptions.

* **Tests**
  * Added coverage for dashboard request invalidation and image fallback behavior.

### DIFF
--- a/.github/workflows/opencode-review.yml
+++ b/.github/workflows/opencode-review.yml
@@ -34,7 +34,7 @@ jobs:
           fi
           echo "$HOME/.opencode/bin" >> "$GITHUB_PATH"
       - name: Run opencode review
-        run: GITHUB_EVENT_NAME=pull_request opencode github run
+        run: GITHUB_EVENT_NAME=pull_request GITHUB_ACTOR=${{ github.repository_owner }} opencode github run
         env:
           OPENCODE_GO_API_KEY: ${{ secrets.OPENCODE_GO_API_KEY }}
           GITHUB_TOKEN: ${{ secrets.OPENCODE_GH_PAT }}
--- a/Conduit/ConduitApp.swift
+++ b/Conduit/ConduitApp.swift
@@ -75,7 +75,7 @@ struct ConduitApp: App {
                     if await appState.openNotificationTarget(target) {
                         notifications.clearPendingTarget(target)
                     } else {
-                        notifications.retryPendingTarget()
+                        notifications.handleFailedNotificationRoute(target)
                     }
                 }
                 .task(id: voiceIntentRouteKey) {
--- a/Conduit/Models/Models.swift
+++ b/Conduit/Models/Models.swift
@@ -274,6 +274,9 @@ struct DelegateAgentActivity: Identifiable, Equatable {
 
 struct SessionSummary: Identifiable, Equatable {
     let id: String
+    /// The durable database identity used by `session.resume`. Hermes stream
+    /// events and notifications may carry the live runtime identity instead.
+    var storedSessionId: String? = nil
     var alternateIds: [String]
     var title: String
     var model: String
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -407,6 +407,8 @@ final class AppState: ObservableObject {
         var resolvedSessionId: String?
         var acceptedSessionIDs: Set<String>
         let acceptsAnySession: Bool
+        let streamTextAtBoundary: String?
+        let streamSessionIDAtBoundary: String?
         var bufferedEvents: [StreamEvent] = []
 
         init(
@@ -415,6 +417,8 @@ final class AppState: ObservableObject {
             automaticSyncOperationID: UUID? = nil,
             acceptedSessionIDs: Set<String> = [],
             acceptsAnySession: Bool = false,
+            streamTextAtBoundary: String? = nil,
+            streamSessionIDAtBoundary: String? = nil,
             bufferedEvents: [StreamEvent] = []
         ) {
             self.token = token
@@ -422,6 +426,8 @@ final class AppState: ObservableObject {
             self.automaticSyncOperationID = automaticSyncOperationID
             self.acceptedSessionIDs = acceptedSessionIDs
             self.acceptsAnySession = acceptsAnySession
+            self.streamTextAtBoundary = streamTextAtBoundary
+            self.streamSessionIDAtBoundary = streamSessionIDAtBoundary
             self.bufferedEvents = bufferedEvents
         }
 
@@ -2010,11 +2016,22 @@ final class AppState: ObservableObject {
     func beginReconciliation() -> UUID {
         let token = UUID()
         let bufferedEvents = reconciliation?.bufferedEvents ?? []
+        let streamTextAtBoundary: String?
+        let streamSessionIDAtBoundary: String?
+        if let existingReconciliation = reconciliation {
+            streamTextAtBoundary = existingReconciliation.streamTextAtBoundary
+            streamSessionIDAtBoundary = existingReconciliation.streamSessionIDAtBoundary
+        } else {
+            streamTextAtBoundary = activeSessionId.map { _ in streamingBuffer }
+            streamSessionIDAtBoundary = activeSessionId
+        }
         reconciliationToken = token
         reconciliation = Reconciliation(
             token: token,
             requestedSessionId: activeSessionId ?? "",
             acceptsAnySession: true,
+            streamTextAtBoundary: streamTextAtBoundary,
+            streamSessionIDAtBoundary: streamSessionIDAtBoundary,
             bufferedEvents: bufferedEvents
         )
         refreshActiveChatScrollSessionIdentity(isReconciling: true)
@@ -2176,14 +2193,15 @@ final class AppState: ObservableObject {
         ), chatViewportTransitionIsCurrent(requiredViewportTransitionGeneration) else {
             return false
         }
-        let bufferedEvents = reconciliation?.token == token
-            ? reconciliation?.bufferedEvents ?? []
-            : []
+        let priorReconciliation = reconciliation?.token == token ? reconciliation : nil
+        let bufferedEvents = priorReconciliation?.bufferedEvents ?? []
         reconciliation = Reconciliation(
             token: token,
             requestedSessionId: sessionId,
             automaticSyncOperationID: automaticSyncOperationID,
             acceptedSessionIDs: acceptedSessionIDs.union([sessionId]),
+            streamTextAtBoundary: priorReconciliation?.streamTextAtBoundary,
+            streamSessionIDAtBoundary: priorReconciliation?.streamSessionIDAtBoundary,
             bufferedEvents: bufferedEvents
         )
         refreshActiveChatScrollSessionIdentity(isReconciling: true)
@@ -2292,13 +2310,65 @@ final class AppState: ObservableObject {
                 }
                 : []
             if result.snapshot.hasLiveProjection {
+                let resumedInflightText = streamingBuffer
+                let boundary = reconciliation?.token == token ? reconciliation : nil
+                let acceptedSessionIDs: Set<String>
+                if let boundary,
+                   let boundarySessionID = boundary.streamSessionIDAtBoundary {
+                    // Do not infer that a newly returned runtime ID belongs
+                    // to this boundary solely because the resume RPC returned
+                    // it. Keep only IDs already accepted for the boundary and
+                    // known as aliases of that catalog session; an empty
+                    // result intentionally disables deduplication rather than
+                    // risking text from a different session.
+                    acceptedSessionIDs = boundary.acceptedSessionIDs.intersection(
+                        knownSessionIDs(for: boundarySessionID)
+                    )
+                    if !acceptedSessionIDs.contains(result.sessionId) {
+                        sessionCatalogLog.debug(
+                            "Skipping buffered delta dedup because resumed session \(result.sessionId, privacy: .public) is not a catalog-confirmed alias of the reconciliation boundary"
+                        )
+                    }
+                } else {
+                    acceptedSessionIDs = [result.sessionId]
+                }
+                let knownPrefix: String?
+                let coveredText: String?
+                if let boundary {
+                    knownPrefix = Self.normalizedReconciliationBoundaryPrefix(
+                        boundaryText: boundary.streamTextAtBoundary,
+                        boundarySessionID: boundary.streamSessionIDAtBoundary,
+                        resumedSessionID: result.sessionId,
+                        acceptedSessionIDs: acceptedSessionIDs,
+                        after: messages
+                    )
+                    coveredText = Self.reconciliationBoundaryCoverageText(
+                        boundaryText: boundary.streamTextAtBoundary,
+                        boundarySessionID: boundary.streamSessionIDAtBoundary,
+                        resumedSessionID: result.sessionId,
+                        acceptedSessionIDs: acceptedSessionIDs,
+                        snapshotInflightText: result.snapshot.inflightAssistantText,
+                        after: messages
+                    )
+                } else {
+                    knownPrefix = nil
+                    coveredText = nil
+                }
+
                 // The live bubble was just seeded from the cumulative inflight
                 // projection, which already includes deltas emitted while this
-                // reconciliation was in flight. Replaying them verbatim repeats
-                // that text in the bubble.
+                // reconciliation was in flight. Replay only the portion beyond
+                // the normalized text captured at the same session boundary.
                 bufferedEvents = Self.deduplicatingBufferedEvents(
                     bufferedEvents,
-                    againstInflight: result.snapshot.inflightAssistantText
+                    againstInflight: resumedInflightText,
+                    knownPrefix: knownPrefix,
+                    sessionID: result.sessionId,
+                    acceptedSessionIDs: acceptedSessionIDs,
+                    coveredText: coveredText,
+                    hasBoundaryAnchor: boundary?.streamTextAtBoundary?
+                        .trimmingCharacters(in: .whitespacesAndNewlines)
+                        .isEmpty == false
                 )
             }
             bufferedEvents.forEach(applyStreamEvent)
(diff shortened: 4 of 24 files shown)


kaishi00:

[github run](/kaishi00/hermes-conduit/actions/runs/31538425797)

kaishi00:
## Summary
This PR fixes a force-unwrap crash on the WebFallbackImage link, plus bundles 8 commits of related hardening: deduplicating buffered stream deltas against the resume inflight snapshot, failing pending WebSocket RPCs on socket death, breaking the WKScriptMessageHandler retain cycle with a weak proxy, caching markdown/highlighting/image rendering, and a regex backtracking fix.

## Findings

### [severity: major] Dedupe reorders assistant text across interleaved non-delta events
**File:** Conduit/Services/AppState.swift:2673
**Issue:** `deduplicatingBufferedEvents` joins all delta text, computes one `covered` offset, then emits the entire uncovered remainder as a single delta in place of the *last* delta event. If a non-delta event (e.g. `.toolStart`) sits between partially-covered deltas, assistant text that was emitted *before* the tool event is moved *after* it, and its original `sessionId` is replaced by the last delta's. Concretely, `delta("abc"), toolStart, delta("def")` with inflight suffix `"a"` produces `[toolStart, delta("def", "bcdef")]` instead of `[delta("bc"), toolStart, delta("def")]` — reordering the transcript across the tool boundary. The added tests only cover adjacent deltas, so this path is unverified.
**Suggestion:** Consume `covered` characters across each delta in order, emitting each non-empty remainder at its original event position with its original `sessionId` (e.g. the `compactMap` approach CodeRabbit suggested). Add the partial-overlap regression test (`delta("abc"), toolStart, delta("def")` vs inflight `"a"`, asserting `delta("bc")` stays before `toolStart`).

### [severity: major] Bridge pending requests are never resumed when the bridge is torn down
**File:** Conduit/Services/DashboardTicketBridge.swift:136 (also AppState.swift:1658, 1757, 1771)
**Issue:** `disconnect()`, `requireSignIn`, and `prepareDashboardBridge(for:)` nil out or replace `dashboardTicketBridge` while `pendingRequests` may hold suspended `CheckedContinuation`s. The bridge then deallocates, `deinit` removes the `"dashboard-response"` handler, and the in-flight continuations are lost in the deallocated dictionary — the awaiting tasks (e.g. `requestJSON`, `mintTicket`) hang forever, even on cancellation, because `cancelPendingRequest` also goes through the now-dead bridge. This PR's weak-proxy change makes this a *regression*: previously the strong cycle kept the bridge/WebKit alive, so the JS-side 12s `AbortController` timeout could still `postMessage` a failure and resume the continuation; now the handler is gone and the JS error goes unhandled.
**Suggestion:** Add a public `invalidate()`/`close()` on `DashboardTicketBridge` that resumes every `pendingRequests` continuation with an error (e.g. `DashboardTicketBridgeError.notReady`) and clears the dictionary, and call it from `disconnect()`, `requireSignIn`, and `prepareDashboardBridge(for:)` before clearing/replacing the property.

### [severity: minor] Image cache bounded by entry count, not byte cost
**File:** Conduit/Views/ChatView.swift:783
**Issue:** `localImageCache` uses `countLimit = 64`, so it can hold 64 fully-decoded attachment images with no memory bound (a 12MP image is ~48MB decoded). NSCache only enforces the count, not the retained memory.
**Suggestion:** Use `totalCostLimit` (e.g. via `cache.totalCostLimit = ...`) and call `setObject(_:forKey:cost:)` with an estimated decoded-byte cost (`width * height * 4` or `image.jpegData`-derived estimate).

### [severity: minor] Markdown/highlight caches have no byte-cost bound and churn during streaming
**File:** Conduit/Views/Components/MarkdownText.swift:91 and :1380
**Issue:** `MarkdownRenderCache` (`countLimit = 256`) and `SyntaxHighlighter.cache` (`countLimit = 128`) store full rendered/attributed outputs keyed by the entire source. Each streaming frame inserts a new prefix entry, so a large streaming code block or message can retain many near-duplicate full copies (key + value) until NSCache evicts under pressure; `countLimit` alone doesn't bound the bytes.
**Suggestion:** Set `totalCostLimit` and pass an estimated byte cost (`selectableText.length` / `highlighted.characters.count`); consider skipping the cache insert while the source is still the transient streaming prefix.

### [severity: minor] "Open image" shown with no destination when URL parsing fails and `alt` is empty
**File:** Conduit/Views/Components/MarkdownText.swift:862
**Issue:** In the non-link branch, when both raw and percent-encoded URLs fail to parse and `alt` is empty, the label reads "Open image" but there is no link — an actionable label with no action.
**Suggestion:** In the `else` branch, use "Image unavailable" when `alt.isEmpty`, keeping the linked variant text unchanged.

### [severity: minor] `overlapLength` is quadratic in the worst case
**File:** Conduit/Services/AppState.swift:2698
**Issue:** `overlapLength` tries `k = min(...)` down to 1, and each candidate compares a `suffix(k)`/`prefix(k)` slice — O(k) per step, O(n²) worst case for large inputs with a long near-matching suffix/prefix. The inflight projection can be a long cumulative transcript; a pathological tail could stall the main thread (this runs in the reconciliation path).
**Suggestion:** Compute the longest suffix/prefix overlap in linear time using the prefix function (KMP/Z-algorithm) on `candidate + "\u{0}" + text`.

### [severity: nit] Percent-encoding fallback may double-encode already-encoded URLs
**File:** Conduit/Views/Components/MarkdownText.swift:868
**Issue:** `addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)` encodes `%` to `%25`, so a raw URL that mixes unencoded non-ASCII with pre-encoded `%XX` sequences gets its encoded parts corrupted when it falls through. `.urlQueryAllowed` is also the wrong charset for path characters.
**Suggestion:** Try `.urlPathAllowed`-based encoding, or reject already-encoded segments; alternatively try decoding then re-encoding. Low priority — the primary crash fix is sound.

### [severity: nit] Cache key separator collision
**File:** Conduit/Views/Components/MarkdownText.swift:105 and :1387
**Issue:** Keys are built with `"|"`-joined components including the raw `source`/`language`, so a source containing `|` could theoretically collide with a different source across a component boundary. Only affects styling cache correctness, never crashes.
**Suggestion:** Use `NSString` keys derived from a delimiter-free encoding (e.g. length-prefixed components) if you want to be strict; otherwise harmless.

## Verdict
REQUEST_CHANGES

The core crash fix (no force-unwrap, dual raw/encoded fallback) and the regex-backtracking fix are correct and well-motivated. However, the delta-deduplication reordering bug and the pending-request hang on bridge teardown (aggravated by the weak-proxy change) are real correctness/lifecycle defects that should be fixed before merge.

<a href="https://opencode.ai/s/hf6iGDnf">New%20session%20-%202026-08-12T01%3A58%3A19.883Z</a>
[opencode session](https://opencode.ai/s/hf6iGDnf)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31555296455)

kaishi00:
## Summary
This PR fixes a force-unwrap crash (`URL(string:)!`) on the WebKit image-fallback link by safely falling back to a percent-encoded URL or a non-linked "Image unavailable" label, and hardens `DashboardTicketBridge` teardown so in-flight `CheckedContinuation`s are resumed with an error (`invalidate()`) before `AppState` disconnects, signs out, or replaces the bridge.

## Findings

### [severity: minor] Percent-encoding fallback double-encodes existing `%` sequences
**File:** Conduit/Views/Components/MarkdownText.swift:905
**Issue:** `url.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)` does not include `%` in the allowed set (verified: `urlQueryAllowed.contains("%") == false`), so a raw URL that both fails `URL(string:)` and contains pre-encoded segments (e.g. `…/100%20photos\n.jpg`) is rewritten with `%20` → `%2520`, producing a destination that points at the wrong resource. `.urlQueryAllowed` is also the wrong charset for path characters (it happens to include `/`, but encodes `#` as `%23` and strips fragment semantics). Since `URL(string:)` is lenient and auto-encodes most inputs, this path is rarely hit, but when it is, the "recovered" link can be corrupted.
**Suggestion:** Use `.urlPathAllowed` for the whole string (preserves `/` and encodes `%`), or detect already-encoded `%XX` segments and skip the encoded variant (falling through to the non-link label). The primary crash fix is sound regardless.

### [severity: minor] No regression test covers the actual crash path (URL parsing fallback)
**File:** ConduitTests/MarkdownFallbackTests.swift:6
**Issue:** The crash was the force-unwrap in `fallbackLabel` when `URL(string: url)` returns nil. The added tests only assert `WebFallbackImageLabel.title(alt:destinationAvailable:)` copy with the flags supplied manually — they never exercise `URL(string: url) ?? URL(string: encoded) ?? nil` that was the crash site, so a future refactor that reintroduces a force-unwrap or a nil-coalescing regression would still pass CI. The PR description claims a "regression for invalid-image fallback copy," which is only the label, not the link resolution.
**Suggestion:** Add a test that runs the same resolution expression against an unparseable URL (e.g. a string with a control character / invalid percent sequence) and asserts it yields `nil` (or factor the resolution into `WebFallbackImageLabel` and test it directly).

### [severity: nit] `DashboardTicketBridgePendingRequests` is a nonisolated mutable class
**File:** Conduit/Services/DashboardTicketBridge.swift:155
**Issue:** The new container holds `CheckedContinuation` values and mutable `storage` but is not `@MainActor`-annotated, even though every access goes through the `@MainActor` bridge or the `@MainActor` test. This compiles cleanly today (SWIFT_VERSION 5.9), but under Swift 6 strict concurrency it would be flagged. Note that `deinit` intentionally calls `rejectAll` nonisolated (so the class must stay nonisolated or the deinit guard must change), so a naive `@MainActor` annotation on the class would break `deinit` — worth handling deliberately if/when the project adopts Swift 6.
**Suggestion:** Document the isolation contract in a comment (as the surrounding code does), and revisit with a `nonisolated(unchecked)`-free design when migrating to strict concurrency.

### [severity: nit] Busy-wait in the new bridge test
**File:** ConduitTests/DashboardTicketBridgeTests.swift:17
**Issue:** `while requests.count == 0 { await Task.yield() }` is a polling loop; it works on the MainActor because the injected task is also `@MainActor`, but it is a minor fragility (no bound, depends on scheduling).
**Suggestion:** Acceptable as-is; a `withCheckedThrowingContinuation`-style handoff or a `DispatchSemaphore`/expectation would be more deterministic, but this is test-only polish.

### [severity: nit] Dropped explanatory comment on the `deinit` invariant
**File:** Conduit/Services/DashboardTicketBridge.swift:207
**Issue:** The old `deinit` carried a detailed comment explaining why the reject loop is effectively a no-op (a suspended request's task frame retains `self`, so `deinit` cannot run while `pendingRequests` is non-empty). That comment was removed. `deinit`'s `rejectAll` is now belt-and-suspenders, but the invariant is subtle — the fix's correctness actually depends on `invalidate()` (resuming continuations to break the bridge→pendingRequests→task-frame→bridge cycle), and a future reader may assume `deinit` alone protects callers.
**Suggestion:** Restore a short comment noting the retain-cycle invariant and that `invalidate()` is the primary resumption path.

## Verdict
APPROVE

The core changes are correct: the force-unwrap crash is eliminated with a safe fallback chain, teardown now resumes all pending continuations before the bridge is replaced or dropped (no lost-resume hang, no double-resume risk since `rejectAll` copies the dictionary and later `removeValue(for:)`/cancel lookups become no-ops), and the label copy fix ("Image unavailable" instead of an actionable "Open image" with no destination) matches the other fallback paths. Remaining findings are minor and non-blocking; the dedupe/cache findings from earlier review rounds concern code already merged to `main` and outside this PR's diff.

<a href="https://opencode.ai/s/ch3nFUnN">New%20session%20-%202026-08-12T22%3A42%3A13.360Z</a>
[opencode session](https://opencode.ai/s/ch3nFUnN)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31647968394)

kaishi00:
## Summary
This PR removes a force-unwrap crash on model-authored image fallback URLs (strict parse → component-aware repair → non-linked "Image unavailable" label) and adds a dashboard bridge `invalidate()` that rejects in-flight `CheckedContinuation`s before `AppState` disconnects/signs out/replaces the bridge, plus tests for both.

## Findings

### [severity: minor] Bridge invalidation test doesn't exercise `invalidate()` or the AppState wiring
**File:** ConduitTests/DashboardTicketBridgeTests.swift:17
**Issue:** The test constructs a bare `DashboardTicketBridgePendingRequests` and calls `rejectAll()` directly. The actual fix lives in `DashboardTicketBridge.invalidate()` and in the three call sites in AppState (`disconnect`, `requireSignIn`, `prepareDashboardBridge`). A regression where a teardown path forgets to call `invalidate()`, or where `invalidate()` forgets to clear `isReady`/`pendingRequests`, would pass CI. The test also busy-waits with `while requests.count == 0 { await Task.yield() }`, which has no bound and depends on the main-actor scheduler handing off to the child task.
**Suggestion:** Add a bridge-level test that creates a real `DashboardTicketBridge` (or a seam), registers a pending `requestJSON`, calls `invalidate()`, and asserts the caller resumes with `.notReady`. Replace the polling loop with an `AsyncStream`/`withCheckedThrowingContinuation` handoff or an `XCTestExpectation`.

### [severity: minor] Repair path can silently rewrite existing percent escapes
**File:** Conduit/Views/Components/MarkdownText.swift:872
**Issue:** On the repair path, `URLComponents` returns the percent-*decoded* `components.path`/`query`/`fragment`, which are then re-encoded. For a URL that fails strict parsing for an unrelated reason, valid escapes can be altered: `…/a%2Fb c.png` decodes to path `/a/b c.png`, and since `/` is in `.urlPathAllowed`, the re-encoded result is `/a/b%20c.png` — `%2F` becomes a path separator, pointing at a different resource. The doc comment ("preserving existing percent escapes … without double-encoding `%XX`") only holds for the strict path, not the repair path. Non-crashing and narrow (only URLs that already fail strict parsing), but it contradicts the stated intent.
**Suggestion:** Operate on `percentEncodedPath`/`percentEncodedQuery`/`percentEncodedFragment` and only repair invalid characters (e.g. via the existing `escapedStrayPercents` plus encoding of characters not in the allowed set) instead of round-tripping through the decoded components, or document the limitation.

### [severity: nit] Nonisolated mutable continuation store
**File:** Conduit/Services/DashboardTicketBridge.swift:155
**Issue:** `DashboardTicketBridgePendingRequests` is a plain (non-`Sendable`, non-isolated) mutable class holding `CheckedContinuation<[String: Any], Error>` values. All mutations happen on the main actor via the `@MainActor` bridge, but `deinit` also calls `rejectAll()` from the (potentially non-main) deallocating thread, and the test captures the instance across a `Task { @MainActor in … }` boundary. This compiles under Swift 5.9 but would be flagged under Swift 6 strict concurrency, and `resume` is being invoked off-main-actor (safe for `CheckedContinuation`, but the pattern is fragile).
**Suggestion:** Keep it as-is for now but document the isolation contract (main-actor mutation, deinit-only read/resume), and revisit with an explicit isolation strategy when the project adopts strict concurrency.

### [severity: nit] Linked fallback label drops alt text for VoiceOver
**File:** Conduit/Views/Components/MarkdownText.swift:970
**Issue:** In the linked branch, `title(alt:destinationAvailable:true)` returns "Open image" / "Image unavailable — open source" and discards the alt description, while the non-linked branch surfaces it ("\(alt) unavailable"). The label is the fallback's accessible content, so a user with VoiceOver hears a generic action label instead of what the image depicts. This predates the PR in the linked branch, but the new code had the opportunity to fix it.
**Suggestion:** Compose the label from alt when present, e.g. append the alt text to the accessible description for both branches.

## Verdict
COMMENT

The primary crash fix is correct (no remaining force-unwrap; the resolution chain is guarded, scheme/host-restricted, and covered by tests for strict, repairable, and unparseable cases), and the teardown fix is sound: `invalidate()` resumes all pending continuations before the bridge is dropped, and `rejectAll` plus the copy-and-clear pattern in `deinit`/`evaluateJavaScript` completion/JS `postMessage` leaves no double-resume or lost-resume path. All findings above are non-blocking; the bridge-level invalidation test coverage is the most worthwhile follow-up.

<a href="https://opencode.ai/s/43yn1658">New%20session%20-%202026-08-12T23%3A07%3A42.761Z</a>
[opencode session](https://opencode.ai/s/43yn1658)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31649648641)

kaishi00:
## Summary
Removes a force-unwrap crash (`URL(string:)!`) on the WebKit image-fallback link by adding strict→component-aware URL resolution and safe non-link fallback copy, and adds `DashboardTicketBridge.invalidate()` so pending `CheckedContinuation`s are resumed before `AppState` disconnects/signs out/replaces the bridge.

## Findings

### [severity: minor] Non-isolated mutable continuation store resumed off the main actor
**File:** Conduit/Services/DashboardTicketBridge.swift:155
**Issue:** `DashboardTicketBridgePendingRequests` is a plain, non-`Sendable` class holding `CheckedContinuation<[String: Any], Error>` in mutable `storage`. All mutation happens on the main actor via the `@MainActor` bridge, but the bridge's `deinit` (line 221) calls `rejectAll` from the deallocating thread — a non-isolated access to a `@MainActor`-isolated stored property that invokes `resume` off-actor. It compiles today (SWIFT_VERSION 5.9), but under Swift 6 strict concurrency this is a compile error and the off-main resume is exactly the kind of unsound access the project's other files guard against.
**Suggestion:** Document the isolation contract (main-actor mutation, deinit-only defensive read/resume) and, when the project adopts strict concurrency, restructure — e.g. make the container explicitly handle isolation (`@unchecked Sendable` with a documented main-actor invariant) or drop the `deinit` resume and rely solely on `invalidate()`, which is already the primary path.

### [severity: minor] Invalidated bridge stays alive for up to the JS timeout after teardown
**File:** Conduit/Services/DashboardTicketBridge.swift:364
**Issue:** The `webView.evaluateJavaScript` completion closure captures `self` strongly. After `invalidate()` + `dashboardTicketBridge = nil`, the bridge and its `WKWebView` remain retained by WebKit until the completion fires — up to the 12s `AbortController` timeout for an in-flight request. Sign-out therefore holds the bridge and the "dashboard-response" message handler for seconds longer than necessary. Pre-existing closure capture, but `invalidate()` is now the natural place to also tear down the handler.
**Suggestion:** In `invalidate()`, capture `self` weakly in the completion closure (or `removeScriptMessageHandler` / cancel the request) so teardown is immediate. Low impact, no user-visible defect.

### [severity: nit] Dropped deinit comment that documented the correctness invariant
**File:** Conduit/Services/DashboardTicketBridge.swift:220
**Issue:** The old `deinit` carried a detailed comment explaining that a suspended request's task frame retains the bridge, so `deinit`'s reject loop is effectively a no-op and teardown correctness depends on resuming continuations before deallocation. That invariant is now satisfied by `invalidate()` at exactly three call sites in AppState (lines 1664, 1764, 1779); a future reader could assume `deinit` alone protects awaiting tasks and drop one of those calls.
**Suggestion:** Restore a short comment noting that `invalidate()` (called before every bridge clear/replacement) is the primary resumption path and `deinit` is belt-and-suspenders.

### [severity: nit] Bridge test spins up a real WKWebView and cookie-restoration task
**File:** ConduitTests/DashboardTicketBridgeTests.swift:15
**Issue:** `DashboardTicketBridge.init` creates a real `WKWebView` and launches the `Task` that restores cookies, though the test only needs the injected `pendingRequests` store plus `invalidate()`. This couples the unit test to an app host and WebKit process availability. The test also doesn't assert the no-double-resume path (a late `postMessage`/evaluateJavaScript completion after `rejectAll`), which the implementation handles structurally via `removeValue(for:)` returning nil.
**Suggestion:** Acceptable as-is; if flakiness appears on a headless host, extract a lightweight invalidation seam that doesn't require WebKit. A follow-up asserting `removeValue` after invalidation returns nil would lock in the double-resume guard.

## Verdict
APPROVE

The crash fix is sound: no force-unwrap remains, `resolve()` is guarded by scheme/host validation, preserves existing `%XX` escapes (verified the raw-component splitting avoids the decoded-path `%2F`→`/` corruption), and repairable/unparseable/missing-host cases are all covered by tests. The teardown fix is correct: `invalidate()` resumes every pending continuation with `.notReady` before the bridge is dropped, `rejectAll` copies-and-clears so there is no double-resume or lost-resume path, and `waitUntilReady`/`requestJSON`/`reload` all bail on `isInvalidated`. Remaining findings are minor and non-blocking.

<a href="https://opencode.ai/s/BXS7RXiX">New%20session%20-%202026-08-13T00%3A08%3A11.944Z</a>
[opencode session](https://opencode.ai/s/BXS7RXiX)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31653323217)

PR #46, opened 2026-08-11
Invalidate the profile session cache on delete, archive, and sign-out

angel12:
`deleteSession` and `setSessionArchived` remove the row from the published arrays but leave it in `profileSessionCache`, and every non-forced catalog load merges that cache back in and re-saves the union — so a deleted or archived conversation resurrects in Recents on the next foreground or send, until a pull-to-refresh clears the exact keys involved.

- `removeSessionFromLiveCatalog` now purges matching rows from the cache (covers both delete and archive; matching is alias-aware via `sessionMatches`).
- `disconnect()` now clears `cronSessions`, `profileSessionCache`, and `loadedFullSessionHistory`, so re-signing-in cannot show sessions deleted from another client while signed out.

Full suite passes (438 tests, 0 failures). Full write-up: [angel12/hermes-conduit#6](https://github.com/angel12/hermes-conduit/issues/6)

## Summary by CodeRabbit

- **Bug Fixes**
  - Disconnecting now clears related session history and cached session data.
  - Deleted or archived sessions are removed from all session views, preventing stale entries from reappearing.
  - Improved session and cron loading to prevent outdated results from overwriting newer changes.
  - Refreshes better preserve accurate catalog data during concurrent updates or failed requests.
  - Session cache cleanup remains consistent during refresh and reconciliation.

- **Documentation**
  - Added documentation covering session-cache behavior, cleanup scenarios, and verification steps.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -175,8 +175,7 @@ struct SessionCatalogCache {
         liveKey: String,
         cronSessions: [SessionSummary]?,
         cronKey: String,
-        loadedFullHistoryKey: String,
-        recordFullHistoryAt: Date?,
+        historyMarkers: [String: Date],
         at generation: UInt64
     ) -> Bool {
         guard generation == mutationGeneration else { return false }
@@ -184,9 +183,9 @@ struct SessionCatalogCache {
         if let cronSessions {
             sessionsByKey[cronKey] = cronSessions
         }
-        if let recordFullHistoryAt {
-            loadedFullHistoryKeys.insert(loadedFullHistoryKey)
-            fullHistoryLoadedAt[loadedFullHistoryKey] = recordFullHistoryAt
+        for (key, loadedAt) in historyMarkers {
+            loadedFullHistoryKeys.insert(key)
+            fullHistoryLoadedAt[key] = loadedAt
         }
         return true
     }
@@ -4871,14 +4870,19 @@ final class AppState: ObservableObject {
                     let merged = uniqueSessions(scoped + cached)
 
                     // Fetch cron sessions separately -- the main query excludes them.
+                    let shouldLoadCron = sessionCatalogCache.shouldLoadFullHistory(
+                        forKey: cronKey,
+                        forceRefresh: forceRefresh
+                    )
                     let cachedCronSessions = sessionCatalogCache.cachedSessions(forKey: cronKey)?.filter {
                         sessionBelongsToProfile($0, profile: profile)
                     }
                     let publishedCronSnapshot = self.cronSessions.filter {
                         sessionBelongsToProfile($0, profile: profile)
                     }
+                    var didFetchCronSessions = false
                     let cronSessions: [SessionSummary]?
-                    if !forceRefresh, let cachedCronSessions {
+                    if !shouldLoadCron, let cachedCronSessions {
                         cronSessions = cachedCronSessions
                     } else {
                         do {
@@ -4888,6 +4892,7 @@ final class AppState: ObservableObject {
                             ).filter {
                                 sessionBelongsToProfile($0, profile: profile)
                             }
+                            didFetchCronSessions = true
                         } catch {
                             // Keep a previous cron snapshot if one exists, but
                             // do not cache an empty result for a failed request.
@@ -4908,13 +4913,19 @@ final class AppState: ObservableObject {
                         merged + (cronSessions ?? cachedCronSessions ?? publishedCronSnapshot)
                     )
                     if !combined.isEmpty || shouldLoadHistory == false {
+                        var historyMarkers: [String: Date] = [:]
+                        if scopedResult.isAuthoritative && !scoped.isEmpty {
+                            historyMarkers[cacheKey] = Date()
+                        }
+                        if didFetchCronSessions {
+                            historyMarkers[cronKey] = Date()
+                        }
                         guard sessionCatalogCache.commit(
                             liveSessions: combined,
                             liveKey: cacheKey,
                             cronSessions: cronSessions,
                             cronKey: cronKey,
-                            loadedFullHistoryKey: cacheKey,
-                            recordFullHistoryAt: scopedResult.isAuthoritative && !scoped.isEmpty ? Date() : nil,
+                            historyMarkers: historyMarkers,
                             at: generation
                         ) else {
                             catalogRetryCount += 1
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -118,6 +118,65 @@ struct ChatResumeLifecycleOperations {
     static let live = ChatResumeLifecycleOperations()
 }
 
+/// Owns the profile-scoped session catalog cache and rejects writes from a
+/// load that started before a destructive cache mutation. AppState is
+/// main-actor isolated, but every dashboard request can re-enter the actor
+/// while it awaits WebKit, so a request must not overwrite a newer purge when
+/// it resumes.
+struct SessionCatalogCache {
+    private(set) var sessionsByKey: [String: [SessionSummary]] = [:]
+    private(set) var loadedFullHistoryKeys = Set<String>()
+    private(set) var mutationGeneration: UInt64 = 0
+
+    func sessions(forKey key: String) -> [SessionSummary] {
+        sessionsByKey[key] ?? []
+    }
+
+    func cachedSessions(forKey key: String) -> [SessionSummary]? {
+        sessionsByKey[key]
+    }
+
+    /// Commits a live and cron snapshot only when no cache mutation occurred
+    /// during the load that produced it.
+    @discardableResult
+    mutating func commit(
+        liveSessions: [SessionSummary],
+        liveKey: String,
+        cronSessions: [SessionSummary],
+        cronKey: String,
+        loadedFullHistoryKey: String,
+        at generation: UInt64
+    ) -> Bool {
+        guard generation == mutationGeneration else { return false }
+        sessionsByKey[liveKey] = liveSessions
+        sessionsByKey[cronKey] = cronSessions
+        loadedFullHistoryKeys.insert(loadedFullHistoryKey)
+        return true
+    }
+
+    mutating func removeAll() {
+        mutationGeneration &+= 1
+        sessionsByKey.removeAll()
+        loadedFullHistoryKeys.removeAll()
+    }
+
+    mutating func removeValue(forKey key: String) {
+        mutationGeneration &+= 1
+        sessionsByKey.removeValue(forKey: key)
+        loadedFullHistoryKeys.remove(key)
+    }
+
+    mutating func removeSession(withIDs sessionIDs: Set<String>) {
+        mutationGeneration &+= 1
+        for key in sessionsByKey.keys {
+            sessionsByKey[key]?.removeAll { cachedSession in
+                let cachedIDs = Set([cachedSession.id] + cachedSession.alternateIds)
+                return !cachedIDs.isDisjoint(with: sessionIDs)
+            }
+        }
+    }
+}
+
 @MainActor
 private func scheduleChatResumeReconnectTask(
     after delay: TimeInterval,
@@ -532,9 +591,8 @@ final class AppState: ObservableObject {
     private let sessionCatalogLoaderOverride: ((Bool) async throws -> [SessionSummary])?
     private var reconnectAttempts = 0
     private var connectedAt: Date?
-    private var profileSessionCache: [String: [SessionSummary]] = [:]
+    private var sessionCatalogCache = SessionCatalogCache()
     private var projectsRequestGeneration = 0
-    private var loadedFullSessionHistory = Set<String>()
     private let sessionPresentationCache: SessionPresentationCache
 
     /// The dashboard's persisted transcript is richer than `session.resume`:
@@ -1237,8 +1295,7 @@ final class AppState: ObservableObject {
         recoverySequence.cancel()
         chatResumeRestorationRequest = nil
         invalidateReconciliation()
-        profileSessionCache.removeAll()
-        loadedFullSessionHistory.removeAll()
+        sessionCatalogCache.removeAll()
         sessions = []
         cronSessions = []
         archivedSessions = []
@@ -1675,10 +1732,9 @@ final class AppState: ObservableObject {
         cronSessions = []
         // Sessions can be deleted from another client while signed out; a
         // stale catalog cache would show those rows again after re-sign-in
-        // (and `loadedFullSessionHistory` would suppress the reload that
-        // could correct them).
-        profileSessionCache.removeAll()
-        loadedFullSessionHistory.removeAll()
+        // (and the full-history marker would suppress the reload that could
+        // correct them).
+        sessionCatalogCache.removeAll()
         pinnedSessionIDs = []
         messages = []
         setActiveSessionState(id: nil, title: "New conversation")
@@ -3504,9 +3560,8 @@ final class AppState: ObservableObject {
 
         let sessionKey = "\(activeProfile):exclude"
         let cronKey = "\(activeProfile):cron"
-        profileSessionCache.removeValue(forKey: sessionKey)
-        profileSessionCache.removeValue(forKey: cronKey)
-        loadedFullSessionHistory.remove(sessionKey)
+        sessionCatalogCache.removeValue(forKey: sessionKey)
+        sessionCatalogCache.removeValue(forKey: cronKey)
         await loadSessions(forceRefresh: true)
     }
 
@@ -3716,9 +3771,9 @@ final class AppState: ObservableObject {
         // the published arrays and re-saves the union. A row left in the
         // cache therefore resurrects a deleted or archived conversation on
         // the next foreground or send, until a pull-to-refresh purges it.
-        for key in profileSessionCache.keys {
-            profileSessionCache[key]?.removeAll { sessionMatches($0, session) }
-        }
+        sessionCatalogCache.removeSession(
+            withIDs: Set([session.id] + session.alternateIds)
+        )
     }
 
     func clearActiveSessionIfNeeded(
@@ -4741,34 +4796,55 @@ final class AppState: ObservableObject {
         if let dashboardTicketBridge {
             do {
                 let cacheKey = "\(profile):exclude"
-                let shouldLoadHistory = forceRefresh || !loadedFullSessionHistory.contains(cacheKey)
-                let scoped = try await dashboardSessions(
-                    profile: profile,
-                    loadFullHistory: shouldLoadHistory,
-                    using: dashboardTicketBridge
-                ).filter { sessionBelongsToProfile($0, profile: profile) }
-
-                let cached = forceRefresh ? [] : (profileSessionCache[cacheKey] ?? []).filter {
-                    sessionBelongsToProfile($0, profile: profile)
-                }
-                let merged = uniqueSessions(scoped + cached)
-                // Fetch cron sessions separately -- the main query excludes them.
                 let cronKey = "\(profile):cron"
-                let cronSessions: [SessionSummary]
-                if !forceRefresh, let cached = profileSessionCache[cronKey] {
-                    cronSessions = cached.filter { sessionBelongsToProfile($0, profile: profile) }
-                } else {
-                    cronSessions = ((try? await dashboardCronSessions(profile: profile, using: dashboardTicketBridge)) ?? []).filter {
+                while true {
+                    let generation = sessionCatalogCache.mutationGeneration
+                    let shouldLoadHistory = forceRefresh
+                        || !sessionCatalogCache.loadedFullHistoryKeys.contains(cacheKey)
+                    let scoped = try await dashboardSessions(
+                        profile: profile,
+                        loadFullHistory: shouldLoadHistory,
+                        using: dashboardTicketBridge
+                    ).filter { sessionBelongsToProfile($0, profile: profile) }
+
+                    let cached = forceRefresh ? [] : sessionCatalogCache.sessions(forKey: cacheKey).filter {
                         sessionBelongsToProfile($0, profile: profile)
                     }
-                    profileSessionCache[cronKey] = cronSessions
-                }
-                let combined = uniqueSessions(merged + cronSessions)
-                if !combined.isEmpty || shouldLoadHistory == false {
-                    loadedFullSessionHistory.insert(cacheKey)
-                    sessionCatalogLog.notice("Dashboard catalog for \(profile, privacy: .public): \(combined.count, privacy: .public) sessions; \(self.sourceSummary(combined), privacy: .public)")
-                    profileSessionCache[cacheKey] = combined
-                    return combined
+                    let merged = uniqueSessions(scoped + cached)
+
+                    // Fetch cron sessions separately -- the main query excludes them.
+                    let cronSessions: [SessionSummary]
+                    if !forceRefresh, let cached = sessionCatalogCache.cachedSessions(forKey: cronKey) {
+                        cronSessions = cached.filter { sessionBelongsToProfile($0, profile: profile) }
+                    } else {
+                        cronSessions = ((try? await dashboardCronSessions(profile: profile, using: dashboardTicketBridge)) ?? []).filter {
+                            sessionBelongsToProfile($0, profile: profile)
+                        }
+                    }
+
+                    // A delete/archive/disconnect can run while either request
+                    // is suspended. Never let this attempt overwrite the
+                    // newer cache state; retry from the authoritative source.
+                    guard sessionCatalogCache.mutationGeneration == generation else {
+                        continue
+                    }
+
+                    let combined = uniqueSessions(merged + cronSessions)
+                    if !combined.isEmpty || shouldLoadHistory == false {
+                        guard sessionCatalogCache.commit(
+                            liveSessions: combined,
+                            liveKey: cacheKey,
+                            cronSessions: cronSessions,
+                            cronKey: cronKey,
+                            loadedFullHistoryKey: cacheKey,
+                            at: generation
+                        ) else {
+                            continue
+                        }
+                        sessionCatalogLog.notice("Dashboard catalog for \(profile, privacy: .public): \(combined.count, privacy: .public) sessions; \(self.sourceSummary(combined), privacy: .public)")
+                        return combined
+                    }
+                    break
                 }
             } catch {
                 sessionCatalogLog.error("Dashboard history failed; using gateway fallback: \(error.localizedDescription, privacy: .public)")
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -139,6 +139,19 @@ struct SessionCatalogCache {
         sessionsByKey[key]
     }
 
+    /// Preserve the previous live snapshot when the dashboard returns no
+    /// usable rows. An empty response is not safe evidence that a populated
+    /// catalog was deleted: it can also represent a transient, malformed, or
+    /// profile-mismatched response.
+    func cachedSessionsForLiveMerge(
+        remoteSessions: [SessionSummary],
+        isAuthoritative: Bool,
+        forKey key: String
+    ) -> [SessionSummary] {
+        guard !isAuthoritative || remoteSessions.isEmpty else { return [] }
+        return sessions(forKey: key)
+    }
+
     func shouldLoadFullHistory(
         forKey key: String,
         forceRefresh: Bool,
@@ -625,6 +638,9 @@ final class AppState: ObservableObject {
     /// state and any in-flight projection.
     private struct DashboardSessionCatalog {
         let sessions: [SessionSummary]
+        /// True only when the dashboard returned a meaningful, terminal
+        /// catalog. Empty or unusable responses must remain mergeable with
+        /// the previous snapshot instead of evicting it.
         let isAuthoritative: Bool
     }
 
@@ -4843,15 +4859,25 @@ final class AppState: ObservableObject {
                         sessionBelongsToProfile($0, profile: profile)
                     }
 
-                    let cached = !scopedResult.isAuthoritative ? sessionCatalogCache.sessions(forKey: cacheKey).filter {
+                    let cached = sessionCatalogCache.cachedSessionsForLiveMerge(
+                        remoteSessions: scoped,
+                        isAuthoritative: scopedResult.isAuthoritative,
+                        forKey: cacheKey
+                    ).filter {
                         sessionBelongsToProfile($0, profile: profile)
-                    } : []
+                    }
                     let merged = uniqueSessions(scoped + cached)
 
                     // Fetch cron sessions separately -- the main query excludes them.
+                    let cachedCronSessions = sessionCatalogCache.cachedSessions(forKey: cronKey)?.filter {
+                        sessionBelongsToProfile($0, profile: profile)
+                    }
+                    let publishedCronSnapshot = self.cronSessions.filter {
+                        sessionBelongsToProfile($0, profile: profile)
+                    }
                     let cronSessions: [SessionSummary]?
-                    if !forceRefresh, let cached = sessionCatalogCache.cachedSessions(forKey: cronKey) {
-                        cronSessions = cached.filter { sessionBelongsToProfile($0, profile: profile) }
+                    if !forceRefresh, let cachedCronSessions {
+                        cronSessions = cachedCronSessions
                     } else {
                         do {
                             cronSessions = try await dashboardCronSessions(
@@ -4875,25 +4901,29 @@ final class AppState: ObservableObject {
                           self.client === client else {
                         throw DashboardTicketBridgeError.notReady
                     }
-                    guard sessionCatalogCache.mutationGeneration == generation else {
-                        catalogRetryCount += 1
-                        guard catalogRetryCount < maximumCatalogRetries else {
-                            break
-                        }
-                        continue
-                    }
 
-                    let combined = uniqueSessions(merged + (cronSessions ?? sessionCatalogCache.sessions(forKey: cronKey)))
+                    let combined = uniqueSessions(
+                        merged + (cronSessions ?? cachedCronSessions ?? publishedCronSnapshot)
+                    )
                     if !combined.isEmpty || shouldLoadHistory == false {
-                        _ = sessionCatalogCache.commit(
+                        guard sessionCatalogCache.commit(
                             liveSessions: combined,
                             liveKey: cacheKey,
                             cronSessions: cronSessions,
                             cronKey: cronKey,
                             loadedFullHistoryKey: cacheKey,
-                            recordFullHistoryAt: scopedResult.isAuthoritative ? Date() : nil,
+                            recordFullHistoryAt: scopedResult.isAuthoritative && !scoped.isEmpty ? Date() : nil,
                             at: generation
-                        )
+                        ) else {
+                            catalogRetryCount += 1
+                            guard catalogRetryCount < maximumCatalogRetries else {
+                                sessionCatalogLog.warning(
+                                    "Dashboard catalog mutation retry budget exhausted for \(profile, privacy: .public); using gateway fallback."
+                                )
+                                break
+                            }
+                            continue
+                        }
                         sessionCatalogLog.notice("Dashboard catalog for \(profile, privacy: .public): \(combined.count, privacy: .public) sessions; \(self.sourceSummary(combined), privacy: .public)")
                         return combined
                     }
@@ -5016,7 +5046,7 @@ final class AppState: ObservableObject {
                 || (total.map { offset + batch.count < $0 } ?? false)
                 || batch.count == 200
             if !hasMore || batch.isEmpty {
-                isAuthoritative = true
+                isAuthoritative = !sessions.isEmpty
                 break
             }
         }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -5042,12 +5042,19 @@ final class AppState: ObservableObject {
             let normalizedBatch = dashboardOwnedSessions(batch, profile: profile)
             sessions += normalizedBatch
 
+            let rawHasMore = response["has_more"] ?? response["hasMore"]
+            let explicitHasMore = rawHasMore.map { booleanValue($0) }
             let nextOffset = integerValue(response["next_offset"] ?? response["nextOffset"])
             let total = integerValue(response["total"])
-            let hasMore = booleanValue(response["has_more"] ?? response["hasMore"])
-                || (nextOffset ?? 0) > offset
-                || (total.map { offset + batch.count < $0 } ?? false)
-                || batch.count == 200
+            let hasExplicitTerminalSignal = explicitHasMore == false
+                || (nextOffset.map { $0 <= offset } ?? false)
+                || (total.map { offset + batch.count >= $0 } ?? false)
+            let hasMore = !hasExplicitTerminalSignal && (
+                explicitHasMore == true
+                    || (nextOffset ?? 0) > offset
+                    || (total.map { offset + batch.count < $0 } ?? false)
+                    || batch.count == 200
+            )
             if batch.isEmpty {
                 // An empty page after a non-empty prefix is not proof that
                 // the full catalog was read. Preserve older rows and retry a
@@ -5055,7 +5062,7 @@ final class AppState: ObservableObject {
                 break
             }
             if !hasMore {
-                isAuthoritative = !normalizedBatch.isEmpty
+                isAuthoritative = hasExplicitTerminalSignal && !normalizedBatch.isEmpty
                 break
             }
         }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -1665,6 +1665,13 @@ final class AppState: ObservableObject {
         showLogin = true
         sessions = []
         archivedSessions = []
+        cronSessions = []
+        // Sessions can be deleted from another client while signed out; a
+        // stale catalog cache would show those rows again after re-sign-in
+        // (and `loadedFullSessionHistory` would suppress the reload that
+        // could correct them).
+        profileSessionCache.removeAll()
+        loadedFullSessionHistory.removeAll()
         pinnedSessionIDs = []
         messages = []
         setActiveSessionState(id: nil, title: "New conversation")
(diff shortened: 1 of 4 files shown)


kaishi00:

[github run](/kaishi00/hermes-conduit/actions/runs/31538428349)

kaishi00:
## Summary
This PR invalidates the profile session cache on delete/archive/sign-out so stale rows can't resurrect in Recents, fails in-flight RPCs when the socket dies, breaks a WKScriptMessageHandler retain cycle, and adds rendering caches (attachments, Markdown, syntax highlighting) plus deduplication of buffered stream deltas after resume.

## Findings

### [severity: major] Partial stream deduplication reorders or drops buffered deltas
**File:** Conduit/Services/AppState.swift:2685
**Issue:** `deduplicatingBufferedEvents` concatenates all buffered deltas, computes a single `covered` prefix, then emits the entire remainder at the position of the *last* delta and silently drops every other delta event. When a non-delta event (e.g. `toolStart`) sits between deltas and the overlap ends mid-way through an earlier delta, the uncovered assistant text that streamed *before* the tool event is replayed *after* it. Example: buffer `[delta("abc def"), toolStart, delta("ghi")]` with `inflight` ending in `"abc"` (covered=3) produces `[toolStart, delta("defghi")]` instead of `[delta("def"), toolStart, delta("ghi")]` — text reordered across the tool event. Even without interleaved events, all deltas but the last are collapsed into one, losing streaming-tail fade granularity. The existing test `testNonDeltaEventsPassThroughInOrder` only covers the fully-covered delta case, so this path is untested.
**Suggestion:** Remove only the covered prefix while iterating the original delta events in order (CodeRabbit's proposed fix): track `remainingCovered`, skip each delta while it's fully covered, emit `String(text.dropFirst(remainingCovered))` for the delta that straddles the boundary, and keep every later delta and all non-delta events in their original order. Add a test with a partially overlapped first delta, an interleaved `toolStart`, and a later delta.

### [severity: minor] Dedup concatenates deltas across different sessions
**File:** Conduit/Services/AppState.swift:2676
**Issue:** `deltaTexts = events.compactMap { ... }` concatenates text from *every* buffered delta regardless of `sessionId`. Reconciliation can use `acceptsAnySession: true` (AppState.swift:2024), so if another session streams while the resume RPC is in flight, its deltas get merged into `combined` and are dropped or folded into the reconciled session's last delta, losing the other session's text in its live bubble.
**Suggestion:** Only run dedup when all buffered deltas share the resolved session's ID, or group deltas by `sessionId` and dedup each session's run independently.

### [severity: minor] Markdown render-cache key uses `String(describing: foregroundStyle)`
**File:** Conduit/Views/Components/MarkdownText.swift:108
**Issue:** `Color` does not offer a stable, unique description; `String(describing:)` of a `Color` is an opaque/debug description that is not a reliable discriminator. Two visually different colors could collide on the same key (serving text styled with the wrong color from the shared static cache) or the description could vary between builds, defeating the cache.
**Suggestion:** Key on a stable discriminator for the fixed style set actually passed in (e.g. an enum/int the caller provides), rather than the string form of `Color`.

### [severity: minor] `WeakScriptMessageHandler` forwards to a `@MainActor` method from a nonisolated context
**File:** Conduit/Services/DashboardTicketBridge.swift:137
**Issue:** The proxy (`NSObject`, nonisolated) forwards `didReceive:` to `delegate` which is `DashboardTicketBridge` (`@MainActor`). WebKit calls this on the main thread so it's correct in practice, but under strict concurrency (Swift 6) this is an error, and the proxy itself is not main-actor isolated.
**Suggestion:** Annotate `WeakScriptMessageHandler` with `@MainActor` (and `MainActor.assumeIsolated` in the forwarding call if needed) to make the guarantee explicit and Swift-6-ready.

### [severity: nit] Attachment image cache counts items, not bytes
**File:** Conduit/Views/ChatView.swift:780
**Issue:** `NSCache.countLimit = 64` bounds the number of decoded `UIImage`s, not memory. Large user photos (tens of MB decoded) × 64 rows could hold hundreds of MB until pressure-based eviction kicks in.
**Suggestion:** Consider `totalCostLimit` with `cost:` based on the image's pixel size when inserting.

The cache-invalidation core (purge on delete/archive via `removeSessionFromLiveCatalog`, full clear on `disconnect()`, and clearing `loadedFullSessionHistory` so history reloads on re-sign-in) is correct and well-reasoned; the RPC-failure and WebKit-cycle changes are sound.

## Verdict
REQUEST_CHANGES — the stream-order bug in `deduplicatingBufferedEvents` is a functional regression in a flow this PR explicitly targets, and the existing tests don't exercise partial overlap with interleaved events.

<a href="https://opencode.ai/s/7E6rXq6z">New%20session%20-%202026-08-12T01%3A58%3A24.997Z</a>
[opencode session](https://opencode.ai/s/7E6rXq6z)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31555299827)

kaishi00:
## Summary
This PR invalidates the profile session catalog cache so deleted/archived conversations can't resurrect in Recents: `removeSessionFromLiveCatalog` now purges matching rows from every `profileSessionCache` entry, and `disconnect()` clears `profileSessionCache` and `loadedFullSessionHistory` so re-sign-in performs an authoritative reload. The salvage plan/spec docs are also added.

## Findings

### [severity: minor] Cache purge can be undone by an in-flight catalog load
**File:** Conduit/Services/AppState.swift:3719-3721 (and 4751-4770)
**Issue:** `removeSessionFromLiveCatalog` only mutates the cache, but `profileSessions` reads the cache *before* the awaits and writes it back *after* them. Interleaving: load reads `cached` (line 4751) containing the row, suspends on the `await dashboardCronSessions(...)` at line 4761 (which runs whenever the cron cache is missing, e.g. first load after launch/profile switch), the delete/archive continuation resumes during that suspension and purges the row, then the load resumes and writes `profileSessionCache[cacheKey] = combined` (line 4770) — resurrecting the row from the pre-purge snapshot. This defeats the exact bug the PR targets.
**Suggestion:** Make the merge resilient to interleaving — e.g. re-read/validate the cache against a persistent tombstone set of deleted/archived session IDs filtered in `profileSessions` before the write-back, or re-apply the purge after the awaits (compare against a mutation generation counter captured before fetching).

### [severity: minor] No regression coverage for the new cache-invalidation contract
**File:** Conduit/Services/AppState.swift:1680-1681, 3719-3721
**Issue:** The two new behaviors — purge on delete/archive and full clear on disconnect/re-sign-in — are untested. The salvage plan (Task 2) explicitly declines to add tests because the caches are private, so the fix relies on the full suite, which doesn't exercise this path (no test references `profileSessionCache`/`loadedFullSessionHistory`).
**Suggestion:** Test the observable contract through existing seams rather than exposing private state — e.g. verify that after `deleteSession`/`archiveSession` followed by a non-forced `loadSessions` the session does not reappear, using the existing `sessionCatalogLoaderOverride` hook if it can represent the cache-merge path.

### [severity: nit] Cross-client deletions while the app stays signed in are not handled
**File:** Conduit/Services/AppState.swift:3719-3721
**Issue:** The purge only fires for deletions/archives performed by this client. If another client deletes a session while this client remains signed in, the stale row still sits in `profileSessionCache`, and a non-forced foreground load (`shouldLoadHistory == false`, cache merge at 4754/4766) re-introduces it until a pull-to-refresh. The PR description only claims coverage for self-deletion/archive and sign-out, so this is a documented limitation rather than a regression.
**Suggestion:** Consider filtering cached rows against a tombstone set (see above) so any deletion, regardless of origin, is honored on the next merge.

### [severity: nit] PR description overstates the disconnect change
**File:** Conduit/Services/AppState.swift:1675-1681
**Issue:** The PR body/plan state disconnect now clears `cronSessions`, but `cronSessions = []` at line 1675 is pre-existing; the new code only adds the two cache clears. Cosmetic, but the docs should match the actual diff.
**Suggestion:** Adjust the description/plan wording to list only the newly added clears.

## Verdict
COMMENT

The core fix is correct, minimal, and well-reasoned: purge on the authoritative delete/archive paths, full cache and history-marker clear on disconnect, matching is alias-aware via `sessionMatches`, and no force-unwraps, retain cycles, or new concurrency hazards in the added code (plain synchronous dictionary mutation on the main actor). The blocking concern is narrow — an in-flight catalog load racing the purge can still resurrect a row — and the remaining items are coverage and documentation gaps, so I'd approve once the cache write-back is made interleaving-safe (or the race is explicitly acknowledged) and ideally with a regression test for the delete-then-reload contract.

<a href="https://opencode.ai/s/amsNveyS">New%20session%20-%202026-08-13T00%3A38%3A14.325Z</a>
[opencode session](https://opencode.ai/s/amsNveyS)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31655020460)

kaishi00:
## Summary
This PR invalidates the profile session catalog cache on delete/archive/sign-out so stale rows can't resurrect in Recents. It wraps the cache in a value type (`SessionCatalogCache`) with a mutation generation counter, guards `profileSessions` commits against loads that started before a destructive mutation (retrying from the authoritative dashboard source), and clears the cache and full-history markers on `disconnect()`.

## Findings

### [severity: minor] Cross-client deletion while signed in still resurrects the row
**File:** Conduit/Services/AppState.swift:3767-3777 (purge) and 4810-4832 (merge)
**Issue:** `removeSessionFromLiveCatalog` only fires for deletions/archives performed by this client. If another client deletes a session while this app stays signed in, the row remains in `sessionCatalogCache`, and the next non-forced load's `merged = uniqueSessions(scoped + cached)` (line 4813) re-introduces it until a pull-to-refresh. This is the exact merge-based resurrection path the PR targets; only the sign-out case is covered.
**Suggestion:** Maintain a persistent tombstone set of deleted/archived session IDs (keyed by profile) and filter `cached` against it during the merge, so deletions from any origin are honored on the next load.

### [severity: minor] Retry loop re-fetches full history and is unbounded
**File:** Conduit/Services/AppState.swift:4800-4848
**Issue:** When a mutation lands during a load, the loop `continue`s and re-runs the entire iteration. For a `loadFullHistory: true` attempt this means re-fetching up to 25 pages of dashboard history (`maximumPages = 25`, line 4944) plus the cron fetch, even though the previously fetched `scoped` batch is discarded. If mutations recur (e.g., rapid deletes), the loop has no iteration bound and will keep re-fetching.
**Suggestion:** Bounded the loop (e.g., a small retry cap falling back to the gateway), and consider re-merging the already-fetched `scoped` data against the fresh cache on retry instead of re-issuing the network calls.

### [severity: minor] No regression coverage for the purge and reload contract
**File:** ConduitTests/SessionCatalogCacheTests.swift:1-46
**Issue:** The single test only exercises commit rejection after `removeSession`. The alias-aware `removeSession` purge itself, `removeAll`, `removeValue`, and the observable delete/archive-then-non-forced-reload contract in AppState are untested (the salvage plan at docs/superpowers/plans/2026-08-12-pr46-session-cache-salvage.md:108-119 explicitly records the AppState flows as unverified).
**Suggestion:** These are cheap to cover at the struct boundary (add assertions that `removeSession` purges by alternateId, and that `removeAll` clears keys and markers). For the AppState contract, delete/archive then call `loadSessions(forceRefresh: false)` through an existing seam and assert the session does not reappear.

### [severity: nit] The inner `guard commit else { continue }` is unreachable
**File:** Conduit/Services/AppState.swift:4834-4842
**Issue:** The outer generation check (line 4828) and the commit's own generation check (SessionCatalogCache.commit, line 150) run with no `await` in between, so on the main actor `commit` can never return false and the `continue` branch is dead code.
**Suggestion:** Either drop the redundant guard or rely solely on the commit result for the retry, keeping only one generation check for clarity.

### [severity: nit] Retry discards a good result on any cache mutation, including unrelated ones
**File:** Conduit/Services/AppState.swift:4801-4828
**Issue:** Any mutation — including `removeValue` for a different profile's key during profile switching (lines 3563-3564) — increments the generation and forces the in-flight load to fully re-fetch even when its data was unaffected. Conservative and correct, but can cause redundant full-history fetches during profile switches.
**Suggestion:** Acceptable as-is; if this shows up in practice, scope generation to the profile's cache keys rather than the whole cache.

## Verdict
APPROVE

The core fix is correct and well-reasoned: purge on the authoritative delete/archive paths, full clear on disconnect, and a generation-guarded commit with retry that closes the earlier load-vs-purge race (no await between the outer guard and commit, so a stale pre-purge snapshot can never be written back). The added code has no force unwraps, retain cycles, or concurrency hazards (all mutations stay on the main actor), and `SessionCatalogCache` is a cleanly testable value type. Remaining findings are minor or documented limitations — most notably the unhandled cross-client-delete-while-signed-in case and thin regression coverage — and none are merge-blocking.

<a href="https://opencode.ai/s/Nfn3GUB4">New%20session%20-%202026-08-13T00%3A57%3A39.925Z</a>
[opencode session](https://opencode.ai/s/Nfn3GUB4)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31656094086)

kaishi00:
## Summary
This PR prevents deleted/archived conversations from resurrecting in Recents by purging matching rows from a new generation-guarded `SessionCatalogCache` on delete/archive, clearing the cache on `disconnect()`, and guarding catalog-load commits against mutations that land while a dashboard request is suspended (with bounded retry and an authoritative-load eviction policy).

## Findings

### [severity: major] An authoritative empty response on a non-forced load wipes the cached catalog and suppresses recovery for 5 minutes
**File:** Conduit/Services/AppState.swift:4846-4896
**Issue:** When an authoritative response is returned, `cached` is forced to `[]` and `merged = uniqueSessions(scoped + cached)`. On a non-forced load within the full-history TTL (`shouldLoadFullHistory == false`, so `loadFullHistory: false` → 1 page), a transient empty or malformed dashboard response (`{"sessions": []}` or a wrong shape → `batch.isEmpty` → `isAuthoritative = true` at line 5019) yields `combined == []`, which passes the guard `!combined.isEmpty || shouldLoadHistory == false` (line 4887) and commits an empty live cache *and* re-records the full-history marker (`recordFullHistoryAt: Date()`, line 4894). Result: every cached session disappears from Recents, and non-forced loads won't re-fetch full history for another 5 minutes. The prior code always merged `cached` back in on non-forced loads, so an empty scoped response never wiped the cache — this is a regression introduced by the authority-based merge.
**Suggestion:** Only treat the response as eviction-authoritative when it is *meaningfully non-empty*, e.g. set `isAuthoritative = true` only when the catalog is complete **and** `!batch.isEmpty` (or require an explicit `has_more == false` on a non-empty batch). On an empty/authoritative response for a key that already has cached rows, fall back to the previous merge-and-keep behavior instead of committing `[]`.

### [severity: minor] Failed cron fetch on a cache-miss round hides cron sessions for that load
**File:** Conduit/Services/AppState.swift:4852-4886
**Issue:** When the cron cache is empty (or `forceRefresh` is set) and `dashboardCronSessions` throws, `cronSessions = nil` and `combined` falls back to `sessionCatalogCache.sessions(forKey: cronKey)` which is `[]` — so the cron sessions are omitted from the returned catalog and the caller publishes an empty `cronSessions` (AppState.swift:3564) for that round. This is transient (the next load retries), and it correctly avoids poisoning the cache, but the UI still flashes "cron conversations missing."
**Suggestion:** This is acceptable as-is; if the flash is undesirable, keep a previously published `cronSessions` snapshot to merge into `combined` when the refresh fails and no cache entry exists.

### [severity: minor] Generation-mismatch retry re-fetches the full history and can be triggered by unrelated cache mutations
**File:** Conduit/Services/AppState.swift:4831-4884
**Issue:** Any destructive mutation — including `removeValue(forKey:)` for an unrelated profile's key during profile switching (lines 3589-3593) — bumps `mutationGeneration` and forces an in-flight load to `continue` and re-run the whole loop body, re-issuing up to 25 pages of dashboard requests per retry (`maximumCatalogRetries = 3`). If mutations keep landing, the retry budget is exhausted and control silently `break`s out to the gateway fallback (line 4914) without logging, even though the dashboard bridge was healthy.
**Suggestion:** Consider scoping generation to the profile's keys rather than the whole cache, and log a warning when the retry budget is exhausted and the load falls back to the gateway. Re-merging the already-fetched `scoped` batch against the fresh cache on retry (instead of re-issuing network calls) would also cut the cost.

### [severity: minor] No integration coverage for the new cache-invalidation contract
**File:** ConduitTests/SessionCatalogCacheTests.swift:4-168
**Issue:** The six tests exercise the `SessionCatalogCache` struct in isolation (stale commit rejection, alias-aware purge, `removeAll`, authoritative overwrite, nil-cron commit, TTL expiry) but none cover the AppState-level flows this PR is fixing: delete/archive followed by a non-forced `loadSessions` must not resurrect the row, and the generation-mismatch retry + owner-guard abort paths in `profileSessions` are untested. The plan at docs/superpowers/plans/2026-08-12-pr46-session-cache-salvage.md:117-128 explicitly records those WebKit-backed flows as outside the harness.
**Suggestion:** Add focused tests through existing seams (e.g. `sessionCatalogLoaderOverride` or a suspended-load hook) asserting that after `deleteSession`/`archiveSession` a non-forced `loadSessions` does not republish the removed session, and that a disconnect mid-request does not commit or publish a stale catalog.

### [severity: nit] `_ = commit(...)` return value is discarded and its guard is dead code
**File:** Conduit/Services/AppState.swift:4878-4896
**Issue:** The outer `guard sessionCatalogCache.mutationGeneration == generation` (line 4878) and the commit's internal generation check (line 167) both run with no `await` between them on the main actor, so `commit` can never return `false` here and the `continue`/`break` retry logic only ever fires from the outer guard. The discarded result implies a stale-write path that cannot actually occur.
**Suggestion:** Keep a single generation check (the outer one drives the retry) and drop the redundant guard inside `commit`, or rely solely on the `commit` result for the retry to avoid the duplicated bookkeeping.

### [severity: nit] Remote deletions still resurrect for large catalogs on non-authoritative loads
**File:** Conduit/Services/AppState.swift:4846-4849
**Issue:** Eviction only happens when the response is `isAuthoritative`. A non-forced load fetches a single page, so for a user with more than 200 sessions (or >5000 when pagination truncates at 25 pages) the response is not authoritative, `cached` rows are merged back, and a session deleted from another client while this app stays signed in reappears until a pull-to-refresh. This is the documented cross-client limitation, not a regression.
**Suggestion:** Document it alongside the PR; a persistent tombstone set filtered during the merge would close it if cross-client deletion-while-signed-in becomes a requirement.

## Verdict
COMMENT

The core fix is correct and well-guarded: purge on the authoritative delete/archive paths, full cache and marker clear on `disconnect()`, a generation-guarded commit with bounded retry, and owner-validation before publish prevent the stale-row resurrection and post-sign-out commits that the PR targets. There are no force unwraps, retain cycles, or concurrency hazards introduced (all cache state is mutated synchronously on the main actor; `mutationGeneration` uses overflow-safe `&+=`). The main item worth addressing before merge is the authoritative-empty-response cache wipe, which is a genuine regression in the refactored merge path; the rest are coverage and robustness gaps.

<a href="https://opencode.ai/s/RdNmdoLN">New%20session%20-%202026-08-13T01%3A12%3A52.102Z</a>
[opencode session](https://opencode.ai/s/RdNmdoLN)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31656922354)

kaishi00:
## Summary
This PR prevents deleted/archived sessions from resurrecting in Recents by purging matching rows from a new generation-guarded `SessionCatalogCache` on delete/archive, clearing the cache (and full-history markers) on `disconnect()`, and rejecting catalog commits from loads that raced a destructive mutation (with bounded retry, owner-validation, and authoritative-response eviction policy). It adds focused struct-level tests and salvage docs.

## Findings

### [severity: minor] Empty batch mid-pagination is treated as authoritative and evicts cached rows
**File:** Conduit/Services/AppState.swift:5048-5050
**Issue:** `if !hasMore || batch.isEmpty { isAuthoritative = !sessions.isEmpty; break }` treats a transient/malformed empty page as terminal whenever an earlier page produced rows. On a full-history load (25 pages), a non-empty page 0 followed by an empty page 1 (older dashboards omit `has_more`, or a transient bad response) yields `isAuthoritative = true` with only the fetched prefix. `cachedSessionsForLiveMerge` then discards all cached rows beyond that prefix and the commit records a fresh full-history marker (AppState.swift:4915), suppressing recovery for 5 minutes. This is the same class of regression the PR's empty-response hardening guards against, but only for the fully-empty case.
**Suggestion:** Only mark the response authoritative when the terminal signal is trustworthy — e.g. `if !hasMore && !batch.isEmpty { isAuthoritative = true; break } else if batch.isEmpty { break }` — so a non-empty prefix followed by an empty batch stays non-authoritative and preserves the cache.

### [severity: minor] Cross-client deletion while signed in still resurrects for partial/truncated catalogs
**File:** Conduit/Services/AppState.swift:4862-4869
**Issue:** Eviction only occurs when the dashboard response is authoritative (full pagination completed). A non-forced load fetches a single page, so for users with >200 sessions (or >5000, where the 25-page cap truncates) the response is non-authoritative, `cachedSessionsForLiveMerge` merges cached rows back, and a session deleted from another client reappears until a pull-to-refresh. Documented in the salvage design as a known limitation, but it is the same merge path the PR is fixing.
**Suggestion:** Accept as a documented limitation, or maintain a tombstone set of deleted/archived IDs (keyed by profile) filtered during the merge so deletions from any origin are honored on the next non-forced load.

### [severity: minor] No AppState-level integration coverage for the new contract
**File:** ConduitTests/SessionCatalogCacheTests.swift:4-214
**Issue:** All six tests exercise `SessionCatalogCache` in isolation. The observable flows this PR fixes — `deleteSession`/`archiveSession` followed by a non-forced `loadSessions` must not republish the row, and a disconnect mid-request must not commit/publish a stale catalog (generation-mismatch retry + owner-guard abort) — are untested. The plan records these paths as unverified.
**Suggestion:** Add tests through existing seams (e.g. `sessionCatalogLoaderOverride` or a suspended-load hook) asserting the deleted/archived session does not reappear after a non-forced reload, and that a disconnect during an in-flight load does not publish stale results.

### [severity: minor] Retry loop is network-expensive and scoped to the whole cache
**File:** Conduit/Services/AppState.swift:4845-4926
**Issue:** Any destructive mutation bumps the global `mutationGeneration`, including `removeValue(forKey:)` for an unrelated profile's keys during profile switching (AppState.swift:3608-3609) or `refreshSessionCatalog`. That forces any in-flight load to discard its fetched `scoped` batch and re-issue up to 25 dashboard pages plus the cron request per retry (bounded at 3, then a silent `break` to the gateway fallback). Conservative and correct, but can cause redundant full-history fetches and, on retry exhaustion, downgrades a healthy dashboard to the gateway without surfacing why.
**Suggestion:** Scope the generation to the affected profile's cache keys rather than the whole cache, and/or re-merge the already-fetched `scoped` batch against the fresh cache on retry instead of re-issuing network calls. A log already exists for budget exhaustion — consider elevating it.

### [severity: nit] Retry re-fetch can still return a just-deleted row
**File:** Conduit/Services/AppState.swift:4917-4926
**Issue:** After a generation mismatch the retry re-queries the dashboard. If the DELETE hasn't propagated to the dashboard read path (eventual consistency), the re-fetch can return the deleted row and the commit at the new generation succeeds, resurrecting it in the cache and published arrays until a later forced refresh. Client-side generation guarding cannot close a server-side read-after-write race.
**Suggestion:** Transient and outside the client's control; acceptable as-is, but a revalidation pass against a just-purged ID set before commit (or relying on the existing forced-refresh path) would harden it.

### [severity: nit] `cachedSessionsForLiveMerge` double-negative reads backwards
**File:** Conduit/Services/AppState.swift:146-153
**Issue:** `guard !isAuthoritative || remoteSessions.isEmpty else { return [] }` means "if authoritative and non-empty, evict; else keep cache." The logic is correct but the double negative obscures intent, and the function name ("ForLiveMerge") doesn't convey that it decides eviction vs. merge.
**Suggestion:** Rename (e.g. `liveCacheForMerge(authoritative:remote:)`) and/or restructure to `if isAuthoritative && !remoteSessions.isEmpty { return [] }; return sessions(forKey: key)`.

No retain cycles, force unwraps, or new concurrency hazards were introduced: the cache is a value type mutated synchronously on the `@MainActor` AppState, `mutationGeneration` uses overflow-safe `&+=`, and the commit/owner guards run with no intervening `await`. The earlier authoritative-empty-response cache-wipe regression is genuinely fixed (`isAuthoritative = !sessions.isEmpty` plus `cachedSessionsForLiveMerge` preserving the cache on empty/non-authoritative responses).

## Verdict
COMMENT

The core fix is correct and well-guarded: purge on the authoritative delete/archive paths, full cache + marker clear on `disconnect()`, generation-guarded commits with bounded retry, and owner validation preventing post-disconnect publishes. The remaining items are narrow robustness edges (mid-pagination empty batch treated as authoritative), documented cross-client limitations, and thin integration coverage — none are merge-blocking, but the AppState-level regression tests and the pagination-authority handling are worth adding.

<a href="https://opencode.ai/s/tqBdoJw7">New%20session%20-%202026-08-13T01%3A26%3A45.778Z</a>
[opencode session](https://opencode.ai/s/tqBdoJw7)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31657659912)

kaishi00:
## Summary
This PR prevents deleted/archived conversations from resurrecting in Recents by purging matching rows from a new generation-guarded `SessionCatalogCache` on delete/archive, clearing the cache (and full-history markers) on `disconnect()`/sign-out, and making catalog-load commits resilient to mutations that land while a dashboard request is suspended (with bounded retry, owner validation, and an authoritative-response eviction policy). It adds struct-level cache tests and salvage planning docs.

## Findings

### [severity: minor] Authoritative eviction on a non-forced load depends on pagination metadata being trustworthy
**File:** Conduit/Services/AppState.swift:5057-5060 and :146-155
**Issue:** On a non-forced load (`loadFullHistory: false`, single page), `hasMore` is computed from `has_more`/`next_offset`/`total`/`batch.count == 200`, and if all are absent it defaults to `false`. `isAuthoritative = true` then evicts the entire cached catalog (`cachedSessionsToMerge` returns `[]`) and the commit records a fresh 5-minute full-history marker (line 4917). On a healthy dashboard this fires only when the catalog genuinely fits one page (≤200 sessions), so eviction is correct; but a dashboard that omits pagination fields while a >200-session catalog page happens to arrive under-filled would truncate the cache and suppress recovery for 5 minutes. The prior code always merged `cached` back on non-forced loads, so this eviction is new.
**Suggestion:** Consider treating the single-page response as eviction-authoritative only when a terminal signal is explicit (e.g. `has_more == false` present), or fall back to merge-and-keep when the cache holds more rows than the fetched page.

### [severity: minor] Cross-client deletion still resurrects rows for non-authoritative catalogs
**File:** Conduit/Services/AppState.swift:4864-4871
**Issue:** Eviction only happens when the response is `isAuthoritative`. A non-forced load fetches a single page; for users with >200 sessions (or >5000, where the 25-page cap truncates), the response is non-authoritative, `cachedSessionsToMerge` merges cached rows back, and a session deleted from another client reappears until a pull-to-refresh. Documented in the salvage design as a known limitation, but it is the same merge path this PR fixes for local deletions.
**Suggestion:** Accept as documented, or maintain a tombstone set of deleted/archived IDs (keyed by profile) filtered during the merge so deletions from any origin are honored on the next non-forced load.

### [severity: minor] A failed cron fetch on a cache-miss round briefly hides cron sessions
**File:** Conduit/Services/AppState.swift:4880-4896 and 4907-4909
**Issue:** When the cron cache is empty (or `forceRefresh`) and `dashboardCronSessions` throws, `cronSessions = nil` and the combined result falls back to `cachedCronSessions ?? publishedCronSnapshot`, which is `[]` on a fresh cache miss. The caller then publishes an empty `cronSessions` for that round (AppState.swift:3582), flashing the cron tab empty until the next load. Transient and correct in that it does not poison the cache, but still a visible regression window per load.
**Suggestion:** Acceptable as-is; if the flash matters, keep the previously published `cronSessions` snapshot (already done via `publishedCronSnapshot`) and only exclude it when it is genuinely stale for this profile.

### [severity: minor] Retry re-fetches the full history and is triggered by unrelated cache mutations
**File:** Conduit/Services/AppState.swift:4849-4928
**Issue:** Any destructive mutation — including `removeValue(forKey:)` for an unrelated profile's keys during `refreshSessionCatalog`/profile switching (lines 3610-3611) — bumps the global `mutationGeneration` and forces an in-flight load to discard its fetched `scoped` batch and re-issue up to 25 dashboard pages plus the cron request per retry (bounded at 3, then a `break` to the gateway fallback). Conservative and correct, but can cause redundant full-history fetches, and on budget exhaustion a healthy dashboard silently downgrades to the gateway (`client.sessions()`), which may not reflect a dashboard delete.
**Suggestion:** Scope the generation to the affected profile's cache keys rather than the whole cache, and/or re-merge the already-fetched `scoped` batch against the fresh cache on retry instead of re-issuing network calls.

### [severity: minor] No integration coverage for the new cache-invalidation contract
**File:** ConduitTests/SessionCatalogCacheTests.swift:4-214
**Issue:** All seven tests exercise `SessionCatalogCache` in isolation (stale-commit rejection, alias purge, `removeAll`, authoritative omission, empty-preservation, nil-cron commit, TTL expiry). The observable flows this PR fixes — `deleteSession`/`archiveSession` followed by a non-forced `loadSessions` must not republish the row, and a disconnect mid-request must not commit/publish a stale catalog — are untested; the salvage plan explicitly records those WebKit-backed paths as outside the harness (docs/superpowers/plans/2026-08-12-pr46-session-cache-salvage.md:117-128).
**Suggestion:** Add tests through existing seams (e.g. `sessionCatalogLoaderOverride`) asserting the deleted/archived session does not reappear after a non-forced reload, and that a disconnect during an in-flight load does not publish stale results.

### [severity: nit] Retry re-fetch can still return a just-deleted row
**File:** Conduit/Services/AppState.swift:4911-4928
**Issue:** After a generation mismatch, the retry re-queries the dashboard. If the DELETE has not propagated to the read path (eventual consistency), the re-fetch can return the deleted row and the commit at the new generation succeeds, resurrecting it in the cache and published arrays until a later forced refresh. Client-side generation guarding cannot close a server-side read-after-write race.
**Suggestion:** Transient and outside the client's control; acceptable as-is, though a revalidation pass against the just-purged ID set before commit would harden it.

### [severity: nit] Retry path on commit failure is reachable (not dead code)
**File:** Conduit/Services/AppState.swift:4919-4927
**Issue:** Note for reviewers: the `else` branch after a failed `commit` is reachable — the generation is captured at line 4850 before the awaits at 4855/4885, so a delete/archive/disconnect landing during either suspension makes `commit` return `false` and the retry `continue` fire. Earlier reviews calling this branch dead code were incorrect; the `generation == mutationGeneration` check in `commit` (line 182) is what makes the retry meaningful.
**Suggestion:** No change needed; keep the owner-guard plus generation-check as-is.

No retain cycles, force unwraps, or new concurrency hazards were introduced: the cache is a value type mutated synchronously on the `@MainActor` AppState, `mutationGeneration` uses overflow-safe `&+=`, and the owner/commit guards run with no intervening `await`. The core fix — purge on the authoritative delete/archive paths, full cache + marker clear on disconnect/sign-out, and generation-guarded commits — is correct and well-reasoned.

## Verdict
COMMENT

The core fix is correct and well-guarded; the remaining items are narrow robustness edges (authoritative single-page eviction when pagination metadata is absent, cross-client deletions on non-authoritative catalogs, transient cron flash, retry cost) and thin integration coverage. None are merge-blocking, but the AppState-level regression tests and the pagination-authority handling are worth addressing.

<a href="https://opencode.ai/s/AzD1Fpnz">New%20session%20-%202026-08-13T01%3A35%3A17.899Z</a>
[opencode session](https://opencode.ai/s/AzD1Fpnz)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31658112348)

kaishi00:
## Summary
This PR prevents deleted/archived conversations from resurrecting in Recents by purging matching rows from a new generation-guarded `SessionCatalogCache` on delete/archive, clearing the cache (and full-history markers) on `disconnect()`, and guarding catalog-load commits against mutations that land while a dashboard request is suspended (bounded retry, owner-validation, and an authoritative-response eviction policy). It adds struct-level cache tests and salvage planning docs.

## Findings

### [severity: minor] Cron cache rows are never revalidated and have no expiry, so remotely deleted cron sessions persist until a forced refresh
**File:** Conduit/Services/AppState.swift:4880-4882
**Issue:** On a non-forced load, `if !forceRefresh, let cachedCronSessions { cronSessions = cachedCronSessions }` skips the network fetch entirely whenever the cron key exists. Unlike the live key, the cron key has no TTL (`shouldLoadFullHistory` only consults the live cache key), so once populated it is reused indefinitely. A cron session deleted from *another* client while this client stays signed in therefore remains visible (and gets recommitted through `combined`) until a pull-to-refresh/`forceRefresh`. Local deletes are handled correctly by `removeSessionFromLiveCatalog`, so this is the cross-client case only, but it is worse than the live-key case because there is no expiry fallback at all.
**Suggestion:** Apply the same TTL/marker treatment to the cron key, or always refresh cron when the cached entry is older than a bounded interval, so stale cron rows eventually self-heal on a normal foreground load.

### [severity: minor] Cross-client live deletions still resurrect for partial/truncated catalogs
**File:** Conduit/Services/AppState.swift:4864-4871
**Issue:** Eviction happens only when `scopedResult.isAuthoritative` is true. A non-forced load fetches a single page, so for users with >200 sessions (or >5000, where the 25-page cap truncates), the response is non-authoritative, `cachedSessionsToMerge` returns the full cached set, and a session deleted from another client reappears until a forced refresh. This is documented in the salvage design as a known limitation, but it is the same merge path this PR fixes for local deletions.
**Suggestion:** Accept as documented, or maintain a persistent tombstone set of deleted/archived IDs (keyed by profile) filtered during the merge so deletions from any origin are honored on the next non-forced load.

### [severity: minor] Retry loop re-fetches full history and the generation is scoped to the whole cache
**File:** Conduit/Services/AppState.swift:4847-4928
**Issue:** Any destructive mutation bumps the global `mutationGeneration` — including `removeValue(forKey:)` for an unrelated profile during `refreshSessionCatalog`/profile switching. That forces an in-flight load to discard its fetched `scoped` batch and re-issue up to 25 dashboard pages plus the cron request per retry (bounded at 3, then a silent `break` to the gateway `client.sessions()` fallback, which is not dashboard-authoritative). Conservative and correct, but can cause redundant full-history fetches and, on budget exhaustion, downgrade a healthy dashboard to the gateway.
**Suggestion:** Scope the generation to the affected profile's cache keys rather than the whole cache, and/or re-merge the already-fetched `scoped` batch against the fresh cache on retry instead of re-issuing network calls. Elevate the budget-exhaustion warning so the fallback is observable.

### [severity: minor] No AppState-level integration coverage for the new cache-invalidation contract
**File:** ConduitTests/SessionCatalogCacheTests.swift:4-214
**Issue:** All seven tests exercise `SessionCatalogCache` in isolation. The observable flows this PR fixes — `deleteSession`/`archiveSession` followed by a non-forced `loadSessions` must not republish the row, and a disconnect mid-request must not commit/publish a stale catalog (the generation-mismatch retry and owner-guard abort paths in `profileSessions`) — are untested; the plan explicitly records those WebKit-backed paths as outside the harness.
**Suggestion:** Add tests through existing seams (e.g. `sessionCatalogLoaderOverride` or a suspended-load hook) asserting the deleted/archived session does not reappear after a non-forced reload, and that a disconnect during an in-flight load does not publish stale results.

### [severity: nit] `removeSession(withIDs:)` bumps the generation even when nothing was removed
**File:** Conduit/Services/AppState.swift:208-216
**Issue:** Every delete/archive increments `mutationGeneration` regardless of whether any cached row actually matched, aborting and retrying any in-flight load even when the session wasn't present in the cache (e.g. a brand-new session deleted before its first load). Cost is one redundant retry per such call.
**Suggestion:** Only bump the generation when at least one row was actually removed (track a `didRemove` flag inside the loop).

### [severity: nit] Retry can recommit a just-deleted row via server read-after-write lag
**File:** Conduit/Services/AppState.swift:4911-4928
**Issue:** After a generation mismatch the retry re-queries the dashboard. If the DELETE has not propagated to the dashboard read path, the re-fetch can return the deleted row and the commit at the new generation succeeds, resurrecting it in the cache and published arrays until a later forced refresh. Client-side generation guarding cannot close a server-side read-after-write race.
**Suggestion:** Transient and outside the client's control; acceptable as-is, though a revalidation pass against the just-purged ID set before commit would harden it.

No retain cycles, force unwraps, or new concurrency hazards were introduced: the cache is a value type mutated synchronously on the `@MainActor` `AppState`, `mutationGeneration` uses overflow-safe `&+=`, and the owner-guard + commit-generation checks run with no intervening `await` (the retry path is genuinely reachable because the generation is captured before the awaiting fetches). The pagination authority logic is conservative and correctly avoids the earlier empty/mid-pagination cache-wipe regressions.

## Verdict
APPROVE

The core fix is correct and well-guarded: purge on the authoritative delete/archive paths, full cache + marker clear on `disconnect()`, generation-guarded commits with bounded retry, and owner-validation preventing post-disconnect publishes. The remaining items are narrow robustness edges (cron cache never revalidated, cross-client deletions on partial catalogs, retry cost) and thin integration coverage — none are merge-blocking, but the AppState-level regression tests and cron revalidation are worth following up on.

<a href="https://opencode.ai/s/rQ5PUUg6">New%20session%20-%202026-08-13T01%3A47%3A27.106Z</a>
[opencode session](https://opencode.ai/s/rQ5PUUg6)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31658754914)

kaishi00:
## Summary
This PR fixes stale Recents entries by wrapping the profile session catalog in a generation-guarded value type (`SessionCatalogCache`): delete/archive purge matching rows (alias-aware) from every cache key, `disconnect()` clears the cache plus full-history markers, and `profileSessions` rejects commits from loads that raced a destructive mutation (bounded retry, owner validation, authoritative-response eviction). Adds 8 struct-level tests and salvage docs.

## Findings

### [severity: minor] Cross-client deletion while signed in still resurrects rows, and eviction never fires for dashboards that omit pagination metadata
**File:** Conduit/Services/AppState.swift:5060-5078 (and :4863-4870)
**Issue:** `isAuthoritative` is only set when an *explicit* terminal signal is present (`has_more == false`, `next_offset ≤ offset`, or `total ≤ offset + batch.count`). For dashboards that omit all three fields (the code itself notes "older servers that ignore the query parameter" at :5147), `hasExplicitTerminalSignal` is always false, so a complete catalog of <200 rows is never authoritative and cached rows are always merged back. A session deleted from another client therefore reappears in Recents on every non-forced load and even on pull-to-refresh (forceRefresh only controls the marker, not the authority check). Self-healing exists only for the `has_more`-aware dashboard after the 5-minute TTL. This is the same merge path the PR fixes for local deletes, so the fix is incomplete for the documented cross-client case.
**Suggestion:** Treat a page as terminal when the batch is under-filled *and* the cache holds no more rows than the fetched catalog (e.g. authoritative when `!hasMore` and `scoped.count` is a superset of the previous cached count minus known deletions), or maintain a persistent tombstone set of deleted/archived IDs (keyed by profile) filtered during the merge so deletions from any origin are honored on the next load.

### [severity: minor] No integration coverage for the exact flows this PR fixes
**File:** ConduitTests/SessionCatalogCacheTests.swift:4-246
**Issue:** All eight tests exercise `SessionCatalogCache` in isolation. The observable contract — `deleteSession`/`archiveSession` followed by a non-forced `loadSessions` must not republish the row, and a `disconnect()` during an in-flight dashboard request must not commit or publish a stale catalog (the generation-retry and owner-guard paths) — is untested. The salvage plan explicitly records those WebKit-backed paths as unverified (docs/superpowers/plans/2026-08-12-pr46-session-cache-salvage.md:117-128).
**Suggestion:** Add tests through existing seams (e.g. `sessionCatalogLoaderOverride` or a suspended-load hook) asserting the deleted/archived session does not reappear after a non-forced reload, and that a disconnect mid-request neither commits to the cache nor publishes stale results.

### [severity: minor] Retry re-issues up to 25 pages plus the cron request and is triggered by unrelated cache mutations
**File:** Conduit/Services/AppState.swift:4930-4938
**Issue:** Any destructive mutation bumps the global `mutationGeneration`, including `removeValue(forKey:)` for an unrelated profile during `refreshSessionCatalog`/profile switching (:3609-3610). That forces an in-flight load to discard its already-fetched `scoped` batch and re-run the whole body per retry (bounded at 3), then silently `break` to the non-dashboard `client.sessions()` gateway fallback (:4957). Correct but network-expensive, and the fallback may not reflect a dashboard delete.
**Suggestion:** Scope the generation to the profile's cache keys rather than the whole cache, and/or re-merge the already-fetched `scoped` batch against the fresh cache on retry instead of re-issuing the network calls. Elevate the budget-exhaustion warning so the gateway downgrade is observable.

### [severity: nit] `removeSession(withIDs:)` bumps the generation even when nothing was removed
**File:** Conduit/Services/AppState.swift:207-215
**Issue:** Every delete/archive increments `mutationGeneration` regardless of whether any cached row matched, aborting and retrying any in-flight load even for a session that was never in the cache (e.g. a brand-new session deleted before its first catalog load).
**Suggestion:** Track a `didRemove` flag inside the loop and only bump the generation when at least one row was actually purged.

### [severity: nit] `cachedSessionsToMerge` double-negative reads backwards
**File:** Conduit/Services/AppState.swift:151-154
**Issue:** `guard !isAuthoritative || remoteSessions.isEmpty else { return [] }` means "if authoritative and non-empty, evict; else keep." Correct but obscures intent, and the name doesn't convey that it decides eviction vs. merge.
**Suggestion:** Restructure to `if isAuthoritative && !remoteSessions.isEmpty { return [] }; return sessions(forKey: key)` and rename (e.g. `liveCacheForMerge(authoritative:remote:)`).

## Verdict
COMMENT

The core fix is correct and well-guarded: purge on the authoritative delete/archive paths, full cache + marker clear on `disconnect()`, alias-aware `removeSession` matching, and a generation-guarded commit whose generation check (captured before the awaiting fetches) genuinely closes the load-vs-purge race. The empty-response wipe and mid-pagination regressions called out in earlier reviews are fixed, and the cron cache now has a TTL. No retain cycles, force unwraps, or concurrency hazards were introduced (cache mutations stay on the `@MainActor` AppState; `mutationGeneration` uses overflow-safe `&+=`). The remaining items — the cross-client-deletion gap that is unconditional on pagination-metadata-less dashboards, thin AppState-level test coverage, and retry cost — are non-blocking but worth following up, with the cross-client tombstone handling being the most impactful.

<a href="https://opencode.ai/s/gyXA1aH6">New%20session%20-%202026-08-13T01%3A57%3A50.014Z</a>
[opencode session](https://opencode.ai/s/gyXA1aH6)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31659287992)

kaishi00 (approved):
Reviewed after the session-cache race, pagination-authority, empty-response, and cron TTL follow-ups. Hosted CI and automated reviews are green; remaining metadata-less dashboard behavior is documented and non-blocking.

PR #47, opened 2026-08-11
Stop re-latching follow-latest while the user is dragging the transcript

angel12:
## Summary

- Prevents geometry callbacks from re-latching `followsLatest` while the user is actively dragging the transcript.
- Defers the drag-end decision to the next main-actor turn so ending near the bottom re-enables following even when no geometry preference changes.
- Captures gesture ownership at drag start and rejects stale completion after a newer drag, viewport transition, restoration, session change, notification handoff, cancellation, or view disappearance.
- Persists only after the relatch decision and resolves equivalent runtime IDs to the current canonical session key.

## Background

A slow upward drag during streaming set `followsLatest = false`, but content-growth and scroll geometry callbacks could immediately set it back to `true` while the finger was still inside the 40 pt near-bottom window. The next streamed delta then pulled the transcript back to the bottom.

The original branch also contained stacked fixes that had since landed on `main`. This branch has been updated with current `main`, so the PR now contains only the scroll interaction repair and its focused tests.

## Verification

- Added regression coverage for:
  - mid-drag relatch suppression;
  - near-bottom, restoration, and notification-handoff guards;
  - deterministic next-main-actor-turn completion;
  - drag-start session and viewport-transition ownership;
  - normal, cancelled, invalidated, and disappeared gesture lifecycles;
  - equivalent-ID canonical persistence;
  - relatch-before-persist ordering.
- Mutation checks:
  - removing the next-turn suspension fails `testDragCompletionWaitsForNextMainActorTurn`;
  - allowing an invalidated gesture to finish fails `testDragLifecycleCapturesStartAndSuppressesInvalidatedGestureUntilFinish`.
- Local full suite on iPhone 17 Pro / iOS 26.5 Simulator: **501 tests passed, 0 failed, 0 skipped**.

Original report: [angel12/hermes-conduit#7](https://github.com/angel12/hermes-conduit/issues/7)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat scrolling during and after drag gestures.
  * Prevented outdated or cancelled drag events from unexpectedly repositioning the conversation.
  * Preserved the latest-message position more reliably during restoration, conversation changes, and notification transitions.
  * Improved viewport persistence when returning to an equivalent conversation.

* **Tests**
  * Added comprehensive coverage for scroll eligibility, drag completion, cancellation, restoration, and persistence.

### DIFF
--- a/Conduit/Services/ChatScrollState.swift
+++ b/Conduit/Services/ChatScrollState.swift
@@ -107,6 +107,10 @@ enum ChatFollowLatestRelatchPolicy {
             && !isDragging
     }
 
+    static func shouldFollowLatestAfterTransition(isDragging: Bool) -> Bool {
+        !isDragging
+    }
+
     static func isCompletionCurrent(
         completed: ChatDragCompletionToken,
         current: ChatDragCompletionToken,
@@ -115,7 +119,11 @@ enum ChatFollowLatestRelatchPolicy {
         hasPendingRestoration: Bool,
         hasNotificationHandoff: Bool
     ) -> Bool {
-        let sameSession = completed.sessionKey == current.sessionKey
+        // A new chat can acquire its first server session ID without replacing
+        // the viewport. The transition generation distinguishes that identity
+        // resolution from an actual transcript transition.
+        let sameSession = completed.sessionKey == nil
+            || completed.sessionKey == current.sessionKey
             || identity.areEquivalent(completed.sessionKey, current.sessionKey)
         return completed.dragGeneration == current.dragGeneration
             && completed.viewportTransitionGeneration == current.viewportTransitionGeneration
--- a/Conduit/Services/ChatScrollState.swift
+++ b/Conduit/Services/ChatScrollState.swift
@@ -39,19 +39,28 @@ enum ChatFollowLatestRelatchPolicy {
     static func shouldRelatch(
         isNearBottom: Bool,
         hasPendingRestoration: Bool,
+        hasNotificationHandoff: Bool,
         isDragging: Bool
     ) -> Bool {
-        isNearBottom && !hasPendingRestoration && !isDragging
+        isNearBottom
+            && !hasPendingRestoration
+            && !hasNotificationHandoff
+            && !isDragging
     }
 
+    /// Defers completion until gesture state from the current actor turn has
+    /// settled. A newer drag or transcript transition invalidates the work via
+    /// `isCurrent`; otherwise persistence runs after the relatch decision.
     @MainActor
-    static func relatchAfterDragEnds(
-        isDragging: () -> Bool,
-        relatch: () -> Void
+    static func completeDragAfterYield(
+        isCurrent: () -> Bool,
+        relatch: () -> Void,
+        persist: () -> Void
     ) async {
         await Task.yield()
-        guard !isDragging() else { return }
+        guard isCurrent() else { return }
         relatch()
+        persist()
     }
 }
 
--- a/Conduit/Services/ChatScrollState.swift
+++ b/Conduit/Services/ChatScrollState.swift
@@ -35,6 +35,26 @@ enum ChatMessageScrollUpdatePolicy {
     }
 }
 
+enum ChatFollowLatestRelatchPolicy {
+    static func shouldRelatch(
+        isNearBottom: Bool,
+        hasPendingRestoration: Bool,
+        isDragging: Bool
+    ) -> Bool {
+        isNearBottom && !hasPendingRestoration && !isDragging
+    }
+
+    @MainActor
+    static func relatchAfterDragEnds(
+        isDragging: () -> Bool,
+        relatch: () -> Void
+    ) async {
+        await Task.yield()
+        guard !isDragging() else { return }
+        relatch()
+    }
+}
+
 struct ChatMessageScrollTargetCache: Equatable {
     private(set) var targets: [ChatMessageScrollTarget] = []
     private(set) var renderingRevision: UInt64 = 0
--- a/Conduit/Services/ChatScrollState.swift
+++ b/Conduit/Services/ChatScrollState.swift
@@ -35,6 +35,65 @@ enum ChatMessageScrollUpdatePolicy {
     }
 }
 
+struct ChatDragCompletionToken: Hashable {
+    let dragGeneration: UInt64
+    let sessionKey: ChatScrollSessionKey?
+    let viewportTransitionGeneration: UInt64
+}
+
+struct ChatDragLifecycleState: Equatable {
+    private(set) var generation: UInt64 = 0
+    private var activeCompletion: ChatDragCompletionToken?
+    private var activeGestureInvalidated = false
+
+    mutating func begin(
+        sessionKey: ChatScrollSessionKey?,
+        viewportTransitionGeneration: UInt64
+    ) -> Bool {
+        guard activeCompletion == nil, !activeGestureInvalidated else { return false }
+        generation &+= 1
+        activeCompletion = ChatDragCompletionToken(
+            dragGeneration: generation,
+            sessionKey: sessionKey,
+            viewportTransitionGeneration: viewportTransitionGeneration
+        )
+        return true
+    }
+
+    mutating func invalidate(hasActiveGesture: Bool) {
+        generation &+= 1
+        if hasActiveGesture || activeCompletion != nil {
+            activeGestureInvalidated = true
+        }
+    }
+
+    mutating func abandon() {
+        generation &+= 1
+        activeCompletion = nil
+        activeGestureInvalidated = false
+    }
+
+    mutating func finish() -> ChatDragCompletionToken? {
+        defer {
+            activeCompletion = nil
+            activeGestureInvalidated = false
+        }
+        guard !activeGestureInvalidated else { return nil }
+        return activeCompletion
+    }
+
+    func currentToken(
+        sessionKey: ChatScrollSessionKey?,
+        viewportTransitionGeneration: UInt64
+    ) -> ChatDragCompletionToken {
+        ChatDragCompletionToken(
+            dragGeneration: generation,
+            sessionKey: sessionKey,
+            viewportTransitionGeneration: viewportTransitionGeneration
+        )
+    }
+}
+
 enum ChatFollowLatestRelatchPolicy {
     static func shouldRelatch(
         isNearBottom: Bool,
@@ -48,17 +107,58 @@ enum ChatFollowLatestRelatchPolicy {
             && !isDragging
     }
 
-    /// Defers completion until gesture state from the current actor turn has
-    /// settled. A newer drag or transcript transition invalidates the work via
-    /// `isCurrent`; otherwise persistence runs after the relatch decision.
+    static func isCompletionCurrent(
+        completed: ChatDragCompletionToken,
+        current: ChatDragCompletionToken,
+        identity: ChatScrollSessionIdentity,
+        isDragging: Bool,
+        hasPendingRestoration: Bool,
+        hasNotificationHandoff: Bool
+    ) -> Bool {
+        let sameSession = completed.sessionKey == current.sessionKey
+            || identity.areEquivalent(completed.sessionKey, current.sessionKey)
+        return completed.dragGeneration == current.dragGeneration
+            && completed.viewportTransitionGeneration == current.viewportTransitionGeneration
+            && sameSession
+            && !isDragging
+            && !hasPendingRestoration
+            && !hasNotificationHandoff
+    }
+
+    static func persistenceSessionKey(
+        currentKey: ChatScrollSessionKey?,
+        identity: ChatScrollSessionIdentity
+    ) -> ChatScrollSessionKey? {
+        guard let currentKey else { return nil }
+        if identity.areEquivalent(currentKey, identity.canonicalSessionKey) {
+            return identity.canonicalSessionKey
+        }
+        return currentKey
+    }
+
     @MainActor
-    static func completeDragAfterYield(
-        isCurrent: () -> Bool,
-        relatch: () -> Void,
-        persist: () -> Void
+    static func waitForNextMainActorTurn() async {
+        await withCheckedContinuation { continuation in
+            DispatchQueue.main.async {
+                continuation.resume()
+            }
+        }
+    }
+
+    /// Defers completion until the next main-actor turn. A newer drag or
+    /// transcript transition invalidates the work via `isCurrent`; otherwise
+    /// persistence runs after the relatch decision.
+    @MainActor
+    static func completeDragAfterNextTurn(
+        suspend: @MainActor () async -> Void = {
+            await ChatFollowLatestRelatchPolicy.waitForNextMainActorTurn()
+        },
+        isCurrent: @MainActor () -> Bool,
+        relatch: @MainActor () -> Void,
+        persist: @MainActor () -> Void
     ) async {
-        await Task.yield()
-        guard isCurrent() else { return }
+        await suspend()
+        guard !Task.isCancelled, isCurrent() else { return }
         relatch()
         persist()
     }
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -634,12 +634,23 @@ struct ChatView: View {
 
     private func updateBottomMarker(_ value: CGFloat?) {
         bottomMarkerMaxY = value
-        if isNearBottom && !hasPendingRestoration { followsLatest = true }
+        relatchFollowsLatestIfSettled()
     }
 
     private func updateViewportBottom(_ value: CGFloat?) {
         scrollViewportMaxY = value
-        if isNearBottom && !hasPendingRestoration { followsLatest = true }
+        relatchFollowsLatestIfSettled()
+    }
+
+    /// Geometry callbacks fire on every content-growth and scroll tick, so
+    /// they must not re-latch while the user's finger is still down: the drag
+    /// sets `followsLatest = false`, but until the finger travels past the
+    /// near-bottom window each streamed delta would yank the scroll back to
+    /// the bottom, fighting the drag.
+    private func relatchFollowsLatestIfSettled() {
+        if isNearBottom && !hasPendingRestoration && !isDraggingChat {
+            followsLatest = true
+        }
     }
 }
 
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -22,8 +22,9 @@ struct ChatView: View {
     @State private var renderedTranscriptRevision: UInt64 = 0
     @State private var renderedViewportTransitionGeneration: UInt64 = 0
     @State private var viewportSnapshotProviderID = UUID()
-    @State private var isDraggingChat = false
-    @State private var chatDragGeneration: UInt64 = 0
+    @GestureState private var isDraggingChat = false
+    @State private var chatDragLifecycle = ChatDragLifecycleState()
+    @State private var chatDragCompletionToken: ChatDragCompletionToken?
     @State private var notificationHandoffPending = false
     @State private var notificationHandoffSessionKey: ChatScrollSessionKey?
     @State private var notificationHandoffHasMeasuredLayout = false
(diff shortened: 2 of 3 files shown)


kaishi00:

[github run](/kaishi00/hermes-conduit/actions/runs/31538429840)

kaishi00:
## Summary
This PR stops the near-bottom geometry callbacks in `ChatView` from re-latching `followsLatest = true` while the user's finger is down (`relatchFollowsLatestIfSettled`), and (as part of its diff surface against the base branch) bundles several previously-stacked fixes: buffered-stream delta deduplication against the resume snapshot, WebKit weak message-handler / WebSocket pending-request cleanup, and bounded `NSCache`-based caching for markdown parsing, syntax highlighting, and local attachment previews.

## Findings

### [severity: major] Buffered-delta deduplication drops events and misplaces/session-tags the uncovered text
**File:** Conduit/Services/AppState.swift:2673-2691
**Issue:** `deduplicatingBufferedEvents` concatenates all delta text, trims the covered prefix, then emits the remainder as a *single* delta at the *last* delta index, dropping every earlier delta. Two correctness problems follow from the documented contract ("Non-delta events pass through", deltas may carry distinct `sessionId`s):
1. If a non-delta event (e.g. `.toolStart`/`.toolComplete`) sits between two deltas and only part of the concatenated text is covered, the uncovered fragment is emitted *after* that event even though part of it originally preceded it — ordering is wrong.
2. The uncovered fragment inherits the *last* delta's `sessionId`, even when the covered prefix came from an earlier delta with a different session. `testRemainderKeepsSessionIdOfBufferedDeltas` only exercises the single-delta case, so the flaw is unguarded.
**Suggestion:** Trim each delta in order instead of merging: walk the events, and for each `.messageDelta` drop up to `covered` characters from its front (dropping the whole delta when fully covered), keeping the tail of *that* delta under *its own* `sessionId`, and pass all non-delta events through untouched. Add a regression test with a partially overlapped first delta, an intervening `.toolStart`, and a later delta with a different `sessionId`.

### [severity: minor] Drag end near the bottom can leave `followsLatest` false with no re-latch tick
**File:** Conduit/Views/ChatView.swift:383-387
**Issue:** `.onEnded` sets `isDraggingChat = false`, but neither `updateBottomMarker` nor `updateViewportBottom` is guaranteed to fire afterward (they only fire when a preference value changes). If the drag ends inside the 40 pt window and the stream is momentarily idle — or the preference pass doesn't run — `followsLatest` stays `false`; the scroll-to-latest button is hidden because `!followsLatest && !isNearBottom` is false, and the first post-drag streamed delta won't auto-scroll because `onChange(of: streamingText)` (line 362) can fire before the geometry relatch. The re-latch then depends on a later content-growth tick, contradicting the PR's claim that "ending a drag inside the near-bottom window still re-latches".
**Suggestion:** After `isDraggingChat = false` in `.onEnded`, schedule a relatch on the next main-actor turn, e.g. `Task { @MainActor in await Task.yield(); guard !isDraggingChat else { return }; relatchFollowsLatestIfSettled() }`, preserving the settled-position checks inside `relatchFollowsLatestIfSettled()`.

### [severity: minor] Decoded-content caches limit entry count, not retained bytes
**File:** Conduit/Views/ChatView.swift:794-806; Conduit/Views/Components/MarkdownText.swift:91-95, 1363-1367
**Issue:** All three caches set only `countLimit`. `localImageCache` can hold up to 64 fully-decoded, full-resolution `UIImage`s (decoded bitmaps are ~W×H×4 bytes, often several MB each); `MarkdownRenderCache` (256) and `SyntaxHighlighter.cache` (128) can each retain many attributed strings. On a long chat this can hold tens-to-hundreds of MB before NSCache pressure eviction kicks in. NSCache will eventually evict under memory warnings, so this is not a leak, but byte-aware limits are the intended control.
**Suggestion:** Set `totalCostLimit` on each cache and insert with explicit costs — `image.size.width * image.size.height * 4` for previews, source + attributed-text length for markdown renderings, source length for highlighted code.

### [severity: nit] Markdown render cache key omits the color scheme
**File:** Conduit/Views/Components/MarkdownText.swift:105-111
**Issue:** The cache key uses `String(describing: foregroundStyle)`, which is identical across light/dark for semantic colors (e.g. `.primary`). The cached `selectableText` bakes in concrete `UIColor`s (`UIColor(foregroundStyle)`, `.link`, `.white`). After a light/dark flip, the cache hit serves the previous scheme's colors until the source text changes, whereas the pre-cache code rebuilt the string every body evaluation and stayed correct.
**Suggestion:** Include the current trait variation in the key (`UITraitCollection.current.userInterfaceStyle` or an `@Environment(\.colorScheme)` value), or store trait-providing dynamic colors instead of resolved `UIColor`s.

### [severity: nit] Local image cache is keyed only by path
**File:** Conduit/Views/ChatView.swift:800-807
**Issue:** `localImageCache` keys on `url.path` only. If a local file at the same path is replaced (e.g. an attachment re-encoded/overwritten between sessions), the preview serves the stale decoded image indefinitely. This is unlikely given unique attachment paths, but there is no invalidation hook.
**Suggestion:** Accept a cache key that also incorporates file modification time/size (e.g. `path + "-" + mtime`), or accept the staleness explicitly with a comment.

## Verdict
REQUEST_CHANGES

The drag-fix itself (`relatchFollowsLatestIfSettled`) is correct and well-documented, but the bundled `deduplicatingBufferedEvents` has a genuine data-integrity flaw (event ordering / `sessionId` attribution for partially-covered, interleaved deltas) that should be addressed before merge, along with the minor drag-end re-latch gap.

<a href="https://opencode.ai/s/rh5Jl53M">New%20session%20-%202026-08-12T01%3A58%3A29.376Z</a>
[opencode session](https://opencode.ai/s/rh5Jl53M)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31555303743)

kaishi00:
## Summary
This PR stops scroll-geometry preference callbacks from re-latching `followsLatest = true` while the user is dragging the transcript, defers the drag-end decision by one main-actor turn so ending near the bottom still re-enables following, and invalidates stale deferred work via a drag/session generation token. It also extracts the logic into a testable `ChatFollowLatestRelatchPolicy`.

## Findings

### [severity: minor] Stuck `isDraggingChat` permanently blocks geometry re-latching if `.onEnded` is missed
**File:** Conduit/Views/ChatView.swift:664-679 (and 386-406)
**Issue:** `relatchFollowsLatestIfSettled()` now gates on `!isDraggingChat`, and `isDraggingChat` is only ever reset in `.onEnded`. Before this PR, geometry callbacks relatched unconditionally (`isNearBottom && !hasPendingRestoration`), so a missed `.onEnded` was harmless. Now a drag interrupted by a view teardown, system gesture, or session switch (none of the `.onChange(of: activeSessionId/isOpeningNotificationSession)` handlers reset `isDraggingChat`) leaves the flag `true` forever, and no geometry callback can ever re-latch — the transcript silently stops following new deltas until the user completes another full drag.
**Suggestion:** Reset `isDraggingChat = false` in `.onDisappear` and in the session-transition/notification-handoff change handlers, or drop the `isDragging` gate from the geometry path and rely solely on `chatDragGeneration`/turn-based invalidation for mid-drag suppression.

### [severity: minor] Unstructured `Task` in `.onEnded` is never cancelled and runs against a possibly-torn-down view
**File:** Conduit/Views/ChatView.swift:390-405
**Issue:** `Task { @MainActor in ... }` is unstructured: it is not cancelled on `onDisappear` and has no lifecycle tie to the view. If the view disappears between drag-end and the yield, the task still runs `relatchFollowsLatestIfSettled()` and `saveChatScrollPosition`, mutating `@State` and persisting a viewport for a view that is no longer installed. The `isCurrent` guard checks drag/session/notification state but not view identity, so a stale instance's task can relatch/persist after a replacement `ChatView` is already on screen.
**Suggestion:** Store the task and cancel it in `.onDisappear`, and/or add an `onDisappear`-scoped flag (e.g. compare `viewportSnapshotProviderID` or a view-generation token) to `isCurrent`.

### [severity: minor] `isCurrent` compares session keys with `==` instead of identity equivalence
**File:** Conduit/Views/ChatView.swift:395
**Issue:** `(renderedScrollSessionKey ?? activeScrollSessionKey) == completedSessionKey` uses struct equality. During reconciliation the coordinator can remap a runtime ID to a catalog canonical ID (`ChatScrollSessionIdentity` equivalence). If that remap lands between drag-end and the deferred run, the guard drops the persist even though it is the same logical session, so the drag-end viewport is not saved.
**Suggestion:** Compare with `appState.activeChatScrollSessionIdentity.areEquivalent(...)` as the other handlers in this view already do.

### [severity: nit] Nonisolated closure parameters on a `@MainActor` static method
**File:** Conduit/Services/ChatScrollState.swift:54-64
**Issue:** `completeDragAfterYield(isCurrent:relatch:persist:)` is `@MainActor` but takes nonisolated `() -> Bool`/`() -> Void` closures. Callers pass closures that capture `@State`/`appState` (main-actor state). This compiles under the project's Swift 5.9 mode but will fail under Swift 6 strict concurrency ("sending main actor-isolated value ... risks causing data races").
**Suggestion:** Declare the closure parameters `@MainActor` (`isCurrent: @MainActor () -> Bool`, etc.) to make the actor contract explicit and future-proof the code.

### [severity: nit] Deferred relatch duplicates work and its test depends on main-actor FIFO scheduling
**File:** ConduitTests/ChatScrollStateTests.swift:33-49
**Issue:** `testDragCompletionRechecksGestureAfterYield` is correct and the yield mutation claim holds, but it relies on the `nextDrag` task running before the current task resumes after `Task.yield()`. That is deterministic on the main actor today but is scheduling-order-sensitive; `nextDrag`'s value is never asserted. Minor: the `relatch()` call is redundant when a geometry tick already re-latched in the gap between `.onEnded` and the deferred run (idempotent, harmless).
**Suggestion:** No change strictly required; optionally assert `await nextDrag.value` ordering by recording events and documenting the FIFO dependency, and note the idempotency of the extra relatch.

## Verdict
APPROVE

The core fix is correct and well-reasoned: the `!isDraggingChat` gate in `relatchFollowsLatestIfSettled` closes the mid-drag relatch race, and the one-turn deferred `relatch → persist` ordering ensures the saved viewport matches the final follow state. The `isDraggingChat`/session-change resilience concerns are edge cases worth hardening but not blockers; the added tests are meaningful and the mutation check (removing `Task.yield()` fails the test) validates the design.

<a href="https://opencode.ai/s/3H1Lxnr3">New%20session%20-%202026-08-13T02%3A02%3A49.081Z</a>
[opencode session](https://opencode.ai/s/3H1Lxnr3)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31659550060)

kaishi00:
## Summary
This PR stops the near-bottom geometry preference callbacks in `ChatView` from re-latching `followsLatest = true` while the user's finger is down, defers the drag-end relatch/persist decision to the next main-actor turn via a generation/session/viewport token, and extracts the logic into a testable `ChatFollowLatestRelatchPolicy` + `ChatDragLifecycleState` with 413 lines of focused tests.

## Findings

### [severity: minor] Mid-drag transitions re-enable stream-following for the remainder of the in-flight gesture
**File:** Conduit/Views/ChatView.swift:512-515 (`invalidateChatDrag`), callers at 291, 298, 318, 330, 368
**Issue:** When a session switch, profile change, notification handoff, or scroll request lands *during* an active drag, `invalidateChatDrag()` sets `activeGestureInvalidated` (ChatScrollState.swift:63-68), which permanently blocks `begin()` for the rest of that gesture. The transition handlers meanwhile set `followsLatest = true` and several call `scrollToLatest` (lines 293, 326, 353). So the remainder of the drag is back to fighting auto-scroll — the exact race this PR eliminates for the ordinary case. The `!isDraggingChat` gate stops geometry from re-latching, but the handlers already re-latched explicitly.
**Suggestion:** When `invalidate(hasActiveGesture: true)` runs, also re-apply the drag opt-out (set `followsLatest = false`), so the ongoing gesture keeps its opt-out until it ends, or let the next `.onChanged` `begin()` succeed after invalidation.

### [severity: minor] Drag that began with an unresolved session can drop its deferred completion
**File:** Conduit/Services/ChatScrollState.swift:118-119; Conduit/Views/ChatView.swift:471-473
**Issue:** `begin()` captures `renderedScrollSessionKey ?? activeScrollSessionKey` at drag start. If a drag starts on a brand-new/unresolved session (key is `nil`) and the session is established before the gesture ends, `isCompletionCurrent` computes `sameSession = (nil == key)` → false and `areEquivalent(nil, key)` → false, so the deferred completion is dropped. `followsLatest` then stays `false` while `isNearBottom` is true — the scroll-to-latest button is suppressed (`!followsLatest && !isNearBottom`), and the transcript stops following until a later content-growth geometry tick relatches. The `nil == nil` branch (line 118) also bypasses identity normalization entirely, though that path is harmless.
**Suggestion:** Treat a completed nil-session token as current if the current key is not yet resolved (or resolve the completed key against the current identity before comparing), so the near-bottom drag on an in-flight session still relatches.

### [severity: nit] Generation counter uses wrapping arithmetic
**File:** Conduit/Services/ChatScrollState.swift:54, 64, 71
**Issue:** `generation &+= 1` wraps after 2^64 gestures, after which a stale token could theoretically pass the generation comparison in `isCompletionCurrent`. Purely theoretical.
**Suggestion:** Accept, or use `generation &+= 1` with a documented overflow rationale.

### [severity: nit] "Next main-actor turn" relies on main queue == main actor executor
**File:** Conduit/Services/ChatScrollState.swift:139-146
**Issue:** `waitForNextMainActorTurn` resumes a `withCheckedContinuation` from a `DispatchQueue.main.async` block, which assumes the main dispatch queue is the main actor's executor (true in the standard runtime). `testDragCompletionWaitsForNextMainActorTurn` asserts FIFO ordering of two queued blocks — deterministic on the main queue today but not a language guarantee, and if the executor configuration ever changes the deferral silently becomes a raw thread hop.
**Suggestion:** This is acceptable given the injectable `suspend` closure for tests; consider documenting the main-queue/main-actor coupling, or deferring via a `Task`-based hop if strictness is desired.

### [severity: nit] Trailing guard in `completeChatDrag` is effectively dead code
**File:** Conduit/Views/ChatView.swift:500-502
**Issue:** After `completeDragAfterNextTurn` returns, `guard !Task.isCancelled, chatDragCompletionToken == completed` runs with no suspension points in between on the main actor, so the equality check always passes when the task isn't cancelled. Harmless but misleading.
**Suggestion:** Keep it as defense-in-depth (it guards against future refactors introducing a suspension), or drop the equality check and clear unconditionally.

## Verdict
COMMENT

The core fix is correct and well-hardened: moving `isDraggingChat` to `@GestureState` guarantees it clears on gesture end/cancel (resolving the earlier "stuck flag" concern), the `isDragging` parameter in `isCompletionCurrent` closes the window between a new gesture's recognition and its first `.onChanged`, and the generation/session/viewport token plus cooperative `.task(id:)` cancellation make stale completions reliably inert. The relatch-before-persist ordering and canonical-key persistence are sound, and the tests (including the controllable `ChatScrollTestGate` suspension) meaningfully validate the design. The two minor findings are edge cases worth a follow-up but not blockers.

<a href="https://opencode.ai/s/R7PpF60G">New%20session%20-%202026-08-13T02%3A42%3A54.325Z</a>
[opencode session](https://opencode.ai/s/R7PpF60G)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31661590885)

kaishi00:
## Summary

This PR prevents the near-bottom scroll-geometry callbacks in `ChatView` from re-latching `followsLatest = true` while the user is dragging the transcript (`@GestureState` gate + `ChatFollowLatestRelatchPolicy`), defers the drag-end relatch/persist decision by one main-actor turn through a generation/session/viewport token (`ChatDragLifecycleState`), and adds 446 lines of focused tests covering the gesture lifecycles and ordering guarantees.

## Findings

### [severity: minor] `chatScrollRequest` handler re-enables follow-latest and scrolls even mid-drag
**File:** Conduit/Views/ChatView.swift:290-295
**Issue:** Unlike the session/profile/notification transitions, which now route through `shouldFollowLatestAfterTransition(isDragging:)`, the `chatScrollRequest` handler sets `followsLatest = true` and calls `scrollToLatest` unconditionally after `invalidateChatDrag()`. If a scroll request ever lands while a gesture is active, the remainder of that in-flight drag reverts to fighting auto-scroll — the exact race this PR eliminates elsewhere. In practice `requestChatScrollToLatest()` is only invoked from `sendMessage` (AppState.swift:4247), when the finger is on the Send button rather than the transcript, so impact is low, but the handler is inconsistent with the rest of the change.
**Suggestion:** Guard with the same policy, e.g. `if ChatFollowLatestRelatchPolicy.shouldFollowLatestAfterTransition(isDragging: isDraggingChat) { followsLatest = true; scrollToLatest(using: proxy) }`.

### [severity: minor] "Next main-actor turn" is implemented via `DispatchQueue.main` and assumes it is the main actor executor
**File:** Conduit/Services/ChatScrollState.swift:147-154
**Issue:** `waitForNextMainActorTurn` resumes a `withCheckedContinuation` from a `DispatchQueue.main.async` block. This only defers to the *main queue*, and correctness of the drag-end relatch (dropping stale completions before they run) leans on the main dispatch queue being the main actor's executor. `testDragCompletionWaitsForNextMainActorTurn` (ChatScrollStateTests.swift:420-436) additionally asserts FIFO ordering of two queued blocks, which is deterministic on the main queue today but is not a language guarantee. If the executor configuration ever diverges, the "deferral" silently becomes a raw thread hop.
**Suggestion:** Acceptable given the injectable `suspend` closure; consider documenting the main-queue/main-actor coupling in a comment, or hop via `await Task.yield()` on the main actor when strictness is desired.

### [severity: nit] Dropped completion leaves a stale `chatDragCompletionToken` behind
**File:** Conduit/Views/ChatView.swift:488-511
**Issue:** When `completeDragAfterNextTurn` returns early because `isCurrent()` is false (and no invalidation/`begin` has run to nil the token — e.g. a viewport-transition generation bump via `beginExplicitChatViewportTransition` that isn't accompanied by an `activeSessionId` change), the trailing guard returns without clearing the token. The stale token then sits until the next drag's `begin` nils it. It is inert (only `.task(id:)` reads it, and the id no longer changes), so this is defensive, but the invariant "token is nil once a completion has been decided" is not maintained.
**Suggestion:** Clear the token unconditionally after the decision, e.g. wrap the call and reset in a `defer { chatDragCompletionToken = nil }`-style cleanup, or add a comment that a stale token is intentionally harmless until the next drag.

### [severity: nit] Trailing guard in `completeChatDrag` is effectively dead code
**File:** Conduit/Views/ChatView.swift:508-510
**Issue:** After `completeDragAfterNextTurn` returns there is no suspension point before `guard !Task.isCancelled, chatDragCompletionToken == completed`, so on the main actor the equality always holds whenever the task wasn't cancelled. Harmless but misleading.
**Suggestion:** Keep as defense-in-depth against a future refactor introducing a suspension, or drop the equality check and clear unconditionally.

### [severity: nit] Generation counter wraps with `&+=`
**File:** Conduit/Services/ChatScrollState.swift:54, 64, 71
**Issue:** After 2^64 drags a stale token's generation could wrap and pass `isCompletionCurrent`'s equality check. Purely theoretical.
**Suggestion:** Accept as-is; the wrapping is consistent with the rest of the codebase's use of `&+=` for revision counters.

## Verdict

APPROVE

The core fix is correct and well-hardened: `@GestureState` guarantees `isDraggingChat` clears on gesture end/cancel, the `!isDragging` gate in `relatchFollowsLatestIfSettled` closes the mid-drag relatch race before `begin` can even run, the generation/session/viewport token plus `.task(id:)`-driven cooperative cancellation make stale completions reliably inert (including the `onDisappear` → `abandon` path), and the relatch-before-persist ordering with canonical-key resolution is sound. The tests meaningfully validate the design, including the controllable `ChatScrollTestGate` suspension that replaced the earlier `Task.yield()`-ordering dependency. The findings above are edge-case hardening items, not blockers.

<a href="https://opencode.ai/s/15KFqtra">New%20session%20-%202026-08-13T02%3A54%3A47.761Z</a>
[opencode session](https://opencode.ai/s/15KFqtra)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31662194762)

kaishi00 (approved):
Reviewed the repaired three-file diff and its lifecycle hardening. Verified 501 local tests plus the required Build & Test, CodeRabbit, and automated review checks on head dbe0482.

PR #49, opened 2026-08-12
docs: correct DashboardTicketBridge.deinit reachability comment

kaishi00:
Follow-up to #42 (merged). The `deinit` pending-request loop shipped with a comment claiming deallocation can happen "mid-request" and leave continuations hanging. A re-review correctly noted that premise is wrong: a non-empty `pendingRequests` implies a `requestJSON` coroutine is suspended awaiting its continuation, and that suspended task retains `self` — so the bridge cannot deallocate until all requests resolve and the dictionary drains. The loop is a no-op today.

This PR keeps the loop as a defensive guard (the retain-cycle fix did make deallocation reachable in general, and a hang is the worst-case failure if a future refactor breaks that invariant) but corrects the comment to state the real invariant and why the guard is worth keeping.

**Comment-only — no behavior change.**

## Summary by CodeRabbit

* **Documentation**
  * Clarified teardown behavior for pending requests and defensive request completion.
  * No user-facing behavior or functionality changed.

### DIFF
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -189,14 +189,17 @@ final class DashboardTicketBridge: NSObject {
     }
 
     deinit {
-        // The retain-cycle fix is what makes deallocation reachable at all, so
-        // `deinit` can now run while requests are still pending. Removing the
-        // message handler alone leaves their continuations awaiting a JS reply
-        // that can never arrive: the proxy's weak delegate is already nil and
-        // silently drops the posted message. Resume them now so callers see
-        // `.notReady` instead of hanging for a response that will never come.
-        // (The dictionary is torn down with the bridge, so only the resumptions
-        // matter here — nothing else can resume them once the bridge is gone.)
+        // Defensive teardown. The retain-cycle fix is what makes bridge
+        // deallocation reachable at all. In practice this loop is a no-op
+        // today: a non-empty `pendingRequests` implies a `requestJSON`
+        // coroutine is suspended awaiting its continuation, and that
+        // suspended task retains `self`, so the bridge cannot deallocate
+        // until every request resolves and the dictionary drains. It is kept
+        // as a guard — if a future refactor breaks that invariant (e.g. the
+        // request task is detached, or continuations are stored somewhere
+        // `self` no longer keeps alive), resuming here is what stops callers
+        // from hanging on a reply the now-nil proxy would silently drop.
+        // Nothing else can resume these once the bridge is gone.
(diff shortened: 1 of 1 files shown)


kaishi00:
## Summary
This PR rewrites the comment in `DashboardTicketBridge.deinit` to correct an earlier, wrong claim that deallocation could occur mid-request; the new comment documents that a non-empty `pendingRequests` implies a suspended `requestJSON` task that retains `self`, making the teardown loop a defensive no-op today. Comment-only, no behavior change.

## Findings

### [nit] Guard is silent — consider asserting when the defensive path actually triggers
**File:** Conduit/Services/DashboardTicketBridge.swift:191-207
**Issue:** The comment now correctly argues the loop is a no-op under the current invariant, which is verified: every resumption path (`didReceive`, `evaluateJavaScript` completion, `rejectPending`, `cancelPendingRequest`) removes the continuation from `pendingRequests` before resuming it, and while any continuation is stored, `self` is retained via continuation → suspended task → operation closure, so deinit cannot run with a non-empty dictionary. The guard therefore only matters for a future refactor — but nothing will observe when it fires, so an invariant break would surface only as a silent divergence from the documented reasoning.
**Suggestion:** Since the loop is intended purely as a tripwire, make it observable, e.g. `assert(pendingRequests.isEmpty, "bridge deallocated with pending requests")` before the loop (or a log when the loop finds entries), so a future refactor that breaks the invariant fails loudly rather than relying on reading this comment. Purely optional given the PR's comment-only scope.

### [nit] Terminology: "coroutine" is not idiomatic Swift
**File:** Conduit/Services/DashboardTicketBridge.swift:194-197
**Issue:** Swift uses "task" / "async function", not "coroutine"; the term reads slightly oddly to a Swift audience and the comment refers to the same construct as a "task" later in the same block.
**Suggestion:** Rephrase "a `requestJSON` coroutine is suspended awaiting its continuation" → "a `requestJSON` call site is suspended awaiting its continuation".

## Verdict
APPROVE

The technical claim in the rewritten comment is accurate and matches the actual code paths (`pendingRequests[id] = continuation` at line 280, removal before every resume in lines 340, 351-353, 372-375, 448-462; retention of `self` by the operation closure at lines 274-342). No retain-cycle, force-unwrap, concurrency, or lifecycle issues are introduced by a comment-only change.

<a href="https://opencode.ai/s/deExG6re">New%20session%20-%202026-08-12T15%3A35%3A42.986Z</a>
[opencode session](https://opencode.ai/s/deExG6re)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31613032574)

PR #50, opened 2026-08-12
Cache markdown parsing, syntax highlighting, and attachment previews

kaishi00:
## Origin

Supersedes the net-new portion of #44 by @angel12. #44 bundled three commits that were already merged to `main` (regex backtracking fix, RPC-fail-on-socket-death, WebKit weak-handler proxy) plus one genuinely new caching commit; that duplication left it conflicting and not up to date. This PR cherry-picks **only** the new caching commit (`473eee3a`, attribution preserved — @angel12 + Claude) onto current `main`, then folds in the review findings in a second commit.

## What it does

Eliminates the per-frame O(visible transcript) re-parse/rebuild that pegged the CPU while a reply streamed. Three `NSCache` layers, all evicted under memory pressure:

- **`MarkdownRenderCache`** — parsed blocks + prebuilt selectable `NSAttributedString`, keyed by (source, style, gateway-media flag, Dynamic Type size). Settled messages render from cache; only genuinely new content pays for parsing. Streaming-tail fade opacities are applied to a *copy* of the cached base each frame instead of forcing a rebuild.
- **`SyntaxHighlighter`** — tokenized code memoized by (language, source). Settled code blocks across the transcript stop re-tokenizing at frame rate.
- **`UserImageAttachmentPreview`** — decoded local attachment previews cached by path instead of re-reading the file every render.

## Review fixes applied (second commit)

1. **Unstable cache key.** Dropped `String(describing: foregroundStyle)` from the render-cache key. Only `.primary` / `.white` are ever passed, each uniquely identified by `usesAccentSurface`, so keying on that is stable — whereas a SwiftUI `Color`'s `description` is undocumented and can drift for adaptive colors.
2. **Transient streaming snapshots.** Added `MarkdownText.isStreaming` (set only by `StreamingText`) and skip the render-cache write while streaming: a streaming message's whole-message `source` changes every frame, so a cached entry is dead on insertion and only evicts reusable settled entries. The message is cached on its first settled render instead.
   - **Deliberate non-change:** the per-block `SyntaxHighlighter` cache still writes while streaming. Only its *single growing block* is transient; stable blocks (earlier in the message and across the transcript) have a constant key and hit every frame, so skipping writes there would discard that benefit. The distinction (whole-message key = transient; per-block key = mostly stable) is documented in a comment.
3. **Unbounded decoded-image memory.** `localImageCache` is now cost-bounded (`totalCostLimit = 64 MB`) and attachments are decoded to the preview's retina size via an ImageIO thumbnail (`CGImageSourceCreateThumbnailAtIndex`) instead of caching full-resolution bitmaps; pixel cost is passed to `setObject(_:forKey:cost:)`.

## Note on the HermesClient socket finding

Both bot reviews on #44 flagged a stale-receive-loop / dead-socket issue in `HermesClient.swift` as blocking. That code is **already on `main`** (from an earlier PR) and is not touched here. It's a real but **latent** defect — `AppState` creates a new `HermesClient` per reconnect and calls `previousClient.disconnect()` first, so the cross-instance clobber can't occur in production today; it only bites on intra-instance double-`connect()`. It will be addressed in a separate PR.

## Testing

Existing `ChatTextSelectionTests` (parser/formatter) are unaffected — caching wraps the same pure functions with identical outputs. Parser/formatter/highlighter behavior is unchanged; only caching, the streaming skip-write, and image downsampling are new. Verifying via CI (macos-26).

## Summary by CodeRabbit

  * Improved image attachment loading with faster, appropriately sized previews and smoother loading states.
  * Added caching for image previews, Markdown rendering, and syntax highlighting to improve display speed.
  * Improved streaming text updates while preserving visual effects during generation.
  * Reduced unnecessary rendering during streamed content updates.

* **Bug Fixes**
  * Improved error reporting by distinguishing local image preview failures from gateway loading failures.

### DIFF
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -777,9 +777,22 @@ private struct UserImageAttachmentPreview: View {
     @State private var gatewayImage: UIImage?
     @State private var gatewayLoadFailed = false
 
+    /// Body re-evaluates at streaming frame rate, and re-reading the file
+    /// from disk each time hitches scrolling. Decoded previews are shared
+    /// across rows; NSCache evicts them under memory pressure.
+    private static let localImageCache: NSCache<NSString, UIImage> = {
+        let cache = NSCache<NSString, UIImage>()
+        cache.countLimit = 64
+        return cache
+    }()
+
     private var localImage: UIImage? {
         guard let url = localFileURL else { return nil }
-        return UIImage(contentsOfFile: url.path)
+        let key = url.path as NSString
+        if let cached = Self.localImageCache.object(forKey: key) { return cached }
+        guard let image = UIImage(contentsOfFile: url.path) else { return nil }
+        Self.localImageCache.setObject(image, forKey: key)
+        return image
     }
 
     private var localFileURL: URL? {
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -7,6 +7,7 @@
 
 import SwiftUI
 import UIKit
+import ImageIO
 
 struct ChatView: View {
     @EnvironmentObject var appState: AppState
@@ -777,12 +778,15 @@ private struct UserImageAttachmentPreview: View {
     @State private var gatewayImage: UIImage?
     @State private var gatewayLoadFailed = false
 
-    /// Body re-evaluates at streaming frame rate, and re-reading the file
-    /// from disk each time hitches scrolling. Decoded previews are shared
-    /// across rows; NSCache evicts them under memory pressure.
+    /// Body re-evaluates at streaming frame rate, and re-reading + fully
+    /// decoding the file each time hitches scrolling. Previews are downsampled
+    /// to the display size and shared across rows; NSCache evicts them under
+    /// memory pressure. Cost-bounded so a handful of large photos can't
+    /// exhaust it before then.
     private static let localImageCache: NSCache<NSString, UIImage> = {
         let cache = NSCache<NSString, UIImage>()
         cache.countLimit = 64
+        cache.totalCostLimit = 64 * 1024 * 1024
         return cache
     }()
 
@@ -790,9 +794,29 @@ private struct UserImageAttachmentPreview: View {
         guard let url = localFileURL else { return nil }
         let key = url.path as NSString
         if let cached = Self.localImageCache.object(forKey: key) { return cached }
-        guard let image = UIImage(contentsOfFile: url.path) else { return nil }
-        Self.localImageCache.setObject(image, forKey: key)
+        // Decode at the preview's retina resolution instead of caching the
+        // full-resolution bitmap — a library photo can decode to tens of MB.
+        guard let image = Self.downsampledPreview(url: url) else { return nil }
+        let pixelWidth = image.cgImage?.width ?? Int(image.size.width * image.scale)
+        let pixelHeight = image.cgImage?.height ?? Int(image.size.height * image.scale)
+        Self.localImageCache.setObject(image, forKey: key, cost: pixelWidth * pixelHeight * 4)
         return image
+    }
+
+    /// Decodes `url` as a thumbnail sized for the 224pt attachment preview
+    /// (~3× retina on the long edge). ImageIO decodes straight to the smaller
+    /// bitmap, so the full-resolution pixels never need to be resident.
+    private static func downsampledPreview(url: URL) -> UIImage? {
+        let options: [CFString: Any] = [
+            kCGImageSourceCreateThumbnailFromImageAlways: true,
+            kCGImageSourceShouldCache: true,
+            kCGImageSourceCreateThumbnailWithTransform: true,
+            kCGImageSourceThumbnailMaxPixelSize: CGFloat(800)
+        ]
+        guard let source = CGImageSourceCreateWithURL(url as CFURL, nil),
+              let thumbnail = CGImageSourceCreateThumbnailAtIndex(source, 0, options as CFDictionary)
+        else { return nil }
+        return UIImage(cgImage: thumbnail)
     }
 
     private var localFileURL: URL? {
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -778,7 +778,8 @@ private struct UserImageAttachmentPreview: View {
     @State private var gatewayImage: UIImage?
     @State private var gatewayLoadFailed = false
     @State private var localPreview: UIImage?
-    @State private var localPreviewFailed = false
+    @State private var localPreviewPath: String?
+    @State private var localPreviewFailedPath: String?
 
     /// Body re-evaluates at streaming frame rate, and re-reading + fully
     /// decoding the file each time hitches scrolling. Previews are downsampled
@@ -799,6 +800,21 @@ private struct UserImageAttachmentPreview: View {
         return Self.localImageCache.object(forKey: url.path as NSString)
     }
 
+    /// The decoded preview, gated on the path it was decoded for so a stale
+    /// value can't leak in if SwiftUI ever reuses this view identity for a
+    /// different `attachment.uri`. Today that can't happen — `Attachment` is
+    /// `Identifiable` so each id gets fresh `@State`, and uris are immutable —
+    /// but this makes the invariant non-load-bearing rather than relying on it.
+    private var currentLocalPreview: UIImage? {
+        guard localPreviewPath == localFileURL?.path else { return nil }
+        return localPreview
+    }
+
+    /// True only for a decode failure of the *current* path.
+    private var localPreviewFailed: Bool {
+        localPreviewFailedPath == localFileURL?.path
+    }
+
     /// Estimated decoded byte cost for NSCache. `bytesPerRow * height` accounts
     /// for the row-alignment padding that a naive `width * height * 4` misses.
     private static func bitmapCost(of image: UIImage) -> Int {
@@ -809,12 +825,16 @@ private struct UserImageAttachmentPreview: View {
     }
 
     /// Decodes `url` as a thumbnail sized for the 224pt attachment preview
-    /// (~3× retina on the long edge). Pure CoreGraphics, so it is safe to call
-    /// off the MainActor — which is how the `.task` below uses it, so a large
-    /// photo can't hitch scrolling on a cache miss. `ShouldCache` is false
-    /// because the resulting `UIImage` is already retained by our cache; the
-    /// default `true` would double-buffer the decoded bitmap inside ImageIO.
-    private static func downsampledPreview(url: URL) -> UIImage? {
+    /// (~3× retina on the long edge). `nonisolated` makes explicit that this is
+    /// pure CoreGraphics and safe to call off the MainActor — which is how the
+    /// `.task` below invokes it, so a large photo can't hitch scrolling on a
+    /// cache miss. `ShouldCache` is false because the resulting `UIImage` is
+    /// already retained by our cache; the default `true` would double-buffer
+    /// the decoded bitmap inside ImageIO. Note: the detached caller returns a
+    /// non-Sendable `UIImage` across the task boundary — fine under the
+    /// project's Swift 5 mode, but it would need a Sendable box if strict
+    /// concurrency is ever enabled.
+    private nonisolated static func downsampledPreview(url: URL) -> UIImage? {
         let options: [CFString: Any] = [
             kCGImageSourceCreateThumbnailFromImageAlways: true,
             kCGImageSourceShouldCache: false,
@@ -835,9 +855,9 @@ private struct UserImageAttachmentPreview: View {
 
     var body: some View {
         Group {
-            if let image = cachedLocalImage ?? localPreview ?? gatewayImage {
+            if let image = cachedLocalImage ?? currentLocalPreview ?? gatewayImage {
                 imagePreview(image)
-            } else if isGatewayImage && !gatewayLoadFailed {
+            } else if isGatewayImage && localFileURL == nil && !gatewayLoadFailed {
                 loadingPlaceholder
             } else if localFileURL != nil && !localPreviewFailed {
                 // Local file decoding off-main on a cache miss.
@@ -859,17 +879,18 @@ private struct UserImageAttachmentPreview: View {
             // frame via `cachedLocalImage`, so only misses reach here. Decode
             // off the MainActor so a large photo can't hitch scrolling.
             if let url = localFileURL {
-                guard localPreview == nil, cachedLocalImage == nil else { return }
+                guard currentLocalPreview == nil, cachedLocalImage == nil else { return }
                 let image = await Task.detached(priority: .userInitiated) { () -> UIImage? in
                     Self.downsampledPreview(url: url)
                 }.value
                 guard !Task.isCancelled else { return }
                 guard let image else {
-                    localPreviewFailed = true
+                    localPreviewFailedPath = url.path
                     return
                 }
                 Self.localImageCache.setObject(image, forKey: url.path as NSString, cost: Self.bitmapCost(of: image))
                 localPreview = image
+                localPreviewPath = url.path
                 // Gateway image, fetched through Hermes. The local-file branch
                 // returns above, so a local URI never reaches here — which also
                 // removes the old `localImage == nil` check that decoded local
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -777,6 +777,8 @@ private struct UserImageAttachmentPreview: View {
     @EnvironmentObject private var appState: AppState
     @State private var gatewayImage: UIImage?
     @State private var gatewayLoadFailed = false
+    @State private var localPreview: UIImage?
+    @State private var localPreviewFailed = false
 
     /// Body re-evaluates at streaming frame rate, and re-reading + fully
     /// decoding the file each time hitches scrolling. Previews are downsampled
@@ -790,26 +792,32 @@ private struct UserImageAttachmentPreview: View {
         return cache
     }()
 
-    private var localImage: UIImage? {
+    /// Cache-only lookup (no decode) so a hit renders on the first frame.
+    /// Misses are filled asynchronously — see the `.task` in `body`.
+    private var cachedLocalImage: UIImage? {
         guard let url = localFileURL else { return nil }
-        let key = url.path as NSString
-        if let cached = Self.localImageCache.object(forKey: key) { return cached }
-        // Decode at the preview's retina resolution instead of caching the
-        // full-resolution bitmap — a library photo can decode to tens of MB.
-        guard let image = Self.downsampledPreview(url: url) else { return nil }
-        let pixelWidth = image.cgImage?.width ?? Int(image.size.width * image.scale)
-        let pixelHeight = image.cgImage?.height ?? Int(image.size.height * image.scale)
-        Self.localImageCache.setObject(image, forKey: key, cost: pixelWidth * pixelHeight * 4)
-        return image
+        return Self.localImageCache.object(forKey: url.path as NSString)
+    }
+
+    /// Estimated decoded byte cost for NSCache. `bytesPerRow * height` accounts
+    /// for the row-alignment padding that a naive `width * height * 4` misses.
+    private static func bitmapCost(of image: UIImage) -> Int {
+        if let cg = image.cgImage {
+            return cg.bytesPerRow * cg.height
+        }
+        return Int(image.size.width * image.scale * image.size.height * image.scale) * 4
     }
 
     /// Decodes `url` as a thumbnail sized for the 224pt attachment preview
-    /// (~3× retina on the long edge). ImageIO decodes straight to the smaller
-    /// bitmap, so the full-resolution pixels never need to be resident.
+    /// (~3× retina on the long edge). Pure CoreGraphics, so it is safe to call
+    /// off the MainActor — which is how the `.task` below uses it, so a large
+    /// photo can't hitch scrolling on a cache miss. `ShouldCache` is false
+    /// because the resulting `UIImage` is already retained by our cache; the
+    /// default `true` would double-buffer the decoded bitmap inside ImageIO.
     private static func downsampledPreview(url: URL) -> UIImage? {
         let options: [CFString: Any] = [
             kCGImageSourceCreateThumbnailFromImageAlways: true,
-            kCGImageSourceShouldCache: true,
+            kCGImageSourceShouldCache: false,
             kCGImageSourceCreateThumbnailWithTransform: true,
             kCGImageSourceThumbnailMaxPixelSize: CGFloat(800)
         ]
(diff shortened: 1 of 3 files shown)


kaishi00:
## Summary
Adds three `NSCache` layers to stop per-frame re-parsing/re-rendering of settled markdown, code highlighting, and attachment previews during streaming, with cost-bounded image downsampling via ImageIO and a documented skip-write for transient streaming snapshots.

## Findings

### [major] Synchronous full-image decode on the main thread on cache miss
**File:** Conduit/Views/ChatView.swift:793-820

**Issue:** `localImage` runs `downsampledPreview` synchronously in `body` (MainActor) whenever the cache misses. The options force a full decode — `kCGImageSourceCreateThumbnailFromImageAlways: true` means ImageIO decodes the *entire* source image (a 48MP library photo can take hundreds of ms) before producing the 800px thumbnail. This hitches scrolling/layout and runs during streaming frame-rate body re-evaluations. `.task(id:)` at line 855 also triggers this decode on the main thread via its `guard localImage == nil` (tasks in `body` inherit MainActor isolation). The previous `UIImage(contentsOfFile:)` was also synchronous, but this PR explicitly targets this exact hitch, so the decode should be moved off the main thread.

**Suggestion:** Load the preview off-main: keep a `@State` for the preview + loading/failed state, and in `.task` perform `downsampledPreview` inside `Task.detached` / a nonisolated helper, then publish the `UIImage` back on the MainActor (cache the result in `localImageCache`). If an embedded thumbnail suffices, prefer `kCGImageSourceCreateThumbnailFromImageIfAbsent` (or use `kCGImageSourceCreateThumbnailWithTransform` with `FromImageAlways` only for small sources) to avoid forcing a full-resolution decode.

### [minor] `.task` guard decodes local images it will never load
**File:** Conduit/Views/ChatView.swift:856

**Issue:** `guard localImage == nil, isGatewayImage` evaluates `localImage` first, so for local (non-gateway) attachments the task synchronously decodes the file on the main thread and then returns early — pure wasted work, and the same decode already happens in `body`.

**Suggestion:** Reorder to `guard isGatewayImage, localImage == nil` so the decode is never performed by the task for local attachments (and disappears entirely once the load moves off-main per the finding above).

### [minor] `localImageCache` cost under-counts real bitmap memory
**File:** Conduit/Views/ChatView.swift:800-802

**Issue:** `pixelWidth * pixelHeight * 4` assumes 4 bytes/pixel and ignores row padding (and color spaces), so the 64 MB `totalCostLimit` is best-effort at best. Also, `kCGImageSourceShouldCache: true` means ImageIO additionally caches the decoded image, double-buffering what NSCache holds.

**Suggestion:** Use `image.cgImage.bytesPerRow * image.cgImage.height` for the cost (fall back to the current estimate when `cgImage` is nil), and consider `kCGImageSourceShouldCache: false` since the resulting `UIImage` is already held by the cache.

### [minor] Markdown/code caches have count limits but no byte-cost limits
**File:** Conduit/Views/Components/MarkdownText.swift:100 and 1395

**Issue:** `countLimit = 256` / `128` bounds entry count, not bytes. A single large message or code block (multi-MB `NSAttributedString`/`AttributedString`) scales the resident memory with entry count; a transcript of big messages can hold many MB in `MarkdownRenderCache`'s prebuilt selectable strings alone.

**Suggestion:** Set `totalCostLimit` on both caches and pass an estimated byte cost (e.g. `source.utf8.count` plus a constant for the attributed structure) to `setObject(_:forKey:cost:)`. Keep the existing count limits.

### [minor] Cache key built by naive `"|"` concatenation without escaping
**File:** Conduit/Views/Components/MarkdownText.swift:121-126

**Issue:** The key joins fixed tokens and `source` with `"|"`. Since `source` is arbitrary user/agent text, a crafted message could theoretically collide with another field-combination's key (e.g. a source beginning with `"1|1|L|"`), serving a stale rendering for a message that would otherwise miss. Practical likelihood is near zero and the worst case is a stale snapshot, but the key format is undocumented and non-injective.

**Suggestion:** Use a delimiter not present in inputs (NUL) or length-prefix the fields, or switch to a composite key type (`[String]` hashing) so fields can't bleed into each other.

### [nit] Nonisolated access to `UIApplication.shared` in a cache key
**File:** Conduit/Views/Components/MarkdownText.swift:124

**Issue:** `MarkdownRenderCache.rendering` is a nonisolated static function that reads `UIApplication.shared.preferredContentSizeCategory`. It only happens to run on the MainActor today (all call sites are `body`/`TimelineView`). Under Swift 6 strict concurrency this won't compile, and a future off-main call site would crash.

**Suggestion:** Mark the function `@MainActor` (or inject the content size category as a parameter from the view, which also makes it testable).

### [nit] `foregroundStyle` deliberately excluded from the cache key
**File:** Conduit/Views/Components/MarkdownText.swift:114-120

**Issue:** Correct today — all call sites pass `.primary`/`.white` tied to `usesAccentSurface` — but it's an implicit invariant guarded only by a comment. If a third style is ever passed, the cache will silently serve strings built with the wrong `foregroundStyle`.

**Suggestion:** Enforce the invariant with an `assert(foregroundStyle == .primary || foregroundStyle == .white)` in `rendering` so the assumption fails loudly rather than silently corrupting renders.

Retain cycles, force unwraps, and SwiftUI lifecycle regressions: none introduced. All cache access is confined to the MainActor and `NSCache` is thread-safe; the shared cached `NSAttributedString` is only ever read (opacity fades go to a copy), so mutation/corruption is not a risk.

## Verdict
COMMENT

The caching design is sound (keys cover Dynamic Type, streaming skip-write is correctly scoped, and the cache-hit path never mutates shared state), and it's a clear win over the pre-PR behavior. The main-thread full-decode on cache miss is the one item worth fixing in this PR given its stated performance goal; the rest are hardening recommendations.

<a href="https://opencode.ai/s/eMJYfO5K">New%20session%20-%202026-08-12T19%3A48%3A24.833Z</a>
[opencode session](https://opencode.ai/s/eMJYfO5K)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31631988000)

kaishi00:
## Summary
Adds three `NSCache` layers (Markdown rendering, syntax highlighting, local image previews) to stop per-frame re-parse/re-render of settled transcript content during streaming, with the local-image decode moved off the main thread and byte-cost limits on all caches.

## Findings

### [minor] Off-main decode relies on unchecked actor isolation and crosses a `Sendable` boundary
**File:** Conduit/Views/ChatView.swift:863-865, 817-828
**Issue:** `downsampledPreview` is called from inside `Task.detached`, which is correct only because `UserImageAttachmentPreview` isn't inferred `@MainActor` under the project's Swift 5.9 / `@preconcurrency` View conformance. Nothing enforces that — if the SDK/Xcode changes inference (or strict concurrency is enabled), this becomes a compile error. Separately, the detached closure returns a non-`Sendable` `UIImage?` across the task boundary, which also fails under Swift 6 strict concurrency. The "safe off the MainActor" claim in the comment is true for CoreGraphics but not compiler-verified.
**Suggestion:** Mark the decoder `nonisolated static func` so its intent is explicit and self-documenting, and box the result (`@unchecked Sendable` value holder) or annotate the helper so the boundary is Swift-6-clean. At minimum, note the strict-concurrency requirement in the comment.

### [minor] Local-image loading branch is ordered behind `isGatewayImage`
**File:** Conduit/Views/ChatView.swift:840-844
**Issue:** `body` checks `isGatewayImage && !gatewayLoadFailed` (shows loading placeholder) before `localFileURL != nil && !localPreviewFailed`. For any local attachment whose URI path starts with `/` (an absolute file path) that fails to decode, the task takes the local branch (so `gatewayLoadFailed` is never set) and the view falls into the gateway branch — showing an indefinite "Loading image..." spinner instead of the "Image unavailable" label. Production URIs from ComposerBar are `file://`-prefixed so this is latent, but the ordering is fragile.
**Suggestion:** Reorder the branches so the local-file check precedes the gateway check (`else if localFileURL != nil && !localPreviewFailed` before `else if isGatewayImage && !gatewayLoadFailed`), or make branch 2 also require `localFileURL == nil`.

### [minor] No single-flight deduplication for concurrent same-file cache misses
**File:** Conduit/Views/ChatView.swift:861-865
**Issue:** Each `UserImageAttachmentPreview` instance checks the cache in `body`/`.task`, but two rows showing the same attachment (or rapid scroll recreation before the cache warms) miss simultaneously and each launches a full `Task.detached` decode of the same file — duplicate file I/O and decode bursts.
**Suggestion:** Consider a per-key in-flight guard (e.g. a small dictionary of `Task` handles) or accept the duplicate work; it's bounded by cache warm-up, so this is optional hardening.

### [nit] Render-cache key joins raw `source` with `|`
**File:** Conduit/Views/Components/MarkdownText.swift:129-134
**Issue:** The key concatenates fixed tokens and arbitrary message text with `"|"`. A crafted message whose text contains `|`-separated segments could theoretically collide with another key layout and serve a stale rendering. Low probability; worst case is a wrong-color stale snapshot.
**Suggestion:** Use a delimiter guaranteed absent from inputs (e.g. `\0`) or a composite hashable key.

### [nit] `foregroundStyle` invariant is debug-only
**File:** Conduit/Views/Components/MarkdownText.swift:120-123
**Issue:** Correct today — all call sites pass `.primary`/`.white` tied to `usesAccentSurface` — but it's enforced by `assert` only. In release, introducing a third style would silently cache strings rendered with the wrong color.
**Suggestion:** Fine to keep, but if this code evolves, add the style token to the key rather than relying on the assert.

### [nit] Dynamic Type key reads the app-level content size category
**File:** Conduit/Views/Components/MarkdownText.swift:132
**Issue:** `UIApplication.shared.preferredContentSizeCategory` reflects the app-wide setting, not a per-scene/per-window override. If the app ever supports scene-specific content sizes, cached attributed strings could be rendered with stale fonts.
**Suggestion:** Pass the size category in from the view's environment (`@Environment(\.sizeCategory)`) — also makes the function testable without `UIApplication.shared`.

Retain cycles, force unwraps, and SwiftUI lifecycle issues: none introduced. Shared cached `NSAttributedString`s are only ever read (`SelectableTextView.configure` and `measureNonWrapping` copy before mutating), fade opacity is applied to a copy, and all `.task` paths guard on `Task.isCancelled` before publishing state, so stale-image overwrite across identity changes is prevented. The prior review's major findings (main-thread decode, unbounded image cache, count-only caches) are all addressed in this PR.

## Verdict
COMMENT

<a href="https://opencode.ai/s/4rkwoKCQ">New%20session%20-%202026-08-12T20%3A00%3A38.121Z</a>
[opencode session](https://opencode.ai/s/4rkwoKCQ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31635373701)

kaishi00:
## Summary
Adds three cost-bounded `NSCache` layers (Markdown rendering, syntax highlighting, local attachment previews) so settled transcript content stops being re-parsed/re-rendered per streaming frame, with local-image decoding moved off the main thread via ImageIO thumbnails. Prior review feedback (main-thread decode, byte-cost limits, stale-state gating by path, off-main `nonisolated` decode) is incorporated.

## Findings

### [minor] `Task.detached` is not cancelled when the parent `.task` dies
**File:** Conduit/Views/ChatView.swift:883-885
**Issue:** `Task.detached` creates an unstructured task that is *not* tied to the parent `.task`'s cancellation. When the row scrolls away, the view disappears, or `appState.activeProfile` changes (restarting `.task(id:)`), the already-launched detached decode keeps running to completion — full file I/O + ImageIO decode for an attachment that is no longer on screen. The `guard !Task.isCancelled` after `await` correctly prevents stale state publication, but the wasted decode is not bounded by cancellation, and concurrent misses (e.g., rapid scroll recreation before the cache warms) launch duplicate decodes of the same file.
**Suggestion:** Check `Task.isCancelled` inside the detached closure before decoding (or gate on a per-path in-flight marker/dictionary), or start the decode as a structured child task so cancellation propagates.

### [minor] `kCGImageSourceCreateThumbnailFromImageAlways` forces a full-resolution decode
**File:** Conduit/Views/ChatView.swift:838-843
**Issue:** `FromImageAlways` makes ImageIO decode the *entire* source bitmap (a 48MP photo can be hundreds of MB transiently) before producing the 800px thumbnail. With `kCGImageSourceShouldCache: false` the result isn't retained, but the peak memory spike during decode is still proportional to the full source, which is off-main but can trip memory warnings on large library photos.
**Suggestion:** Use `kCGImageSourceCreateThumbnailFromImageIfAbsent` (or a size gate that falls back to `FromImageAlways` only for small sources) so embedded thumbnails avoid the full decode.

### [minor] Render-cache key omits the trait collection while the cached string bakes colors at build time
**File:** Conduit/Views/Components/MarkdownText.swift:129-134
**Issue:** `MarkdownSelectionFormatter.attributedText` bakes `.foregroundColor = UIColor(foregroundStyle)` (and `.link` colors) into the cached `selectableText` for every run, and `SelectableTextView.configure` only fills *missing* attributes (SelectableTextView.swift:172). The key covers source, media flag, accent surface, and content-size category, but not `userInterfaceStyle`/trait collection. If `UIColor(Color.primary)` bridges to a statically-resolved color rather than a dynamic one, a light↔dark change will serve stale-colored text until the entry is evicted. Worth verifying the bridge is dynamic; it is currently an unstated invariant.
**Suggestion:** Confirm `UIColor(.primary)` returns a dynamic color, or add a trait key (e.g., `UITraitCollection.current.userInterfaceStyle`) to the render-cache key.

### [minor] Cache keys are non-injective `|`-joins of arbitrary text
**File:** Conduit/Views/Components/MarkdownText.swift:129-134 and 1413
**Issue:** Both caches join fixed tokens and user/agent-supplied `source` with `"|"`. Because `source` may contain `|`, a crafted message can equal a different (flags, category, source) combination and be served a stale rendering. Probability is near-zero and the worst case is a wrong snapshot, but the format is undocumented and non-injective in both caches.
**Suggestion:** Length-prefix the fields, use a NUL delimiter, or switch to a composite key type.

### [minor] Non-`Sendable` `UIImage` crosses the detached task boundary
**File:** Conduit/Views/ChatView.swift:883-885
**Issue:** The `Task.detached` closure returns `UIImage?` across an actor boundary. This compiles under the project's Swift 5 mode and is documented in a comment, but it is a compile error under Swift 6 strict concurrency (`Task.value` requires `Sendable` success), so the off-main guarantee is comment-enforced rather than compiler-enforced.
**Suggestion:** Wrap the result in a small `@unchecked Sendable` box or return `Data`/`CGImage`, and/or gate with a build setting once strict concurrency is adopted.

### [nit] `MarkdownRendering` retains both `blocks` and `selectableText`
**File:** Conduit/Views/Components/MarkdownText.swift:80-88, 138-146
**Issue:** For the selectable path the entry keeps the parsed `[MarkdownBlock]` *and* the built attributed string, roughly doubling resident memory per entry; the blocks are only needed when `selectableText == nil`.
**Suggestion:** Store the parsed blocks only for the non-selectable path (or nil them out when `selectableText` is non-nil).

No retain cycles, force unwraps, or SwiftUI lifecycle regressions were introduced. Cache access is confined to the MainActor, `NSCache` is thread-safe, and the shared cached `NSAttributedString` is only ever read — the streaming fade goes to an `NSMutableAttributedString` copy (MarkdownText.swift:357-361), and `SelectableTextView` copies before mutating — so shared-state corruption is not a risk.

## Verdict
COMMENT

The design is sound and the prior major findings (main-thread decode, unbounded caches, stale local-preview state) are genuinely resolved. The remaining items are non-blocking hardening: cancel-aware detached decoding, avoiding forced full-resolution decodes, and two cache-key robustness concerns (trait-collection coverage and delimiter collisions).

<a href="https://opencode.ai/s/SB9tewYY">New%20session%20-%202026-08-12T20%3A40%3A31.620Z</a>
[opencode session](https://opencode.ai/s/SB9tewYY)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31637852406)

kaishi00:
Closing the loop on the two review rounds. Everything actionable is addressed in `f5658a9` and `dcf31fd`; the rest are skipped with reasoning.

## Addressed

**Image preview decode (`f5658a9`)**
- *Sync full-image decode on cache miss* — local attachments now decode off the MainActor via `Task.detached` into `@State`. Cache hits still render on the first frame (sync `cachedLocalImage` lookup), so only a first-view miss shows the brief loading placeholder.
- *`.task` guard decoded local images it never loaded* — the local-file branch short-circuits before the gateway path.
- *Image cache cost under-counts / double-buffers* — `bitmapCost` uses `cgImage.bytesPerRow * height`; `kCGImageSourceShouldCache` is `false`.
- *Unbounded image cache* — `totalCostLimit` (64 MB) + per-image cost.

**Render / highlight caches (`f5658a9`)**
- *Count-only limits* — `totalCostLimit` (32 MB render / 16 MB highlight) + byte-cost `setObject`.
- *foregroundStyle invariant* — enforced with a debug `assert`.
- *Nonisolated `UIApplication.shared`* — `MarkdownRenderCache.rendering` marked `@MainActor`.

**State hardening (`dcf31fd`)**
- *Path-key local-preview state* — `localPreview` / failure now carry their path; `currentLocalPreview` / `localPreviewFailed` only return a value on a path match, so a stale image can't render if a view identity is ever reused for a different uri (today unreachable — `Attachment` is `Identifiable`, uris immutable — but the invariant is no longer load-bearing).
- *Gateway vs local-file branch overlap* — gateway loading branch now requires `localFileURL == nil`, closing a latent infinite-spinner on a `/`-prefixed local file that fails to decode.
- *Off-main decoder isolation* — `downsampledPreview` marked `nonisolated`; comment documents the Swift-6 `UIImage`-Sendable caveat.

## Considered and skipped
- *Naive `|`-joined cache key* — the first three fields are fixed-format and pipe-free and `source` is the trailing field, so the joined key is effectively injective; no collision is reachable.
- *No single-flight dedup for concurrent same-file misses* — optional; duplicate decode is bounded to cache warm-up and self-resolving.
- *foregroundStyle assert is debug-only* — appropriate as an invariant guard; release trusts the invariant.
- *App-level `preferredContentSizeCategory`* — single-scene app, so app-level == scene-level. Injecting `@Environment(\.sizeCategory)` would be cleaner/testable but is a moderate refactor for no current benefit; noted as a possible future cleanup.

Build & Test green across all three pushes.

kaishi00:
## Summary
Adds three `NSCache` layers (`MarkdownRenderCache`, `SyntaxHighlighter`, `UserImageAttachmentPreview.localImageCache`) so settled transcript content stops re-parsing/re-rendering at streaming frame rate, plus an off-main ImageIO downsampled decode path for local attachment previews with path-gated `@State`, and a `isStreaming` flag that skips render-cache writes for the transient streaming snapshot.

## Findings

### [minor] Detached image decode ignores the parent task's cancellation
**File:** Conduit/Views/ChatView.swift:883-885
**Issue:** `Task.detached` is unstructured and not tied to the `.task(id:)` that launched it. When the row scrolls off, the view disappears, or `appState.activeProfile` changes (restarting the task), the already-launched decode keeps doing full file I/O + ImageIO work for an attachment that is no longer visible. The `guard !Task.isCancelled` after the `await` correctly prevents stale state publication, but the decode itself is not bounded, and simultaneous misses on the same path (multiple rows / rapid scroll recreation) launch duplicate decodes of the same file.
**Suggestion:** Check `Task.isCancelled` inside the detached closure before decoding, or gate on a per-path in-flight marker so concurrent misses share one decode.

### [minor] `kCGImageSourceCreateThumbnailFromImageAlways` forces a full-resolution decode
**File:** Conduit/Views/ChatView.swift:839
**Issue:** `FromImageAlways` decodes the *entire* source bitmap (a 48MP photo can spike hundreds of MB transiently) before producing the 800px thumbnail. It's off the main thread so it won't hitch scrolling, but the peak-memory spike can trip a memory warning on large library photos — precisely the failure mode the cache is meant to prevent.
**Suggestion:** Use `kCGImageSourceCreateThumbnailFromImageIfAbsent` (or a size gate that falls back to `FromImageAlways` only for small sources) so embedded thumbnails avoid the full decode.

### [minor] Render-cache key omits trait collection while colors/fonts are baked at build time
**File:** Conduit/Views/Components/MarkdownText.swift:129-134
**Issue:** `MarkdownSelectionFormatter.attributedText` bakes `.foregroundColor = UIColor(foregroundStyle)` / `UIColor(Color.conduitAccent)` (list markers, MarkdownText.swift:262) and `UIFont.preferredFont(...)` into the cached string. The key covers source, media flag, accent surface, and content-size category, but not `userInterfaceStyle`. If `UIColor(Color.primary)` bridges to a statically-resolved color rather than a dynamic one, a light↔dark switch serves stale-colored text until eviction. Separately, the Bold Text accessibility setting changes `preferredFont` metrics without changing `preferredContentSizeCategory`, so bold-text toggles also wouldn't invalidate the key.
**Suggestion:** Confirm `UIColor(Color.primary)` is dynamic; otherwise add `UITraitCollection.current.userInterfaceStyle` to the key. Consider whether bold-text font drift matters for this app.

### [minor] Streaming selectable path copies the entire freshly-built attributed string every frame
**File:** Conduit/Views/Components/MarkdownText.swift:47-51, 352-361
**Issue:** While streaming, the render cache is always missed (source changes per frame), so `rendering.selectableText` is freshly built every frame; `applyingTrailingCharacterOpacities` then copies that whole new string just to fade the last ~18 characters. This is net-new per-frame work vs. the pre-PR code, which applied opacities while building (MarkdownText.swift:221). Most streamed replies are selectable-flow blocks, so this is the common path.
**Suggestion:** When the cache misses, pass `newestCharacterOpacities` straight into the builder (as before) and only copy when the base actually comes from the cache — the copy exists to protect a shared cached instance, which a fresh build isn't.

### [minor] `SyntaxHighlighter` churn from the still-growing code block during streaming
**File:** Conduit/Views/Components/MarkdownText.swift:1412-1417
**Issue:** The growing block writes a new entry every frame, each key being a fresh O(block-size) `NSString` join, at 30fps. Over a long streaming code block this continuously churns the 128-entry/16MB cache, evicting genuinely reusable stable entries (other messages' code) while the stale growing-block snapshots hold slots. The behavior is documented, but the write side still defeats part of the cache's purpose during the exact scenario the PR targets.
**Suggestion:** Consider skipping the write for the trailing block (e.g., callers pass an `isLastBlock`/growing flag) so only stable blocks populate the cache.

### [nit] `SyntaxHighlighter` key joins arbitrary `language` and `source` with `|`
**File:** Conduit/Views/Components/MarkdownText.swift:1413
**Issue:** A fenced-code language tag containing `|` (user/agent-supplied text) can collide with a different `(language, source)` pair, serving one block's tokenization for another. Unlike the render-cache key — whose fixed single-char flags and pipe-free size category make it effectively injective — this one is non-injective. Worst case is a wrong-colored snapshot.
**Suggestion:** Use a composite key type or a NUL/length-prefixed delimiter.

### [nit] `MarkdownRendering` retains both `blocks` and `selectableText`
**File:** Conduit/Views/Components/MarkdownText.swift:80-88
**Issue:** For the selectable path, the entry keeps the parsed `[MarkdownBlock]` *and* the built attributed string, roughly doubling resident memory; the blocks are only used when `selectableText == nil`.
**Suggestion:** Nil out `blocks` (or store them only for the non-selectable path).

### [nit] Local-file branch precedes the gateway branch in `.task`
**File:** Conduit/Views/ChatView.swift:881
**Issue:** `if let url = localFileURL` runs before `else if isGatewayImage`. A gateway URI like `/media/...` whose path coincidentally exists on disk (e.g., a legacy stored attachment) would be decoded locally and never fetched through Hermes. Today gateway URIs don't collide with local paths, so this is latent.
**Suggestion:** Require `!isGatewayImage` on the local branch, or reorder to check gateway first.

### [nit] Non-`Sendable` `UIImage` crosses the detached-task boundary
**File:** Conduit/Views/ChatView.swift:883-885
**Issue:** The `@Sendable` detached closure returns `UIImage?` and it's read back via `.value`. This compiles in the project's Swift 5 mode (and is documented in the comment), but it is a compile error under Swift 6 strict concurrency — the off-main guarantee is comment-enforced, not compiler-enforced.
**Suggestion:** Wrap the result in a small `@unchecked Sendable` box or return `CGImage`/`Data`; acceptable to defer until strict concurrency is adopted.

Confirmed non-issues from earlier review rounds: the `|`-joined `MarkdownRenderCache` key is actually injective (fixed single-char flags + pipe-free size category before the trailing source), so no collision is reachable there. No retain cycles, force unwraps, or SwiftUI lifecycle regressions were introduced. The shared cached `NSAttributedString` is only ever read — the streaming fade goes to an `NSMutableAttributedString` copy (MarkdownText.swift:358) and `SelectableTextView.configure` copies before mutating (SelectableTextView.swift:172) — so shared-state corruption is not a risk. Cache access is confined to the MainActor, all cache limits are cost-bounded, and the `isStreaming` skip-write is correctly scoped.

## Verdict
COMMENT

The design is sound and the prior major findings (main-thread decode, unbounded caches, stale local-preview state, `@MainActor` isolation) are genuinely resolved. Remaining items are non-blocking hardening: cancel-aware detached decoding, avoiding forced full-resolution decodes, verifying the color-scheme/bold-text behavior against the render-cache key, and trimming the redundant streaming-path copy.

<a href="https://opencode.ai/s/z4GF3XJT">New%20session%20-%202026-08-12T20%3A49%3A06.833Z</a>
[opencode session](https://opencode.ai/s/z4GF3XJT)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31639435474)

PR #51, opened 2026-08-12
Guard WebSocket receive failure by socket identity

kaishi00:
## Problem

`HermesClient.receiveLoop()`'s catch block mutated shared state — `isConnected`, the pending-RPC map, and `onDisconnected` — with **no socket-identity check**, unlike `didOpen`/`didClose`/`failSocketOpen`, which all guard on `self.socket === socket`.

If a reconnect ever replaced `self.socket` and spawned a new loop while an old `socket.receive()` was still suspended, the old loop's late catch would run unguarded on the `@MainActor` and:
- set `isConnected = false` on the healthy new connection,
- fail the *new* connection's in-flight RPCs via `failAllPendingRequests`,
- fire `onDisconnected?()`.

A prior change (failing pending requests in this catch) made that clobbering strictly more damaging. Separately, the catch left the dead socket installed, and since `URLSessionWebSocketTask.closeCode` stays `.invalid` after a receive error (it means "no close frame," not "healthy"), `rpc`'s `closeCode == .invalid` precondition would let a new RPC ride the dead socket to its timeout.

## Severity: latent

This does **not** reproduce in production today. `AppState` creates a *new* `HermesClient` for each reconnect and calls `previousClient.disconnect()` first (which cancels the old `receiveTask`), so the cross-instance clobber can't happen. These changes harden against an intra-instance double-`connect()`, which no caller currently performs. Flagged by the bot reviews on #44/#50 against already-merged code.

## Fix

1. **`receiveLoop()` catch** — guard the entire teardown on `guard self.socket === socket else { break }` (the loop already captures its socket in a local). A superseded loop becomes a no-op.
2. **`receiveLoop()` catch** — nil out the dead socket so `rpc` can't re-issue against it.
3. **`connect()`** — cancel any prior `receiveTask` before spawning a new one (defense-in-depth, mirroring `disconnect()`).
4. **`rpc()` guard** — require `isConnected` in addition to a non-nil socket, so the window between socket creation and completed handshake is also covered.

## Testing

No new regression test: a real lifecycle test needs a `URLSessionWebSocketTask`/`URLSession` seam, which is a sizable refactor for a latent, unreachable-in-production defect. Verified by CI build + the existing suite (unchanged behavior on the happy path and the normal single-connection disconnect/reconnect flow). Adding a socket seam is a reasonable follow-up if we want this covered long-term.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved WebSocket connection handling during reconnects and disconnects.
  - Prevented requests from using failed or superseded connections.
  - Improved cleanup after receive failures and invalidated obsolete connections.
  - Increased reliability when multiple connection attempts overlap.

- **Tests**
  - Added coverage for handshakes, connection failures, reconnects, cancellation, receive-loop behavior, and concurrent connection changes.

### DIFF
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -358,6 +358,14 @@ final class HermesClient: ObservableObject {
     // MARK: - Connection
 
     func connect() async throws {
+        // If a previous receive loop is still alive, cancel it now so it can't
+        // outlive the socket it was reading from and later mutate this client's
+        // state. (AppState makes a fresh HermesClient per reconnect and
+        // disconnects the old one first, so this is defense-in-depth against
+        // an intra-instance reconnect — which no caller does today.)
+        receiveTask?.cancel()
+        receiveTask = nil
+
         closedIntentionally = false
         socketHasOpened = false
         let url: URL
@@ -452,11 +460,21 @@ final class HermesClient: ObservableObject {
             } catch {
                 // Socket closed or errored
                 logger.error("WebSocket receive failed: \(error.localizedDescription, privacy: .public)")
+                // Only tear down if this loop still owns the current socket.
+                // A reconnect replaces `self.socket` and spawns a new loop; an
+                // unguarded late catch from the superseded loop would mark the
+                // healthy new connection disconnected and fail its in-flight
+                // RPCs. didOpen/didClose/failSocketOpen guard the same way.
+                guard self.socket === socket else { break }
                 isConnected = false
                 // No response can ever arrive on a dead socket; failing the
                 // in-flight requests here keeps awaiting UI from hanging for
                 // the remainder of each request's timeout.
                 failAllPendingRequests(with: HermesError.connectionClosed)
+                // Drop the dead socket: its `closeCode` stays `.invalid` after
+                // a receive error, so leaving it installed would let `rpc`
+                // re-issue against it and hang until its own timeout.
+                self.socket = nil
                 if !closedIntentionally {
                     onDisconnected?()
                 }
@@ -527,7 +545,10 @@ final class HermesClient: ObservableObject {
     // MARK: - RPC
 
     private func rpc(_ method: String, params: [String: Any]? = nil, timeout: TimeInterval = requestTimeout) async throws -> AnyCodable {
-        guard let socket, socket.closeCode == .invalid else {
+        // Require both a live socket and a completed handshake. A receive
+        // error leaves the socket installed with `closeCode == .invalid`, so
+        // closeCode alone would let an RPC ride a dead socket to its timeout.
+        guard let socket, socket.closeCode == .invalid, isConnected else {
             throw HermesError.notConnected
         }
 
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -299,12 +299,75 @@ private struct PendingRequest {
     let timer: Timer?
 }
 
+// MARK: - WebSocket seam
+
+/// The slice of `URLSessionWebSocketTask` that `HermesClient` depends on, so
+/// the client can be driven against a fake transport in tests without a real
+/// network. `AnyObject`-bound so the `===` identity checks (which tell a
+/// superseded receive loop apart from the current socket) keep working on the
+/// existential type.
+protocol HermesWebSocket: AnyObject {
+    var closeCode: URLSessionWebSocketTask.CloseCode { get }
+    func resume()
+    func cancel(with closeCode: URLSessionWebSocketTask.CloseCode, reason: Data?)
+    func receive() async throws -> URLSessionWebSocketTask.Message
+    func send(_ message: URLSessionWebSocketTask.Message, completionHandler: @escaping (Error?) -> Void)
+}
+
+extension URLSessionWebSocketTask: HermesWebSocket {}
+
+/// Produces sockets for a `HermesClient` and owns the underlying session.
+/// `makeSocket` wires the handshake callbacks (`onOpen` / `onCloseBeforeOpen`,
+/// mirroring `URLSessionWebSocketDelegate`); `invalidate` tears the session
+/// down. The production impl is a faithful extraction of the client's former
+/// inline `URLSession` logic; tests supply a fake. All methods are invoked from
+/// `HermesClient` on the MainActor; the protocol is left nonisolated so the
+/// `URLSessionWebSocketTransport` default doesn't force closure-isolation hoops.
+protocol HermesWebSocketTransport: AnyObject {
+    func makeSocket(
+        request: URLRequest,
+        onOpen: @escaping (any HermesWebSocket) -> Void,
+        onCloseBeforeOpen: @escaping (any HermesWebSocket) -> Void
+    ) -> any HermesWebSocket
+
+    func invalidate()
+}
+
+/// Production transport: one `URLSession` per socket, delegate-driven
+/// handshake. Behaviour is identical to the client's pre-seam connect().
+final class URLSessionWebSocketTransport: HermesWebSocketTransport {
+    private var session: URLSession?
+    private var delegate: WebSocketOpenDelegate?
+
+    func makeSocket(
+        request: URLRequest,
+        onOpen: @escaping (any HermesWebSocket) -> Void,
+        onCloseBeforeOpen: @escaping (any HermesWebSocket) -> Void
+    ) -> any HermesWebSocket {
+        let delegate = WebSocketOpenDelegate()
+        delegate.onOpen = onOpen
+        delegate.onCloseBeforeOpen = onCloseBeforeOpen
+        let config = URLSessionConfiguration.default
+        config.waitsForConnectivity = true
+        let session = URLSession(configuration: config, delegate: delegate, delegateQueue: nil)
+        self.session = session
+        self.delegate = delegate
+        return session.webSocketTask(with: request)
+    }
+
+    func invalidate() {
+        session?.invalidateAndCancel()
+        session = nil
+        delegate = nil
+    }
+}
+
 /// URLSession only confirms the WebSocket handshake through its delegate. The
 /// client must not issue session RPCs or paint a green indicator before this
 /// callback, otherwise a just-opened socket can race its first `session.resume`.
 private final class WebSocketOpenDelegate: NSObject, URLSessionWebSocketDelegate {
-    var onOpen: ((URLSessionWebSocketTask) -> Void)?
-    var onCloseBeforeOpen: ((URLSessionWebSocketTask) -> Void)?
+    var onOpen: ((any HermesWebSocket) -> Void)?
+    var onCloseBeforeOpen: ((any HermesWebSocket) -> Void)?
 
     func urlSession(_ session: URLSession, webSocketTask: URLSessionWebSocketTask, didOpenWithProtocol protocol: String?) {
         onOpen?(webSocketTask)
@@ -326,9 +389,9 @@ final class HermesClient: ObservableObject {
     @Published private(set) var isConnected = false
 
     // Non-published internal state
-    private var socket: URLSessionWebSocketTask?
-    private var session: URLSession?
-    private var socketDelegate: WebSocketOpenDelegate?
+    private var socket: (any HermesWebSocket)?
+    private var transport: (any HermesWebSocketTransport)?
+    private let transportFactory: () -> any HermesWebSocketTransport
     private var requestId = 0
     private var pending = [Int: PendingRequest]()
     private var closedIntentionally = false
@@ -349,22 +412,36 @@ final class HermesClient: ObservableObject {
     static let promptSubmitTimeout: TimeInterval = 180 // 3 minutes
     static let titleGenerationTimeout: TimeInterval = 90
 
-    init(connection: HermesConnection, profile: String? = nil, cloudflareAccess: CloudflareAccessCredentials? = nil) {
+    init(
+        connection: HermesConnection,
+        profile: String? = nil,
+        cloudflareAccess: CloudflareAccessCredentials? = nil,
+        transportFactory: @escaping () -> any HermesWebSocketTransport = { URLSessionWebSocketTransport() }
+    ) {
         self.connection = connection
         self.profile = profile
         self.cloudflareAccess = cloudflareAccess
+        self.transportFactory = transportFactory
     }
 
     // MARK: - Connection
 
     func connect() async throws {
-        // If a previous receive loop is still alive, cancel it now so it can't
-        // outlive the socket it was reading from and later mutate this client's
-        // state. (AppState makes a fresh HermesClient per reconnect and
-        // disconnects the old one first, so this is defense-in-depth against
-        // an intra-instance reconnect — which no caller does today.)
+        // Tear down any prior connection in full before installing a new one.
+        // Cancelling the receive task alone is not enough: URLSessionWebSocketTask
+        // .receive() does not honor cooperative cancellation, so a superseded
+        // loop stays alive until its socket actually errors. Cancelling the
+        // socket and invalidating the prior session forces that error, after
+        // which the loop's identity guard turns its late catch into a no-op.
+        // (AppState makes a fresh HermesClient per reconnect and disconnects the
+        // old one first, so this is defense-in-depth against an intra-instance
+        // reconnect — which no caller does today.)
         receiveTask?.cancel()
         receiveTask = nil
+        socket?.cancel(with: .goingAway, reason: nil)
+        socket = nil
+        transport?.invalidate()
+        transport = nil
 
         closedIntentionally = false
         socketHasOpened = false
@@ -379,22 +456,16 @@ final class HermesClient: ObservableObject {
             throw HermesError.invalidUrl
         }
 
-        let config = URLSessionConfiguration.default
-        config.waitsForConnectivity = true
-        let socketDelegate = WebSocketOpenDelegate()
-        socketDelegate.onOpen = { [weak self] task in
-            Task { @MainActor in self?.didOpen(task) }
-        }
-        socketDelegate.onCloseBeforeOpen = { [weak self] task in
-            Task { @MainActor in self?.didClose(task) }
-        }
-        let session = URLSession(configuration: config, delegate: socketDelegate, delegateQueue: nil)
-        self.session = session
-        self.socketDelegate = socketDelegate
+        let transport = transportFactory()
+        self.transport = transport
 
         var request = URLRequest(url: url)
         request = cloudflareAccess?.applying(to: request) ?? request
-        let socket = session.webSocketTask(with: request)
+        let socket = transport.makeSocket(
+            request: request,
+            onOpen: { [weak self] socket in Task { @MainActor in self?.didOpen(socket) } },
+            onCloseBeforeOpen: { [weak self] socket in Task { @MainActor in self?.didClose(socket) } }
+        )
         self.socket = socket
         socket.resume()
         try await waitForSocketOpen(socket)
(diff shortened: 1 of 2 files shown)


kaishi00:
## Summary
Hardens `HermesClient`'s receive-failure teardown and RPC guard against a superseded receive loop clobbering a new connection, by gating the catch teardown on socket identity, dropping the dead socket, cancelling the prior `receiveTask` in `connect()`, and requiring `isConnected` in `rpc()`.

## Findings

### [minor] Superseded receive loop's success path is unguarded; old socket/session never torn down
**File:** Conduit/Services/HermesClient.swift:360
**Issue:** The new `receiveTask?.cancel()` defense-in-depth is only partially effective. `URLSessionWebSocketTask.receive()` does not respond to cooperative task cancellation, so a superseded loop suspended in `socket.receive()` stays alive until the *socket* errors/closes. `connect()` replaces `self.socket`/`self.session` without canceling the old socket or invalidating the old session (unlike `disconnect()`, which does both). Consequences in the exact intra-instance scenario this PR hardens: (a) the old loop's `handleMessage` success path (HermesClient.swift:452-459) has no socket-identity check, so a stale stream event on the old connection can still fire `onEvent` on the new connection; (b) if the server keeps the old connection open, the old socket + URLSession are retained indefinitely.
**Suggestion:** In `connect()`, cancel the previous socket and invalidate the previous session before installing new ones, e.g. `socket?.cancel(with: .goingAway, reason: nil); session?.invalidateAndCancel()`. Optionally gate `handleMessage`/event dispatch on the loop still owning the current socket.

### [nit] Concurrent `connect()` leaks the first open continuation
**File:** Conduit/Services/HermesClient.swift:410
**Issue:** Pre-existing, but relevant since this PR now hardens `connect()` for re-entrancy: if a second `connect()` runs while a first is still inside `waitForSocketOpen`, the second overwrites `openContinuation` (line 412), leaving the first continuation never resumed — the first `connect()` call hangs forever and its `isConnected`/`receiveTask` state is then set by the wrong task.
**Suggestion:** Guard `openContinuation`/handshake against re-entry (e.g. a `connecting` flag or assert), or at minimum note it as out of scope in the comment.

### [nit] Socket-identity guard relies on reference identity semantics
**File:** Conduit/Services/HermesClient.swift:468
**Issue:** The `guard self.socket === socket` correctly uses `===` on `URLSessionWebSocketTask` (an NSObject subclass), and the ordering is sound since the catch body runs synchronously on the MainActor after the `await`. No defect — just confirming the reasoning holds only because the catch runs to completion without further `await`s.
**Suggestion:** None required.

### [nit] No regression coverage
**File:** Conduit/Services/HermesClient.swift:360-551
**Issue:** The PR intentionally ships without a test; the behavior changed in `rpc()` (the `isConnected` requirement) and the catch teardown are untested paths. Acceptable given the stated seam refactor cost, but the `isConnected` guard change would be cheaply testable if a seam existed.
**Suggestion:** Fine to defer; a `URLSessionWebSocketTask` seam is a worthwhile follow-up.

No issues found with retain cycles (all new/affected closures use `[weak self]`), force unwraps, concurrency (all mutation remains on `@MainActor`), or accessibility.

## Verdict
APPROVE — the change is correct, strictly safer than the status quo, and the remaining gaps are confined to the same latent intra-instance scenario the PR acknowledges no caller exercises.

<a href="https://opencode.ai/s/Q4GNfyVg">New%20session%20-%202026-08-12T19%3A29%3A12.181Z</a>
[opencode session](https://opencode.ai/s/Q4GNfyVg)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31632785293)

kaishi00:
## Summary
Hardens `HermesClient`'s WebSocket lifecycle for the latent intra-instance reconnect case: gates the `receiveLoop()` catch teardown on socket identity, drops the dead socket, fully tears down the prior connection in `connect()`, requires `isConnected` in `rpc()`, and adds a transport seam plus regression tests.

## Findings

### [minor] Superseded loop's message-handling path is still unguarded
**File:** Conduit/Services/HermesClient.swift:524-534 (and `handleMessage` at 561)
**Issue:** The PR only guards the *catch* of the receive loop. The success path (`case .data/.string` → `handleMessage`) has no socket-identity check. Since `URLSessionWebSocketTask.receive()` does not honor cooperative cancellation, a superseded loop survives until its socket actually errors; a stale frame delivered in that window still runs `handleMessage`, which can fire `onEvent` on the new connection (the same clobbering class this PR aims to prevent). In production, AppState's per-client `activeClientEpoch` guard masks this, but the intra-instance scenario this PR hardens has no epoch change.
**Suggestion:** In `receiveLoop`, after each `socket.receive()` succeeds, check `guard self.socket === socket` before `handleMessage` (or check once at the top of the loop body).

### [minor] `connect()` teardown can leak a superseded `openContinuation`
**File:** Conduit/Services/HermesClient.swift:439-444 (and `waitForSocketOpen` at 481-495)
**Issue:** The teardown cancels the old `receiveTask`/socket/transport but does not resume a still-pending `openContinuation`, nor cancel `openTimeoutTask`. The superseded continuation is only resumed later inside `waitForSocketOpen`. If a second `connect()` throws *between* the teardown and `waitForSocketOpen` (e.g. `ConnectionURLPolicy.webSocketURL` → `HermesError.invalidUrl`, or `transportFactory()` throwing), the first `connect()`'s continuation is never resumed and that call hangs forever; its 15s `openTimeoutTask` would also no-op (identity guard) without resuming it.
**Suggestion:** Mirror `disconnect()` in the teardown: `openContinuation?.resume(throwing: HermesError.connectionClosed)` + `openContinuation = nil`, and `openTimeoutTask?.cancel()`.

### [minor] `isConnected` left stale across `connect()` teardown and on failed handshake
**File:** Conduit/Services/HermesClient.swift:429, 472, 511-519
**Issue:** The new teardown nils the socket but does not reset `isConnected`. If an intra-instance reconnect's new handshake fails, `failSocketOpen` leaves `isConnected == true` with `socket == nil` — the published connection indicator disagrees with reality. `rpc()` is protected (nil socket → `.notConnected`), so this is a state-consistency/UI concern rather than a correctness bug, and AppState's disconnect-first flow avoids it today.
**Suggestion:** Set `isConnected = false` in `connect()`'s teardown (and consider invalidating the transport in `failSocketOpen` so a failed handshake's `URLSession` is torn down rather than lingering until the next connect).

### [nit] Tests exercise the seam with a shared transport instance
**File:** ConduitTests/HermesClientTests.swift:128-137, 227-235
**Issue:** `transportFactory` returns the same `FakeTransport` for every `makeClient`, and `testConnectInvalidatesPriorTransport` relies on the shared instance still being usable after `invalidate()` — whereas production creates a fresh transport per connect. The test validates that `invalidate()` is *called*, which is good, but the shared-instance model can't catch a stale-transport reuse bug.
**Suggestion:** Have the fake return a new instance (or track a factory count) so the tests more faithfully model the production transport-per-connect behavior.

No issues found with retain cycles (all new closures use `[weak self]`), force unwraps, concurrency (all mutation stays on `@MainActor`; delegate callbacks hop onto the MainActor correctly), or accessibility (no UI changes).

## Verdict
APPROVE — the core identity-guard fix and `rpc()` hardening are correct and strictly safer than the status quo; remaining gaps are minor and confined to the acknowledged latent intra-instance scenario.

<a href="https://opencode.ai/s/EEvBOoRl">New%20session%20-%202026-08-12T21%3A26%3A44.848Z</a>
[opencode session](https://opencode.ai/s/EEvBOoRl)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31642465000)

kaishi00:
## Summary
Hardens `HermesClient` against a latent intra-instance double-`connect()` by gating the `receiveLoop()` failure teardown on socket identity, dropping dead sockets, tearing down the prior connection in `connect()`, adding an `isConnected` requirement to `rpc()`, and introducing a WebSocket transport seam with regression tests.

## Findings

### [major] New tests race `transport.open()` against `makeSocket()` registration, so the handshake-open callback is dropped
**File:** ConduitTests/HermesClientTests.swift:47-49 (and 69-70, 79-80, 104-105, 127)
**Issue:** `transport.open(socket)` is called synchronously immediately after spawning `Task { try? await client.connect() }`. An unstructured `Task` body does not run until the caller's synchronous section yields, so `makeSocket` (which appends to `socketsMade` and registers the open callback) has not executed when `open()` runs. `FakeTransport.open` at ConduitTests/HermesClientTests.swift:231-233 then `guard`s on `socketsMade.firstIndex` and silently drops the open. The handshake never completes, `connect()` blocks on the 15s open-timeout, and the `XCTAssertTrue(client.isConnected)` assertions (lines 50, 85, 132) fail. `testConcurrentConnectSupersedesPriorWithoutHanging` (lines 121-128) has the same problem for `socketB`. This makes the new regression tests either ~15s-per-test failures or scheduling-dependent.
**Suggestion:** Make `FakeSocket`/`FakeTransport` queue opens that arrive before registration (store unmatched sockets, then invoke their open callback from `makeSocket`), or have the tests `await` until the socket is registered (e.g. reusing `Gate`/`flushMainActor`) before calling `open(_:)`.

### [minor] `connect()` teardown does not fail in-flight pending RPCs
**File:** Conduit/Services/HermesClient.swift:439-444
**Issue:** `connect()` cancels the old receive task/socket and invalidates the transport, but leaves `pending` untouched. After this PR, the superseded loop's catch is a no-op by identity (it can no longer reach `failAllPendingRequests`), so any RPC issued on the old connection is stranded until its own timer fires — up to 180s for `prompt.submit`. `disconnect()` already calls `failAllPendingRequests(with:)` for exactly this reason.
**Suggestion:** Call `failAllPendingRequests(with: HermesError.connectionClosed)` in the `connect()` teardown before clearing state, mirroring `disconnect()`.

### [minor] `isConnected` left stale across reconnect teardown and failed handshakes
**File:** Conduit/Services/HermesClient.swift:439-444, 511-519
**Issue:** The teardown nils the socket and invalidates the transport but never resets `isConnected`. If an intra-instance reconnect's new handshake fails (`didClose`-before-open or the 15s timeout route to `failSocketOpen`), `isConnected` remains `true` with `socket == nil`. `rpc()` is protected (nil socket → `.notConnected`), but the published connection indicator disagrees with reality. During the handshake window `isConnected` is also `true`, which is the state the new `rpc()` guard is meant to reject.
**Suggestion:** Set `isConnected = false` at the top of the `connect()` teardown (and in `failSocketOpen`), so the published state and the `rpc()` guard stay consistent.

### [minor] Superseded loop's success path is unguarded; stale socket/session can linger
**File:** Conduit/Services/HermesClient.swift:524-534, 561
**Issue:** Only the catch is identity-guarded. The success path (`case .data/.string` → `handleMessage`) has no `self.socket === socket` check. Because `URLSessionWebSocketTask.receive()` does not honor cooperative cancellation, a superseded loop stays alive until the socket errors; a frame delivered in that window still runs `handleMessage`, which can fire `onEvent` on the new connection's behalf or resolve a `pending` id. `connect()`'s `socket.cancel(...)` + `transport.invalidate()` normally force the error quickly, but if the server keeps the connection open the window persists.
**Suggestion:** Check `guard self.socket === socket` after each successful `receive()` (before `handleMessage`) so a superseded loop drops stale frames.

### [minor] Superseded `openContinuation` leaks if a second `connect()` throws before `waitForSocketOpen`
**File:** Conduit/Services/HermesClient.swift:439-457, 481-495
**Issue:** The teardown cancels the old receive task/socket/transport but neither resumes a pending `openContinuation` nor cancels `openTimeoutTask`. The resume happens only inside `waitForSocketOpen` (line 486). If the second `connect()` throws *before* reaching it — `ConnectionURLPolicy.webSocketURL` → `HermesError.invalidUrl` (line 450-457) or `transportFactory()` throwing (line 459) — the first `connect()`'s continuation is never resumed and that call hangs forever.
**Suggestion:** In the `connect()` teardown, mirror `disconnect()`: `openContinuation?.resume(throwing: HermesError.connectionClosed)`, `openContinuation = nil`, and `openTimeoutTask?.cancel()`.

### [nit] `failSocketOpen` does not invalidate the transport
**File:** Conduit/Services/HermesClient.swift:511-519
**Issue:** On a failed/timed-out handshake, `socket.cancel` + `self.socket = nil` runs but `transport` stays installed, keeping the `URLSession` and its `WebSocketOpenDelegate` alive until the next `connect()`/`disconnect()`.
**Suggestion:** `transport?.invalidate(); transport = nil` in `failSocketOpen`.

### [nit] Test doubles have fidelity gaps: shared transport per connect, Gate signal loss, missing `closeCode`
**File:** ConduitTests/HermesClientTests.swift:137-164, 182-185, 209-228
**Issue:** (1) `makeClient` reuses one `FakeTransport` across connects, whereas production creates a fresh transport per connect, so stale-transport reuse can't be modeled. (2) `Gate.signal()` drops a signal that arrives before `wait()` suspends, making `testRPCGuardRejectsBeforeHandshake` depend on task-scheduling order. (3) `FakeSocket.cancel` leaves `closeCode` at `.invalid` instead of the supplied value, diverging from `URLSessionWebSocketTask`.
**Suggestion:** Return a distinct transport per connect (or track a factory count); give `Gate` a `signalled` flag; assign `closeCode` in `cancel`.

No issues found with retain cycles (all new closures use `[weak self]`, and the transport→delegate→closure chain terminates on weak self), force unwraps, or concurrency (all mutation stays on `@MainActor`; the `@Sendable` send-completion pattern is pre-existing). No UI/accessibility surface changed.

## Verdict
REQUEST_CHANGES — the production hardening is sound and strictly safer than the status quo, but the newly added regression tests appear to drop their handshake-open callbacks (an execution-order race), which would make the suite fail or flake; the `connect()` teardown gaps (pending RPCs not failed, `isConnected` not reset, continuation leak) should also be closed for the hardening to be complete.

<a href="https://opencode.ai/s/P0ZCF0p5">New%20session%20-%202026-08-12T21%3A34%3A52.778Z</a>
[opencode session](https://opencode.ai/s/P0ZCF0p5)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31643089201)

kaishi00:
## Summary
Hardens `HermesClient`'s WebSocket lifecycle against a latent intra-instance double-`connect()`: the `receiveLoop()` catch and success paths are now gated on socket identity, `connect()` fully tears down the prior connection, `rpc()` additionally requires `isConnected`, and a `HermesWebSocket`/`HermesWebSocketTransport` seam plus regression tests were added.

## Findings

### [minor] Receive-failure teardown leaves the transport (dead URLSession + delegate) installed
**File:** Conduit/Services/HermesClient.swift:552-574
**Issue:** The catch nils `self.socket` so `rpc()` can't re-issue on the dead socket, but it does not invalidate `self.transport` — unlike `failSocketOpen`, which does (`transport?.invalidate(); transport = nil` at HermesClient.swift:525-526). After a receive error, the `URLSession` and its `WebSocketOpenDelegate` inside the transport are retained until the next `connect()`/`disconnect()`. This is not a regression (the old code retained the session too), but it's an inconsistency with the handshake-failure path and leaves the dead session alive for an unbounded time if the caller neither disconnects nor reconnects.
**Suggestion:** Mirror `failSocketOpen`: `transport?.invalidate(); transport = nil` in the catch before/after `self.socket = nil`.

### [minor] Test doubles share one transport across connects, masking transport-per-connect semantics
**File:** ConduitTests/HermesClientTests.swift:180-184
**Issue:** `makeClient` returns the same `FakeTransport` for every `connect()`, whereas production uses `transportFactory()` → a fresh `URLSessionWebSocketTransport` per connect. `testConnectInvalidatesPriorTransport` (HermesClientTests.swift:105-118) passes because the teardown's `invalidate()` flags the very instance that is then immediately reused to build the new socket; a real invalidated `URLSession` cannot serve `makeSocket` again. The suite therefore can't catch a stale-transport-reuse bug.
**Suggestion:** Have the factory return a queue of fresh `FakeTransport` instances (tracking which one was invalidated) so the doubles match production's transport-per-connect behavior.

### [nit] `URLSessionWebSocketTransport.makeSocket` overwrites a live session without invalidating it
**File:** Conduit/Services/HermesClient.swift:342-356
**Issue:** Calling `makeSocket` twice on one transport replaces `self.session`/`self.delegate` without invalidating the prior session, leaking the old `URLSession`, its delegate, and its socket. Unreachable today because `connect()` builds a fresh transport each time and the seam's protocol doc even notes callbacks run on the MainActor, but the injected seam now permits the shape and the `FakeTransport` exercises it (two sockets from one instance).
**Suggestion:** Call `invalidate()` at the top of `makeSocket` (or document the single-use invariant and enforce it in the fake).

### [nit] Stale `earlyOpenRequests` entry in `FakeTransport` if `open(_:)` is called twice
**File:** ConduitTests/HermesClientTests.swift:225-235
**Issue:** If `open(_:)` is invoked after the socket's callback was already consumed (i.e., a second open for the same socket), the key is inserted into `earlyOpenRequests` and never removed. Test-only and not exercised today, but it makes the double's bookkeeping diverge from `makeSocket`'s expectations.
**Suggestion:** Guard `open(_:)` against re-insertion (e.g., check a `fired` set or only buffer when the socket hasn't been registered).

### [nit] Non-Sendable delegate callbacks invoked from a background queue (Swift 6 migration hazard)
**File:** Conduit/Services/HermesClient.swift:368-379
**Issue:** `WebSocketOpenDelegate` calls the non-Sendable `onOpen`/`onCloseBeforeOpen` closures from `URLSession`'s background delegate queue; those closures hop onto the MainActor via `Task { @MainActor ... }`. No data race today (callbacks are written once before the session starts, transport replaced per connect), and it's pre-existing, but under Swift 6 strict concurrency this is a violation. Same for `rpc()`'s `socket.send` completion touching `self?.logger` off the MainActor (HermesClient.swift:680).
**Suggestion:** Consider marking the callback types `@Sendable` or wrapping the delegate hop explicitly; low priority given Swift 5.9 target.

No issues found with retain cycles (all new closures use `[weak self]`; the transport→delegate→weak-self chain terminates correctly), force unwraps, or data races (all state mutation stays on `@MainActor`). The identity guards in the receive loop are correctly placed — the `break`-on-supersession in both the success path (HermesClient.swift:543) and catch (HermesClient.swift:560) is sound because a superseded loop cannot re-acquire the same identity, and the test `Gate`/`FakeTransport.open` buffering now removes the scheduling race that earlier revisions had.

## Verdict
APPROVE — the identity-guard fix, `connect()` teardown (resume-continuation, pending-RPC fail, `isConnected` reset), and `rpc()` hardening are correct and strictly safer than the status quo; remaining findings are minor resource/fidelity gaps confined to the acknowledged latent intra-instance scenario and the test doubles.

<a href="https://opencode.ai/s/ISCJKm4G">New%20session%20-%202026-08-12T22%3A17%3A55.630Z</a>
[opencode session](https://opencode.ai/s/ISCJKm4G)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31646262870)

kaishi00:
Closing the loop on the round-2 review. Everything is addressed in `8ed181d`; Build & Test is green (7m3s).

## Addressed
- **[major] Tests raced `transport.open()` against `makeSocket()` registration (the stuck CI)** — `FakeTransport.open(_:)` now buffers opens that arrive before the socket is registered and fires them once it is, so `Task { connect() }; transport.open(socket)` completes the handshake deterministically instead of dropping the open and hanging on the 15s timeout.
- **[minor] `connect()` teardown didn't fail in-flight pending RPCs** — teardown now mirrors `disconnect()`: `failAllPendingRequests(.connectionClosed)`.
- **[minor] `isConnected` stale across teardown / failed handshake** — teardown sets `isConnected = false`; `failSocketOpen` does too.
- **[minor] Superseded loop's success path was unguarded** — `receiveLoop` now checks `self.socket === socket` after each successful `receive()` (not just in the catch), so a stale frame is dropped before `handleMessage`. Covered by `testSupersededLoopDropsStaleFrame` (delivers a real `message.start` event to a superseded socket).
- **[minor] `openContinuation` leaked if a second `connect()` threw before `waitForSocketOpen`** — teardown now resumes the prior continuation and cancels the open timeout (mirroring `disconnect()`); the resume-prior that lived in `waitForSocketOpen` is now redundant and was removed.
- **[nit] `failSocketOpen` didn't invalidate the transport** — it now does (`transport?.invalidate(); transport = nil`).
- **[nit] Test-double fidelity** — `Gate` gained a sticky `signalled` flag (no lost signal); `FakeSocket.cancel` records the supplied `closeCode` (with a `cancelErrorsReceive` toggle so the stale-frame test can deliver after cancel); tests await deterministic resume/receive-pending signals and clean up via `disconnect()`.

## Considered and kept
- *Shared `FakeTransport` across connects* (production creates a fresh one per connect). Kept as a test simplification — it doesn't model stale-transport reuse, but `testConnectInvalidatesPriorTransport` still validates that `connect()` calls `invalidate()`. Not worth the per-test plumbing for the current suite.

PR #27, opened 2026-08-09
Fix iOS image paste from keyboard providers

kaishi00:
Closes #26

## Summary

- Restore image insertion from iOS 27's keyboard “Photo to paste” action.
- Load image data through UIKit's `NSItemProvider` paste path.
- Declare text and image paste support during UIKit preflight.
- Surface provider load failures through the composer's existing error UI.
- Reuse the existing composer attachment pipeline; no backend changes are required.
- Preserve ordinary text-provider and legacy paste behavior.

## Root cause

`ImagePasteTextView` only handled the legacy `paste(_:)` path and inspected
`UIPasteboard.general`. The iOS 27 keyboard action supplies an image through
UIKit's item-provider paste API, so the existing `onPastedImage` callback was
never reached.

## Validation

- Regression test was red before the fix: image callback expectation timed out.
- Focused composer tests pass: 4 tests, 0 failures.
- Full iOS test suite passes: 222 tests, 0 failures.
- Physical iOS 27 validation remains pending because no device was connected
  to the development workspace; the reported keyboard action should be
  verified on-device before merging.

## Summary by CodeRabbit

* **New Features**
  * Restored support for pasting images from the iOS keyboard and photo picker.
  * Added asynchronous image loading with fallback conversion and improved attachment metadata.

* **Bug Fixes**
  * Preserved text-paste behavior when pasted content is not an image.
  * Added visible error messages and haptic feedback when image pasting fails.

* **Tests**
  * Added coverage for image pasting, fallback handling, capability detection, and failures.

* **Documentation**
  * Added design and implementation documentation for image-paste support.

Issue #26, opened 2026-08-09
[BUG] iOS 27 Photo to paste action blinks but adds no image


Brandon168:
## Bug

On iOS 27.0 beta, the iOS keyboard shows the new **Photo to paste** clipboard action above the Hermes Conduit composer. Tapping it briefly blinks, but Hermes Conduit does not add an attachment, change the composer, show an error, or send anything.

## Environment

- App: Hermes Conduit
- App version: 0.1.2 (current App Store release)
- OS: iOS 27.0 beta
- Device: not recorded
- Backend: Hermes gateway

## Steps to reproduce

1. Copy an image or screenshot to the iOS clipboard.
2. Open a Hermes Conduit chat.
3. Focus the message composer so the iOS keyboard is visible.
4. Tap the **Photo to paste** clipboard action.
5. Observe the result.

## Actual behavior

The paste control blinks as if the action was accepted, then remains in place. No image preview or attachment chip appears, the composer does not change, and no error is shown.

## Expected behavior

The image should be imported into the composer and shown as an attachment before the message is sent.

Please advise whether this is an iOS 27 beta compatibility issue or a Hermes Conduit clipboard-handling bug.

kaishi00:
I just tried it myself and seems like there was a regression at some point. I'll look into it, thanks for the report.

PR #53, opened 2026-08-13
Offer Paste in composer edit menu for image pasteboards

kaishi00:
## Summary

Follow-up to #27 / refs #26. Pasting an image into the composer still did not work: long-pressing the composer showed only "Autofill", with no "Paste" item, so the existing image-paste code was unreachable from the edit menu.

- Override `canPerformAction(_:withSender:)` on `ImagePasteTextView` to surface the standard "Paste" edit-menu item whenever the general pasteboard advertises image content.
- Add a privacy-safe `pasteboardContainsImage()` helper that uses metadata-only checks (`contains(pasteboardTypes:)` + `NSItemProvider` conformance) so it does **not** raise the iOS "Paste from Other App" banner.
- Add regression coverage and a `tearDown` that clears `UIPasteboard.general`.

## Root cause

Image paste in `ComposerPasteTextView.swift` is implemented through three paths — `canPaste(_:)`/`paste(itemProviders:)` (the responder paste system, for external triggers like iOS 27's keyboard "Photo to paste") and the legacy `paste(_:)` selector. **None of those gate the long-press edit menu**, whose items are controlled by `canPerformAction(_:withSender:)`. That was never overridden (no `UIEditMenuInteraction`, no `UIPasteControl` anywhere in the app).

For an image-only pasteboard, UITextView's default `canPerformAction(paste:)` returns false — it cannot paste an image as text — so "Paste" is dropped, and iOS surfaces system items like "Autofill" instead. PR #27 fixed the item-provider code path but never addressed edit-menu gating, so the menu never offered Paste.

With this change, tapping "Paste" dispatches the existing `paste(_:)` selector, which reads `UIPasteboard.general` and delivers the image through the existing `onPastedImage` → `addAttachment` path. No backend, model, or UX-layout changes.

## Validation

- Regression test mirrors the failing behavior: `UITextView` drops "Paste" for an image-only pasteboard; the new test asserts `canPerformAction(#selector(paste(_:)))` is `true`.
- Focused composer tests cover pasteboard detection (image vs. text-only) and the `canPerformAction` regression.
- ⚠️ **Not yet built/run from this Windows dev environment.** Needs `xcodegen` + `ComposerPasteTextViewTests` on a Mac or the macos-26 CI runner before merge.
- On-device check: copy a screenshot → long-press the composer → "Paste" should now appear and add the image attachment.

## Considered alternative

A dedicated `UIPasteControl` button would guarantee availability without the edit menu, but it changes the composer's UX (new visible control). The `canPerformAction` override is the minimal root-cause fix that preserves the existing long-press → Paste gesture.

## Summary by CodeRabbit

- **New Features**
  - Added support for pasting images into editable text areas using the standard Paste action.

- **Bug Fixes**
  - Improved paste detection so image content is recognized even when text is unavailable, making image pasting more reliable.

- **Tests**
  - Added coverage for image and text paste behavior, including a regression test confirming image pasting remains available through the standard menu action.

### DIFF
--- a/Conduit/Views/Components/ComposerPasteTextView.swift
+++ b/Conduit/Views/Components/ComposerPasteTextView.swift
@@ -146,6 +146,45 @@ final class ImagePasteTextView: UITextView {
         return super.canPaste(itemProviders)
     }
 
+    /// Surfaces the standard "Paste" edit-menu item when an image is on the
+    /// pasteboard. UITextView drops "Paste" for image-only content (it cannot
+    /// paste an image as text), leaving system items such as "Autofill" as the
+    /// only options. `canPaste`/`paste(itemProviders:)` route external paste
+    /// triggers but never gate the long-press menu, so without this override
+    /// the image paste code is unreachable from the edit menu.
+    override func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
+        if action == #selector(paste(_:)), shouldOfferImagePaste() {
+            return true
+        }
+        return super.canPerformAction(action, withSender: sender)
+    }
+
+    /// Whether the edit menu should offer "Paste" for an image. Extracted as a
+    /// pure function over `isEditable` and `pasteboardContainsImage()` so the
+    /// gate is unit-testable in isolation, without depending on `super` or the
+    /// view's first-responder/window state.
+    func shouldOfferImagePaste() -> Bool {
+        isEditable && pasteboardContainsImage()
+    }
+
+    /// Reports whether the general pasteboard advertises image content using
+    /// only banner-safe metadata access: `hasImages` (Apple's conformance-aware
+    /// detector) plus a conformance scan over `pasteboard.types` as a
+    /// belt-and-suspenders fallback. Both read type metadata rather than
+    /// contents, so they do NOT raise the iOS "Paste from Other App" prompt —
+    /// important because `canPerformAction` runs before the user consents to a
+    /// paste and is called repeatedly while the edit menu is displayed. This
+    /// deliberately avoids `pb.image`/`pb.data(...)`, `pb.url`, and
+    /// materializing `pb.itemProviders`, all of which can trigger the banner.
+    /// Note: an image *URL* without image data (e.g. copied from Safari) is not
+    /// detected here, so "Paste" is not offered for that case; the legacy
+    /// `pb.url` path in `paste(_:)` stays reachable only via external triggers.
+    func pasteboardContainsImage() -> Bool {
+        let pasteboard = UIPasteboard.general
+        if pasteboard.hasImages { return true }
+        return pasteboard.types.contains { UTType($0)?.conforms(to: .image) ?? false }
+    }
+
     override func layoutSubviews() {
         super.layoutSubviews()
         let contentHeight = bounds.width > 0 && contentSize.height.isFinite
@@ -234,6 +273,10 @@ final class ImagePasteTextView: UITextView {
         }
     }
 
+    /// When the pasteboard holds an image it is delivered as an attachment via
+    /// `onPastedImage`; any accompanying text is not inserted, since this
+    /// attachment composer treats the image as the payload. Non-image
+    /// pasteboards fall through to the standard text behavior.
     override func paste(_ sender: Any?) {
         let pb = UIPasteboard.general
 
--- a/ConduitTests/ComposerPasteTextViewTests.swift
+++ b/ConduitTests/ComposerPasteTextViewTests.swift
@@ -7,6 +7,95 @@ import XCTest
 
 @MainActor
 final class ComposerPasteTextViewTests: XCTestCase {
+    private var savedPasteboardItems: [[String: Any]] = []
+
+    override func setUp() {
+        super.setUp()
+        // Snapshot the system pasteboard so tearDown can restore it instead of
+        // clobbering a real clipboard (relevant when the suite runs on a device).
+        savedPasteboardItems = UIPasteboard.general.items
+    }
+
+    override func tearDown() {
+        UIPasteboard.general.items = savedPasteboardItems
+        super.tearDown()
+    }
+
+    func testPasteboardContainsImageTrueForImagePasteboard() {
+        let view = ImagePasteTextView()
+        UIPasteboard.general.image = Self.fixtureImage()
+
+        XCTAssertTrue(view.pasteboardContainsImage())
+    }
+
+    func testPasteboardContainsImageFalseForTextPasteboard() {
+        let view = ImagePasteTextView()
+        UIPasteboard.general.string = "just text"
+
+        XCTAssertFalse(view.pasteboardContainsImage())
+    }
+
+    func testCanPerformActionOffersPasteForImagePasteboard() {
+        let view = ImagePasteTextView()
+        view.isEditable = true
+        UIPasteboard.general.image = Self.fixtureImage()
+
+        // Regression: UITextView drops "Paste" for an image-only pasteboard, so
+        // the long-press edit menu only offered system items like "Autofill".
+        // canPerformAction must surface paste: so the existing paste(_:) path
+        // becomes reachable from the menu.
+        XCTAssertTrue(view.canPerformAction(#selector(UIResponder.paste(_:)), withSender: nil))
+    }
+
+    func testShouldOfferImagePasteTrueForImagePasteboard() {
+        let view = ImagePasteTextView()
+        view.isEditable = true
+        UIPasteboard.general.image = Self.fixtureImage()
+
+        XCTAssertTrue(view.shouldOfferImagePaste())
+    }
+
+    func testShouldOfferImagePasteFalseWhenNotEditable() {
+        let view = ImagePasteTextView()
+        view.isEditable = false
+        UIPasteboard.general.image = Self.fixtureImage()
+
+        // A disabled composer must not offer image paste even with an image on
+        // the pasteboard.
+        XCTAssertFalse(view.shouldOfferImagePaste())
+    }
+
+    func testShouldOfferImagePasteFalseForTextOnlyPasteboard() {
+        let view = ImagePasteTextView()
+        view.isEditable = true
+        UIPasteboard.general.string = "just text"
+
+        // Text-only content must not be treated as an image paste.
+        XCTAssertFalse(view.shouldOfferImagePaste())
+    }
+
+    func testPasteSelectorDeliversImageFromPasteboard() async {
+        // End-to-end check of the exact path a menu tap now dispatches: the
+        // legacy paste(_:) selector reads UIPasteboard.general and fires
+        // onPastedImage. This is independent of the canPerformAction gate.
+        let view = ImagePasteTextView()
+        UIPasteboard.general.image = Self.fixtureImage()
+
+        let callback = expectation(description: "paste(_:) delivered image")
+        view.onPastedImage = { pastedImage in
+            XCTAssertFalse(pastedImage.data.isEmpty)
+            XCTAssertEqual(pastedImage.typeIdentifier, UTType.png.identifier)
+            callback.fulfill()
+        }
+        view.onPastedImageError = { message in
+            XCTFail("paste(_:) should deliver the image, got: \(message)")
+        }
+
+        view.paste(nil as Any?)
+
+        await fulfillment(of: [callback], timeout: 5.0)
+    }
+
     func testPasteItemProvidersDeliversImageData() async {
         let view = ImagePasteTextView()
         let expectedData = Data([0x89, 0x50, 0x4E, 0x47])
@@ -253,6 +342,13 @@ final class ComposerPasteTextViewTests: XCTestCase {
             ComposerBar.pastedImageErrorMessage("The image provider failed."),
             "Could not paste image: The image provider failed."
         )
+    }
+
+    private static func fixtureImage() -> UIImage {
+        UIGraphicsImageRenderer(size: CGSize(width: 2, height: 2)).image { context in
+            UIColor.systemOrange.setFill()
+            context.fill(CGRect(x: 0, y: 0, width: 2, height: 2))
+        }
     }
 
     private func encodedImageData(_ image: UIImage, type: UTType) -> Data? {


kaishi00:
## Summary
This PR makes image pasting reachable from the composer's long-press edit menu by overriding `canPerformAction(_:withSender:)` on `ImagePasteTextView` so "Paste" appears whenever the general pasteboard advertises image content, using metadata-only checks to avoid the iOS paste banner, plus regression tests and pasteboard cleanup.

## Findings

### [severity: minor] Repeated, unbounded pasteboard queries from `canPerformAction`
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:155-172
**Issue:** UIKit calls `canPerformAction(_:withSender:)` frequently (while the edit menu is displayed, on selection changes, keystrokes, etc.). Every `paste:` query now hits `UIPasteboard.general`, and when `contains(pasteboardTypes:)` misses it materializes `pasteboard.itemProviders` and runs provider metadata queries. With multi-item pasteboards this work repeats on every query with no caching.
**Suggestion:** Use the cheap, banner-safe `UIPasteboard.general.hasImages` check as the primary gate, falling back to item-provider conformance only when needed, and consider memoizing the result per menu-display run (or invalidate on `UIPasteboard.changedNotification`).

### [severity: minor] Image-URL paste path remains unreachable from the edit menu
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:166-172 vs 272-283
**Issue:** `paste(_:)` supports pasting an image via `pb.url` (e.g. copied from Safari/Photos), but `pasteboardContainsImage()` only recognizes pasteboards advertising `public.image` or image-conforming item providers. A pasteboard holding only an image URL returns `false`, so "Paste" is still suppressed and that legacy path stays dead from the menu.
**Suggestion:** Optionally extend `pasteboardContainsImage()` to also return `true` when `pb.url`'s extension is in the image-extensions list, or drop the URL path if it's intentionally unreachable.

### [severity: minor] "Metadata-only, no banner" claim is unverified
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:162-172
**Issue:** The design hinges on `contains(pasteboardTypes:)` and `NSItemProvider` conformance checks not raising the "Paste from Other App" banner. `contains(pasteboardTypes:)` and `hasItemConformingToTypeIdentifier` are documented as banner-safe, but `pasteboard.itemProviders` access is a gray area across iOS versions, and the PR is explicitly unbuilt/unrun, so the claim hasn't been validated on-device.
**Suggestion:** Verify on a device/simulator (copy a screenshot from another app, long-press, confirm no banner before tapping Paste). Prefer `hasImages` + conformance checks and minimize `itemProviders` materialization.

### [severity: nit] Mixed image+text pasteboard silently drops the text
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:262-268
**Issue:** For a pasteboard containing both image and text (or when text is selected in the composer), `canPerformAction(paste:)` returns `true`, but `paste(_:)` delivers only the image and never replaces the selected text. This is pre-existing `paste(_:)` behavior, but the new menu gating now exposes it more readily.
**Suggestion:** Document this behavior or have the image paste path clear/replace the current selection so the result matches user expectation.

### [severity: nit] Negative-path test coverage missing
**File:** ConduitTests/ComposerPasteTextViewTests.swift:31-41
**Issue:** The regression test covers the positive case (image pasteboard → `canPerformAction(paste:)` == true) but not the inverse: no test asserts Paste is *not* offered for a text-only pasteboard or when `isEditable == false`, nor the mixed image+text case.
**Suggestion:** Add assertions that `canPerformAction(paste:)` is `false` for text-only content and for a non-editable view to lock in the gating logic.

## Verdict
APPROVE

The change is small, well-documented, and correctly targets the root cause (edit-menu gating). No retain cycles, force unwraps, or concurrency violations were introduced in the new code. Findings are minor/nit; the primary residual risk is the unverified banner claim, which should be checked on-device before or shortly after merge.

<a href="https://opencode.ai/s/hLcBeygM">New%20session%20-%202026-08-13T04%3A46%3A14.472Z</a>
[opencode session](https://opencode.ai/s/hLcBeygM)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31668008215)

kaishi00:
## Summary

Follow-up to #27: overrides `canPerformAction(_:withSender:)` on `ImagePasteTextView` so the standard "Paste" edit-menu item appears when the pasteboard advertises image content, with metadata-only detection to (claimedly) avoid the iOS "Paste from Other App" banner, plus regression tests and pasteboard cleanup.

## Findings

### [severity: major] `itemProviders` access undermines the stated "no banner" design
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:166-172
**Issue:** The helper's doc comment claims metadata-only, banner-safe access, but the fallback `pasteboard.itemProviders.contains { canLoadImageProvider($0) }` materializes the pasteboard's item providers. Per Apple's guidance (WWDC "Meet paste controls"), the banner-safe checks are the metadata properties (`hasStrings`/`hasImages`/`hasURLs`/`contains(pasteboardTypes:)`); reading content — including `itemProviders` — is what raises the "Paste from Other App" prompt. Since `canPerformAction` is invoked repeatedly (menu display, selection changes, keystrokes), a provider-conformance pasteboard would not only prompt but prompt repeatedly, defeating the entire purpose of this change. The author's own note confirms this was never validated on-device.
**Suggestion:** Gate primarily on `pasteboard.hasImages || pasteboard.contains(pasteboardTypes: [UTType.image.identifier])` (both banner-safe). Either drop the `itemProviders` conformance fallback or verify on-device that it does not prompt; if it does, accept the tradeoff explicitly or find a non-materializing conformance check (e.g. inspect `pasteboard.types` for conforming UTIs).

### [severity: minor] Expensive, unbounded pasteboard queries from `canPerformAction`
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:155-160
**Issue:** UIKit calls `canPerformAction(_:withSender:)` very frequently, and every `paste:` query now runs `pasteboardContainsImage()`. When the cheap `contains` check misses, it materializes `itemProviders` and runs `imageTypeIdentifier(for:)` (including `provider.hasItemConformingToTypeIdentifier`, a potentially remote/provider-backed query) with no caching or invalidation. This is repeated work per query, on the main thread, every time the responder menu is evaluated.
**Suggestion:** Rely on the cheap metadata check first; avoid materializing providers per query, and memoize the result across the menu-display lifetime (e.g. cache and invalidate on `UIPasteboard.changedNotification`).

### [severity: minor] Image-URL paste path remains unreachable from the edit menu
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:166-172 vs 262-283
**Issue:** `paste(_:)` supports pasting an image via `pb.url` (e.g. copied from Safari/Photos), but `pasteboardContainsImage()` only recognizes pasteboards advertising `public.image` or image-conforming providers. A pasteboard holding only an image URL returns `false`, so "Paste" is still suppressed and the URL path stays dead from the menu — the same class of bug this PR fixes.
**Suggestion:** Have `pasteboardContainsImage()` also return `true` when `pb.url` has an image file extension, or drop the URL branch if it's intentionally unreachable.

### [severity: nit] Mixed image+text paste silently drops the text / doesn't replace selection
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:262-268
**Issue:** For a pasteboard with both image and text (or with text selected in the composer), `canPerformAction(paste:)` now returns `true` (the new gating exposes this more readily), but `paste(_:)` delivers only the image and never replaces the selected text.
**Suggestion:** Document the behavior, or have the image path replace/clear the current selection to match user expectations.

### [severity: nit] Missing negative-path coverage for `canPerformAction`
**File:** ConduitTests/ComposerPasteTextViewTests.swift:31-41
**Issue:** The regression test covers the positive case only. There is no assertion that Paste is *not* offered for a text-only pasteboard or for a non-editable view, so the `isEditable`/content gating logic is unverified (`pasteboardContainsImage` false-for-text is tested separately, but not through `canPerformAction`).
**Suggestion:** Add cases asserting `canPerformAction(#selector(paste(_:)))` is `false` for a text-only pasteboard and for `isEditable == false`.

## Verdict
REQUEST_CHANGES

The change correctly targets the root cause (edit-menu gating) with no force unwraps, retain cycles, or concurrency violations introduced; the `[weak self]` capture discipline and `@MainActor` tests are fine. However, the central design claim — banner-safe detection — is very likely violated by the `itemProviders` fallback in the exact conformance path the fallback exists for, and `canPerformAction`'s high call frequency makes that a repeated prompt, not a one-off. This should be resolved or validated on-device before merge.

<a href="https://opencode.ai/s/NsTIt5KK">New%20session%20-%202026-08-13T12%3A48%3A36.735Z</a>
[opencode session](https://opencode.ai/s/NsTIt5KK)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31701711922)

kaishi00:
Thanks for the reviews — addressed in `308b650`. Calibrating each finding against the code:

**[major] \`itemProviders\` access undermines the "no banner" design — fixed.** Valid and important. The \`pasteboard.itemProviders.contains { … }\` fallback materializes providers, which is outside Apple's banner-safe set and could prompt repeatedly (since \`canPerformAction\` fires on every menu/selection/keystroke eval). Replaced detection with the banner-safe APIs only:

\`\`\`swift
func pasteboardContainsImage() -> Bool {
    let pasteboard = UIPasteboard.general
    return pasteboard.hasImages
        || pasteboard.contains(pasteboardTypes: [UTType.image.identifier])
}
\`\`\`

\`hasImages\` / \`contains(pasteboardTypes:)\` read type metadata, not contents, so they're safe to call from the pre-consent hot path. \`itemProviders\` is no longer touched in the gate. (It's still used in \`paste(_:)\` line 268, but that runs *after* the user consents to a paste, so prompting there is expected.)

**[minor] Expensive/unbounded queries — resolved by the above.** \`hasImages\`/\`contains\` are the cheap, Apple-blessed detectors built exactly for these hot paths. Skipped memoization as it would add \`changedNotification\` bookkeeping for no real gain now that each call is a metadata read.

**[minor] Image-URL paste path unreachable from the menu — documented, not expanded.** Correct that \`pb.url\`-only pasteboards (Safari "copy image address") won't surface "Paste". Reading \`pb.url\` in the gate to detect that would re-introduce exactly the banner risk the major finding flags, so I kept the gate banner-safe and documented the limitation in the doc comment. The legacy \`pb.url\` branch stays for external paste triggers; widening the gate to cover it is a separate, banner-risk tradeoff.

**[nit] Mixed image+text drops the text — documented, behavior kept.** For an attachment composer, image→attachment (text ignored) matches the composer's model; clearing/replacing the selection isn't desired. Left as-is.

**[nit] Missing negative-path coverage — added.** Extracted the gate into \`shouldOfferImagePaste() -> Bool\` (\`isEditable && pasteboardContainsImage()\`) so both gating dimensions are unit-testable without depending on \`super\` or first-responder/window state. Added three cases: image+editable (true), image+not-editable (false), text-only (false), alongside the existing \`canPerformAction\` integration test.

Still flagging the same caveat from the PR body: this is **not yet validated on-device** — CI proves compile + unit assertions, but the real "does Paste appear without a banner" check needs a simulator/device pass (copy a screenshot from another app, long-press the composer) before merge.

kaishi00:
## Summary
Overrides `canPerformAction(_:withSender:)` on `ImagePasteTextView` so the standard "Paste" edit-menu item appears whenever the general pasteboard advertises image content, using only banner-safe metadata APIs (`hasImages` / `contains(pasteboardTypes:)`), plus regression tests with pasteboard cleanup.

## Findings

### [severity: minor] `contains(pasteboardTypes:)` matches exact UTIs, not conforming types
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:181-185
**Issue:** `UIPasteboard.contains(pasteboardTypes:)` matches the pasteboard's *exact* type-identifier strings, not conformance. Most image pasteboards register concrete types (`public.png`, `public.jpeg`, `com.apple.uikit.image`), so the `contains(pasteboardTypes: [UTType.image.identifier])` clause is effectively dead code in those cases, and all real-world detection is carried by `hasImages` alone (which is conformance-aware). The OR of the two gives a false impression of broader coverage.
**Suggestion:** Keep `hasImages` as the primary gate and either drop the `contains` clause or scan the (also banner-safe) `pasteboard.types` with `UTType($0)?.conforms(to: .image)` if broader UTI coverage is intended.

### [severity: minor] Menu "Paste" now surfaces for mixed image+text pasteboards but drops the text
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:155-160, 275-312
**Issue:** With an image and text both on the pasteboard, `pasteboardContainsImage()` returns `true`, so "Paste" appears; tapping it routes to `paste(_:)`, which checks `pb.image` first and delivers only the image — the text is silently dropped and any selected composer text is left untouched. This is pre-existing `paste(_:)` behavior, but the new gating makes it the common outcome for mixed boards.
**Suggestion:** Acceptable for an attachment composer, but consider documenting it in the method comment (the "Paste" menu item implies replace-selection semantics to most users), or clear/replace the selection when an image is delivered.

### [severity: minor] Image-URL pasteboards still never offer "Paste" from the menu
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:177-180 vs 284-296
**Issue:** The `pb.url` image branch in `paste(_:)` (e.g. Safari "Copy Link") remains unreachable from the edit menu because `pasteboardContainsImage()` ignores URL-only pasteboards. This is deliberately documented and the tradeoff against banner risk is reasonable, but it means a common user action (copy image link → long-press) still shows no Paste.
**Suggestion:** Accept as documented; if the URL path is worth supporting, note that `pasteboard.types` is banner-safe metadata, so checking URL-extension heuristics via `types` (not `pb.url`) could widen the gate without re-introducing the banner risk.

### [severity: nit] No test for the full menu → `paste(_:)` → `onPastedImage` path
**File:** ConduitTests/ComposerPasteTextViewTests.swift:31-41
**Issue:** `testCanPerformActionOffersPasteForImagePasteboard` verifies the menu gate, and `paste(itemProviders:)` is tested, but there is no test that invokes the legacy `paste(_:)` selector with an image on `UIPasteboard.general` and asserts `onPastedImage` fires. Since that is the exact path a menu tap now dispatches, the wiring between the new gate and the delivery code is untested end-to-end.
**Suggestion:** Add a test calling `view.paste(nil)` with `UIPasteboard.general.image` set and asserting the `onPastedImage` callback receives the expected PNG data.

### [severity: nit] `paste(_:)` URL task is uncancellable and swallows failures
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:289-293
**Issue:** Pre-existing, not introduced by this PR: the `URLSession.shared.dataTask` for image URLs has no `[weak self]` guard on error (silent `return` on nil data), no cancellation on view deallocation, and never calls `onPastedImageError`. Now that the menu path is wired, this branch is slightly more reachable.
**Suggestion:** Surface failures via `onPastedImageError` and consider `retain`/`cancel` bookkeeping tied to the view's lifetime.

## Verdict
APPROVE

The change correctly targets the root cause (edit-menu gating), the prior major concern about `itemProviders` materialization triggering the "Paste from Other App" banner was resolved by gating solely on `hasImages`/`contains(pasteboardTypes:)`, and negative-path test coverage was added. No force unwraps, retain cycles, or concurrency violations were introduced; the remaining findings are minor/nit. The one residual risk is the unvalidated-on-device banner claim, which should be confirmed on a simulator/device before or shortly after merge.

<a href="https://opencode.ai/s/Upepit4a">New%20session%20-%202026-08-13T13%3A07%3A07.213Z</a>
[opencode session](https://opencode.ai/s/Upepit4a)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31703195587)

kaishi00:
## Summary
This PR makes image paste reachable from the composer's long-press edit menu by overriding `canPerformAction(_:withSender:)` on `ImagePasteTextView` so the standard "Paste" item appears when the general pasteboard advertises image content, gated on a banner-safe metadata-only check (`hasImages` + a conformance scan of `pasteboard.types`), plus regression tests and pasteboard snapshot/restore in `setUp`/`tearDown`.

## Findings

### [severity: minor] Core "no banner" claim remains unvalidated on-device
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:170-186
**Issue:** The entire PR (per the author's own note) has never been built or run — only compile + unit-test assertions. The central design claim — that `hasImages`/`pasteboard.types` metadata reads do not raise the "Paste from Other App" banner when `canPerformAction` runs pre-consent — is documented as Apple-blessed but is not verified in the real long-press flow. Since `canPerformAction` is the hot path where a banner would be most disruptive, this is the main residual risk.
**Suggestion:** Before merge (or immediately after), validate on a simulator/device: copy a screenshot from another app, long-press the composer, confirm "Paste" appears and no banner shows until the user actually taps Paste. This can be a manual gate or an XCUITest.

### [severity: minor] Unbounded pasteboard reads from the `canPerformAction` hot path
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:155-186
**Issue:** UIKit calls `canPerformAction(_:withSender:)` frequently (menu display, selection changes, keystrokes). Every `paste:` query invokes `pasteboardContainsImage()`, and when `hasImages` is false it scans `pasteboard.types` (a cross-process read on the main thread) with no caching or invalidation. It's metadata-only so it's cheap relative to the pre-fix `itemProviders` materialization, but the work is still repeated per query for the common text/empty-pasteboard case.
**Suggestion:** Acceptable as-is given `hasImages` short-circuits image boards; if profiling shows it matters, memoize per menu-display run and invalidate on `UIPasteboard.changedNotification`. At minimum, confirm `hasImages` covers the common copy-from-Photos/Screenshot cases so the `types` fallback is rarely hit.

### [severity: nit] Negative-path `canPerformAction` coverage is indirect
**File:** ConduitTests/ComposerPasteTextViewTests.swift:31-75
**Issue:** The negative cases (`testShouldOfferImagePasteFalseWhenNotEditable`, `testShouldOfferImagePasteFalseForTextOnlyPasteboard`) exercise `shouldOfferImagePaste()` directly, but the `canPerformAction` integration test only covers the positive case. Since `canPerformAction` delegates to `shouldOfferImagePaste` for `paste:`, the logic is covered, but the selector-path wiring for the negative branches is not asserted.
**Suggestion:** Optional: add `XCTAssertFalse(view.canPerformAction(#selector(UIResponder.paste(_:)), withSender: nil))` for the text-only and non-editable cases to lock in the full gate through the responder API.

### [severity: nit] Non-editable view still routes external paste triggers into `paste(_:)`
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:280-317
**Issue:** `canPerformAction`/`shouldOfferImagePaste` gate the menu on `isEditable`, but `paste(_:)` itself has no `isEditable` guard. An external trigger (iOS 27 keyboard "Photo to paste", or a paste configuration delivery) can still invoke `paste(_:)` on a disabled composer and deliver an attachment. The menu path is correctly gated, so this is only reachable via non-menu triggers.
**Suggestion:** Consider a `guard isEditable else { return super.paste(sender) }` at the top of `paste(_:)` so the disabled state is enforced at delivery time, not just in the menu.

## Verdict
APPROVE

The change is small, well-documented, and correctly targets the root cause (edit-menu gating). The prior major concern about `itemProviders` materialization triggering the paste banner was properly resolved — the gate now uses only metadata APIs (`hasImages` + `pasteboard.types` conformance scan), the pasteboard is snapshot/restored in tests rather than cleared, negative-path coverage was added, and end-to-end `paste(_:)` → `onPastedImage` delivery is tested. No force unwraps, retain cycles, or concurrency violations were introduced; `[weak self]` discipline and `@MainActor` usage are correct. The one real residual risk is that the "no banner" behavior is unverified on-device, so confirm that before or shortly after merge.

<a href="https://opencode.ai/s/Euv2YlRS">New%20session%20-%202026-08-13T13%3A21%3A22.269Z</a>
[opencode session](https://opencode.ai/s/Euv2YlRS)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31704393570)

PR #54, opened 2026-08-13
Keep pending decisions visible after background resume

kaishi00:
## Summary

- keep unresolved clarification and approval cards visible and answerable after a background resume, including when the gateway omits the one-shot decision event
- keep an explicit `running == false` session idle so the composer remains a normal send action instead of showing false busy/stop/steer state
- use shared pending-decision keys and reconciliation so gateway-resolved records replace cached pending cards without duplicates
- persist gateway-provided pending cards with the existing bounded unconfirmed expiry safeguard, including across subsequent cache flushes
- add regression coverage for cached and gateway-provided clarification and approval cards

## Root cause

The resume path treated pending approvals as restorable only while the gateway reported an active turn, and explicitly pruned gateway approvals when `running == false`. Clarifications were restored by forcing `turnState` to `.running`, which made the UI report a false busy state and route composer input through steer/interrupt behavior. A background resume can omit the one-shot decision event while still reporting the preceding turn as settled, so both decision types need independent presentation restoration and the settled turn state must remain separate from card actionability.

Gateway-provided pending cards also had no restoration guard. A later background or streaming cache flush passed an empty unconfirmed-key set to the cache, clearing the 24-hour expiry marker and allowing a stale card to persist indefinitely. Ordinary cache writes now derive the pending decision keys directly from the current presentation so the original marker survives until the card is resolved or expires.

## Validation

- Focused `SessionPresentationCacheTests`: 33 passed, 0 failed
- Full iOS Simulator XCTest suite: 510 passed, 0 failed
- New regression confirms a background cache flush preserves the original pending-decision expiry marker
- `git diff --check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending clarification and approval cards now persist when resuming from the background, even if the session appears inactive.
  * Resolved decisions are correctly removed when authoritative updates are available.
  * Pending cards remain available during background cache updates and can expire safely when no longer confirmed.
  * Restored cards now retain composer availability and decision context more reliably.

* **Documentation**
  * Added design and implementation guidance for background session resume behavior.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2680,33 +2680,46 @@ final class AppState: ObservableObject {
             for: result.sessionId,
             fallbackSessionId: reconciliation?.requestedSessionId
         )
-        // Hermes can omit `running` on versions that still support paused
-        // clarify/approval turns. Only an explicit false means the turn is
-        // settled, so nil remains eligible for restoring locally observed
-        // pending decision cards.
-        let restorePendingCards = Self.shouldRestorePendingCards(running: result.snapshot.running)
+        // A clarification request is a user-actionable event, not merely a
+        // projection of the gateway's current turn state. Hermes can omit the
+        // one-shot clarify event from a resume, and `running == false` can
+        // describe the preceding text even while that clarification remains
+        // unresolved. Keep clarification restoration independent of running;
+        // approvals retain the older explicit-running eligibility rule.
+        let restorePendingClarifications = true
+        let restorePendingApprovals = Self.shouldRestorePendingCards(running: result.snapshot.running)
         let sessionIDs = [result.sessionId, reconciliation?.requestedSessionId].compactMap { $0 }
         let gatewayDecisionKeys = Set(result.messages.compactMap(SessionPresentationCache.decisionKey(for:)))
-        let retainedMessages = restorePendingCards
-            ? retainedRestoredMessages.filter {
-                guard let key = SessionPresentationCache.decisionKey(for: $0) else { return false }
-                return !gatewayDecisionKeys.contains(key)
+        let retainedMessages = retainedRestoredMessages.filter { message in
+            guard let key = SessionPresentationCache.decisionKey(for: message),
+                  !gatewayDecisionKeys.contains(key) else {
+                return false
             }
-            : []
+            if message.clarify != nil { return restorePendingClarifications }
+            return restorePendingApprovals
+        }
         let restored = sessionPresentationCache.merge(
             result.messages + retainedMessages,
             profile: activeProfile,
             sessionIDs: sessionIDs,
-            includePendingClarifications: restorePendingCards,
-            includePendingApprovals: restorePendingCards
+            includePendingClarifications: restorePendingClarifications,
+            includePendingApprovals: restorePendingApprovals
         )
         messages = mergeCachedReviews(into: restored, sessionId: result.sessionId)
         noteChatViewportTranscriptReplacement()
         let gatewayPendingDecisionKeys = SessionPresentationCache.pendingDecisionKeys(in: result.messages)
-        if result.snapshot.running == false && !gatewayPendingDecisionKeys.isEmpty {
+        let gatewayPendingClarificationKeys = Set(result.messages.compactMap { message -> String? in
+            guard let clarify = message.clarify,
+                  SessionPresentationCache.isPendingDecision(clarify.status) else {
+                return nil
+            }
+            return "clarify:\(clarify.requestId)"
+        })
+        let gatewayPendingApprovalKeys = gatewayPendingDecisionKeys.subtracting(gatewayPendingClarificationKeys)
+        if result.snapshot.running == false && !gatewayPendingApprovalKeys.isEmpty {
             messages = SessionPresentationCache.removingPendingDecisionPresentation(
                 from: messages,
-                matching: gatewayPendingDecisionKeys
+                matching: gatewayPendingApprovalKeys
             )
         }
         let restoredPendingDecisionKeys = SessionPresentationCache
@@ -2736,6 +2749,7 @@ final class AppState: ObservableObject {
         // marker until Hermes confirms the turn or the user interacts with it.
         let gatewayConfirmsActiveTurn = result.snapshot.running == true
             || (result.snapshot.running != false && gatewaySentPendingDecision)
+            || !gatewayPendingClarificationKeys.isEmpty
         let unconfirmedPendingDecisionKeys = result.snapshot.running != true
             ? restoredPendingDecisionKeys
             : []
@@ -2779,11 +2793,15 @@ final class AppState: ObservableObject {
         receivedReasoningForCurrentTurn = false
         applyRuntime(result.snapshot)
 
-        // A paused clarification or approval can have neither a live text
-        // projection nor an explicit `running` field. The restored card is
-        // actionable evidence for this UI, so keep the composer enabled for
-        // it instead of treating the gateway as unsupported.
-        if result.snapshot.running == nil && (result.snapshot.hasLiveProjection || hasPendingDecision) {
+        // A pending clarification can survive a resume that reports the
+        // preceding turn as settled. Its answer controls remain actionable,
+        // so keep the composer enabled even when `running == false`.
+        let hasPendingClarification = messages.contains { message in
+            guard let clarify = message.clarify else { return false }
+            return SessionPresentationCache.isPendingDecision(clarify.status)
+        }
+        if hasPendingClarification
+            || (result.snapshot.running == nil && (result.snapshot.hasLiveProjection || hasPendingDecision)) {
             turnState = .running
         } else if TurnState.fromGatewayRunning(result.snapshot.running) == .unsupportedGateway {
             turnState = .unsupportedGateway
@@ -2795,10 +2813,10 @@ final class AppState: ObservableObject {
         return true
     }
 
-    /// Testable helper: derives whether pending clarify/approval cards should
-    /// be restored from the presentation cache on resume. An explicit false
-    /// is the only settled-turn signal; nil is still eligible because older
-    /// Hermes gateways omit `running` for paused decision turns.
+    /// Testable helper: derives whether pending approval cards should be
+    /// restored from the presentation cache on resume. Clarification cards
+    /// have independent eligibility because an unresolved request can survive
+    /// a resume that reports `running == false`.
     static func shouldRestorePendingCards(running: Bool?) -> Bool {
         running != false
     }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -604,11 +604,12 @@ final class AppState: ObservableObject {
     /// Coalesces presentation-cache flushes during streaming so we
     /// don't serialize and write UserDefaults on every WebSocket frame.
     private var presentationCacheFlushTask: Task<Void, Never>?
-    /// A nil resume can restore a decision card from local presentation data.
-    /// Keep the card in memory for this AppState so another foreground resume
-    /// can still show it, but strip it from cache writes until Hermes confirms
-    /// the turn. Scope the guard to the session/profile that produced it so a
-    /// session switch cannot affect another session's presentation.
+    /// A resume without explicit active-turn confirmation can restore a
+    /// decision card from local presentation data. Keep the card in memory for
+    /// this AppState so another foreground resume can still show it, but strip
+    /// it from cache writes until Hermes confirms the turn. Scope the guard to
+    /// the session/profile that produced it so a session switch cannot affect
+    /// another session's presentation.
     private struct PendingDecisionRestorationGuard {
         let profile: String
         let sessionID: String
@@ -2680,14 +2681,14 @@ final class AppState: ObservableObject {
             for: result.sessionId,
             fallbackSessionId: reconciliation?.requestedSessionId
         )
-        // A clarification request is a user-actionable event, not merely a
-        // projection of the gateway's current turn state. Hermes can omit the
-        // one-shot clarify event from a resume, and `running == false` can
-        // describe the preceding text even while that clarification remains
-        // unresolved. Keep clarification restoration independent of running;
-        // approvals retain the older explicit-running eligibility rule.
-        let restorePendingClarifications = true
-        let restorePendingApprovals = Self.shouldRestorePendingCards(running: result.snapshot.running)
+        // Pending clarifications and approvals are user-actionable decision
+        // events, not merely projections of the gateway's current turn state.
+        // Hermes can omit a one-shot decision event from a resume, and
+        // `running == false` can describe the preceding text even while the
+        // decision remains unresolved. Restore both card types independently
+        // of the turn snapshot; the gateway decision key remains authoritative
+        // when it includes a resolved record.
+        let restorePendingDecisionCards = true
         let sessionIDs = [result.sessionId, reconciliation?.requestedSessionId].compactMap { $0 }
         let gatewayDecisionKeys = Set(result.messages.compactMap(SessionPresentationCache.decisionKey(for:)))
         let retainedMessages = retainedRestoredMessages.filter { message in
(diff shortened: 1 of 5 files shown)


kaishi00:
## Summary

This PR fixes a background-resume bug where unresolved clarification cards were dropped when the resume snapshot reported `running == false`, by decoupling clarification-card restoration/actionability from the settled-turn signal while keeping settled-approval pruning intact, plus regression tests and design docs.

## Findings

### [severity: major] Forcing `turnState = .running` for a settled resume misrepresents the gateway as busy
**File:** Conduit/Services/AppState.swift:2796-2812
**Issue:** When `running == false` and a pending clarification exists, `turnState` is forced to `.running`. This makes `isBusy` (`turnState.isRunning`, AppState.swift:390) true even though the gateway explicitly reported the turn settled. Downstream side effects: ChatView.swift:109-111 renders a `TypingIndicator` ("assistant working") with no active turn; ComposerBar.swift:357 disables the attach button and the composer action shows a Stop button that calls `cancelCurrent()` → `interruptForReplacement()` (AppState.swift:4791) which issues `client.cancel` to an idle gateway and can surface "Could not interrupt the active response"; `submitComposer` routes typed sends through the `steer`/`redirect` interrupt path (AppState.swift:4233-4245) instead of normal `sendMessage`. None of this is needed: `TurnState.fromGatewayRunning(false)` is `.idle`, and `.idle` already satisfies `acceptsComposerActions` (Models.swift:390-392), so the composer and the card's own answer UI (`canRespond`, ChatView.swift:1639) remain fully usable without the false-busy state.
**Suggestion:** For the `running == false` case, let `turnState` resolve to `.idle` (which keeps the composer enabled and the card answerable) and reserve the forced-`.running` override for the genuinely ambiguous `running == nil` path. Update the test assertions (`XCTAssertEqual(appState.turnState, TurnState.running)` at SessionPresentationCacheTests.swift:1001 and :1310) to `.idle`, and assert `composerIsEnabled` instead of turn state.

### [severity: minor] Gateway-provided pending clarifications bypass the bounded unconfirmed expiry safeguard
**File:** Conduit/Services/AppState.swift:2750-2764
**Issue:** When the gateway itself supplies a pending clarification while `running == false`, `gatewayConfirmsActiveTurn` is set true, so `save` is called with `preservePendingDecisionCards: true` but with empty `unconfirmedPendingDecisionKeys` (because `restoredPendingDecisionKeys` subtracts the gateway keys). This persists the pending card with `unconfirmedPendingDecisionAt == nil`. Unlike the cache-restored path (which gets the 24h marker), a gateway-provided pending clarification is then restored on every resume/cold-launch indefinitely — and since one-shot clarify events are never replayed on resume, the card can stay forever even if the gateway auto-resolved it (timeout) or it was answered elsewhere. The design doc claims "Existing expiry and confirmation safeguards remain in force" (spec line 56), which is only true for the cached-restored branch.
**Suggestion:** When a gateway-provided pending clarification is persisted with a settled `running == false`, stamp an `unconfirmedPendingDecisionAt` marker for its keys as well, or explicitly document the deliberate unbounded lifetime for gateway-authoritative cards.

### [severity: minor] No regression test for cached pending approval pruning when `running == false`
**File:** ConduitTests/SessionPresentationCacheTests.swift:1326
**Issue:** The PR's stated goal includes "preserving settled-approval pruning," and the implementation now drops cache-restored pending approvals when `running == false` (`restorePendingApprovals = false`, includePendingApprovals: false). The only coverage for that path is the gateway-supplied approval case (`testApplyChatResumePrunesGatewayPendingApprovalWhenRunningIsFalse`). The pre-existing test that covered the cache-restored settled case (`testApplyChatResumeDoesNotRestorePendingCardsWhenRunningIsFalse`) used a clarification card and was rewritten to assert the opposite behavior, leaving the cache-restored approval path untested.
**Suggestion:** Add a focused test seeding a cached pending `ApprovalActivity` and asserting it is removed (and not re-persisted) after a resume with `running == false`.

### [severity: nit] Duplicated pending-clarification key construction
**File:** Conduit/Services/AppState.swift:2711-2718
**Issue:** `gatewayPendingClarificationKeys` rebuilds `"clarify:\(clarify.requestId)"` by hand, duplicating the format already produced by `SessionPresentationCache.pendingDecisionKey(for:)` (SessionPresentationCache.swift:42-43). If the key format ever changes in one place but not the other, the subtract (`gatewayPendingApprovalKeys`) and the pruning would silently break.
**Suggestion:** Derive the clarification keys by filtering `result.messages` for `message.clarify != nil` using `SessionPresentationCache.pendingDecisionKey(for:)`, or add a shared helper that extracts pending clarification keys so the format is defined once.

## Verdict

REQUEST_CHANGES

The core reconciliation logic is correct and the regression tests are well-scoped, but forcing `.running` on an explicitly settled resume introduces real user-facing side effects (false typing indicator, Stop button / steer-interrupt routing on an idle gateway) when `.idle` would satisfy the same requirement. The unbounded persistence of gateway-provided pending clarifications and the missing cache-restored-approval test should also be addressed.

<a href="https://opencode.ai/s/jV0Ugd9h">New%20session%20-%202026-08-13T13%3A06%3A09.867Z</a>
[opencode session](https://opencode.ai/s/jV0Ugd9h)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31703144567)

kaishi00:
## Summary

This PR fixes a background-resume bug where pending clarification/approval cards were dropped or unanswerable when the resume snapshot reported `running == false` (a settled turn), by decoupling decision-card restoration/actionability from the settled-turn signal, using shared pending-decision keys for gateway reconciliation, and persisting restored/gateway cards with the bounded unconfirmed-expiry safeguard, plus extensive regression tests and design docs.

## Findings

The prior review's four findings are addressed in this iteration: `running == false` now resolves to `.idle` (no false busy/stop/steer), gateway-provided pending cards are stamped with the unconfirmed marker (via the full `pendingDecisionKeys(in: messages)` set, AppState.swift:2739-2741), and new tests cover cached + gateway-provided cards for both decision types. I verified the resume/save/merge reconciliation logic end-to-end (resolved-record suppression, retained-card re-resume, grace-period expiry, submitting→pending reset). Remaining issues below.

### [severity: minor] Cache flush clears the 24h expiry marker for gateway-provided pending cards
**File:** Conduit/Services/AppState.swift:685-693 (with SessionPresentationCache.swift:332-342)
**Issue:** The unconfirmed marker stamped for a gateway-provided pending card at resume (`unconfirmedPendingDecisionKeys = pendingDecisionKeys(in: messages)`) is only *honored on the next write while a `PendingDecisionRestorationGuard` exists*. The guard is created only from `restoredPendingDecisionKeys` (restored cards, gateway keys subtracted, AppState.swift:2751-2762), so a gateway-authoritative card has no guard. The next `cacheMessagePresentation()` with `restorationKeys == nil` calls `save(..., unconfirmedPendingDecisionKeys: [])`, and `save` unconditionally sets `unconfirmedPendingDecisionAt = nil` when the key set is empty (SessionPresentationCache.swift:335-337). This clears the marker while the card is still pending. Consequence: if a flush occurs (common during streaming/typing), the card loses its expiry, and a later cold-launch resume after the gateway stopped replaying the pending row (one-shot events aren't replayed) restores it with marker `nil` → `isUnconfirmedPendingDecisionExpired(nil)` is false → re-stamped fresh. The bounded-expiry guarantee the design doc claims for gateway-provided cards (spec line 56-61) is therefore only enforced until the first guard-less flush.
**Suggestion:** In `cacheMessagePresentation`, when no guard is active, derive `unconfirmedPendingDecisionKeys` from the currently pending keys in `messages` (mirroring `applyChatResume`), so gateway-provided pending cards keep their marker across flushes. Alternatively, make `save` preserve an existing `unconfirmedPendingDecisionAt` when pending decision records remain, rather than unconditionally nil-ing it.

### [severity: minor] `resetSubmittingRestoredDecisions` hand-builds decision keys instead of using the shared helpers
**File:** Conduit/Services/AppState.swift:2805-2826 (key construction at 2812 and 2820)
**Issue:** This PR's stated goal is "use shared pending-decision keys and reconciliation," and the new code correctly uses `SessionPresentationCache.pendingDecisionKeys(in:)` / `decisionKey(for:)`, but this function — invoked on every resume with restored cards — still concatenates `"clarify:\(requestId)"` and `"approval:\(sessionId)"` by hand, duplicating `SessionPresentationCache.pendingDecisionKey(for:)` (SessionPresentationCache.swift:41-49). If the key format changes in one place, the submitting→pending reset silently stops matching the restored keys.
**Suggestion:** Derive the keys from the restored `ChatMessage`s via `SessionPresentationCache.pendingDecisionKey(for:)` (or add a shared key-builder helper) and pass them in, keeping the format defined once.

### [severity: nit] `gatewayConfirmsActiveTurn == true` while a 24h marker is still stamped
**File:** Conduit/Services/AppState.swift:2737-2741
**Issue:** For `running == nil` with a gateway-provided pending decision, `gatewayConfirmsActiveTurn` is true, but `unconfirmedPendingDecisionKeys` is still the full pending set (because the branch keyed off `running != true`), so a "confirmed active turn" card is persisted with a 24h expiry marker. The comment at 2732-2736 says markers are stamped "until Hermes confirms the turn," which is not what the code does here. Harmless in practice (the gateway re-sends pending rows each resume), but the naming/behavior mismatch is confusing.
**Suggestion:** Either exclude keys from `unconfirmedPendingDecisionKeys` when `gatewayConfirmsActiveTurn` is true, or rename the variable / update the comment to reflect that markers are stamped for any resume without an explicit `running == true`.

### [severity: nit] `running == nil` still forces `.running`, re-creating the false-busy state the PR removes for `running == false`
**File:** Conduit/Services/AppState.swift:2785-2786
**Issue:** When `running == nil` and a pending decision (or live projection) is present, `turnState` is forced to `.running`, producing the same side effects this PR eliminates for `running == false` — the typing indicator, a Stop button, and steer/interrupt routing on a turn that is actually paused on a user decision. Since `.idle` already satisfies `acceptsComposerActions` (Models.swift:390-392) and the card's own answer UI (`canRespond`, ChatView.swift:1639) is independent of `turnState`, the forced `.running` adds no answerability. The design doc (spec line 51-55) explicitly retains this as conservative, so it's a documented tradeoff rather than a regression.
**Suggestion:** Consider resolving `.idle` when the only driver is a restored/gateway pending decision and there is no live projection, reserving `.running` for genuine `hasLiveProjection`; at minimum, ensure the doc comment explains why the conservative path stays.

## Verdict

COMMENT

The core fix is correct and well-tested: settled `running == false` now keeps cards visible and answerable while remaining `.idle`, gateway-resolved records correctly suppress cached pending cards without duplicates, and all prior review findings are resolved. The remaining items are non-blocking, but the expiry-marker clearing in `cacheMessagePresentation` is worth addressing because the PR's design doc explicitly promises a bounded lifetime for gateway-provided pending cards that the current code doesn't fully deliver.

<a href="https://opencode.ai/s/dVFJlcx0">New%20session%20-%202026-08-13T13%3A36%3A21.010Z</a>
[opencode session](https://opencode.ai/s/dVFJlcx0)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31705671832)

kaishi00:
## Summary

This PR keeps unresolved clarification/approval cards visible and answerable after a background-to-foreground resume even when the snapshot reports `running == false`, decoupling decision-card restoration/actionability from the settled-turn signal, persisting gateway-provided pending cards with the bounded unconfirmed-expiry marker (including across cache flushes), and adding comprehensive regression coverage for cached and gateway-provided cards of both decision types.

## Findings

### [severity: minor] `running == nil` still forces `.running`, recreating the false-busy UI the PR removes for `running == false`
**File:** Conduit/Services/AppState.swift:2792-2793
**Issue:** When `running == nil` and a pending decision (or live projection) is present, `turnState` is forced to `.running`. `TurnState.isRunning` (Models.swift:394-396) makes `isBusy` true, so ChatView renders a typing indicator, the composer shows Stop/steer (Models.swift:409-413) which routes through `interruptForReplacement()`/`client.cancel` against a gateway that is actually paused on a user decision. `.idle` already satisfies `acceptsComposerActions` (Models.swift:390-392) and card answerability (`canRespond`) is independent of `turnState`, so the forced `.running` buys no answerability. This is documented as conservative in the design spec (lines 51-55), so it is not a regression — but it is inconsistent with the rationale applied to `running == false`.
**Suggestion:** Resolve `.idle` when the only driver is a restored/gateway pending decision without a live projection, reserving `.running` for genuine `hasLiveProjection`. If kept, ensure the doc comment explains why the paused-turn path deliberately shows busy/stop/steer.

### [severity: minor] `restorePendingDecisionCards = true` is a constant, making the retention filter dead conditional logic
**File:** Conduit/Services/AppState.swift:2696 and 2704
**Issue:** `let restorePendingDecisionCards = true` combined with `return restorePendingDecisionCards` inside the `retainedMessages` filter means the guard clause always returns `true`; the conditional is dead. It also reads as if some path could still prune restored cards, which the new intent explicitly rejects, and the two `includePendingClarifications/Approvals` args that consume it are likewise always true.
**Suggestion:** Remove the constant and keep the filter as just the key-membership check (`!gatewayDecisionKeys.contains(key)`), or replace the constant with a documented intent comment.

### [severity: minor] Marker is session-scoped, so a new gateway card under an active guard can inherit a near-expired timestamp
**File:** Conduit/Services/AppState.swift:685-689 with SessionPresentationCache.swift:332-337
**Issue:** `unconfirmedPendingDecisionAt` is a single timestamp per `CachedSession`, not per card. `pendingDecisionKeysToPreserve = restorationKeys ?? pendingDecisionKeys` drops the keys of a newly arrived gateway-provided card whenever a restoration guard is active (the guard's keys win). The card still survives (the removal step is skipped when `preservePendingDecisionCards` is true), but the shared marker means a fresh gateway card can inherit an old marker from a long-lived guarded card and be dropped prematurely by `isUnconfirmedPendingDecisionExpired` (SessionPresentationCache.swift:147-150) on a later `running == false` resume. This is a pre-existing session-scoped design, but this PR newly extends the marker scheme to gateway-provided cards, widening exposure.
**Suggestion:** Pass the union `restorationKeys ∪ pendingDecisionKeys` as `unconfirmedPendingDecisionKeys`, and consider tracking per-key timestamps if the premature-expiry edge matters.

### [severity: nit] `gatewayConfirmsActiveTurn == true` while an unconfirmed marker is still stamped
**File:** Conduit/Services/AppState.swift:2744-2748
**Issue:** For `running == nil` with a gateway-provided pending decision, `gatewayConfirmsActiveTurn` is true (2745) while `unconfirmedPendingDecisionKeys` is simultaneously the full pending set (2747), so a "confirmed active turn" card is persisted with a 24h expiry marker. The comment at 2740-2743 documents this deliberately, but the name/behavior mismatch is confusing for future maintainers.
**Suggestion:** Rename to something like `gatewayReportsActiveTurn` or add a brief comment that markers are intentionally independent of the active-turn signal.

## Verdict

APPROVE

The core fix is correct and well-tested: `running == false` now resolves to `.idle` (no false busy/stop/steer), gateway-resolved records correctly suppress cached pending cards without duplicates, gateway-provided pending cards keep the bounded expiry marker across guard-less cache flushes, and both cached and gateway-provided cards are covered for clarification and approval. All findings from the prior review iterations are resolved. The remaining items are non-blocking; the `running == nil` busy-state tradeoff is documented, and the marker-scoping edge is minor.

<a href="https://opencode.ai/s/DqJ0sNm2">New%20session%20-%202026-08-13T14%3A02%3A38.426Z</a>
[opencode session](https://opencode.ai/s/DqJ0sNm2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31707967031)

kaishi00:
## Summary

This PR fixes a background-resume bug where pending clarification/approval cards were dropped or unanswerable when the resume snapshot reported `running == false`, by decoupling decision-card restoration/actionability from the settled-turn signal, persisting restored and gateway-provided cards with the bounded unconfirmed-expiry marker (including across guard-less cache flushes), and adding regression coverage for both decision types and both sources.

## Findings

### [severity: minor] Fresh gateway pending card under an active restoration guard is excluded from the unconfirmed-key set and can be prematurely expired
**File:** Conduit/Services/AppState.swift:685-698
**Issue:** `pendingDecisionKeysToPreserve = restorationKeys ?? pendingDecisionKeys` lets the guard's keys win. When a restoration guard is active and a *new* gateway-provided pending card arrives in `messages` (e.g. a second clarify while the first is pending), the flush still persists the card (`preservePendingDecisionCards` is true because `pendingDecisionKeys` is non-empty), but the new card's key is dropped from `unconfirmedPendingDecisionKeys`. Because `CachedSession.unconfirmedPendingDecisionAt` is a single session-scoped timestamp, the fresh card inherits the guarded card's marker; when that marker exceeds `maxUnconfirmedPendingDecisionAge`, a later `running == false` resume strips all pending cards from the cache via `isUnconfirmedPendingDecisionExpired` (SessionPresentationCache.swift:147-150) even though the gateway never confirmed the fresh card's resolution.
**Suggestion:** Pass the union `restorationKeys ∪ pendingDecisionKeys` as `unconfirmedPendingDecisionKeys`, or track per-key timestamps if premature expiry of gateway-authoritative cards matters.

### [severity: minor] `running == nil` still forces `.running`, re-creating the false-busy UI this PR removes for `running == false`
**File:** Conduit/Services/AppState.swift:2792-2793
**Issue:** When `running == nil` with a pending decision (or live projection), `turnState` is forced to `.running`. `TurnState.isRunning` makes `isBusy` true, so ChatView shows a typing indicator and the composer surfaces Stop/steer routing (`interruptForReplacement()` → `client.cancel` on a turn that is actually paused on a user decision). `.idle` already satisfies `acceptsComposerActions` (Models.swift:390-392) and the card's own answer UI (`canRespond`) is independent of `turnState`, so the forced `.running` buys no answerability. This is documented as conservative in the design spec, so it is a deliberate tradeoff rather than a regression — but it is inconsistent with the rationale applied to the `running == false` path.
**Suggestion:** Resolve to `.idle` when the only driver is a restored/gateway pending decision without a live projection, reserving `.running` for genuine `hasLiveProjection`. If kept, ensure the doc comment explains why the paused-turn path deliberately shows busy/stop/steer.

### [severity: nit] `restorePendingDecisionCards` is a constant `true`, making the retention filter dead conditional logic
**File:** Conduit/Services/AppState.swift:2696, 2704
**Issue:** `let restorePendingDecisionCards = true` combined with `return restorePendingDecisionCards` inside the `retainedMessages` filter means the guard always yields `true`; the conditional is dead. It also reads as if a path could still prune restored cards, which the new intent explicitly rejects.
**Suggestion:** Remove the constant and keep the filter as just the key-membership check (`!gatewayDecisionKeys.contains(key)`), or replace the constant with a documented intent comment.

### [severity: nit] `gatewayConfirmsActiveTurn == true` while a 24h unconfirmed marker is still stamped
**File:** Conduit/Services/AppState.swift:2744-2748
**Issue:** For `running == nil` with a gateway-provided pending decision, `gatewayConfirmsActiveTurn` is true while `unconfirmedPendingDecisionKeys` is simultaneously the full pending set, so a "confirmed active turn" card is persisted with an expiry marker. The comment at 2740-2743 documents this, but the name/behavior mismatch is confusing for future maintainers.
**Suggestion:** Rename to something like `gatewayReportsActiveTurn`, or add a brief comment that markers are intentionally independent of the active-turn signal.

## Verdict

COMMENT

The core fix is correct and well-tested: `running == false` now resolves to `.idle` (no false busy/stop/steer), gateway-resolved records suppress cached pending cards without duplicates, gateway-provided pending cards keep the bounded expiry marker across guard-less cache flushes, and both cached and gateway-provided cards are covered for clarification and approval (14 focused `applyChatResume` tests). No retain cycles, force unwraps, or concurrency issues were introduced in the changed code. The remaining items are non-blocking; the marker-scoping edge in `cacheMessagePresentation` is the most actionable since the design doc promises a bounded lifetime for gateway-provided pending cards that the current union-less key set does not fully guarantee.

<a href="https://opencode.ai/s/ugCC3RRA">New%20session%20-%202026-08-13T14%3A18%3A23.770Z</a>
[opencode session](https://opencode.ai/s/ugCC3RRA)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31709368639)

PR #55, opened 2026-08-13
Persist session-scoped YOLO overrides across turns

kaishi00:
## Summary
- persist explicit session YOLO choices as normalized profile/session overrides with distinct true, false, and absent states
- treat an explicit session snapshot `yolo` from Hermes as authoritative during fresh resume reconciliation when no newer local write occurred after that resume began
- preserve the local override when live pushes or an older in-flight resume response may be stale
- replay buffered `session.info` fields while anchoring only their YOLO value to fresh resume authority
- avoid writing a session YOLO override when Model Picker Apply leaves the YOLO selection unchanged
- seed Model Picker YOLO draft state synchronously before asynchronous model loading, without reseeding an existing draft on repeated appearances
- resolve both canonical and runtime session IDs across resume and catalog updates, migrating raw runtime aliases to canonical keys
- evict session YOLO overrides for canonical and alternate IDs after successful archive or delete mutations
- write the override and update visible runtime only after the gateway accepts the setting
- log and surface corrupt, unsupported-version, and encode persistence failures
- add isolated store, relaunch, session-switching, precedence, success, failure, model-picker, alias, server-authority, stale-ordering, and buffered-replay coverage

## Root cause
`setYoloMode` waits for the gateway RPC before persisting the local override, while WebSocket `session.info` pushes and `session.resume` responses can be delivered independently and out of order. A snapshot captured before the toggle could arrive after the RPC succeeded; the conflict handler treated that stale `yolo` value as authoritative, cleared the new override, and flipped the UI back. During foreground reconciliation, buffered `session.info` events were then replayed after the fresh resume snapshot; a contradictory buffered snapshot could reapply the old YOLO value, and the old whole-event filter also discarded newer non-YOLO runtime fields.

## Review follow-ups addressed
- model-only Apply no longer creates a spurious persisted YOLO override
- early Model Picker YOLO changes survive asynchronous model loading
- repeated Model Picker appearances no longer overwrite an in-progress YOLO draft or its comparison baseline
- explicit gateway session YOLO state can replace a stale local override during a fresh resume
- stale live session-info pushes cannot erase a newer local write
- stale resume responses are compared with a per-session local-write baseline and cannot erase a newer successful write
- buffered session-info replay preserves model, provider, running, context, and other runtime fields while fresh resume authority controls YOLO
- overrides written under a runtime session ID migrate to the canonical catalog ID
- archived and deleted sessions no longer retain local YOLO overrides
- persistence decode/version/encode failures produce OSLog diagnostics

## Validation
- Focused `SessionYoloPersistenceTests`: 11 passed, 0 failed
- Focused `ModelPickerTests`: 6 passed, 0 failed
- Full iOS Simulator XCTest suite: 532 passed, 0 failed
- `git diff --check` clean

## Summary by CodeRabbit

- **New Features**
  - YOLO settings now persist independently for each chat session, including across app relaunches and session switches.
  - Session-specific settings are restored consistently and support canonical session identity changes.
- **Bug Fixes**
  - Prevented stale resume data or live updates from overwriting newer local YOLO changes.
  - Model Picker preserves existing YOLO selections instead of unexpectedly resetting them.
  - Failed YOLO updates no longer leave the interface or saved state inconsistent.
- **Documentation**
  - Added design and implementation documentation for session-level YOLO persistence and reconciliation.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -55,6 +55,7 @@ struct ChatResumeLifecycleOperations {
         String
     ) async throws -> AnyCodable)?
     var setBusyInputMode: (@MainActor (HermesClient, BusyInputMode) async throws -> Void)?
+    var setSessionYolo: (@MainActor (HermesClient, String, Bool) async throws -> Void)?
     var loadProfiles: (@MainActor () async -> Void)?
     var loadBusyInputMode: (@MainActor (HermesClient) async -> Void)?
     var loadProfileDisplayPreferences: (@MainActor () async -> Void)?
@@ -90,6 +91,7 @@ struct ChatResumeLifecycleOperations {
             String
         ) async throws -> AnyCodable)? = nil,
         setBusyInputMode: (@MainActor (HermesClient, BusyInputMode) async throws -> Void)? = nil,
+        setSessionYolo: (@MainActor (HermesClient, String, Bool) async throws -> Void)? = nil,
         loadProfiles: (@MainActor () async -> Void)? = nil,
         loadBusyInputMode: (@MainActor (HermesClient) async -> Void)? = nil,
         loadProfileDisplayPreferences: (@MainActor () async -> Void)? = nil,
@@ -109,6 +111,7 @@ struct ChatResumeLifecycleOperations {
         self.executeSlash = executeSlash
         self.dispatchCommand = dispatchCommand
         self.setBusyInputMode = setBusyInputMode
+        self.setSessionYolo = setSessionYolo
         self.loadProfiles = loadProfiles
         self.loadBusyInputMode = loadBusyInputMode
         self.loadProfileDisplayPreferences = loadProfileDisplayPreferences
@@ -632,6 +635,7 @@ final class AppState: ObservableObject {
     private var sessionCatalogCache = SessionCatalogCache()
     private var projectsRequestGeneration = 0
     private let sessionPresentationCache: SessionPresentationCache
+    private let sessionYoloStore: SessionYoloStore
 
     /// The dashboard's persisted transcript is richer than `session.resume`:
     /// it retains database timestamps, complete tool-call inputs, and other
@@ -800,10 +804,12 @@ final class AppState: ObservableObject {
         reconnectScheduler: ChatResumeReconnectScheduler? = nil,
         reconnectExecutor: ChatResumeReconnectExecutor? = nil,
         chatResumeLifecycleOperations: ChatResumeLifecycleOperations = .live,
-        sessionPresentationCache: SessionPresentationCache = .shared
+        sessionPresentationCache: SessionPresentationCache = .shared,
+        sessionYoloStore: SessionYoloStore? = nil
     ) {
         self.defaults = defaults
         self.sessionPresentationCache = sessionPresentationCache
+        self.sessionYoloStore = sessionYoloStore ?? SessionYoloStore(defaults: defaults)
         self.chatResumeCoordinator = chatResumeCoordinator
             ?? ChatResumeCoordinator(store: ChatResumeStore(defaults: defaults))
         self.recoverySequence = recoverySequence
@@ -2777,7 +2783,7 @@ final class AppState: ObservableObject {
         activeAssistantMessageId = nil
         activeReasoningMessageId = nil
         receivedReasoningForCurrentTurn = false
-        applyRuntime(result.snapshot)
+        applyRuntime(result.snapshot, for: result.sessionId)
 
         // A paused clarification or approval can have neither a live text
         // projection nor an explicit `running` field. The restored card is
@@ -3137,7 +3143,19 @@ final class AppState: ObservableObject {
         return recovered
     }
 
-    private func applyRuntime(_ snapshot: SessionRuntimeSnapshot) {
+    private func canonicalSessionID(for sessionID: String?) -> String? {
+        ChatSessionPersistenceIdentity.canonicalID(
+            for: sessionID,
+            identity: activeChatScrollSessionIdentity,
+            catalog: sessions + cronSessions,
+            activeProfile: activeProfile
+        )
+    }
+
+    private func applyRuntime(
+        _ snapshot: SessionRuntimeSnapshot,
+        for sessionID: String? = nil
+    ) {
         if let model = snapshot.model { runtime.model = model }
         if let provider = snapshot.provider { runtime.provider = provider }
         if let cwd = snapshot.cwd { runtime.cwd = cwd }
@@ -3149,7 +3167,13 @@ final class AppState: ObservableObject {
             runtime.reasoningEffort = reasoningEffort.lowercased() == "none" ? "" : reasoningEffort
         }
         if let fast = snapshot.fast { runtime.fast = fast }
-        if let yolo = snapshot.yolo {
+        if let canonicalSessionID = canonicalSessionID(for: sessionID ?? activeSessionId),
+           let storedOverride = sessionYoloStore.storedOverride(
+               for: activeProfile,
+               sessionID: canonicalSessionID
+           ) {
+            runtime.yolo = storedOverride
+        } else if let yolo = snapshot.yolo {
             runtime.yolo = yolo
         } else if let approvalsMode = snapshot.approvalsMode {
             runtime.yolo = approvalsMode.lowercased() == "off"
@@ -4293,7 +4317,17 @@ final class AppState: ObservableObject {
     func setYoloMode(_ enabled: Bool) async -> Bool {
         guard let client, let sessionId = activeSessionId else { return false }
         do {
-            try await client.setSessionYolo(sessionId, enabled: enabled)
+            if let setSessionYolo = chatResumeLifecycleOperations.setSessionYolo {
+                try await setSessionYolo(client, sessionId, enabled)
+            } else {
+                try await client.setSessionYolo(sessionId, enabled: enabled)
+            }
+            let persistedSessionID = canonicalSessionID(for: sessionId) ?? sessionId
+            sessionYoloStore.setOverride(
+                enabled,
+                for: activeProfile,
+                sessionID: persistedSessionID
+            )
             runtime.yolo = enabled
             return true
         } catch {
@@ -6349,8 +6383,8 @@ final class AppState: ObservableObject {
                 scheduleResponseHapticConclusion(after: 180)
             }
 
-        case .sessionInfo(_, let snapshot):
-            applyRuntime(snapshot)
+        case .sessionInfo(let sessionID, let snapshot):
+            applyRuntime(snapshot, for: sessionID)
             if let running = snapshot.running {
                 setRunning(running)
             }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -3167,14 +3167,24 @@ final class AppState: ObservableObject {
             runtime.reasoningEffort = reasoningEffort.lowercased() == "none" ? "" : reasoningEffort
         }
         if let fast = snapshot.fast { runtime.fast = fast }
-        if let canonicalSessionID = canonicalSessionID(for: sessionID ?? activeSessionId),
-           let storedOverride = sessionYoloStore.storedOverride(
-               for: activeProfile,
-               sessionID: canonicalSessionID
-           ) {
-            runtime.yolo = storedOverride
-        } else if let yolo = snapshot.yolo {
+        let requestedSessionID = sessionID ?? activeSessionId
+        let resolvedCanonicalSessionID = canonicalSessionID(for: requestedSessionID)
+        let sessionIDsForOverride = [resolvedCanonicalSessionID, requestedSessionID]
+            .compactMap { $0 }
+        let storedOverride = sessionYoloStore.storedOverride(
+            for: activeProfile,
+            sessionIDs: sessionIDsForOverride
+        )
+        if let yolo = snapshot.yolo {
+            if let storedOverride, storedOverride != yolo {
+                sessionYoloStore.clearOverride(
+                    for: activeProfile,
+                    sessionIDs: sessionIDsForOverride
+                )
+            }
             runtime.yolo = yolo
+        } else if let storedOverride {
+            runtime.yolo = storedOverride
         } else if let approvalsMode = snapshot.approvalsMode {
             runtime.yolo = approvalsMode.lowercased() == "off"
         }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2783,7 +2783,11 @@ final class AppState: ObservableObject {
         activeAssistantMessageId = nil
         activeReasoningMessageId = nil
         receivedReasoningForCurrentTurn = false
-        applyRuntime(result.snapshot, for: result.sessionId)
+        applyRuntime(
+            result.snapshot,
+            for: result.sessionId,
+            reconcileExplicitYolo: true
+        )
 
         // A paused clarification or approval can have neither a live text
         // projection nor an explicit `running` field. The restored card is
@@ -3154,7 +3158,8 @@ final class AppState: ObservableObject {
 
     private func applyRuntime(
         _ snapshot: SessionRuntimeSnapshot,
-        for sessionID: String? = nil
+        for sessionID: String? = nil,
+        reconcileExplicitYolo: Bool = false
     ) {
         if let model = snapshot.model { runtime.model = model }
         if let provider = snapshot.provider { runtime.provider = provider }
@@ -3185,13 +3190,22 @@ final class AppState: ObservableObject {
             sessionIDs: sessionIDsForOverride
         )
         if let yolo = snapshot.yolo {
-            if let storedOverride, storedOverride != yolo {
-                sessionYoloStore.clearOverride(
-                    for: activeProfile,
-                    sessionIDs: sessionIDsForOverride
-                )
+            if let storedOverride {
+                if reconcileExplicitYolo, storedOverride != yolo {
+                    sessionYoloStore.clearOverride(
+                        for: activeProfile,
+                        sessionIDs: sessionIDsForOverride
+                    )
+                    runtime.yolo = yolo
+                } else {
+                    // Live session.info pushes can be older than a client
+                    // initiated config.set RPC. Keep the local choice until a
+                    // fresh resume snapshot can reconcile server authority.
+                    runtime.yolo = storedOverride
+                }
+            } else {
+                runtime.yolo = yolo
             }
-            runtime.yolo = yolo
         } else if let storedOverride {
             runtime.yolo = storedOverride
         } else if let approvalsMode = snapshot.approvalsMode {
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2420,7 +2420,8 @@ final class AppState: ObservableObject {
                 presentationResult,
                 automaticWorkToken: automaticWorkToken,
                 automaticSyncOperationID: automaticSyncOperationID,
-                yoloWriteBaseline: yoloWriteBaseline
+                yoloWriteBaseline: yoloWriteBaseline,
+                reconcileExplicitYolo: reconcileExplicitYolo
             ) else {
                 settleReconciliation(token, automaticSyncOperationID: automaticSyncOperationID)
                 return false
@@ -2507,18 +2508,10 @@ final class AppState: ObservableObject {
                         .isEmpty == false
                 )
             }
-            if reconcileExplicitYolo {
-                // A session.info received while resume was in flight may be a
-                // pre-resume snapshot. Do not replay a contradictory YOLO
-                // value after fresh resume reconciliation clears an override;
-                // pushes received after reconciliation remain authoritative.
-                bufferedEvents = Self.filteringBufferedSessionInfo(
-                    bufferedEvents,
-                    against: result.snapshot,
-                    acceptedSessionIDs: reconciliation?.acceptedSessionIDs ?? Set([result.sessionId])
-                )
+            let bufferedYoloAuthority = reconcileExplicitYolo ? result.snapshot.yolo : nil
+            bufferedEvents.forEach { event in
+                applyStreamEvent(event, authoritativeYolo: bufferedYoloAuthority)
             }
-            bufferedEvents.forEach(applyStreamEvent)
             return settleReconciliationAndPublish(
                 token,
                 automaticWorkToken: automaticWorkToken,
@@ -2717,7 +2710,8 @@ final class AppState: ObservableObject {
         _ result: SessionResumeResult,
         automaticWorkToken: ChatResumeAutomaticWorkToken? = nil,
         automaticSyncOperationID: UUID? = nil,
-        yoloWriteBaseline: SessionYoloWriteBaseline?
+        yoloWriteBaseline: SessionYoloWriteBaseline?,
+        reconcileExplicitYolo: Bool? = nil
     ) -> Bool {
         guard automaticChatResumeWorkIsCurrent(
             automaticWorkToken,
@@ -2830,7 +2824,7 @@ final class AppState: ObservableObject {
         applyRuntime(
             result.snapshot,
             for: result.sessionId,
-            reconcileExplicitYolo: yoloWriteBaseline.map {
+            reconcileExplicitYolo: reconcileExplicitYolo ?? yoloWriteBaseline.map {
                 !hasNewerSessionYoloWrite(
                     since: $0,
                     sessionIDs: [result.sessionId, reconciliation?.requestedSessionId].compactMap { $0 }
@@ -3122,24 +3116,6 @@ final class AppState: ObservableObject {
         return deduplicated
     }
 
-    nonisolated static func filteringBufferedSessionInfo(
-        _ events: [StreamEvent],
-        against resumeSnapshot: SessionRuntimeSnapshot,
-        acceptedSessionIDs: Set<String>
-    ) -> [StreamEvent] {
-        guard let resumeYolo = resumeSnapshot.yolo else { return events }
-        return events.filter { event in
-            guard case .sessionInfo(let sessionID, let snapshot) = event,
-                  acceptedSessionIDs.contains(sessionID) else {
-                return true
-            }
-            let eventYolo = snapshot.yolo
-                ?? snapshot.approvalsMode.map { $0.lowercased() == "off" }
-            guard let eventYolo else { return true }
-            return eventYolo == resumeYolo
-        }
-    }
-
     /// Returns every non-empty prefix of `buffered` that is also a suffix of
     /// `covered`. The prefix-function scan stays linear in the cumulative
     /// projection size; callers can reject repeated-content ambiguity when
@@ -3257,7 +3233,8 @@ final class AppState: ObservableObject {
     private func applyRuntime(
         _ snapshot: SessionRuntimeSnapshot,
         for sessionID: String? = nil,
-        reconcileExplicitYolo: Bool = false
+        reconcileExplicitYolo: Bool = false,
+        authoritativeYolo: Bool? = nil
     ) {
         if let model = snapshot.model { runtime.model = model }
         if let provider = snapshot.provider { runtime.provider = provider }
@@ -3287,7 +3264,7 @@ final class AppState: ObservableObject {
             for: activeProfile,
             sessionIDs: sessionIDsForOverride
         )
-        if let yolo = snapshot.yolo {
+        if let yolo = authoritativeYolo ?? snapshot.yolo {
             if let storedOverride {
                 if reconcileExplicitYolo, storedOverride != yolo {
                     sessionYoloStore.clearOverride(
@@ -6462,7 +6439,7 @@ final class AppState: ObservableObject {
         return reconciliation.acceptedSessionIDs.contains(sessionId)
     }
 
-    private func applyStreamEvent(_ event: StreamEvent) {
+    private func applyStreamEvent(_ event: StreamEvent, authoritativeYolo: Bool? = nil) {
         let streamSessionId = sessionID(for: event)
         guard eventBelongsToActiveSession(streamSessionId) else { return }
         defer { schedulePresentationCacheFlush(for: streamSessionId) }
@@ -6524,7 +6501,7 @@ final class AppState: ObservableObject {
             }
 
         case .sessionInfo(let sessionID, let snapshot):
-            applyRuntime(snapshot, for: sessionID)
+            applyRuntime(snapshot, for: sessionID, authoritativeYolo: authoritativeYolo)
             if let running = snapshot.running {
                 setRunning(running)
             }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -620,6 +620,10 @@ final class AppState: ObservableObject {
         let messages: [ChatMessage]
     }
 
+    private struct SessionYoloWriteBaseline {
+        let revisions: [ChatScrollSessionKey: UInt64]
+    }
+
     private var restoredPendingDecisionCardsAwaitingConfirmation: PendingDecisionRestorationGuard?
     /// Timestamp of the last successful coalesced cache flush; used to
     /// enforce a maximum 5-second interval even during continuous streaming.
@@ -636,6 +640,8 @@ final class AppState: ObservableObject {
     private var projectsRequestGeneration = 0
     private let sessionPresentationCache: SessionPresentationCache
     private let sessionYoloStore: SessionYoloStore
+    private var sessionYoloWriteRevision: UInt64 = 0
+    private var sessionYoloWriteRevisions: [ChatScrollSessionKey: UInt64] = [:]
 
     /// The dashboard's persisted transcript is richer than `session.resume`:
     /// it retains database timestamps, complete tool-call inputs, and other
@@ -2325,6 +2331,10 @@ final class AppState: ObservableObject {
         refreshActiveChatScrollSessionIdentity(isReconciling: true)
         turnState = .synchronizing
         let profile = activeProfile
+        // The resume RPC can overlap a user-initiated config.set. Capture the
+        // local-write position before launching either request so a response
+        // from the older snapshot cannot clear the newer override.
+        let yoloWriteBaseline = sessionYoloWriteBaseline(for: sessionId)
 
         do {
             // Match Hermes Desktop: fetch the durable transcript and resume the
@@ -2402,7 +2412,8 @@ final class AppState: ObservableObject {
             guard applyChatResume(
                 presentationResult,
                 automaticWorkToken: automaticWorkToken,
-                automaticSyncOperationID: automaticSyncOperationID
+                automaticSyncOperationID: automaticSyncOperationID,
+                yoloWriteBaseline: yoloWriteBaseline
             ) else {
                 settleReconciliation(token, automaticSyncOperationID: automaticSyncOperationID)
                 return false
@@ -2675,6 +2686,21 @@ final class AppState: ObservableObject {
         automaticWorkToken: ChatResumeAutomaticWorkToken? = nil,
         automaticSyncOperationID: UUID? = nil
     ) -> Bool {
+        applyChatResume(
+            result,
+            automaticWorkToken: automaticWorkToken,
+            automaticSyncOperationID: automaticSyncOperationID,
+            yoloWriteBaseline: nil
+        )
+    }
+
+    @discardableResult
+    private func applyChatResume(
+        _ result: SessionResumeResult,
+        automaticWorkToken: ChatResumeAutomaticWorkToken? = nil,
+        automaticSyncOperationID: UUID? = nil,
+        yoloWriteBaseline: SessionYoloWriteBaseline?
+    ) -> Bool {
         guard automaticChatResumeWorkIsCurrent(
             automaticWorkToken,
             syncOperationID: automaticSyncOperationID
@@ -2786,7 +2812,12 @@ final class AppState: ObservableObject {
         applyRuntime(
             result.snapshot,
             for: result.sessionId,
-            reconcileExplicitYolo: true
+            reconcileExplicitYolo: yoloWriteBaseline.map {
+                !hasNewerSessionYoloWrite(
+                    since: $0,
+                    sessionIDs: [result.sessionId, reconciliation?.requestedSessionId].compactMap { $0 }
+                )
+            } ?? true
         )
 
         // A paused clarification or approval can have neither a live text
@@ -3154,6 +3185,37 @@ final class AppState: ObservableObject {
             catalog: sessions + cronSessions,
             activeProfile: activeProfile
         )
+    }
+
+    private func sessionYoloKeys(for sessionIDs: [String]) -> Set<ChatScrollSessionKey> {
+        Set(sessionIDs.map { ChatScrollSessionKey(profile: activeProfile, sessionID: $0) })
+            .filter(\.isValid)
+    }
+
+    private func sessionYoloWriteBaseline(for sessionID: String) -> SessionYoloWriteBaseline {
+        let sessionIDs = [sessionID, canonicalSessionID(for: sessionID)].compactMap { $0 }
+        let keys = sessionYoloKeys(for: sessionIDs)
+        return SessionYoloWriteBaseline(
+            revisions: Dictionary(uniqueKeysWithValues: keys.map { key in
+                (key, sessionYoloWriteRevisions[key] ?? 0)
+            })
+        )
+    }
+
+    private func hasNewerSessionYoloWrite(
+        since baseline: SessionYoloWriteBaseline,
+        sessionIDs: [String]
+    ) -> Bool {
+        sessionYoloKeys(for: sessionIDs).contains { key in
+            sessionYoloWriteRevisions[key, default: 0] > baseline.revisions[key, default: 0]
+        }
+    }
+
+    private func recordSessionYoloWrite(for sessionIDs: [String]) {
+        sessionYoloWriteRevision &+= 1
+        for key in sessionYoloKeys(for: sessionIDs) {
+            sessionYoloWriteRevisions[key] = sessionYoloWriteRevision
+        }
     }
 
     private func applyRuntime(
@@ -4369,6 +4431,7 @@ final class AppState: ObservableObject {
                 for: activeProfile,
                 sessionID: persistedSessionID
             )
+            recordSessionYoloWrite(for: [sessionId, persistedSessionID])
             runtime.yolo = enabled
             return true
         } catch {
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2409,6 +2409,13 @@ final class AppState: ObservableObject {
                 presentationResult = result
             }
 
+            let resumeSessionIDs = [result.sessionId, reconciliation?.requestedSessionId]
+                .compactMap { $0 }
+            let reconcileExplicitYolo = !hasNewerSessionYoloWrite(
+                since: yoloWriteBaseline,
+                sessionIDs: resumeSessionIDs
+            )
+
             guard applyChatResume(
                 presentationResult,
                 automaticWorkToken: automaticWorkToken,
@@ -2498,6 +2505,17 @@ final class AppState: ObservableObject {
                     hasBoundaryAnchor: boundary?.streamTextAtBoundary?
                         .trimmingCharacters(in: .whitespacesAndNewlines)
                         .isEmpty == false
+                )
+            }
+            if reconcileExplicitYolo {
+                // A session.info received while resume was in flight may be a
+                // pre-resume snapshot. Do not replay a contradictory YOLO
+                // value after fresh resume reconciliation clears an override;
+                // pushes received after reconciliation remain authoritative.
+                bufferedEvents = Self.filteringBufferedSessionInfo(
+                    bufferedEvents,
+                    against: result.snapshot,
+                    acceptedSessionIDs: reconciliation?.acceptedSessionIDs ?? Set([result.sessionId])
                 )
             }
             bufferedEvents.forEach(applyStreamEvent)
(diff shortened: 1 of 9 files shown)


kaishi00:
## Summary
This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in an injectable `SessionYoloStore`, applies them with precedence (local override > snapshot `yolo` > profile `approvals_mode`), writes them only after the gateway accepts the setting, and adds store/AppState test coverage. The core mechanism is sound and well-tested.

## Findings

### [severity: major] Model picker "Apply" persists a spurious YOLO override without a user change
**File:** Conduit/Views/ModelPickerView.swift:405 (with Conduit/Services/AppState.swift:4325-4330)
**Issue:** `applyModel()` calls `appState.setYoloMode(yoloEnabled)` unconditionally on every Apply, and `yoloEnabled` was seeded from `appState.runtime.yolo`. Now that `setYoloMode` writes a permanent override after a successful gateway call, a user who only changes the model (never touching the YOLO toggle) will freeze that session's YOLO state in the store. If the server later changes the session's YOLO (gateway reset, another client), the spurious override shadows it indefinitely — even though the user never expressed intent to lock it.
**Suggestion:** Only call `setYoloMode` when the value actually changed (e.g., `guard yoloEnabled != appState.runtime.yolo` before persisting), or split the gateway `config.set` from the override write so the override is only recorded on an explicit toggle.

### [severity: major] Local override permanently shadows server-reported YOLO with no way to clear it
**File:** Conduit/Services/AppState.swift:3170-3180
**Issue:** Once written, an override wins over an explicit `snapshot.yolo` on every `applyRuntime` call, forever, and there is no API to remove an override. If the session's YOLO state legitimately changes server-side (another client, `/approvals` executed against the gateway, gateway restart), the app can display YOLO enabled while the server requires approval — or show "auto-approve enabled" while the server will auto-approve, which is a safety-relevant UI divergence. Note this is documented in the design spec, but the divergence risk and the absence of any reset/expiry path deserve review.
**Suggestion:** Consider clearing the override when a snapshot explicitly carries a session-level `yolo` that differs from the stored value (trusting explicit server truth over a possibly-stale local record), or add a mechanism to remove/reset the override.

### [severity: minor] Write-time vs read-time canonicalization can disagree when the catalog isn't populated
**File:** Conduit/Services/AppState.swift:4325 (write), 3146-3153/3170 (read)
**Issue:** `setYoloMode` keys the override by `canonicalSessionID(for:)`, which resolves through `sessions + cronSessions` and `activeChatScrollSessionIdentity`. If the toggle happens before the catalog is available (e.g., model picker right after launch, or a failed catalog fetch), the override is stored under the raw runtime ID. Later reads, once the catalog is present, canonicalize to the stored ID and miss the override, so the choice appears lost. The common launch path loads the catalog before resume (AppState.swift:2034), so this is an edge case, not a regression in the headline flow.
**Suggestion:** Persist under both the canonical and raw IDs on write, or re-key overrides when the catalog/identity is refreshed.

### [severity: minor] Overrides accumulate without eviction
**File:** Conduit/Services/SessionYoloStore.swift:41-49, 9-12
**Issue:** `setOverride` only ever adds/updates entries; archived or deleted sessions leave their overrides in `UserDefaults` indefinitely, so the payload grows monotonically with every session ever toggled.
**Suggestion:** Prune entries when a session is evicted/archived from the catalog, or cap/tombstone stale profile entries.

### [severity: nit] Silent persistence failures
**File:** Conduit/Services/SessionYoloStore.swift:56-59 (and 25-31)
**Issue:** `persist()` swallows encode failures and `init` silently drops corrupt/mismatched-version payloads, so an override loss is invisible. This matches codebase conventions, but the loss of user choice on a corrupt store is silent.
**Suggestion:** Log on failure (e.g., `os.Logger`) so store corruption is diagnosable.

### [severity: nit] Profile key normalization differs from the rest of the app
**File:** Conduit/Services/SessionYoloStore.swift:51-53
**Issue:** `ChatScrollIdentityNormalization.profile` lowercases the profile, while elsewhere the app treats profile names case-sensitively (e.g., `profile == activeProfile`). A profile named `"Work"` and `"work"` would be collapsed into one override bucket here but kept separate by the app.
**Suggestion:** Confirm profile names are always lowercase/normalized at the source; if so, note it, otherwise keep a single normalization policy for profile identity.

## Verdict
COMMENT

The change is well-structured, the precedence logic is correctly ordered, the write-after-gateway-success behavior is right, and the tests meaningfully cover true/false/absent, isolation, relaunch, switching, and failure paths. The two concerns worth resolving before merge are the spurious override created by an unchanged model-picker apply (a behavior regression introduced here) and the indefinite shadowing of server-reported YOLO with no reset path.

<a href="https://opencode.ai/s/GbL3kAUq">New%20session%20-%202026-08-13T13%3A06%3A11.481Z</a>
[opencode session](https://opencode.ai/s/GbL3kAUq)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31703149956)

kaishi00:
## Summary
This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in a new injectable `SessionYoloStore`, applies them with the precedence snapshot.yolo > local override > profile `approvals_mode`, writes them only after the gateway accepts the change, and adds isolated store/AppState/model-picker coverage. The two prior "major" findings (spurious override from unchanged model-picker Apply, and indefinite shadowing of server-reported YOLO) are both addressed: `applyModel` now only persists on an actual toggle change (ModelPickerView.swift:412), and `applyRuntime` clears a conflicting local override when a snapshot explicitly carries `yolo` (AppState.swift:3178-3184).

## Findings

### [severity: minor] YOLO toggle change silently dropped if Apply happens before/without model load
**File:** Conduit/Views/ModelPickerView.swift:392-396, 412-414
**Issue:** `initialYoloEnabled` is seeded only inside `loadModels()` and remains `nil` on failure (the `catch` swallows the error) or if the user taps Apply before the async load returns. `sessionYoloSelectionChanged(from: nil, to: …)` always returns `false`, so an explicit user toggle is silently not applied while the sheet still dismisses. Additionally, until `loadModels` completes the toggle displays the `@State` default `false`, misrepresenting `runtime.yolo == true`. (Strictly better than the old unconditional write of `false`, but the silent drop is a real UX gap.)
**Suggestion:** Seed `initialYoloEnabled = appState.runtime.yolo` and `yoloEnabled = appState.runtime.yolo` synchronously in `onAppear`/init (or in the button handler before the network call), so the comparison is valid even when `loadModels` hasn't finished.

### [severity: minor] Overrides are never evicted when sessions are archived or deleted
**File:** Conduit/Services/SessionYoloStore.swift:50-58 (with Conduit/Services/AppState.swift:3664, 3789)
**Issue:** `setOverride` only adds/updates entries; `archiveSession` and `deleteSession` never touch the store, so every session a user ever toggles leaves a permanent `UserDefaults` entry and the payload grows monotonically. Toggling a YOLO value also leaves a dead entry under the old raw runtime ID if the catalog later resolves a different canonical ID (the read path prefers canonical, so the raw entry becomes orphaned).
**Suggestion:** Clear the override in `archiveSession`/`deleteSession` (and on catalog canonicalization), or prune entries during snapshot application when a session is no longer present.

### [severity: nit] Persistence failures are fully silent
**File:** Conduit/Services/SessionYoloStore.swift:25-31, 87-90
**Issue:** `persist()` swallows encode errors and `init` silently discards corrupt or mismatched-version payloads (resetting to empty), so a user's saved YOLO choices can vanish without any trace.
**Suggestion:** Log failures with `os.Logger` (used elsewhere in the codebase, e.g., `import OSLog` in AppState.swift) so store corruption is diagnosable.

### [severity: nit] PR description precedence wording contradicts the implementation
**File:** PR body "Precedence" claim vs Conduit/Services/AppState.swift:3178-3184
**Issue:** The PR summary says "local override > snapshot `yolo`", but the code makes an explicit snapshot `yolo` authoritative and clears a conflicting override — the correct behavior per the "server-authority" follow-up and the tests (`testExplicitGatewayYoloReplacesConflictingLocalOverride`). The summary line is misleading.
**Suggestion:** Fix the summary to state the implemented precedence: explicit snapshot `yolo` > local override > profile `approvals_mode`.

## Verdict
COMMENT

The mechanism is sound, well-isolated, and thoroughly tested; both previously-flagged major issues are resolved. Remaining findings are minor/nit and don't block merge, but the apply-before-load silent drop and the unbounded override growth are worth a follow-up.

<a href="https://opencode.ai/s/jajGiRrV">New%20session%20-%202026-08-13T14%3A34%3A31.489Z</a>
[opencode session](https://opencode.ai/s/jajGiRrV)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31710810502)

kaishi00:
## Summary
This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in a new `@MainActor` `SessionYoloStore`, applies them with precedence (snapshot `yolo` > local override > profile `approvals_mode`), writes the override only after the gateway accepts the change, migrates raw runtime IDs to canonical catalog IDs, evicts overrides on archive/delete, and adds store/AppState/model-picker test coverage.

## Findings

### [severity: major] Stale in-flight snapshot can silently delete a just-persisted override and revert the toggle
**File:** Conduit/Services/AppState.swift:3187-3193
**Issue:** The clear-on-conflict path runs on every `applyRuntime`, including live `.sessionInfo` pushes (AppState.swift:6413-6414). If the user toggles YOLO via `setYoloMode`, a session snapshot that was captured *before* the gateway applied the change (and carries the old `yolo`) can be delivered after the toggle RPC returns — WebSocket pushes do not wait for client-initiated RPCs. That snapshot differs from `storedOverride`, so the code clears the override and sets `runtime.yolo` back to the old value. The user's explicit, persisted choice is silently destroyed and the UI flips back with no recovery path. Since `setYoloMode` also updates `runtime.yolo` directly, a later snapshot omitting `yolo` (older gateway) then falls back to the profile `approvals_mode`, so the choice is genuinely lost rather than merely delayed. `sessionInfo` is emitted frequently (running/busy/status changes all push snapshots), so the window is not purely theoretical.
**Suggestion:** Guard the clear so a locally-written value isn't immediately clobbered — e.g., record the last locally-persisted value + timestamp per session and ignore a conflicting snapshot clear for a short grace window, or only honor the clear when the snapshot is confirmed to reflect post-toggle gateway state (e.g., on resume, not on every live push).

### [severity: minor] Write-time vs read-time canonicalization can diverge, leaving orphaned raw-ID entries
**File:** Conduit/Services/AppState.swift:4352 (write) vs 3171-3186 (read)
**Issue:** `setYoloMode` persists under `canonicalSessionID(for:) ?? rawID`, computed against `sessions + cronSessions` at write time. If the catalog isn't populated yet (toggle in the model picker immediately after launch, or a failed catalog fetch), the override lands under the raw runtime ID. Reads do check both `[canonical, requested]`, so the value is still found, but `canonicalizeOverride` only migrates (and removes) the raw entry when a later snapshot actually flows through `applyRuntime`. If the catalog later resolves the raw ID to a different canonical key than where an entry was migrated, or no further snapshot arrives, the raw entry survives as a permanent orphan and the payload grows.
**Suggestion:** Run the canonicalize/read with the same fallback consistently (already mostly done) and prune orphaned keys during catalog reconciliation rather than only on snapshot application.

### [severity: minor] Overrides accumulate for sessions that leave the catalog without archive/delete
**File:** Conduit/Services/SessionYoloStore.swift:44-58 (with AppState.swift:3701-3706, 3817-3822)
**Issue:** `setOverride` only adds/updates, and `clearOverride` is invoked only on archive/delete and during snapshot canonicalization. A session evicted server-side (dropped from the catalog without an explicit archive/delete) leaves its entry in `UserDefaults` forever, so the payload grows monotonically with every session ever toggled.
**Suggestion:** Prune overrides for session IDs no longer present in the catalog during snapshot/catalog application, in addition to the existing archive/delete paths.

### [severity: nit] `applyModel` partially applies settings before the YOLO write can fail
**File:** Conduit/Views/ModelPickerView.swift:430-432
**Issue:** If the YOLO toggle actually changed and `setYoloMode` fails, the `guard ... else { return }` returns without dismissing the sheet, but model/provider/reasoning/fast were already applied to both the gateway and `runtime`. The sheet stays open showing stale state while runtime was mutated, with no rollback. This ordering predates the PR, but the new override-persistence makes the divergence more visible.
**Suggestion:** Apply YOLO (or all settings) transactionally — either persist all-or-nothing, or dismiss/refresh the sheet consistently on failure.

## Verdict
REQUEST_CHANGES

The mechanism is well-structured, correctly ordered (snapshot `yolo` authoritative, override otherwise, then `approvals_mode`), write-after-gateway-success is right, store isolation and versioning are solid, and the test coverage (true/false/absent, relaunch, switching, precedence, failure, canonicalization) is good. The blocking item is the stale-snapshot race in `applyRuntime`, which can silently destroy the persisted override this PR exists to protect — that needs an ordering/grace guard before merge. The canonicalization divergence and unbounded override growth are worth a follow-up.

<a href="https://opencode.ai/s/ZVniW6pg">New%20session%20-%202026-08-13T15%3A37%3A03.522Z</a>
[opencode session](https://opencode.ai/s/ZVniW6pg)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31716255787)

kaishi00:
## Summary
This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in an injectable `@MainActor SessionYoloStore`, applies them with precedence (explicit snapshot `yolo` > local override > profile `approvals_mode`), writes the override only after the gateway accepts the change, migrates raw runtime IDs to canonical catalog IDs, evicts overrides on archive/delete, and adds focused store/AppState/model-picker tests.

## Findings

### [severity: major] Stale resume snapshot still destroys a just-persisted override
**File:** Conduit/Services/AppState.swift:3194-3199 (resume call site at 2786)
**Issue:** The clear-on-conflict path (`reconcileExplicitYolo: true`) clears the stored override whenever a fresh-resume snapshot carries an explicit `yolo` that differs from it. This closes the live-push race but relocates it to the resume boundary: if a resume RPC was sent *before* the user's `config.set` completed (foreground/reconnect while the Model Picker or composer toggle is in flight), the resume snapshot can predate the toggle and carry the old `yolo`. When that stale snapshot is delivered after `setYoloMode` succeeded, the override is cleared and `runtime.yolo` is flipped back — exactly the "snapshot captured before the toggle clears the new override and reverts the UI" root cause this PR describes in its own summary. Because `setYoloMode` also writes `runtime.yolo` directly, the value is then genuinely lost, not merely delayed. There is no timestamp/ordering guard to distinguish a stale resume snapshot from genuine server authority.
**Suggestion:** Record the last local write time per (profile, session) in the store, and skip the clear-on-conflict reconciliation for a short grace window after a successful local write (or only reconcile when the resume was initiated after the last local write). At minimum, assert/verify the ordering assumption with a test that applies a stale resume snapshot after `setYoloMode` succeeds.

### [severity: minor] Overrides accumulate without eviction for sessions that leave the catalog without archive/delete
**File:** Conduit/Services/SessionYoloStore.swift:83-91 (with AppState.swift:3718, 3834)
**Issue:** `setOverride` only ever adds/updates, and `clearOverride` is invoked only on explicit archive/delete and during snapshot canonicalization. A session evicted server-side (dropped from the catalog with no archive/delete) keeps its `UserDefaults` entry forever, and the payload grows monotonically with every session ever toggled. Orphaned raw-runtime-ID entries also survive if the catalog never resolves them to a canonical ID and no later snapshot flows through `applyRuntime`.
**Suggestion:** Prune overrides for session IDs no longer present in the catalog during catalog/snapshot reconciliation, in addition to the existing archive/delete paths.

### [severity: minor] Server-side YOLO changes are shadowed until the next fresh resume
**File:** Conduit/Services/AppState.swift:3192-3205
**Issue:** When a stored override exists, live `session.info` pushes always keep `runtime.yolo = storedOverride` even when the server reports a differing explicit `yolo` (e.g., changed by another client or `/approvals` against the gateway). The divergence persists for the entire foreground session, which is safety-relevant: the UI can show "auto-approve enabled" while the server will require approval (or vice-versa). The clear-on-conflict only runs on fresh resume, so this is bounded but unbounded-in-time within a session.
**Suggestion:** Acceptable as documented, but consider reconciling on a server-initiated change signal (e.g., clear the override when the gateway applies an external YOLO change), or surface the divergence.

### [severity: nit] Model Picker `onAppear` re-seeds the YOLO draft and can silently drop an in-progress toggle
**File:** Conduit/Views/ModelPickerView.swift:85-89
**Issue:** `onAppear` unconditionally re-seeds `yoloEnabled`/`initialYoloEnabled` from `appState.runtime.yolo`. If `onAppear` fires again for the same presentation (known SwiftUI behavior on sheet re-presentation/detent changes), a user's in-progress toggle is reset to the runtime value; since `initialYoloEnabled` is re-seeded to the same value, `sessionYoloSelectionChanged` returns false and the change is silently dropped on Apply.
**Suggestion:** Seed only when `initialYoloEnabled == nil` (matching the guard already used in `loadModels`), so an existing draft is never clobbered.

### [severity: nit] Module-level helper declarations
**File:** Conduit/Views/ModelPickerView.swift:10-22
**Issue:** `sessionYoloSelectionChanged` and `ModelPickerYoloDraft` are top-level `internal` declarations in the app target, polluting the module namespace; the nil-`initial` guard they encode is also effectively dead now that `onAppear` always seeds the draft.
**Suggestion:** Scope them to the file (`private`) or nest them, and consider dropping the nil branch once seeding is guaranteed.

## Verdict
REQUEST_CHANGES

The mechanism is well-structured, the write-after-gateway-success ordering is correct, canonicalization/eviction and store versioning are solid, and the test coverage (true/false/absent, relaunch, switching, precedence, failure, canonicalization, stale-push) is good. The blocking item is that the stale-snapshot race this PR exists to fix still exists at the resume boundary: a resume snapshot captured before a completed toggle clears the persisted override and reverts the UI. A last-write grace window closes it cleanly and is worth landing before merge. The remaining findings are minor/nit follow-ups.

<a href="https://opencode.ai/s/wzQqIDzk">New%20session%20-%202026-08-13T15%3A49%3A38.850Z</a>
[opencode session](https://opencode.ai/s/wzQqIDzk)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31717386965)

kaishi00:
## Summary

This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in an injectable `@MainActor SessionYoloStore`, applies them with precedence (local override vs. explicit snapshot `yolo` vs. profile `approvals_mode`), writes only after the gateway accepts the change, and adds a per-session local-write revision baseline so stale live pushes and in-flight resume responses cannot clear a newer override. Previous review findings (spurious model-picker override, indefinite shadowing, stale resume race) are addressed.

## Findings

### [severity: minor] Stale buffered `sessionInfo` replay can re-flip the UI after a fresh-resume override clear
**File:** Conduit/Services/AppState.swift:3256-3261 (clear path) with 2503 (buffered replay)
**Issue:** The server-authority clear only runs on a fresh resume (`reconcileExplicitYolo: true`) and does not bump `sessionYoloWriteRevisions`. A `.sessionInfo` push buffered during reconciliation (AppState.swift:6375-6384) is replayed at AppState.swift:2503 *after* the clear. If that buffered push predates the server-side change and still carries the old `yolo`, `storedOverride` is now nil so `applyRuntime` applies the stale value directly (AppState.swift:3268-3269) — the UI flips back to the pre-change state, exactly what this PR exists to prevent. It self-corrects on the next push and no persisted data is lost, but the divergence window (including the safety-relevant "auto-approve enabled" display) contradicts the stated guarantee.
**Suggestion:** After a server-authority clear, also bump the write revision (or record a "reconciled" marker) so buffered/live pushes carry the reconcile guard, or filter stale `sessionInfo` from the replay set when it contradicts the just-applied resume snapshot.

### [severity: minor] Live-push override protection hides genuine server-side YOLO changes for the entire foreground session
**File:** Conduit/Services/AppState.swift:3262-3267
**Issue:** When a stored override exists, every live `session.info` push keeps `runtime.yolo = storedOverride` even when the server reports a differing explicit `yolo` (external change by another client or `/approvals`). The UI can show "auto-approve enabled" while the server requires approval, or vice-versa, for the whole session — the clear-on-conflict only runs on a fresh resume. This is documented in the design spec, so it's acceptable, but the divergence is safety-relevant and has no in-session signal to trigger reconciliation.
**Suggestion:** Consider reconciling on a server-initiated change signal (e.g., when a `sessionInfo` push's `yolo` differs and no client RPC was recently recorded for that session), or surface the divergence in the UI.

### [severity: minor] Overrides never evicted for sessions dropped from the catalog without archive/delete
**File:** Conduit/Services/SessionYoloStore.swift:83-91, 132-152
**Issue:** `setOverride` only adds/updates, and `clearOverride` is invoked only on archive/delete (AppState.swift:3780, 3893) and during alias canonicalization. A session removed server-side without an explicit archive/delete keeps its `UserDefaults` entry forever, and an override written under a raw runtime ID that never gets canonicalized (catalog empty at write time, no later snapshot) survives as a permanent orphan. The payload grows monotonically with every session ever toggled.
**Suggestion:** Prune overrides for session IDs absent from the catalog during catalog/snapshot reconciliation, in addition to the existing archive/delete paths.

### [severity: minor] `applyModel` partially applies settings before a YOLO write failure
**File:** Conduit/Views/ModelPickerView.swift:420-433
**Issue:** If the YOLO toggle changed and `setYoloMode` fails, the `guard ... else { return }` at line 431 exits without dismissing the sheet, but model/provider/reasoning/fast were already applied to the gateway and `runtime`. The sheet stays open showing a YOLO toggle that was not applied while runtime was otherwise mutated, with no rollback.
**Suggestion:** Apply YOLO (or all settings) transactionally — persist all-or-nothing, or dismiss/refresh consistently on failure.

### [severity: nit] Model Picker `onAppear` unconditionally re-seeds the YOLO draft
**File:** Conduit/Views/ModelPickerView.swift:85-89
**Issue:** `onAppear` re-seeds both `yoloEnabled` and `initialYoloEnabled` from `appState.runtime.yolo` on every appear. If `onAppear` fires again for the same presentation (detent/layout re-layout behavior), an in-progress toggle is reset and, since `initialYoloEnabled` is re-seeded to the same value, `sessionYoloSelectionChanged` returns false so the change is silently dropped on Apply. Also, seeding from `runtime.yolo` before resume reconciliation has settled could capture a stale baseline.
**Suggestion:** Seed only when `initialYoloEnabled == nil`, matching the guard already used in `loadModels`.

### [severity: nit] Module-level helper declarations pollute the module namespace
**File:** Conduit/Views/ModelPickerView.swift:10-23
**Issue:** `sessionYoloSelectionChanged` and `ModelPickerYoloDraft` are top-level `internal` declarations in the app target, and the nil-`initial` branch of `sessionYoloSelectionChanged` is effectively dead now that `onAppear` always seeds the draft.
**Suggestion:** Scope them to the file (`private`) or nest them, and drop the nil branch once seeding is guaranteed.

### [severity: nit] Design doc precedence list contradicts the implemented precedence
**File:** docs/superpowers/specs/2026-08-13-session-yolo-persistence-design.md:49-54
**Issue:** The spec's three-item precedence lists explicit `snapshot.yolo` as highest, but the implementation gives the local override precedence over explicit snapshot `yolo` on live pushes (AppState.swift:3262-3267) and only lets explicit `yolo` win during fresh-resume reconciliation with no newer local write.
**Suggestion:** Update the precedence list to match the implemented behavior (explicit snapshot `yolo` authoritative on fresh resume; local override otherwise; `approvals_mode` fallback).

## Verdict

COMMENT

The core mechanism is sound: the write-revision baseline correctly closes the previously-flagged stale resume and live-push races, the store is well-isolated and versioned with diagnostics, canonicalization/eviction on archive/delete is in place, and the tests meaningfully cover true/false/absent, relaunch, switching, precedence, failure, canonicalization, and stale-ordering. Remaining findings are minor/nit and non-blocking; the buffered-replay gap and the unbounded override growth are the most worthwhile follow-ups.

<a href="https://opencode.ai/s/yGQIvXS2">New%20session%20-%202026-08-13T16%3A05%3A20.700Z</a>
[opencode session](https://opencode.ai/s/yGQIvXS2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31718770503)

kaishi00:
## Summary
Persists explicit session-scoped YOLO choices as normalized profile/session overrides in an injectable `@MainActor SessionYoloStore`, applies them with precedence (explicit snapshot `yolo` authoritative on fresh resume > local override > profile `approvals_mode`), writes only after gateway acceptance, and adds a per-session write-revision baseline plus buffered-event filtering so stale resume/live snapshots cannot clear a newer override. Prior blocking races (stale resume, stale live push, buffered replay) are addressed and the logic is well-isolated and tested.

## Findings

### [severity: minor] Server-authoritative YOLO changes remain shadowed for the entire foreground session
**File:** Conduit/Services/AppState.swift:3290-3303
**Issue:** When a stored override exists, every live `session.info` push (delivered via `.sessionInfo` at AppState.swift:6526-6527) keeps `runtime.yolo = storedOverride` even when the server explicitly reports a differing `yolo` (external change by another client, `/approvals`, or a gateway reset). The clear-on-conflict runs only on fresh resume (`reconcileExplicitYolo`), so the UI can display "auto-approve enabled" while the server will require approval — or vice-versa — for the whole session with no in-session signal to reconcile. This is documented in the design spec, but it is safety-relevant and the divergence window is unbounded within a session.
**Suggestion:** Reconcile on a server-initiated signal, e.g. treat a live push's explicit `yolo` as authoritative when no client RPC for that session was recorded recently (use the existing `sessionYoloWriteRevisions` as a timestamp proxy), or surface the divergence in the UI.

### [severity: minor] Buffered sessionInfo filter drops non-YOLO fields and can discard genuinely newer events
**File:** Conduit/Services/AppState.swift:3122-3140 (filter), 2510-2520 (call site)
**Issue:** `filteringBufferedSessionInfo` drops whole `sessionInfo` events whose derived YOLO contradicts the resume snapshot, discarding that event's other fields (`running`, `model`, `context`) from replay. It also derives `eventYolo` from `approvalsMode`, so an event with `approvalsMode: "on"` is dropped against a resume `yolo: true` even though it may carry a newer `running: false`. Additionally the filter is gated on the outer `reconcileExplicitYolo` (computed at line 2414, before `applyChatResume`'s awaits), which can diverge from the inner flag that actually drove the clear (line 2833) — in the worst case the filter runs when no clear happened and drops a genuinely post-write buffered event.
**Suggestion:** Filter only on the event's explicit `yolo` field (not `approvalsMode`), replay YOLO-filtered events for their other fields, and compute the filter guard from the same flag instance used by the clear.

### [severity: minor] Override entries and the in-memory revision table grow unbounded for catalog-evicted sessions
**File:** Conduit/Services/SessionYoloStore.swift:83-91 (setOverride), AppState.swift:3250-3255 (recordSessionYoloWrite)
**Issue:** `setOverride` only adds/updates; `clearOverride` runs only on archive/delete (AppState.swift:3816, 3932) and alias canonicalization. A session that leaves the catalog server-side without an explicit archive/delete keeps its `UserDefaults` entry forever, and an override written under a raw runtime ID that never gets canonicalized survives as an orphan. `sessionYoloWriteRevisions` likewise retains one entry per session ever toggled for the app's lifetime.
**Suggestion:** Prune overrides (and revision keys) for session IDs absent from the catalog during catalog/snapshot reconciliation, in addition to the existing archive/delete and canonicalization paths.

### [severity: minor] Model Picker `applyModel` partially applies settings when the YOLO RPC fails
**File:** Conduit/Views/ModelPickerView.swift:420-433
**Issue:** If the YOLO toggle changed and `setYoloMode` fails, the `guard ... else { return }` at line 431 exits without dismissing, but model/provider/reasoning/fast were already applied to both the gateway and `runtime`. The sheet stays open showing a YOLO toggle that was not applied while runtime was otherwise mutated, with no rollback or refresh.
**Suggestion:** Apply the YOLO setting before the other settings (or persist all-or-nothing), and dismiss or refresh the sheet consistently on failure.

### [severity: nit] Model Picker `onAppear` unconditionally re-seeds the YOLO draft
**File:** Conduit/Views/ModelPickerView.swift:85-89
**Issue:** `onAppear` re-seeds both `yoloEnabled` and `initialYoloEnabled` from `appState.runtime.yolo` on every appear, whereas `loadModels` guards with `initialYoloEnabled == nil` (line 407). If `onAppear` fires again for the same presentation (SwiftUI sheet re-presentation/detent behavior), an in-progress toggle is reset and — because `initialYoloEnabled` is re-seeded to the same value — `sessionYoloSelectionChanged` returns false, silently dropping the change on Apply.
**Suggestion:** Seed only when `initialYoloEnabled == nil`, matching the `loadModels` guard, so an existing draft is never clobbered.

### [severity: nit] Module-level helper declarations pollute the module namespace
**File:** Conduit/Views/ModelPickerView.swift:10-23
**Issue:** `sessionYoloSelectionChanged` and `ModelPickerYoloDraft` are top-level `internal` declarations in the app target; the nil-`initial` branch of `sessionYoloSelectionChanged` is effectively dead now that `onAppear` always seeds the draft.
**Suggestion:** Mark them `fileprivate` (the tests can keep their own copies or test through the store/applyModel path) and drop the nil branch once seeding is guaranteed.

### [severity: nit] Monotonic write-revision counter can wrap and is never compacted
**File:** Conduit/Services/AppState.swift:3251
**Issue:** `sessionYoloWriteRevision &+= 1` wraps on overflow, and ordering comparisons against a baseline would then be wrong (theoretically, after 2^64 writes). The `[ChatScrollSessionKey: UInt64]` table is also never pruned, so it grows with every unique session toggled.
**Suggestion:** Use a monotonic non-wrapping counter or rely on a per-key timestamp comparison; prune keys alongside the catalog-based override eviction.

## Verdict
COMMENT

The core mechanism is sound: the write-revision baseline closes the previously-flagged stale-resume and stale-push races, write-after-gateway-success and the true/false/absent triple state are correct, canonicalization/eviction on archive/delete is in place, and the tests meaningfully cover relaunch, switching, precedence, failure, canonicalization, stale-ordering, and buffered-replay. No blocking issue remains; the shadowing of external server changes, the buffered-filter field loss, and the unbounded store/revision growth are the most worthwhile follow-ups.

<a href="https://opencode.ai/s/b2xoXTcN">New%20session%20-%202026-08-13T16%3A26%3A38.549Z</a>
[opencode session](https://opencode.ai/s/b2xoXTcN)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31720601075)

kaishi00:
## Summary
This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in an injectable `@MainActor SessionYoloStore`, applies them with precedence (explicit snapshot `yolo` authoritative on fresh resume > local override > profile `approvals_mode`), writes only after gateway acceptance, and adds a per-session write-revision baseline plus buffered-event YOLO anchoring so stale resume/live snapshots cannot clear a newer override. The previously-flagged stale-resume, stale-push, and buffered-replay races are closed and covered by focused tests.

## Findings

### [severity: minor] Stored override shadows server-reported YOLO for the entire foreground session
**File:** Conduit/Services/AppState.swift:3268-3280
**Issue:** When a stored override exists, every live `session.info` push keeps `runtime.yolo = storedOverride` even when the snapshot explicitly reports a differing `yolo` (change by another client, `/approvals` against the gateway, or a gateway reset). The clear-on-conflict runs only on fresh resume (`reconcileExplicitYolo`), so the UI can show "auto-approve enabled" while the server will require approval — or vice versa — for the whole foreground session with no in-session reconcile signal. This is documented design, but the divergence is safety-relevant and unbounded within a session.
**Suggestion:** Reconcile on a server-initiated signal — e.g., treat a live push's explicit `yolo` as authoritative when no client `setYoloMode` for that session was recorded recently (the existing `sessionYoloWriteRevisions` can serve as the timestamp proxy) — or surface the divergence in the UI.

### [severity: minor] Overrides and the revision table grow unbounded for catalog-evicted sessions
**File:** Conduit/Services/SessionYoloStore.swift:83-91 (with Conduit/Services/AppState.swift:3793, 3909)
**Issue:** `setOverride` only ever adds/updates, and `clearOverride` is invoked only on explicit archive/delete and alias canonicalization. A session dropped server-side without an archive/delete keeps its `UserDefaults` entry forever, and an override written under a raw runtime ID that never flows through `applyRuntime`'s `canonicalizeOverride` (catalog empty at write time, no later snapshot) survives as a permanent orphan. `sessionYoloWriteRevisions` (AppState.swift:3226-3231) likewise retains one entry per session ever toggled for the app's lifetime.
**Suggestion:** Prune override entries and revision keys for session IDs absent from the catalog during catalog/snapshot reconciliation, in addition to the existing archive/delete and canonicalization paths.

### [severity: minor] `applyModel` partially applies settings when the YOLO RPC fails
**File:** Conduit/Views/ModelPickerView.swift:429-445
**Issue:** Model/provider/reasoning/fast are applied to both the gateway and `runtime` before `setYoloMode` runs. If the YOLO toggle changed and `setYoloMode` fails, the `guard ... else { return }` at line 440 exits without dismissing the sheet, but the other settings were already mutated — the sheet stays open showing a YOLO toggle that was not applied, with no rollback or refresh.
**Suggestion:** Apply the YOLO setting first (or before the mutating calls), or apply all settings all-or-nothing and dismiss/refresh the sheet consistently on failure.

### [severity: nit] Monotonic write-revision counter wraps and the table is never compacted
**File:** Conduit/Services/AppState.swift:3227
**Issue:** `sessionYoloWriteRevision &+= 1` wraps on overflow, after which ordering comparisons against a baseline are wrong (theoretically, after 2^64 writes). The `[ChatScrollSessionKey: UInt64]` table is also never pruned, so it grows with every unique session toggled.
**Suggestion:** Use a monotonic non-wrapping counter or a per-key timestamp comparison, and prune keys alongside the catalog-based override eviction.

### [severity: nit] Module-level helper declarations pollute the module namespace
**File:** Conduit/Views/ModelPickerView.swift:10-23
**Issue:** `sessionYoloSelectionChanged` and `ModelPickerYoloDraft` are top-level `internal` declarations in the app target; the nil-`initial` branch of `sessionYoloSelectionChanged` is effectively dead in the real view since `onAppear` now seeds the draft synchronously.
**Suggestion:** Mark them `fileprivate` (the tests exercise them directly) or nest them, and drop the nil branch once seeding is guaranteed.

## Verdict
COMMENT

The core mechanism is sound: the write-revision baseline correctly closes the stale-resume and stale-live-push races, fresh-resume server authority plus buffered YOLO anchoring handles out-of-order delivery, write-after-gateway-success and the true/false/absent triple state are correct, canonicalization/eviction on archive/delete is in place, and the store/AppState/model-picker tests meaningfully cover relaunch, switching, precedence, failure, stale-ordering, and buffered-replay. No blocking issue remains; the shadowing of external server changes, unbounded override/revision growth, and the partial `applyModel` application are the most worthwhile follow-ups.

<a href="https://opencode.ai/s/7oHp7S2K">New%20session%20-%202026-08-13T16%3A53%3A37.088Z</a>
[opencode session](https://opencode.ai/s/7oHp7S2K)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31722867255)

kaishi00:
## Summary
This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in an injectable `@MainActor SessionYoloStore`, applies them with precedence (explicit snapshot `yolo` authoritative on fresh resume > local override > profile `approvals_mode`), writes only after the gateway accepts the change, migrates raw runtime IDs to canonical catalog IDs, evicts on archive/delete, and adds a per-session local-write revision baseline plus buffered-event YOLO anchoring to close stale-resume/live-push/buffered-replay races. Prior review rounds' blocking issues are addressed and covered by focused tests.

## Findings

### [severity: minor] Resume-authority clear leaves a window where a stale live push can still revert the UI
**File:** Conduit/Services/AppState.swift:3269-3280
**Issue:** When `reconcileExplicitYolo` is true and a conflicting override is cleared, the clear does not record any revision/marker (unlike `setYoloMode`). Live `session.info` pushes after the reconciliation settles take the `else` branch — `storedOverride` is now nil, so a push captured server-side before the YOLO change (the same out-of-order delivery this PR's root-cause describes) applies its stale `yolo` directly and flips `runtime.yolo` back. It self-corrects on the next push and no persisted data is lost, but the divergence window (including the safety-relevant "auto-approve enabled" display) contradicts the stated guarantee. There is no test covering a stale live push after a resume-authority clear.
**Suggestion:** On a server-authority clear, record a short-lived "reconciled value" per session (or bump the write revision and have live pushes consult the same per-session last-write/reconcile timestamp) so pushes within the transition window keep the reconciled value, and add a test for the clear-then-stale-push ordering.

### [severity: minor] Overrides and the in-memory revision table grow unbounded for catalog-evicted sessions
**File:** Conduit/Services/SessionYoloStore.swift:44-58, Conduit/Services/AppState.swift:3226-3231
**Issue:** `setOverride` only ever adds/updates; `clearOverride` runs only on archive/delete (AppState.swift:3793, 3906) and alias canonicalization. A session dropped server-side without an explicit archive/delete keeps its `UserDefaults` entry forever, and an override written under a raw runtime ID that never flows through `applyRuntime`'s canonicalize path survives as a permanent orphan. `sessionYoloWriteRevisions` likewise retains one key per session ever toggled for the app's lifetime. The persisted payload grows monotonically with every session ever toggled.
**Suggestion:** Prune override entries (and revision keys) for session IDs absent from the catalog during catalog/snapshot reconciliation, in addition to the existing archive/delete and canonicalization paths.

### [severity: minor] Stored override shadows server-reported YOLO for the entire foreground session
**File:** Conduit/Services/AppState.swift:3267-3280
**Issue:** When a stored override exists, every live `session.info` push keeps `runtime.yolo = storedOverride` even when the snapshot explicitly reports a differing `yolo` (external change by another client or `/approvals` against the gateway). The clear-on-conflict runs only on fresh resume, so the UI can display "auto-approve enabled" while the server will require approval — or vice-versa — for the whole session with no in-session reconcile signal. This is documented design, but the divergence is safety-relevant and unbounded within a session.
**Suggestion:** Reconcile on a server-initiated signal, e.g. treat a live push's explicit `yolo` as authoritative when no client `setYoloMode` for that session was recorded recently (the existing `sessionYoloWriteRevisions` can serve as the timestamp proxy), or surface the divergence in the UI.

### [severity: minor] `applyModel` still partially applies when a non-YOLO setting fails after YOLO succeeds
**File:** Conduit/Views/ModelPickerView.swift:434-446
**Issue:** The YOLO-first ordering correctly prevents YOLO from being applied after other settings, but if `setModel`/`setReasoning`/`setFast` throws after `setYoloMode` succeeded, the YOLO override is already persisted and `runtime.yolo` updated while the sheet stays open (errorMessage set, no dismissal) with the model unchanged — no rollback or refresh. The sheet can linger showing a YOLO toggle whose change was applied while other edits were not.
**Suggestion:** On any failure after a successful YOLO write, either dismiss the sheet or explicitly surface which settings were applied, so the visible state doesn't silently diverge from the persisted state.

### [severity: nit] Monotonic write-revision counter wraps and the table is never compacted
**File:** Conduit/Services/AppState.swift:3227
**Issue:** `sessionYoloWriteRevision &+= 1` wraps on overflow, after which ordering comparisons against a baseline are wrong (theoretically, after 2^64 writes). The `[ChatScrollSessionKey: UInt64]` table is also never pruned.
**Suggestion:** Use a monotonic non-wrapping counter or a per-key timestamp comparison, and prune keys alongside the catalog-based override eviction.

### [severity: nit] Module-level helper declarations pollute the module namespace
**File:** Conduit/Views/ModelPickerView.swift:10-23
**Issue:** `sessionYoloSelectionChanged` and `ModelPickerYoloDraft` are top-level `internal` declarations in the app target; the nil-`initial` branch of `sessionYoloSelectionChanged` is effectively dead in the real view since `onAppear` now seeds the draft synchronously.
**Suggestion:** Mark them `fileprivate` (or nest them) and drop the nil branch once seeding is guaranteed.

## Verdict
COMMENT

The core mechanism is sound: the write-revision baseline correctly closes the stale-resume and stale-live-push races, fresh-resume server authority plus buffered YOLO anchoring handles out-of-order delivery while preserving buffered non-YOLO fields, write-after-gateway-success and the true/false/absent triple state are correct, canonicalization and archive/delete eviction are in place, and the store/AppState/model-picker tests meaningfully cover relaunch, switching, precedence, failure, canonicalization, stale-ordering, and buffered-replay. No blocking issue remains; the residual stale-live-push-after-clear window, the shadowing of external server changes, and the unbounded override/revision growth are the most worthwhile follow-ups.

<a href="https://opencode.ai/s/plmugGVh">New%20session%20-%202026-08-13T20%3A23%3A51.573Z</a>
[opencode session](https://opencode.ai/s/plmugGVh)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31740130124)

kaishi00:
## Summary
This PR persists explicit, session-scoped YOLO choices as normalized profile/session overrides in an injectable `@MainActor SessionYoloStore`, applies them with precedence (explicit snapshot `yolo` authoritative on fresh resume > local override > profile `approvals_mode`), writes only after the gateway accepts the change, migrates raw runtime IDs to canonical catalog IDs, evicts on archive/delete, and adds a per-session write-revision baseline plus buffered-event YOLO anchoring to close stale-resume/live-push/buffered-replay races. The mechanism is sound and well-tested; the remaining issues are minor/nit.

## Findings

### [severity: minor] `setYoloMode` reads `activeProfile` after the awaited RPC, keying the override to the wrong profile on a mid-flight switch
**File:** Conduit/Services/AppState.swift:4433-4454
**Issue:** `sessionId` and `client` are captured before the `await`, but `activeProfile` is read fresh after the RPC completes (line 4442 for `setOverride`, 4447 for `recordSessionYoloWrite`, 4448 for `runtime.yolo`). `AppState` is `@MainActor`, so the await suspends the actor and a profile or session switch can run concurrently. If the user switches profile while the YOLO RPC is in flight, the override is persisted under the *new* profile while the session ID belongs to the old profile, and `runtime.yolo` is mutated for whatever session is now active — a silent wrong-profile/wrong-session write with no way to detect it.
**Suggestion:** Capture `let profile = activeProfile` before the await (mirroring the `sessionId` capture) and use that for the store write and revision recording.

### [severity: minor] `canonicalizeOverride` discards a newer alias write when a canonical value already exists
**File:** Conduit/Services/SessionYoloStore.swift:105-124
**Issue:** When a toggle is written while the catalog doesn't yet resolve the canonical ID (e.g., right after launch), it lands under the raw runtime ID. If a canonical entry already exists from an earlier toggle, `canonicalizeOverride` keeps the old canonical value and removes the alias entry (lines 114-124) — the user's most recent choice is silently lost rather than migrated.
**Suggestion:** On conflict, prefer the more recently written value, or at minimum log the collision and keep the canonical entry explicit about which value won. Consider writing under both keys in `setYoloMode` (canonical + raw) so migration is never the deciding path.

### [severity: minor] Residual stale-live-push window after a resume-authority clear
**File:** Conduit/Services/AppState.swift:3267-3280 (clear path at 3269-3274)
**Issue:** The fresh-resume clear that runs with `reconcileExplicitYolo: true` does not bump `sessionYoloWriteRevisions`. Once the override is nil, any later live `.sessionInfo` push captured server-side before the change applies its old `yolo` directly (`runtime.yolo = yolo`, line 3282), flipping the UI back. It self-corrects on the next push and no persisted data is lost, but the divergence window contradicts the PR's guarantee.
**Suggestion:** After a server-authority clear, record a short-lived reconcile marker (or bump the write revision) so live pushes within the transition window keep the reconciled value; add a test for clear-then-stale-push ordering.

### [severity: minor] Stored override shadows genuine server-side YOLO changes for the whole foreground session
**File:** Conduit/Services/AppState.swift:3275-3280
**Issue:** When a stored override exists, every live `session.info` push keeps `runtime.yolo = storedOverride` even when the server explicitly reports a differing `yolo` (external change by another client, `/approvals`, or gateway reset). The clear only runs on fresh resume, so the UI can display "auto-approve enabled" while the server will require approval (or vice versa) for the entire session with no in-session reconcile signal.
**Suggestion:** Reconcile on a server-initiated signal — e.g., treat a live push's explicit `yolo` as authoritative when no client `setYoloMode` for that session was recorded recently — or surface the divergence in the UI.

### [severity: minor] Overrides and the revision table grow unbounded for catalog-evicted sessions
**File:** Conduit/Services/SessionYoloStore.swift:83-91, Conduit/Services/AppState.swift:3226-3231
**Issue:** `setOverride` only ever adds/updates, and `clearOverride` runs only on archive/delete and alias canonicalization. A session dropped server-side without an explicit archive/delete keeps its `UserDefaults` entry forever; an override written under a raw runtime ID that never flows through `applyRuntime`'s canonicalize path survives as a permanent orphan. `sessionYoloWriteRevisions` likewise retains one key per session ever toggled for the app's lifetime.
**Suggestion:** Prune override entries and revision keys for session IDs absent from the catalog during catalog/snapshot reconciliation, in addition to the existing archive/delete and canonicalization paths.

### [severity: minor] `applyModel` applies YOLO first, then partially applies if a later setting fails
**File:** Conduit/Views/ModelPickerView.swift:429-450
**Issue:** The YOLO-first ordering correctly bails before mutating other settings when `setYoloMode` fails, but if `setModel`/`setReasoning`/`setFast` throws after `setYoloMode` succeeded, the override is already persisted and `runtime.yolo` updated while the sheet stays open with the other edits unapplied — no rollback or refresh, so the visible sheet diverges from persisted state.
**Suggestion:** On any failure after a successful YOLO write, dismiss the sheet or explicitly surface which settings were applied.

### [severity: nit] Write-revision counter wraps and is never compacted
**File:** Conduit/Services/AppState.swift:3227
**Issue:** `sessionYoloWriteRevision &+= 1` wraps on overflow, after which ordering comparisons against a baseline are wrong (theoretically, after 2^64 writes), and the `[ChatScrollSessionKey: UInt64]` table is never pruned.
**Suggestion:** Use a monotonic non-wrapping counter or per-key timestamp comparisons, and prune keys alongside catalog-based override eviction.

### [severity: nit] Module-level helper declarations pollute the namespace and carry a dead branch
**File:** Conduit/Views/ModelPickerView.swift:10-28
**Issue:** `sessionYoloSelectionChanged` and `ModelPickerYoloDraft` are top-level `internal` declarations in the app target, and the nil-`initial` branch of `sessionYoloSelectionChanged` is effectively dead since `onAppear` seeds the draft synchronously before any interaction.
**Suggestion:** Mark them `fileprivate` or nest them; consider dropping the nil branch once seeding is guaranteed.

## Verdict
COMMENT

The core mechanism is sound: the write-revision baseline closes the stale-resume and stale-live-push races, fresh-resume server authority plus buffered YOLO anchoring handles out-of-order delivery while preserving buffered non-YOLO fields, write-after-gateway-success and the true/false/absent triple state are correct, canonicalization and archive/delete eviction are in place, and the store/AppState/model-picker tests meaningfully cover relaunch, switching, precedence, failure, canonicalization, stale-ordering, and buffered-replay. No blocking issue remains; the mid-RPC profile-switch write and the canonicalize-discards-newer-alias edge are the most worthwhile follow-ups alongside the previously-noted unbounded growth and external-change shadowing.

<a href="https://opencode.ai/s/vgsONQWe">New%20session%20-%202026-08-13T20%3A56%3A13.137Z</a>
[opencode session](https://opencode.ai/s/vgsONQWe)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31743207656)

PR #56, opened 2026-08-13
fix: dismiss login keyboard when tapping outside text fields

kaishi00:
## Problem
On the login screen, tapping outside a text field doesn't dismiss the keyboard, so it stays up and covers the **Connect** button. The only workaround today is to press Enter through the `server → username → password` submit chain, which eventually drops it.

## Fix
`LoginView` already manages focus with `@FocusState private var focusedField: LoginField?`. Added a tap gesture on the login root that clears `focusedField`, which dismisses the keyboard so the Connect button is reachable. Text fields, toggles, and the button still consume their own taps, so their interaction is unchanged.

```swift
.onTapGesture {
    // Tapping outside the text fields clears focus so the keyboard
    // dismisses and the Connect button stays reachable. Interactive
    // controls (fields, toggles, the button) still consume their own taps.
    focusedField = nil
}
```

## Verification
- `xcodebuild` Debug build for the iOS Simulator (generic destination): **BUILD SUCCEEDED**.
- Worth a quick interactive check on device/sim: tapping a field still focuses on the first tap (no two-tap regression), and tapping the backdrop / empty area dismisses the keyboard.

## Summary by CodeRabbit

* **New Features**
  * Added Cloudflare client ID and client secret fields to the login screen.
  * Cloudflare fields now support focus navigation.

* **Bug Fixes**
  * Tapping outside interactive controls on the login screen now dismisses the keyboard and clears the focused text field.
  * Control-specific taps continue to work as expected.

### DIFF
--- a/Conduit/Views/LoginView.swift
+++ b/Conduit/Views/LoginView.swift
@@ -170,6 +170,12 @@ struct LoginView: View {
             }
             .padding(.horizontal, 24)
         }
+        .onTapGesture {
+            // Tapping outside the text fields clears focus so the keyboard
+            // dismisses and the Connect button stays reachable. Interactive
+            // controls (fields, toggles, the button) still consume their own taps.
+            focusedField = nil
+        }
         .onAppear {
             guard serverUrl.isEmpty else { return }
             serverUrl = appState.lastDashboardURL
--- a/Conduit/Views/LoginView.swift
+++ b/Conduit/Views/LoginView.swift
@@ -28,11 +28,21 @@ struct LoginView: View {
 
     private enum LoginField {
         case server, username, password
+        case cloudflareClientID, cloudflareClientSecret
     }
 
     var body: some View {
         ZStack {
             ConduitBackdrop()
+                .contentShape(Rectangle())
+                .onTapGesture {
+                    // Tapping the backdrop (outside any field or control) clears
+                    // focus so the keyboard dismisses and the Connect button is
+                    // reachable. The gesture lives on the background layer — not
+                    // the root — so it never competes with a text field's
+                    // first-responder touch (no two-tap-to-focus regression).
+                    focusedField = nil
+                }
 
             VStack(spacing: 28) {
                 Spacer(minLength: 36)
@@ -125,9 +135,11 @@ struct LoginView: View {
                         if cloudflareEnabled {
                             TextField("Cloudflare Client ID", text: $cloudflareClientID)
                                 .textInputAutocapitalization(.never).autocorrectionDisabled()
+                                .focused($focusedField, equals: .cloudflareClientID)
                                 .padding(.horizontal, 14).frame(height: 50)
                                 .conduitGlassSurface(cornerRadius: 17, tint: .conduitAura.opacity(0.06))
                             SecureField("Cloudflare Client Secret", text: $cloudflareClientSecret)
+                                .focused($focusedField, equals: .cloudflareClientSecret)
                                 .padding(.horizontal, 14).frame(height: 50)
                                 .conduitGlassSurface(cornerRadius: 17, tint: .conduitAura.opacity(0.06))
                             Text("Used only to reach this Cloudflare-protected dashboard; the secret stays in Keychain.")
@@ -169,12 +181,6 @@ struct LoginView: View {
                     .padding(.bottom, 16)
             }
             .padding(.horizontal, 24)
-        }
-        .onTapGesture {
-            // Tapping outside the text fields clears focus so the keyboard
-            // dismisses and the Connect button stays reachable. Interactive
-            // controls (fields, toggles, the button) still consume their own taps.
-            focusedField = nil
         }
         .onAppear {
             guard serverUrl.isEmpty else { return }


kaishi00:
## Summary
This PR adds an `.onTapGesture` to `LoginView`'s root `ZStack` that clears `@FocusState focusedField` when tapping outside interactive controls, so the keyboard dismisses and the Connect button stays reachable.

## Findings

### [severity: major] Container `onTapGesture` risks blocking text-field focus (two-tap regression)
**File:** Conduit/Views/LoginView.swift:173
**Issue:** Attaching `.onTapGesture` to the root `ZStack` (which wraps the `TextField`s) is a known SwiftUI pitfall: the container's tap gesture can consume the touch intended for a `TextField`, preventing it from becoming first responder, so the first tap only dismisses/keyboard-clears and the field needs a second tap to focus. This is precisely the regression the PR's own verification notes ("no two-tap regression") but hasn't confirmed interactively. The three bound fields (`server`, `username`, `password`) would be the affected ones.
**Suggestion:** Attach the dismiss gesture to the backdrop/background layer only instead of the whole root — e.g., put `.onTapGesture { focusedField = nil }` on `ConduitBackdrop()` (or add a `.contentShape(Rectangle())` background that sits behind the controls). This keeps field taps, toggles, and the button fully intact while still dismissing on empty-area taps. Verify on device/simulator with a first-tap focus check.

### [severity: minor] Cloudflare credential fields aren't focus-managed, so the keyboard still won't dismiss
**File:** Conduit/Views/LoginView.swift:126,130
**Issue:** The "Cloudflare Client ID" and "Cloudflare Client Secret" fields have no `.focused($focusedField, equals:)` binding, and `LoginField` only covers `.server`, `.username`, `.password`. When one of these fields is active, tapping outside sets `focusedField = nil`, which has no effect on the unbound field — the keyboard stays up and the stated goal ("tap outside dismisses the keyboard") is only partially met.
**Suggestion:** Extend `LoginField` with `cloudflareClientID` / `cloudflareClientSecret` cases and bind those fields via `.focused(...)`, so clearing focus dismisses the keyboard for every text input on the screen.

### [severity: nit] `onTapGesture` also fires over the static header/footer, not just "outside" fields
**File:** Conduit/Views/LoginView.swift:173
**Issue:** The gesture covers the entire root, including the icon/title area and the caption text. This is harmless functionally, but the intent described ("tapping outside the text fields") is broader than the actual hit area; scoping the gesture to the backdrop avoids surprising tap targets.
**Suggestion:** Scope the gesture to the backdrop element as suggested in the first finding.

## Verdict
REQUEST_CHANGES — the gesture should be attached to the background layer rather than the root container to eliminate the first-tap focus risk, and the Cloudflare fields should be focus-managed so keyboard dismissal works for all inputs.

<a href="https://opencode.ai/s/HuCEulOM">New%20session%20-%202026-08-13T21%3A26%3A36.694Z</a>
[opencode session](https://opencode.ai/s/HuCEulOM)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31745570043)

kaishi00:
## Summary
This PR adds a tap-to-dismiss-keyboard gesture on the login screen's background layer and extends `LoginField`/`.focused(...)` bindings to the Cloudflare credential fields so clearing focus dismisses the keyboard for every text input.

## Findings
### [severity: minor] `focusedField` can reference a Cloudflare field after it's removed from the hierarchy
**File:** Conduit/Views/LoginView.swift:138-142
**Issue:** The Cloudflare `TextField`/`SecureField` are conditionally shown behind `if cloudflareEnabled`. If the user toggles `cloudflareEnabled` off while a Cloudflare field is first responder, `focusedField` is left set to `.cloudflareClientID`/`.cloudflareClientSecret`. SwiftUI generally clears focus when a bound view is removed, so this is mostly harmless today, but the stale value is fragile (e.g., if the binding logic changes or a submit chain is added later).
**Suggestion:** Set `focusedField = nil` (or guard) when the Cloudflare toggle turns off, e.g. `.onChange(of: cloudflareEnabled) { _, enabled in if !enabled { focusedField = nil } }`, mirroring the existing pattern for `useFaceID`/`saveCredentials`.

### [severity: nit] Backdrop gesture only covers single-tap on the background; taps in the form card's caption/header areas behave identically to "outside" taps
**File:** Conduit/Views/LoginView.swift:38-44
**Issue:** The intent ("tapping outside a text field dismisses") is met, but taps on the header icon/title and the footer caption also fall through to the backdrop and clear focus, since non-interactive `Text`/`Image` views don't consume touches. Functionally harmless and arguably desirable, just slightly broader than the stated hit area.
**Suggestion:** No change required; if narrower semantics are ever wanted, attach the gesture to a dedicated background layer sized behind the form card. Current behavior is fine.

### [severity: nit] Accessibility of the dismiss action
**File:** Conduit/Views/LoginView.swift:38-44
**Issue:** The keyboard-dismiss tap is not discoverable to VoiceOver users. This is a pre-existing gap (return-key submit already dismisses), so it's non-blocking.
**Suggestion:** Optional — could expose a `.accessibilityAction(named: "Dismiss Keyboard") { focusedField = nil }` on the container for parity.

## Verdict
APPROVE

The original two concerns were addressed in the follow-up commit: the gesture now lives on `ConduitBackdrop()` behind the content rather than the root `ZStack`, so a text field's first-responder tap isn't stolen (no two-tap regression), and the Cloudflare fields are now focus-bound so `focusedField = nil` dismisses the keyboard for all inputs. No force unwraps, retain cycles, or concurrency issues in the change.

<a href="https://opencode.ai/s/7UaLRPoN">New%20session%20-%202026-08-13T21%3A33%3A57.789Z</a>
[opencode session](https://opencode.ai/s/7UaLRPoN)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31746126534)

PR #60, opened 2026-08-14
Render push-delivered approval cards after background arrival

kaishi00:
## Summary

- Approval cards now appear (and are answerable) when the approval was raised **while the app was backgrounded** and the user foregrounds via the notification — the case PR #54 could not cover, because the card was never observed or cached locally.
- Parses the structured `decision` content forwarded by the notifier plugin + relay (companion PR: kaishi00/hermes-conduit-notifier#5) and caches it via a new **non-destructive** `SessionPresentationCache.recordPendingDecision` upsert, so the existing PR #54 background-resume merge restores an answerable card.
- Adds a design doc covering the full root-cause analysis (verified against Hermes source), the payload contract, and the phased plan for clarify.

## Root cause

A pending decision can only reach the client through the **one-shot** `approval.request` gateway stream event. When the app is backgrounded, that event is missed, and there is no fallback: `session.resume` returns only the stored transcript (decisions are live blocking prompts, not persisted rows), the snapshot carries no decision content (`status: "waiting"` at best), no RPC fetches pending decisions, and `SessionPresentationCache` is empty because the card was never observed. Verified against `nousresearch/hermes-agent`: the gateway emits `*.request` exactly once, holds it in-memory only, has no replay-on-reconnect, and plugins cannot register RPCs — so the content must ride the existing notification pipeline.

## Changes

- `ConduitNotificationTarget` gains an optional `decision` (`PendingDecisionPayload`); nil for non-decision notifications, existing callers unchanged via an explicit defaulted init.
- `PushNotificationService` parses and bounds the payload, dropping approvals without an answerable session key.
- `SessionPresentationCache.recordPendingDecision`: upserts a single card without disturbing the cached transcript, dedupes by decision key, and stamps the bounded unconfirmed-expiry marker (same safeguard as PR #54).
- `AppState.openNotificationTarget` records the card under **both** the notification's runtime session id and the resolved stored id, before opening, so the resume merge finds it regardless of which identity the gateway resumes against.
- Card is answered through the existing `respondToApproval(sessionId:choice:)` path; the live foreground `approval.request` event remains authoritative and supersedes a cached card by decision key.

## Design notes (flagged for review)

- The push payload carries `description` + the always-valid `["once","deny"]` choices; the raw `command` is deliberately omitted upstream (privacy — the relay/APNs path is an extra egress transport). The foreground card still shows the full command and full choice set.
- **Clarify is intentionally deferred** (phases 2–3 in the design doc): `clarify.respond` is request-id-only and that id is minted inside the gateway, so a clarify card cannot be made answerable from a push payload without gateway changes.
- Backward-compatible: a payload without `decision` changes nothing.

## Validation

- Full iOS Simulator XCTest suite: **615 passed, 0 failed** (incl. 3 new `recordPendingDecision` tests and 3 new payload-parsing tests).
- Focused `SessionPresentationCacheTests` (36) + `MessageNormalizerTests` (34): all green.
- `git diff --check` clean.
- Companion plugin/relay tests: 8 plugin + 5 relay green (hermes-conduit-notifier#5).

**To go live:** deploy the relay change to `push.milim.dev`; self-hosters update the plugin + relay. The iOS side degrades gracefully until then.

## Summary by CodeRabbit

* **New Features**
  * Approval decisions received while the app is in the background now appear as answerable decision cards when the related session is opened or resumed.
  * Decision cards persist across session restoration, preserve conversation history, reconcile duplicates, and expire stale items automatically.
  * Added an “Approval cards in pushes” setting to control whether approval details appear in notifications.
  * The setting is enabled by default and remains compatible with existing notification preferences.

* **Documentation**
  * Added design documentation covering background decision delivery, notification handling, caching, privacy safeguards, and rollout considerations.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -4161,17 +4161,21 @@ final class AppState: ObservableObject {
     /// it. The card is recorded as pending and answerable through the existing
     /// `respondToApproval` path; the bounded unconfirmed marker is stamped by
     /// `SessionPresentationCache` so a stale card expires rather than lingering.
+    /// The decision's session key must match one of the routed session
+    /// identities — a mismatched key could not be answered via
+    /// `approval.respond` and would only duplicate or contradict the live card.
     private func recordNotificationDecision(
         _ decision: PendingDecisionPayload,
         sessionIDs: [String]
     ) {
         switch decision {
         case let .approval(sessionKey, description, choices):
+            guard sessionIDs.contains(sessionKey) else { return }
             let activity = ApprovalActivity(
                 sessionId: sessionKey,
                 command: "",
                 description: description,
-                choices: choices.isEmpty ? nil : choices,
+                choices: choices,
                 allowPermanent: false,
                 smartDenied: false,
                 status: .pending,
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -5275,8 +5275,12 @@ final class AppState: ObservableObject {
             guard let updatedIndex = messages.firstIndex(where: { $0.clarify?.requestId == requestId }) else { return }
             messages[updatedIndex].clarify?.status = .error
             messages[updatedIndex].clarify?.answer = nil
-            messages[updatedIndex].clarify?.error = "Hermes did not accept that answer."
-            errorMessage = error.localizedDescription
+            if Self.isExpiredPromptError(error) {
+                messages[updatedIndex].clarify?.error = "This question is no longer active — Hermes timed it out and continued."
+            } else {
+                messages[updatedIndex].clarify?.error = "Hermes did not accept that answer."
+                errorMessage = error.localizedDescription
+            }
             cacheMessagePresentation()
         }
     }
@@ -5781,6 +5785,20 @@ final class AppState: ObservableObject {
         }
     }
 
+    /// Hermes blocks a clarify/approval prompt for only ~5 minutes server-side
+    /// (JSON-RPC error 4009, "no pending … request"), while a restored or
+    /// push-delivered card can legitimately outlive it — a notification opened
+    /// an hour later is the feature's ordinary case, not an error. Treat that
+    /// outcome as "the decision is no longer active" instead of a generic
+    /// failure the user can retry forever.
+    static func isExpiredPromptError(_ error: Error) -> Bool {
+        if let rpcError = error as? RpcError {
+            if rpcError.code == 4009 { return true }
+            return rpcError.message.lowercased().contains("no pending")
+        }
+        return error.localizedDescription.lowercased().contains("no pending")
+    }
+
     func respondToApproval(messageId: String, choice: String) async {
         guard let index = messages.firstIndex(where: { $0.id == messageId }),
               let current = messages[index].approval,
@@ -5808,8 +5826,12 @@ final class AppState: ObservableObject {
             guard let updatedIndex = messages.firstIndex(where: { $0.id == messageId }) else { return }
             messages[updatedIndex].approval?.status = .error
             messages[updatedIndex].approval?.choice = nil
-            messages[updatedIndex].approval?.error = "Hermes did not accept that decision."
-            errorMessage = error.localizedDescription
+            if Self.isExpiredPromptError(error) {
+                messages[updatedIndex].approval?.error = "This approval is no longer active — Hermes timed it out and continued."
+            } else {
+                messages[updatedIndex].approval?.error = "Hermes did not accept that decision."
+                errorMessage = error.localizedDescription
+            }
             cacheMessagePresentation()
         }
     }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -4138,6 +4138,14 @@ final class AppState: ObservableObject {
             for: requestedID,
             in: sessions + cronSessions
         )
+        // A decision raised while the app was backgrounded is delivered as a
+        // structured payload on the notification (the one-shot gateway stream
+        // event was missed). Cache it under both the runtime and resolved
+        // stored IDs so the upcoming resume's `merge` restores the card
+        // regardless of which identity the gateway resumes against.
+        if let decision = target.decision {
+            recordNotificationDecision(decision, sessionIDs: [resumableID, requestedID])
+        }
         let opened = await openSession(
             resumableID,
             reusing: transitionGeneration
@@ -4147,6 +4155,42 @@ final class AppState: ObservableObject {
             transitionGeneration: transitionGeneration
         ) else { return false }
         return opened
+    }
+
+    /// Caches a push-delivered decision card so the resume merge can restore
+    /// it. The card is recorded as pending and answerable through the existing
+    /// `respondToApproval` path; the bounded unconfirmed marker is stamped by
+    /// `SessionPresentationCache` so a stale card expires rather than lingering.
+    private func recordNotificationDecision(
+        _ decision: PendingDecisionPayload,
+        sessionIDs: [String]
+    ) {
+        switch decision {
+        case let .approval(sessionKey, description, choices):
+            let activity = ApprovalActivity(
+                sessionId: sessionKey,
+                command: "",
+                description: description,
+                choices: choices.isEmpty ? nil : choices,
+                allowPermanent: false,
+                smartDenied: false,
+                status: .pending,
+                choice: nil,
+                error: nil
+            )
+            let message = ChatMessage(
+                id: "approval-\(sessionKey)",
+                role: .approval,
+                content: description,
+                timestamp: Self.localTimestamp(),
+                approval: activity
+            )
+            sessionPresentationCache.recordPendingDecision(
+                message,
+                profile: activeProfile,
+                sessionIDs: sessionIDs
+            )
+        }
     }
 
     private func notificationOpenAttemptIsCurrent(
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -704,10 +704,19 @@ final class AppState: ObservableObject {
         let pendingDecisionKeys = SessionPresentationCache.pendingDecisionKeys(in: cacheableMessages)
         // Gateway-provided cards do not create a restoration guard, so keep
         // their bounded expiry marker across ordinary cache flushes as well.
+        // Push-recorded cards (recordPendingDecision) live only in the store —
+        // the in-memory transcript has never seen them — so union in the
+        // store's pending keys or the flush would drop the card before the
+        // notification-open resume merge could restore it.
+        let storedPendingDecisionKeys = sessionPresentationCache.storedPendingDecisionKeys(
+            profile: activeProfile,
+            sessionIDs: ids
+        )
         let pendingDecisionKeysToPreserve = restorationKeys
-            ?? pendingDecisionKeys
+            ?? pendingDecisionKeys.union(storedPendingDecisionKeys)
         let preservePendingDecisionCards = restorationKeys == nil
             || !pendingDecisionKeys.isEmpty
+            || !storedPendingDecisionKeys.isEmpty
         sessionPresentationCache.save(
             cacheableMessages,
             profile: activeProfile,
@@ -4170,7 +4179,13 @@ final class AppState: ObservableObject {
     ) {
         switch decision {
         case let .approval(sessionKey, description, choices):
-            guard sessionIDs.contains(sessionKey) else { return }
+            // Compare trimmed on both sides: the payload parser trims the
+            // session key, but the routed ids arrive as the notification
+            // delivered them.
+            let knownKeys = sessionIDs.map {
+                $0.trimmingCharacters(in: .whitespacesAndNewlines)
+            }
+            guard knownKeys.contains(sessionKey) else { return }
             let activity = ApprovalActivity(
                 sessionId: sessionKey,
                 command: "",
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -6,7 +6,33 @@ struct ConduitNotificationTarget: Equatable, Identifiable {
     let profile: String?
     let sessionId: String
     let type: String?
+    /// Structured decision content carried alongside a decision notification.
+    /// Lets Conduit render an answerable card from the push payload alone when
+    /// the one-shot gateway stream event was missed while the app was
+    /// backgrounded. Nil for non-decision notifications.
+    let decision: PendingDecisionPayload?
     var id: String { "\(profile ?? "default"):\(sessionId):\(type ?? "")" }
+
+    init(
+        profile: String?,
+        sessionId: String,
+        type: String?,
+        decision: PendingDecisionPayload? = nil
+    ) {
+        self.profile = profile
+        self.sessionId = sessionId
+        self.type = type
+        self.decision = decision
+    }
+}
+
+/// The structured card content for a decision notification. Approval is
+/// session-keyed (`approval.respond { choice, session_id }`), so a payload
+/// carrying the session key is fully answerable. Clarify is keyed by a
+/// gateway-generated request id that no plugin hook exposes, so it is not
+/// representable here yet (see the background-arrival design doc).
+enum PendingDecisionPayload: Equatable {
+    case approval(sessionKey: String, description: String, choices: [String])
 }
 
 enum NotificationSessionResolver {
@@ -237,8 +263,37 @@ final class PushNotificationService: ObservableObject {
         return ConduitNotificationTarget(
             profile: profile?.isEmpty == false ? profile : nil,
             sessionId: sessionId,
-            type: type?.isEmpty == false ? type : nil
+            type: type?.isEmpty == false ? type : nil,
+            decision: pendingDecision(from: payload)
         )
+    }
+
+    /// Parses the structured decision content the relay forwards alongside a
+    /// decision notification (see the background-arrival design doc). Returns
+    /// nil for non-decision notifications or malformed/unknown payloads so the
+    /// notification degrades to its ordinary routing target.
+    private func pendingDecision(from payload: [String: Any]) -> PendingDecisionPayload? {
+        guard let decision = payload["decision"] as? [String: Any],
+              let kind = (decision["kind"] as? String)?.trimmingCharacters(in: .whitespacesAndNewlines).lowercased(),
+              !kind.isEmpty else {
+            return nil
+        }
+        switch kind {
+        case "approval":
+            guard let sessionKey = (decision["session_key"] as? String)?
+                .trimmingCharacters(in: .whitespacesAndNewlines),
+                !sessionKey.isEmpty else {
+                return nil
+            }
+            let description = (decision["description"] as? String)?
+                .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
+            let choices = ((decision["choices"] as? [Any])?
+                .compactMap { ($0 as? String)?.trimmingCharacters(in: .whitespacesAndNewlines) }
+                .filter { !$0.isEmpty }) ?? []
+            return .approval(sessionKey: sessionKey, description: description, choices: choices)
+        default:
+            return nil
+        }
     }
 
     private func requestDeviceToken() async throws -> String {
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -64,6 +64,11 @@ struct ConduitNotificationPreferences: Codable, Equatable {
     var backgroundTaskFinished = true
     var completionSound = true
     var showPreviews = false
+    /// Independent of `showPreviews`: controls whether pushes carry structured
+    /// decision content (answerable approval cards). Defaults on because the
+    /// feature's audience is exactly the approval-gate crowd; privacy-focused
+    /// users can turn just this off.
+    var decisionCards = true
 
     enum CodingKeys: String, CodingKey {
         case enabled
@@ -74,6 +79,7 @@ struct ConduitNotificationPreferences: Codable, Equatable {
         case backgroundTaskFinished = "background_task_finished"
         case completionSound = "completion_sound"
         case showPreviews = "show_previews"
+        case decisionCards = "decision_cards"
     }
 }
 
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -271,7 +271,10 @@ final class PushNotificationService: ObservableObject {
     /// Parses the structured decision content the relay forwards alongside a
     /// decision notification (see the background-arrival design doc). Returns
     /// nil for non-decision notifications or malformed/unknown payloads so the
-    /// notification degrades to its ordinary routing target.
+    /// notification degrades to its ordinary routing target. An approval
+    /// requires a session key to answer, a description to display, and at
+    /// least one usable choice — otherwise a cached card could render the
+    /// approval view's default action set, which the payload never promised.
     private func pendingDecision(from payload: [String: Any]) -> PendingDecisionPayload? {
         guard let decision = payload["decision"] as? [String: Any],
               let kind = (decision["kind"] as? String)?.trimmingCharacters(in: .whitespacesAndNewlines).lowercased(),
@@ -282,14 +285,17 @@ final class PushNotificationService: ObservableObject {
         case "approval":
             guard let sessionKey = (decision["session_key"] as? String)?
                 .trimmingCharacters(in: .whitespacesAndNewlines),
-                !sessionKey.isEmpty else {
+                !sessionKey.isEmpty,
+                let description = (decision["description"] as? String)?
+                    .trimmingCharacters(in: .whitespacesAndNewlines),
+                !description.isEmpty,
+                let rawChoices = decision["choices"] as? [Any] else {
                 return nil
             }
-            let description = (decision["description"] as? String)?
-                .trimmingCharacters(in: .whitespacesAndNewlines) ?? ""
-            let choices = ((decision["choices"] as? [Any])?
+            let choices = rawChoices
                 .compactMap { ($0 as? String)?.trimmingCharacters(in: .whitespacesAndNewlines) }
-                .filter { !$0.isEmpty }) ?? []
+                .filter { !$0.isEmpty }
+            guard !choices.isEmpty else { return nil }
             return .approval(sessionKey: sessionKey, description: description, choices: choices)
         default:
             return nil
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -83,6 +83,27 @@ struct ConduitNotificationPreferences: Codable, Equatable {
     }
 }
 
+extension ConduitNotificationPreferences {
+    /// Decoding must tolerate registrations persisted by older builds, which
+    /// predate later-added keys — a `keyNotFound` failure would make the
+    /// `try?` in the init drop the whole stored registration and silently
+    /// disable push for an upgrading user. Every key falls back to its
+    /// default when absent. (Declared in an extension so the synthesized
+    /// `init()` is preserved.)
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        enabled = try container.decodeIfPresent(Bool.self, forKey: .enabled) ?? true
+        approvalNeeded = try container.decodeIfPresent(Bool.self, forKey: .approvalNeeded) ?? true
+        inputNeeded = try container.decodeIfPresent(Bool.self, forKey: .inputNeeded) ?? true
+        responseReady = try container.decodeIfPresent(Bool.self, forKey: .responseReady) ?? true
+        turnFailed = try container.decodeIfPresent(Bool.self, forKey: .turnFailed) ?? true
+        backgroundTaskFinished = try container.decodeIfPresent(Bool.self, forKey: .backgroundTaskFinished) ?? true
+        completionSound = try container.decodeIfPresent(Bool.self, forKey: .completionSound) ?? true
+        showPreviews = try container.decodeIfPresent(Bool.self, forKey: .showPreviews) ?? false
+        decisionCards = try container.decodeIfPresent(Bool.self, forKey: .decisionCards) ?? true
+    }
+}
+
 @MainActor
 final class PushNotificationService: ObservableObject {
     static let shared = PushNotificationService()
--- a/Conduit/Services/SessionPresentationCache.swift
+++ b/Conduit/Services/SessionPresentationCache.swift
@@ -274,6 +274,51 @@ final class SessionPresentationCache {
         return merged
     }
 
+    /// Records a pending decision card observed outside the live stream —
+    /// today, from a push notification's structured payload, which is the only
+    /// source for a decision raised while the app was backgrounded and missed
+    /// the one-shot gateway event. Upserts the single card into whatever is
+    /// already cached for each session (the transcript is otherwise the
+    /// gateway's source of truth), dedupes by decision key, and stamps the
+    /// bounded unconfirmed marker so a stale card expires rather than
+    /// lingering. `merge(includePendingApprovals:)` restores it on resume.
+    func recordPendingDecision(
+        _ message: ChatMessage,
+        profile: String,
+        sessionIDs: [String]
+    ) {
+        guard Self.pendingDecisionKey(for: message) != nil,
+              let targetKey = Self.decisionKey(for: message) else { return }
+        let ids = Set(sessionIDs.map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }.filter { !$0.isEmpty })
+        guard !ids.isEmpty else { return }
+
+        var store = load()
+        let record = CachedMessage(message)
+        let stampedAt = now()
+        var changed = false
+        for id in ids {
+            let cacheKey = key(profile: profile, sessionID: id)
+            var session = store[cacheKey] ?? CachedSession(
+                updatedAt: stampedAt,
+                messages: [],
+                unconfirmedPendingDecisionAt: nil
+            )
+            // Replace any prior card for this exact decision, then append the
+            // fresh one. Other cached rows (the transcript) are untouched.
+            session.messages.removeAll { existing in
+                decisionKey(for: existing) == targetKey
+            }
+            session.messages.append(record)
+            session.unconfirmedPendingDecisionAt = session.unconfirmedPendingDecisionAt ?? stampedAt
+            session.updatedAt = stampedAt
+            store[cacheKey] = session
+            changed = true
+        }
+        guard changed else { return }
+        trim(&store)
+        persist(store)
+    }
+
     func save(
         _ messages: [ChatMessage],
         profile: String,
--- a/Conduit/Services/SessionPresentationCache.swift
+++ b/Conduit/Services/SessionPresentationCache.swift
@@ -274,6 +274,24 @@ final class SessionPresentationCache {
         return merged
     }
 
+    /// Pending decision keys currently held in the store for the given
+    /// sessions, regardless of what the live in-memory transcript contains.
+    /// A presentation-cache flush rebuilds its records from the in-memory
+    /// transcript, so without this a flush for a session whose store holds a
+    /// push-recorded card (`recordPendingDecision`) would silently drop that
+    /// card — most notably the open-from-notification path, which records and
+    /// then immediately flushes when the notified session is already active.
+    func storedPendingDecisionKeys(
+        profile: String,
+        sessionIDs: [String]
+    ) -> Set<String> {
+        let stored = load()
+        let keys = sessionIDs
+            .compactMap { stored[key(profile: profile, sessionID: $0)] }
+            .flatMap { session in session.messages.compactMap(pendingDecisionKey(for:)) }
+        return Set(keys)
+    }
+
     /// Records a pending decision card observed outside the live stream —
     /// today, from a push notification's structured payload, which is the only
     /// source for a decision raised while the app was backgrounded and missed
--- a/Conduit/Services/SessionPresentationCache.swift
+++ b/Conduit/Services/SessionPresentationCache.swift
@@ -309,7 +309,14 @@ final class SessionPresentationCache {
                 decisionKey(for: existing) == targetKey
             }
             session.messages.append(record)
-            session.unconfirmedPendingDecisionAt = session.unconfirmedPendingDecisionAt ?? stampedAt
+            // Restart the bounded unconfirmed window from this observation.
+            // The marker is per session, and expiry strips every pending card
+            // in the session at once, so preserving an old marker would let a
+            // near-24h stamp immediately expire a decision that just arrived.
+            // (The card is also written under each session identity; an
+            // answered card can linger under an identity the answer flow did
+            // not update, but this same bounded window retires that copy.)
+            session.unconfirmedPendingDecisionAt = stampedAt
             session.updatedAt = stampedAt
             store[cacheKey] = session
             changed = true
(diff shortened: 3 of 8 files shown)


kaishi00:
## Summary

Adds a `decision` payload to the APNs `conduit` notification contract, parses it in `PushNotificationService`, and caches an answerable approval card via a new non-destructive `SessionPresentationCache.recordPendingDecision` upsert so the PR #54 resume merge restores it after a background-gap approval. Includes a root-cause design doc and tests. The core flow is sound and well-reasoned; findings below are all minor/nit-level.

## Findings

### [minor] Card is cached at notification-open time, not push-receipt time
**File:** Conduit/Services/AppState.swift:4146
**Issue:** `recordNotificationDecision` runs inside `openNotificationTarget`, which is only reached when the user *taps* the notification (`ConduitApp.swift:73-80`). The design doc's central guarantee — "the card is captured on the device the moment the push lands, so it is visible even if the user does not immediately open the notification" (design doc line 116-120, acceptance tests lines 302-305) — is not implemented. If the user dismisses the banner and foregrounds manually, `pendingTarget` is never set, nothing is cached, and the resume merge finds no card.
**Suggestion:** Either record the decision in `PushNotificationService.receiveNotificationPayload` (already MainActor) so the card is cached on receipt, or update the design doc to explicitly scope the behavior to "user foregrounds via the notification" (the PR's stated scope). The current doc oversells the behavior.

### [minor] A fresh card inherits the session's oldest unconfirmed marker and can expire immediately
**File:** Conduit/Services/SessionPresentationCache.swift:312
**Issue:** `session.unconfirmedPendingDecisionAt = session.unconfirmedPendingDecisionAt ?? stampedAt` preserves the *first* unconfirmed stamp. If a session already holds a near-24h-old unconfirmed card and a new push-delivered approval is recorded for that session, the fresh card shares the old marker. At the next resume, `isUnconfirmedPendingDecisionExpired` strips *all* pending cards in the session (line 168-171), silently dropping a brand-new decision only an hour old.
**Suggestion:** Stamp the marker per decision card (store the stamp on `CachedMessage`) or refresh it to the newest pending card's `stampedAt`, so expiry is computed per card rather than per session. Note this strips all pending cards at once regardless, so consider filtering expired cards individually.

### [minor] Stale pending card lingers under the alternate session ID after it is answered
**File:** Conduit/Services/SessionPresentationCache.swift:299-311
**Issue:** `recordPendingDecision` writes the same `.pending` card under both `[resumableID, requestedID]`. When the user answers, `respondToApproval` → `cacheMessagePresentation` only updates the resumed ID's entry; the alternate identity's entry still holds the `.pending` card (with its marker). A later resume that resolves against the alternate identity can re-surface an answered decision as pending and answerable, which would then fail against `approval.respond`.
**Suggestion:** When a decision is answered (or superseded by a live event), supersede/clear the matching decision key across all cached identities for that session, or document that the alternate copy is expiry-bounded.

### [minor] Empty `choices` degrades to a wider answer surface than the payload guarantees
**File:** Conduit/Services/PushNotificationService.swift:290-292, Conduit/Services/AppState.swift:4174
**Issue:** If `choices` is absent or all-empty, `recordNotificationDecision` stores `choices: nil`. In `ChatView.allows` (ChatView.swift:2009-2013) a nil `choices` with `smartDenied == false` yields the default `["once","session","always","deny"]`, so an "Allow session" button appears even though the payload only promised `["once","deny"]`. Sending `"session"` to the gateway would be rejected (non-permanent approval), surfacing a "TRY AGAIN" card.
**Suggestion:** Default the fallback to `["once","deny"]` (matching the documented payload) rather than nil, or forward `smartDenied`/`allow_permanent` in the payload.

### [nit] Design doc payload contract no longer matches the shipped implementation
**File:** docs/superpowers/specs/2026-08-13-decision-cards-after-background-arrival-design.md:224-234
**Issue:** The doc's approval payload still documents a redacted `command`, `allow_permanent`, and the full choice set, but the implementation deliberately omits `command` and hardcodes `allowPermanent: false` with `["once","deny"]` choices (a privacy-motivated deviation noted in the PR body). The doc also describes caching via `SessionPresentationCache.save` on receipt (lines 251-255), which differs from `recordPendingDecision` at open time.
**Suggestion:** Update the doc's payload contract and "iOS changes" section to match the shipped contract so the spec stays a reliable reference.

### [nit] No validation that `decision.session_key` matches the routed session id
**File:** Conduit/Services/PushNotificationService.swift:283-287
**Issue:** The card is cached under `[resumableID, requestedID]` but its `ApprovalActivity.sessionId` is the payload `session_key`. The supersede/dedupe/answer flow (AppState.swift:6875-6891, SessionPresentationCache merge by `approval.sessionId`) all depend on `session_key ==` the gateway approval session id. A stale or mismatched `session_key` from the plugin silently yields a card that either duplicates the live one or fails on `approval.respond`.
**Suggestion:** Add a defensive check that `session_key` is non-empty and matches (or is contained within) the routed session's known IDs before recording, or at least document the invariant.

### [nit] Synchronous full-store encode on the main actor
**File:** Conduit/Services/SessionPresentationCache.swift:295-320
**Issue:** `recordPendingDecision` loads, mutates, trims, and `persist`s the entire store (up to 32 sessions × 320 messages) synchronously on the main actor during the notification-open path, without the coalescing used for streaming flushes.
**Suggestion:** Acceptable given small payloads, but consider routing through the same debounced flush path or noting the cost if multiple decision notifications arrive in sequence.

## Verdict

COMMENT

The implementation is careful and correct for its stated scope (tap-to-foreground), with clean upsert semantics, decision-key dedupe, and a well-documented root-cause analysis. The findings are non-blocking; the most valuable fixes are recording the card at receipt time to match the design doc's guarantee, and per-card expiry so a fresh decision is never truncated by an older session marker.

<a href="https://opencode.ai/s/RbcxfEq6">New%20session%20-%202026-08-14T08%3A44%3A23.389Z</a>
[opencode session](https://opencode.ai/s/RbcxfEq6)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31785065203)

kaishi00:
Addressed in 7862074 — full suite 618 passed, 0 failed (73 focused). Both bots' overlapping findings on the parser were fixed as suggested.

**[minor] Card cached at open time, not push-receipt time** — Took the review's second option: the design doc now states the scope explicitly. Alert-style pushes cannot wake a suspended app, so the payload only reaches the app on tap or cold-launch-from-notification — `receiveNotificationPayload` itself only runs in those cases, so moving the recording there wouldn't extend coverage. The doc's "Scope of receipt" section explains this and lists silent (`content-available`) pushes / Phase 3 gateway work as the path to banner-time capture.

**[minor] Fresh card inherits the session's oldest unconfirmed marker** — Fixed. `recordPendingDecision` now restarts the bounded window (`unconfirmedPendingDecisionAt = stampedAt`) on each observation, so a decision recorded at hour 25 is never expired by an hour-0 stamp. Regression test uses an injected clock: record at t0, re-record at t0+25h, merge at t0+26h → card restored. Per-card stamps noted as a follow-up refinement in the doc.

**[minor] Empty `choices` → wider default answer surface** — Fixed at the parser (same as CodeRabbit's suggestion): an approval decision requires a non-empty `description` **and** at least one usable choice, else it degrades to the plain routing target. `recordNotificationDecision` now passes choices straight through (never `nil`), so the push card can only ever offer `once`/`deny`. Tests: absent choices, all-invalid choices, whitespace-only description.

**[nit] Doc contract vs implementation** — Fixed. The payload contract now matches the shipped shape (approval-only, no `command`, no `allow_permanent`, `once`/`deny` choices, `show_previews`-gated at the relay, APNs size-capped), the component table and iOS section describe `recordPendingDecision` at notification-open, and the acceptance criteria are approval-only.

**[nit] `session_key` not validated against the routed session** — Fixed: `recordNotificationDecision` now guards that the decision's session key is one of the routed identities (`requestedID`/`resumableID`) before recording.

**[minor] Stale pending card under the alternate session id** — Documented as a known bound: the alternate-identity copy is retired by the same 24h unconfirmed window (the code comment and doc both call this out); cross-identity supersede on answer is listed as a follow-up.

**[nit] Synchronous store encode on the main actor** — Left as-is: it matches the existing `save` path's behavior and payload size (bounded at ≤32 sessions), noted in the doc's follow-ups.

kaishi00:
## Summary

This PR delivers answerable approval cards after a background-arrival gap by carrying a structured `decision` payload through the existing APNs notification pipeline, parsing it in `PushNotificationService`, and caching the card via a non-destructive `SessionPresentationCache.recordPendingDecision` upsert that the PR #54 resume merge then restores. Includes a well-researched design doc and solid unit coverage of the parser and cache.

## Findings

### [major] Recorded card is wiped by the pre-resume flush when the notified session is already active
**File:** Conduit/Services/AppState.swift:4146
**Issue:** `openNotificationTarget` records the decision at line 4146 and then calls `openSession` (line 4149). Inside `openSession`, `flushPendingPresentationCache()` → `cacheMessagePresentation()` → `save(...)` runs at AppState.swift:4057 **before** the resume, and it rebuilds the cache entry for `[activeSessionId, reconciliation?.requestedSessionId, ...]` from the in-memory `messages` array. When the tapped notification targets the session that is already active (the most common "warm resume from a backgrounded session" case), `activeSessionId == requestedID == resumableID` (for ordinary sessions `storedSessionId` is nil, so all three collapse to one key). That `save` call does not contain the just-recorded card in `messages`, `pendingDecisionKeys(in: messages)` is empty, and `appendUnconfirmedPendingDecisionRecords` preserves nothing — so the store entry is overwritten without the card and its `unconfirmedPendingDecisionAt` marker is reset to nil. The subsequent `merge` finds nothing and the card never appears — the exact failure this PR sets out to fix. Only the cold-launch-from-notification and "active session differs from notified session" cases are safe.
**Suggestion:** Make the flush during a notification-open preserve store-only pending cards for the ids being written (e.g., pass the recorded decision key through `save`'s `unconfirmedPendingDecisionKeys`), or flush the old session state *before* `recordNotificationDecision`, and verify with an AppState-level integration test that taps a notification while the target session is active (none of the existing `AppStateChatResumeTests.openNotificationTarget` cases carry a `decision`, so this path is untested).

### [minor] Restored card looks answerable for 24h but the gateway rejects answers after ~5 min
**File:** Conduit/Services/SessionPresentationCache.swift:16
**Issue:** `maxUnconfirmedPendingDecisionAge` is 24h, but the gateway's in-memory `_block` prompt timeout is ~5 minutes. A user who taps the notification within the 24h window but after the timeout gets a fully rendered, tappable `once`/`deny` card that resolves to "Hermes did not accept that decision" via `respondToApproval` (AppState.swift:5796). The design doc acknowledges this, but the delivered UX still surfaces an error for the majority of the window.
**Suggestion:** Consider shortening the pending-card window for push-restored cards to match the server-side prompt lifetime, or on answer failure distinguish "expired" from "rejected" so the card degrades gracefully instead of presenting an error.

### [minor] `session_key`↔routed-identity invariant is silently assumed
**File:** Conduit/Services/AppState.swift:4173
**Issue:** `recordNotificationDecision` guards `sessionIDs.contains(sessionKey)`, so a card is dropped when the plugin's `session_key` is a durable identity the catalog did not resolve into `[resumableID, requestedID]`. The design doc asserts `session_key == session_id`, but if the plugin ever emits the stored id and the catalog fails to resolve, the feature silently degrades with no signal. The guard also compares raw values (no trimming) — fine in practice since the parser trims, but the invariant itself is untested at the AppState layer.
**Suggestion:** Add a defensive normalization (trim both sides) and a test for the guard path (matching and non-matching `session_key`), or log a debug-level note when a decision is dropped so a mismatched plugin contract is detectable.

### [nit] Alternate-identity pending copy after answer
**File:** Conduit/Services/SessionPresentationCache.swift:316
**Issue:** The card is written under `[resumableID, requestedID]`. After answering, `cacheMessagePresentation()` normally rewrites both identities with the resolved card, but if a later resume/answer flow only covers one identity (or the gateway resumes against a third id), the other copy stays `.pending` and can re-surface as answerable until the 24h window retires it. Documented as a known bound in the code comment and design doc.
**Suggestion:** On resolve, supersede the decision key across all cached identities for the session; per-card stamps (also noted as follow-up) would make this exact.

### [nit] Synchronous full-store encode on the main actor during notification open
**File:** Conduit/Services/SessionPresentationCache.swift:295
**Issue:** `recordPendingDecision` loads, mutates, trims, and JSON-encodes the entire store (≤32 sessions × 320 messages) synchronously on the MainActor, on top of the resume's own `save` — two full-store writes per decision open. Bounded and consistent with the existing `save` path, so acceptable, but the open path bypasses the debounced flush used for streaming.
**Suggestion:** Fine to leave; if multiple decision notifications are expected in sequence, route through the same coalescing path.

### [nit] Decision payload contract rides on a text-only `show_previews` gate
**File:** docs/superpowers/specs/2026-08-13-decision-cards-after-background-arrival-design.md:149
**Issue:** The structured `decision` (including the answerable `session_key`) is gated on `show_previews` at the relay, the same flag that controls `title`/`body` preview text. Users with previews disabled therefore never receive answerable cards, and there is no separate opt-in for the "actionable but body-hiding" case. Documented, but it couples a privacy display preference to functional answerability.
**Suggestion:** Consider a distinct preference (e.g., "actionable approval notifications") so answerability is not tied to body-preview visibility.

## Verdict

REQUEST_CHANGES

The design, parsing strictness, dedupe semantics, and expiry handling are careful and well-tested at the cache/parser level, and the major finding's mechanism is the only genuine functional defect — but it strikes the PR's primary "warm resume via notification" path and is currently invisible because no AppState-level test exercises the record→flush→merge ordering. Fixing the flush clobber (and adding the integration test) should be sufficient; the remaining items are minor/nit and mostly already documented as follow-ups.

<a href="https://opencode.ai/s/n36wUA2m">New%20session%20-%202026-08-14T09%3A13%3A55.055Z</a>
[opencode session](https://opencode.ai/s/n36wUA2m)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31786925762)

kaishi00:
Findings addressed in dfe3b99 — full suite 620 passed, 0 failed.

**[major] Flush clobber on the warm-resume path** — Excellent catch, confirmed and fixed. `cacheMessagePresentation` now unions the store's pending decision keys into the preserve set via a new `SessionPresentationCache.storedPendingDecisionKeys(profile:sessionIDs:)`, so `save`'s existing `appendUnconfirmedPendingDecisionRecords` path re-appends store-only cards and keeps their marker — the flush can no longer drop a push-recorded card it was never told about, on this path or any other. Added the requested AppState-level integration test (`testNotificationDecisionSurvivesOpenWhenNotifiedSessionIsActive`) with the notified session already active and a non-empty in-memory transcript; **A/B-verified it fails without the fix** (card wiped → `nil`) and passes with it. `makeHarness` gained an injectable cache so the test runs against an isolated store.

**[minor] 24h window vs ~5-min server prompt lifetime** — Deliberate tradeoff, documented in the design doc: shortening the client window to ~5 min would make the card vanish almost immediately (worse for the primary complaint), and the same window already governs PR #54's restored cards. Distinguishing "expired" from "rejected" in the answer-failure UX is a good follow-up, but it touches the shared `respondToApproval` error handling for all approval cards, so I'd rather do it as its own change than widen this PR.

**[minor] session_key invariant untested/unnormalized** — Guard now compares trimmed identities, and `testNotificationDecisionWithMismatchedSessionKeyIsNotRecorded` covers both paths at the AppState layer.

**[nit] alternate-identity copy after answer / sync main-actor encode** — Both remain documented as bounded follow-ups (unchanged).

**[nit] `show_previews` couples answerability to preview visibility** — Agreed this is a product decision rather than a code fix: it's the direct consequence of gating decision content on the previews preference (the round-1 privacy request), and the same observation came up on the companion notifier PR. Surfaced to the maintainer: if cards should work for previews-off users, the clean path is a dedicated installation preference (e.g. `decision_cards`) rather than reusing `show_previews` — happy to implement that as a follow-up across relay + iOS settings.

kaishi00:
## Summary
Delivers answerable approval cards after a background-arrival gap by carrying a structured `decision` payload through the APNs notification pipeline, parsing it strictly in `PushNotificationService`, and caching the card via a non-destructive `SessionPresentationCache.recordPendingDecision` upsert that PR #54's resume merge restores. The pre-resume flush clobber (previous review's major finding) is fixed via `storedPendingDecisionKeys` unioning store-only pending keys into the preserve set, with an AppState-level integration test.

## Findings

### [minor] Approval dedup/supersede key is per-session, not per-decision
**File:** Conduit/Services/SessionPresentationCache.swift:261
**Issue:** `decisionKey(for:)` for approvals is `"approval:\(sessionId)"`, so all decisions for a session collapse to one key. `merge` skips appending the cached push card when `merged.contains { $0.approval?.sessionId == approval.sessionId }`, and `applyChatResume` (AppState.swift:2760) filters retained cards against `gatewayDecisionKeys` from the resume transcript — regardless of decision status. If the gateway transcript ever contains a *resolved* approval row for the same session, a genuinely new push-delivered pending card is suppressed and never appears. This is currently safe only because the design doc asserts transcripts never persist decision rows.
**Suggestion:** Key approvals by a per-decision id (e.g., a `decision_id`/prompt id from the payload) rather than the session id, or explicitly filter the suppression to *pending* status, so a new decision can't be masked by an old resolved row.

### [minor] A stale card remains answerable and shows a confusing error for most of its window
**File:** Conduit/Services/AppState.swift:5807
**Issue:** The push-restored card is rendered answerable for the full 24h `maxUnconfirmedPendingDecisionAge`, but the gateway holds the prompt in memory ~5 minutes. A user tapping within 24h but after the timeout gets a fully rendered `once`/`deny` card that resolves to "Hermes did not accept that decision." Deliberate and documented, but it surfaces an error for the majority of the window.
**Suggestion:** When `respondToApproval` fails, distinguish a gateway "no such prompt" (expired) from a real rejection and degrade the card gracefully (e.g., mark expired/remove) instead of a generic error.

### [minor] Dropped decisions are silent — no diagnostic when the session-key guard fails
**File:** Conduit/Services/AppState.swift:4188
**Issue:** `recordNotificationDecision` returns silently when `session_key` isn't in the routed identities. If the plugin/relay ever emits a durable id the catalog doesn't resolve, the feature degrades with zero signal, making a contract drift undetectable in the field.
**Suggestion:** Emit a debug-level log (or debug `errorMessage`) when a decision is dropped for a key mismatch so a misbehaving plugin is discoverable.

### [nit] No client-side check that notification `type` matches the decision kind
**File:** Conduit/Services/PushNotificationService.swift:267
**Issue:** `pendingDecision(from:)` only checks `kind == "approval"`; `target.type` is never cross-checked. The relay enforces the kind↔type bound, but a mismatched payload (e.g., `type: "response.ready"` with a `decision` blob) would still record and render a card the notification never promised.
**Suggestion:** Guard `target.type == "approval.needed"` when a decision is present (defense in depth at the client trust boundary).

### [nit] Extra full-store decode on every presentation-cache flush
**File:** Conduit/Services/AppState.swift:711
**Issue:** `cacheMessagePresentation` now calls `storedPendingDecisionKeys` (a full `load()` decode) and then `save` (another `load()` + encode) on every flush, including the 2-second-debounced streaming path. That doubles the main-actor JSON I/O for the whole bounded store (≤32 sessions × 320 messages) even when no push cards exist.
**Suggestion:** Read the store once and pass the keys into `save` (or short-circuit `storedPendingDecisionKeys` when no decision content was ever recorded for the profile).

### [nit] Alternate-identity copy of an answered card lingers as `.pending`
**File:** Conduit/Services/SessionPresentationCache.swift:334
**Issue:** The card is written under `[resumableID, requestedID]`; after answering, `cacheMessagePresentation` rewrites only the resumed identity, so the other copy stays `.pending` until the bounded window retires it. A later resume resolving against the alternate identity can re-surface an already-answered decision as answerable, which then fails on `approval.respond`. Documented as a known bound.
**Suggestion:** On resolve, supersede the decision key across all cached identities for the session (per-card stamps, already listed as a follow-up, would make this exact).

## Verdict
COMMENT

The mechanism is coherent and the major defect (flush clobber) is correctly fixed with a regression test that A/B-fails without the fix. Parser strictness, dedupe, and expiry-restart semantics are well-tested. Remaining items are minor/nit-level, mostly already acknowledged in the design doc as follow-ups; the two most worthwhile actions are adding a diagnostic when a decision is silently dropped and ensuring a new decision can't be suppressed by an old resolved approval row for the same session.

<a href="https://opencode.ai/s/TDy2BBFa">New%20session%20-%202026-08-14T09%3A31%3A30.570Z</a>
[opencode session](https://opencode.ai/s/TDy2BBFa)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31788350606)

kaishi00:
Follow-up in 5727bb3: the `show_previews` coupling (flagged by both reviewers as a product decision) is resolved — decision content is now gated on a **dedicated `decision_cards` preference**, default **on**, independent of previews.

- `ConduitNotificationPreferences.decisionCards` (wire key `decision_cards`) with a Settings toggle: **Settings > Notifications > "Approval cards in pushes"** — *"Include approval details so cards work from notifications. Disable for maximum privacy."*
- Codable round-trip test asserts the snake_case wire key the relay normalizes.
- Design doc updated to describe the dedicated preference.

Rationale from the maintainer: the people running approval gates are exactly the cards' audience, and they're often the same people who keep previews off for privacy — so functional answerability shouldn't ride on a banner-text setting.

Companion relay change (kaishi00/hermes-conduit-notifier#5, `5ea1315`) adds the preference server-side with a full gating-matrix test: previews off + cards on → answerable card with generic banner text; cards off → content-free payload; legacy installations default on.

Full suite: 621 passed, 0 failed.

kaishi00:
## Summary
This PR makes approval cards raised while the app was backgrounded appear and be answerable by carrying structured `decision` content through the APNs notification pipeline, parsing it strictly in `PushNotificationService`, and caching the card via a non-destructive `SessionPresentationCache.recordPendingDecision` upsert that the PR #54 resume merge restores. It also adds a `decision_cards` notification preference and a root-cause design doc.

## Findings

### [major] New non-optional Codable field silently breaks decoding of persisted registrations on upgrade
**File:** Conduit/Services/PushNotificationService.swift:58-83, 125-129
**Issue:** `ConduitNotificationPreferences` is `Codable` with 9 non-optional fields and only a `CodingKeys` enum, so the synthesized `init(from:)` calls `decode(_:forKey:)` for every key. `StoredRegistration.preferences` is persisted to the keychain (line 360-363), and an existing registration saved by an older build will not contain the new `decision_cards` key. On upgrade, `try? JSONDecoder().decode(StoredRegistration.self, ...)` throws, the failure is swallowed, `registration` becomes `nil`, and the user's push registration is silently dropped — notifications stop working until the user re-enables, with no diagnostic. The PR body claims backward compatibility, but only the notification payload (not the keychain payload) is backward-compatible.
**Suggestion:** Add a custom `init(from:)` that decodes every field with `decodeIfPresent` + its default (or make the field optional / use a lenient `StoredRegistration` decoder). Add a regression test that decodes a payload missing `decision_cards` (and ideally missing all optional keys).

### [minor] Cached push card is answerable for 24h while the gateway rejects answers after ~5 min
**File:** Conduit/Services/SessionPresentationCache.swift:16, Conduit/Services/AppState.swift:5807-5811
**Issue:** `maxUnconfirmedPendingDecisionAge` is 24h, but the gateway holds the in-memory prompt ~5 minutes. A user who taps within the window but after the timeout gets a fully rendered `once`/`deny` card that resolves to the generic "Hermes did not accept that decision" error. The design doc acknowledges this, but it surfaces an error for the majority of the window.
**Suggestion:** On `respondToApproval` failure, distinguish a gateway "no such prompt" (expired) from a real rejection and degrade the card (mark expired / remove) rather than a generic error.

### [minor] Extra synchronous full-store decode on every presentation-cache flush
**File:** Conduit/Services/AppState.swift:711, Conduit/Services/SessionPresentationCache.swift:284-293
**Issue:** `cacheMessagePresentation` now calls `storedPendingDecisionKeys` (a full `load()` decode) on every flush — including the 2-second streaming debounce — on top of `save`'s own `load()` + encode, doubling main-actor JSON I/O for the whole bounded store (≤32 sessions × 320 messages) even when no push card was ever recorded.
**Suggestion:** Read the store once and pass the keys into `save`, or short-circuit `storedPendingDecisionKeys` when no decision was ever recorded for the profile.

### [minor] New pending card can be suppressed by any same-session approval row in the transcript
**File:** Conduit/Services/SessionPresentationCache.swift:31-39, 257-273; Conduit/Services/AppState.swift:2760-2763
**Issue:** `decisionKey(for:)` for approvals is `"approval:\(sessionId)"`, and `merge` skips appending the cached push card whenever the transcript already contains *any* approval with the same `sessionId` (`merged.contains { $0.approval?.sessionId == ... }`). Likewise `applyChatResume` filters retained cards against `gatewayDecisionKeys` regardless of status. If the gateway transcript ever contains a resolved approval row (or a generic approval tool row) for the session, a genuinely new pending card is silently never restored. This is safe today only via the documented "transcripts never persist decision rows" invariant.
**Suggestion:** Key approvals by a per-decision id (e.g., a payload-provided `decision_id`), or restrict the suppression to *pending* rows so a new decision can't be masked by an old resolved one.

### [minor] Client never cross-checks `type` against the decision kind
**File:** Conduit/Services/PushNotificationService.swift:261-275
**Issue:** `pendingDecision(from:)` only checks `kind == "approval"`; `target.type` is not validated. A mismatched payload (e.g., `type: "response.ready"` carrying a `decision` blob) would still record and render an answerable card the notification never promised. The relay enforces the bound, but the client is the trust boundary for rendering.
**Suggestion:** Guard `target.type == "approval.needed"` when a decision is present (defense in depth).

### [nit] Dropped decisions are silent — no diagnostic on session-key mismatch
**File:** Conduit/Services/AppState.swift:4188
**Issue:** `recordNotificationDecision` returns silently when the payload `session_key` isn't in the routed identities. If the plugin ever emits a durable id the catalog doesn't resolve, the feature degrades with zero signal, making contract drift undetectable in the field.
**Suggestion:** Log a debug-level note when a decision is dropped so a misbehaving plugin is discoverable.

### [nit] `recordPendingDecision` restarts the session's single marker and removes prior same-session approval presentation
**File:** Conduit/Services/SessionPresentationCache.swift:326-337
**Issue:** The unconfirmed marker is per session, and `recordPendingDecision` stamps it to `now()` on every observation — extending the expiry window for any other cached pending card in that session. Additionally, the dedup `removeAll { decisionKey == targetKey }` removes *all* cached approval presentation for the session (including a previously resolved card retained for UI), not just a stale pending one.
**Suggestion:** Stamp the marker per decision card (as the doc's follow-up notes) and dedupe against pending status only.

### [nit] Alternate-identity copy of an answered card lingers as `.pending`
**File:** Conduit/Services/SessionPresentationCache.swift:334-337
**Issue:** The card is written under `[resumableID, requestedID]`; after answering, `cacheMessagePresentation` rewrites only the identities in `ids`, so the other copy stays `.pending` until the bounded window retires it, and a later resume against the alternate identity can re-surface an already-answered decision (which then fails on `approval.respond`). Documented as a known bound.
**Suggestion:** On resolve, supersede the decision key across all cached identities for the session.

## Verdict
REQUEST_CHANGES

The core mechanism is well-reasoned and carefully tested: strict payload parsing, decision-key dedupe, expiry-window restart, and the pre-resume flush fix (`storedPendingDecisionKeys` union) are all correct, with the flush fix A/B-verified by an AppState integration test. The blocking item is the Codable keychain-regression: adding the non-optional `decision_cards` field breaks decoding of `StoredRegistration` from prior builds, silently disabling push notifications for existing users on upgrade. Everything else is minor/nit-level and mostly already documented as deliberate or follow-up work.

<a href="https://opencode.ai/s/24cPn3iD">New%20session%20-%202026-08-14T20%3A03%3A21.631Z</a>
[opencode session](https://opencode.ai/s/24cPn3iD)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31835997116)

kaishi00:
Latest findings addressed in f340323 — full suite 622 passed, 0 failed.

**[major] Codable regression on legacy Keychain registrations** — Real catch, fixed. `ConduitNotificationPreferences` now decodes via `decodeIfPresent` with per-key defaults (declared in an extension so the synthesized `init()` survives), so a registration persisted before `decision_cards` existed decodes cleanly instead of throwing `keyNotFound` → `try?` dropping the registration → silently disabling push on upgrade. Regression test `testNotificationPreferencesDecodeLegacyRegistrationWithoutDecisionCardsKey` decodes a legacy payload and asserts persisted values survive + `decisionCards` falls back to on.

**[CodeRabbit] stale `show_previews` gate naming in the design doc** — Fixed in both spots (component table + acceptance criteria); they now name `decision_cards`, matching the shipped relay gate.

**[minor] 24h window vs ~5-min prompt lifetime** — unchanged from the previous reply: deliberate, documented tradeoff; graceful "expired" degradation is queued as its own follow-up since it touches shared approval error handling.

kaishi00:
## Summary
This PR makes approval cards raised while the app was backgrounded appear and be answerable by carrying structured `decision` content through the APNs pipeline, parsing it strictly in `PushNotificationService`, and caching it via a non-destructive `SessionPresentationCache.recordPendingDecision` upsert that the PR #54 resume merge restores. It also adds a dedicated `decision_cards` preference (independent of `show_previews`), tolerant legacy Codable decoding, and a root-cause design doc.

## Findings

### [minor] A stale restoration-guard card can mask a newer push-delivered card for the same session
**File:** Conduit/Services/SessionPresentationCache.swift:261-263, Conduit/Services/AppState.swift:2744-2768
**Issue:** `merge` dedupes approvals by `sessionId`, and the approval-replacement in the per-message match is gated on `message.approval == nil`. If a first notification restores a card (which installs the `PendingDecisionRestorationGuard` with `retainedRestoredMessages`), and then a *second, different* approval for the same session arrives via push while the app is still alive and the guard is set, the retained old card (same `approval-\(sessionKey)` id) bestMatches the cached new card but keeps its own stale activity (because `message.approval != nil`), and the same-sessionId dedupe then suppresses the newer card entirely. The user sees the stale, already-resolved approval as pending/answerable, which fails on `approval.respond`; the genuinely current card is hidden.
**Suggestion:** When appending pending approvals, prefer the newest cached record for a decision key and drop older same-key duplicates, and/or retire the guard's retained card when a fresh card is recorded for the same key in `recordPendingDecision`.

### [minor] No client-side cross-check of notification `type` against the decision kind
**File:** Conduit/Services/PushNotificationService.swift:290-330
**Issue:** `pendingDecision(from:)` only inspects the `decision` blob; the target's `type` is never validated against it. The relay enforces the kind↔type bound, but a mismatched payload (e.g. `type: "response.ready"` carrying a `decision`) would still be parsed and rendered as an answerable card the notification never promised.
**Suggestion:** Guard `type == "approval.needed"` when a decision is present (defense in depth at the client trust boundary).

### [minor] Restored card stays answerable for 24h while the gateway prompt lives ~5 min
**File:** Conduit/Services/SessionPresentationCache.swift:16, Conduit/Services/AppState.swift:5807-5811
**Issue:** `maxUnconfirmedPendingDecisionAge` is 24h, but the gateway holds the prompt in memory ~5 minutes. A user tapping within the window but after the timeout gets a fully rendered `once`/`deny` card that resolves to the generic "Hermes did not accept that decision" error for most of the window. Deliberate and documented, but the UX still surfaces an error.
**Suggestion:** On `respondToApproval` failure, distinguish a "no such prompt" (expired) from a real rejection and degrade/remove the card instead of a generic error (acknowledged as a follow-up).

### [minor] Per-session decision key drops resolved presentation and can suppress a new pending card
**File:** Conduit/Services/SessionPresentationCache.swift:326-328, 261-263
**Issue:** `decisionKey(for:)` for approvals is `"approval:\(sessionId)"`, so `recordPendingDecision`'s `removeAll` discards *all* cached approval presentation for the session (including a previously resolved card retained for UI), and `merge` skips a new pending card whenever the transcript/retained messages contain any same-session approval row regardless of status. Safe today only via the documented "transcripts never persist decision rows" invariant.
**Suggestion:** Key approvals by a payload-provided decision id, or restrict suppression to *pending* rows so a new decision can't be masked by an old resolved one.

### [nit] Extra full-store decode on every presentation-cache flush
**File:** Conduit/Services/AppState.swift:711-714
**Issue:** `cacheMessagePresentation` now calls `storedPendingDecisionKeys` (a full `load()` decode) on every flush — including the 2-second debounced streaming path — on top of `save`'s own `load()` + encode, doubling main-actor JSON I/O for the whole bounded store even when no push card was ever recorded.
**Suggestion:** Read the store once and pass the keys into `save`, or short-circuit `storedPendingDecisionKeys` when no decision content was recorded for the profile.

## Verdict
COMMENT

The core mechanism is well-reasoned and carefully tested: strict payload parsing, decision-key dedupe, the pre-resume flush fix (`storedPendingDecisionKeys` union), and the legacy `decision_cards` Codable decoding are all correct, with the flush fix A/B-verified by an AppState integration test. All previously blocking issues (flush clobber, Codable keychain regression, parser strictness) are resolved. The remaining items are minor/nit-level edge cases and documented tradeoffs; the two most worth acting on are the stale-guard masking a newer same-session card and the client-side `type`↔kind cross-check.

<a href="https://opencode.ai/s/qTMAy12T">New%20session%20-%202026-08-14T20%3A13%3A53.528Z</a>
[opencode session](https://opencode.ai/s/qTMAy12T)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31836842637)

kaishi00:
Your 24h-window-vs-5-min-prompt finding prompted a triage pass: of the remaining minors, it was the only one users are *likely* to hit — a push opened after ~5 minutes is the feature's ordinary case, and until now it surfaced as a generic "Hermes did not accept that decision" error with a forever-retryable card.

Fixed in 5f499a3: both answer paths classify the gateway's expired-prompt outcome (JSON-RPC 4009 / "no pending …") and show "This decision is no longer active — Hermes timed it out and continued" without the error banner; genuine failures keep the existing treatment. This also neutralizes the stale alternate-identity copy finding — that card fails the same way when answered, and now degrades the same way.

Deliberately **not** fixing (assessed as unlikely to be hit, or subsumed): main-actor store encode during notification open (one-off, bounded ≤32×320), per-card expiry stamps (subsumed by graceful expiry), and the relay double-validation nit (deliberate defense-in-depth). These stay documented as follow-ups.

Full suite: 623 passed, 0 failed.

kaishi00:
## Summary

Carries a structured `decision` approval payload through the existing APNs notification pipeline, parses it strictly in `PushNotificationService`, and caches an answerable card via a non-destructive `SessionPresentationCache.recordPendingDecision` upsert so PR #54's resume merge restores it after an approval raised during a background gap. Adds a dedicated `decision_cards` preference, tolerant legacy Codable decoding, graceful handling of gateway prompt expiry, and a root-cause design doc.

## Findings

### [minor] Stale restoration-guard card can mask a newer push-delivered card for the same session
**File:** Conduit/Services/AppState.swift:2761-2763, Conduit/Services/SessionPresentationCache.swift:261-263
**Issue:** When the first approval times out server-side (~5 min) but the locally restored card is still within the 24h unconfirmed window, the guard's `retainedRestoredMessages` holds that stale card in memory. If a *new* approval is then raised for the same session, `recordNotificationDecision` correctly replaces the store copy (same `decisionKey`), but on the next resume the retained stale card and the fresh cached card share the id `approval-<sessionKey>` and the same decision key: `bestMatch` keeps the retained card's non-nil `approval`, and the same-sessionId dedupe at SessionPresentationCache.swift:261 skips appending the fresh card. The user sees the stale, already-expired approval as answerable while the genuinely current card never appears. This scenario is not covered by the expired-prompt fix and was not explicitly addressed.
**Suggestion:** When appending pending approvals, prefer the newest cached record for a decision key and drop older same-key duplicates, and/or retire the guard's retained card when `recordPendingDecision` records a fresh card for that key.

### [minor] `pendingDecision(from:)` never cross-checks the notification `type` against the decision kind
**File:** Conduit/Services/PushNotificationService.swift:305-330
**Issue:** The parser inspects only the `decision` blob; `target.type` is never validated. A mismatched payload (e.g., `type: "response.ready"` carrying a structurally valid approval decision) would still be parsed and rendered as an answerable card the notification never promised. The relay enforces the kind↔type bound, but the client is the rendering trust boundary.
**Suggestion:** Guard `target.type == "approval.needed"` when a decision is present (defense in depth).

### [minor] Per-session decision key conflates approvals: strips resolved presentation and suppresses new cards on any same-session row
**File:** Conduit/Services/SessionPresentationCache.swift:31-39, 326-328, 261-263
**Issue:** `decisionKey(for:)` for approvals is `"approval:\(sessionId)"`. `recordPendingDecision`'s `removeAll` discards *all* cached approval presentation for the session (including a previously resolved card retained for UI), and `merge` skips a new pending card whenever the transcript/retained messages contain any same-session approval regardless of status. Safe today only via the documented "transcripts never persist decision rows" invariant.
**Suggestion:** Key approvals by a payload-provided per-decision id, or restrict the suppression to *pending* rows so a new decision can't be masked by an old resolved one.

### [minor] Cached card is rendered answerable for 24h while the gateway prompt lives ~5 minutes
**File:** Conduit/Services/SessionPresentationCache.swift:16, Conduit/Services/AppState.swift:5829-5834
**Issue:** `maxUnconfirmedPendingDecisionAge` is 24h; the gateway holds the prompt in memory ~5 min. The 5f499a3 fix classifies the expired outcome gracefully, but the card still shows `status = .error` and `canRespond(.error)` keeps the Run/Reject buttons enabled, so a user tapping within the window but after the timeout can retry forever against a dead prompt. Deliberate and documented, but the UX still exposes a retryable-looking card for most of the window.
**Suggestion:** On expiry classification, degrade to a terminal state (e.g., mark expired and disable response controls) instead of a retryable `.error`.

### [minor] Every presentation-cache flush now does an extra full-store decode
**File:** Conduit/Services/AppState.swift:711-714
**Issue:** `cacheMessagePresentation` calls `storedPendingDecisionKeys` (a full `load()` decode) on every flush — including the 2-second debounced streaming path — on top of `save`'s own `load()` + encode, doubling main-actor JSON I/O for the whole bounded store even when no push card was ever recorded.
**Suggestion:** Read the store once and pass the keys into `save`, or short-circuit `storedPendingDecisionKeys` when no decision content was recorded for the profile.

### [nit] Dropped decisions are silent — no diagnostic on session-key mismatch
**File:** Conduit/Services/AppState.swift:4188
**Issue:** `recordNotificationDecision` returns silently when the payload `session_key` isn't in the routed identities. If the plugin/relay ever emits a durable id the catalog fails to resolve, the feature degrades with zero signal, making contract drift undetectable in the field.
**Suggestion:** Emit a debug-level log when a decision is dropped for a key mismatch.

### [nit] `isExpiredPromptError` falls back to a `localizedDescription` substring heuristic
**File:** Conduit/Services/AppState.swift:5794-5800
**Issue:** For non-`RpcError` failures (transport, `HermesError`, etc.), classification relies on `error.localizedDescription.lowercased().contains("no pending")`. A genuine non-expiry error whose text happens to include "no pending" would be mislabeled as an expired prompt and suppress the real error banner. Low risk but heuristic.
**Suggestion:** Restrict the message-substring match to `RpcError` (which is the gateway failure type) and let non-RPC errors keep the generic path.

## Verdict

COMMENT

The core mechanism is coherent and now well-tested: strict payload parsing, decision-key dedupe, expiry-window restart, the pre-resume flush fix (`storedPendingDecisionKeys` union, A/B-verified at the AppState level), tolerant legacy `decision_cards` decoding, and the expired-prompt classification are all correct. All previously blocking issues (flush clobber, Codable keychain regression, parser strictness) are resolved. The remaining items are minor/nit-level edge cases and documented tradeoffs; the two most worth acting on are the stale-guard masking of a newer same-session card and the client-side `type`↔kind cross-check.

<a href="https://opencode.ai/s/ihHkvxPd">New%20session%20-%202026-08-14T20%3A41%3A07.841Z</a>
[opencode session](https://opencode.ai/s/ihHkvxPd)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31838938047)

PR #62, opened 2026-08-14
Show plugin and relay compatibility in notification settings

kaishi00:
## Summary

- Settings > Notifications gains a **Compatibility** section so a stale notifier plugin or relay is visible instead of silently degrading decision cards to plain notifications.
- Backed by the relay's `GET /v1/meta` (device credential — the app already holds it): a relay row (version + decision-card support) and one row per paired Hermes profile (plugin version + approval/clarify card support), with copy-paste `hermes plugins update conduit_push` / `hermes gateway restart` commands when the notifier is outdated.
- Capability flags drive the checks, not version parsing. Graceful degradation on both ends: a relay without the endpoint (older or self-hosted) renders "compatibility unknown"; a gateway that has never reported renders "waiting for the first notification from this profile" rather than a false outdated state.

## Validation

- Decoding + capability-check unit test (current gateway, never-reported gateway, relay flags).
- Full iOS Simulator suite: **649 passed, 0 failed**.
- Companion plugin/relay PR: kaishi00/hermes-conduit-notifier `compatibility-version-reporting` (e2e covers the endpoint contract end to end).

**Rollout:** the section shows "unknown" until the relay is redeployed at 0.2.0 — which is exactly the signal it exists to give.

## Summary by CodeRabbit

* **New Features**
  * Added relay and gateway compatibility details to notification settings, including versions and decision-card support.
  * Added gateway capability indicators for approval and clarification actions.
  * Added guidance and available update or restart actions when a plugin is outdated.
  * Compatibility information now refreshes when notification tasks load or a custom relay URL is submitted.
  * Unsupported or unavailable compatibility information is handled gracefully without disrupting valid gateway details.

### DIFF
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -52,7 +52,6 @@ struct RelayMetaInfo: Decodable, Equatable {
         let name: String
         let pluginVersion: String?
         let pluginCapabilities: [String]
-        let lastEventAt: String?
 
         var supportsApprovalCards: Bool { pluginCapabilities.contains("approval-decisions") }
         var supportsClarifyCards: Bool { pluginCapabilities.contains("clarify-loop") }
@@ -62,7 +61,6 @@ struct RelayMetaInfo: Decodable, Equatable {
             case name
             case pluginVersion = "plugin_version"
             case pluginCapabilities = "plugin_capabilities"
-            case lastEventAt = "last_event_at"
         }
     }
 
@@ -71,6 +69,33 @@ struct RelayMetaInfo: Decodable, Equatable {
     let gateways: [Gateway]
 
     var supportsDecisionCards: Bool { capabilities.contains("decisions") }
+
+    /// One malformed gateway record must not hide the whole section: decode
+    /// gateway rows lossily so a single incompatible record degrades to just
+    /// its row while the relay and other gateways still render.
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        version = try container.decode(String.self, forKey: .version)
+        capabilities = try container.decode([String].self, forKey: .capabilities)
+        var gateways: [Gateway] = []
+        var array = try container.nestedUnkeyedContainer(forKey: .gateways)
+        while !array.isAtEnd {
+            if let gateway = try? array.decode(Gateway.self) {
+                gateways.append(gateway)
+            } else {
+                _ = try? array.decode(Empty.self)
+            }
+        }
+        self.gateways = gateways
+    }
+
+    private struct Empty: Decodable {}
+
+    private enum CodingKeys: String, CodingKey {
+        case version
+        case capabilities
+        case gateways
+    }
 }
 
 enum NotificationSessionResolver {
@@ -155,6 +180,7 @@ final class PushNotificationService: ObservableObject {
     @Published private(set) var navigationAttempt = 0
     @Published var preferences = ConduitNotificationPreferences()
     @Published private(set) var relayMeta: RelayMetaInfo?
+    @Published private(set) var isFetchingMeta = false
 
     private var relayURL: URL {
         if let saved = UserDefaults.standard.string(forKey: "conduit.relayURL"),
@@ -195,12 +221,18 @@ final class PushNotificationService: ObservableObject {
     }
 
     /// Loads relay + plugin compatibility state for Settings > Notifications.
-    /// Nil (also on 404 from older/self-hosted relays) renders as unknown.
+    /// Nil after a completed fetch (also on 404 from older/self-hosted relays)
+    /// renders as unknown; `isFetchingMeta` distinguishes that from an
+    /// in-flight request so the UI never diagnoses "predates version
+    /// reporting" while still loading.
     func refreshMeta() async {
         guard let registration else {
             relayMeta = nil
+            isFetchingMeta = false
             return
         }
+        isFetchingMeta = true
+        defer { isFetchingMeta = false }
         var request = URLRequest(url: relayURL.appending(path: "/v1/meta"))
         request.setValue("Bearer \(registration.credential)", forHTTPHeaderField: "Authorization")
         do {
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -41,6 +41,38 @@ enum PendingDecisionPayload: Equatable {
     static let relayRequestPrefix = "conduit-push-"
 }
 
+/// Relay + paired-plugin compatibility state (`GET /v1/meta`), rendered in
+/// Settings > Notifications so users can see when the notifier plugin or the
+/// relay needs an update for decision cards to work. Capability flags (not
+/// version parsing) drive the checks; a relay without the endpoint (older or
+/// self-hosted pre-0.2) surfaces as `nil` and renders an unknown state.
+struct RelayMetaInfo: Decodable, Equatable {
+    struct Gateway: Decodable, Equatable, Identifiable {
+        let id: String
+        let name: String
+        let pluginVersion: String?
+        let pluginCapabilities: [String]
+        let lastEventAt: String?
+
+        var supportsApprovalCards: Bool { pluginCapabilities.contains("approval-decisions") }
+        var supportsClarifyCards: Bool { pluginCapabilities.contains("clarify-loop") }
+
+        enum CodingKeys: String, CodingKey {
+            case id
+            case name
+            case pluginVersion = "plugin_version"
+            case pluginCapabilities = "plugin_capabilities"
+            case lastEventAt = "last_event_at"
+        }
+    }
+
+    let version: String
+    let capabilities: [String]
+    let gateways: [Gateway]
+
+    var supportsDecisionCards: Bool { capabilities.contains("decisions") }
+}
+
 enum NotificationSessionResolver {
     /// Hermes notifications identify a live runtime session, while
     /// `session.resume` is keyed by the durable stored session. Catalog rows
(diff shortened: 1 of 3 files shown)


kaishi00:
## Summary
Adds a "Compatibility" section to Settings > Notifications backed by a new `GET /v1/meta` fetch (`PushNotificationService.refreshMeta()`), surfacing relay/plugin versions and capability-driven decision-card support, with copy-paste update commands for stale notifiers.

## Findings

### [severity: major] Compatibility section never refreshes after enabling notifications (or changing relay URL) within the screen
**File:** Conduit/Views/AuxiliaryViews.swift:1451
**Issue:** `.task { await notifications.refresh(); await notifications.refreshMeta() }` runs once on appear. If the user enters this screen with notifications disabled, `refreshMeta()` hits the `guard let registration` and sets `relayMeta = nil`. Toggling "Enable notifications" in the same screen creates the registration, but `.task` never re-runs — so the freshly-shown Compatibility section renders the "predates version reporting" unknown state permanently until the user navigates away and back. The same staleness applies if the user edits the "Push relay" URL below and the section was already fetched from the old relay.
**Suggestion:** Refresh when the section actually becomes visible, e.g. attach `.task { await notifications.refreshMeta() }` to the Compatibility `ConduitSettingsSection` itself (the section appears only when `isEnabled` becomes true), and/or `.onChange(of: customRelayURL)` to refetch after the relay URL changes.

### [severity: minor] All-or-nothing decode collapses the whole section on one bad gateway record
**File:** Conduit/Services/PushNotificationService.swift:212
**Issue:** A single gateway whose `id`/`name`/`plugin_capabilities` are missing or malformed makes `JSONDecoder().decode(RelayMetaInfo.self, ...)` throw, and `catch` sets the entire `relayMeta` to nil — the valid relay row (and other gateway rows) are hidden, and the user sees the blanket "unknown" message. This is the opposite of the graceful-degradation intent: one partial record hides information that was fully available.
**Suggestion:** Decode `gateways` lossily (e.g. `decode([Gateway?].self)` filtering nils, or decode each record individually and drop failures) so a single incompatible gateway degrades to just that row while the relay/other gateways still render.

### [severity: minor] No loading state — in-flight fetch misdiagnosed as "relay predates version reporting"
**File:** Conduit/Views/AuxiliaryViews.swift:1390
**Issue:** `relayMeta` starts nil, so while the request is in flight (or if the relay is slow/unreachable for up to the URLSession timeout), the UI states categorically "Your relay predates version reporting; decision cards may not be available" — a diagnosis the code can't yet know. That's misleading right when the section is most useful (post-update rollout).
**Suggestion:** Track a loading flag (e.g. `@Published private(set) var isMetaLoading`), show "Checking compatibility…" while fetching, and only render the "predates version reporting" copy once a fetch has actually completed and returned non-200/no-data.

### [severity: minor] Accessibility: status icon relies on implicit SF Symbol labels
**File:** Conduit/Views/AuxiliaryViews.swift:1466
**Issue:** The `Image(systemName: ...)` in `compatibilityRow` has no `accessibilityLabel`; VoiceOver reads the raw SF Symbol name, and the status is additionally conveyed only by green/orange color. The trailing detail text does carry meaning, but the row's semantic status isn't announced as a unit.
**Suggestion:** Add `.accessibilityElement(children: .combine)` on the row and/or an explicit `.accessibilityLabel("\(isSupported ? "Supported" : "Update needed")")` on the icon.

### [severity: nit] `lastEventAt` decoded but never used
**File:** Conduit/Services/PushNotificationService.swift:55
**Issue:** The field is decoded into `Gateway` but nothing reads it. Either drop it or use it (e.g. render "last seen" time, which would also strengthen the "waiting for first notification" case).
**Suggestion:** Remove the property or surface it in the row detail.

### [severity: nit] Hardcoded version strings will drift
**File:** Conduit/Views/AuxiliaryViews.swift:1371, 1381
**Issue:** "0.2.0 or newer" is hardcoded in two user-visible strings; when the relay/plugin next bumps, these go stale. The code deliberately avoids version parsing, so the literal thresholds are inconsistent with that design.
**Suggestion:** Drive the copy from the capability check result without embedding a version number (e.g. "Decision cards need a relay update"), or centralize the constant.

## Verdict
COMMENT

The core design (capability flags, unknown-vs-outdated degradation, unit test) is sound and concurrency-correct (`@MainActor` class, no new retain cycles). The main thing to fix before this delivers its purpose is the refresh lifecycle: the section shows a false "unknown" on the common first-time-enable path. The all-or-nothing decode and the missing loading state are worth addressing alongside it.

<a href="https://opencode.ai/s/o9lqlLDZ">New%20session%20-%202026-08-14T23%3A40%3A56.468Z</a>
[opencode session](https://opencode.ai/s/o9lqlLDZ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31851119138)

kaishi00:
Findings addressed in eef6241 — full suite 650 passed, 0 failed.

**[major] refresh lifecycle** — Fixed: the Compatibility section carries its own `.task` (it appears exactly when the section becomes visible, i.e. the moment notifications get enabled in-place), and the relay URL field refetches on submit. The screen-level task still covers the already-enabled entry case.

**[minor] loading state** — Added: `isFetchingMeta` renders "Checking compatibility…" so an in-flight or slow request is never misdiagnosed as "predates version reporting".

**[minor] all-or-nothing decode** — Fixed: gateway rows decode lossily; one malformed record drops just its row while the relay and other gateways still render (unit-tested with a bad row between good ones).

**[minor] accessibility** — Icons carry explicit labels and rows combine into single accessibility elements. **[nits]** `last_event_at` dropped from the model (the relay still sends it; decoding ignores it); the version numbers are removed from the user-facing copy so nothing drifts on the next bump.

kaishi00:
## Summary
Adds a "Compatibility" section to Settings > Notifications backed by a new `GET /v1/meta` fetch (`PushNotificationService.refreshMeta()`), surfacing relay/plugin versions and capability-driven decision-card support with copy-paste update commands for stale notifiers, plus unit tests for decoding and capability checks.

## Findings

### [severity: minor] Non-404 failures are still misdiagnosed as "relay predates version reporting"
**File:** Conduit/Services/PushNotificationService.swift:240
**Issue:** Every non-200 response and every thrown error (401, 5xx, timeout, transport failure, decode error) maps to `relayMeta = nil`, which renders the categorical copy "Compatibility unknown. Your relay predates version reporting; decision cards may not be available until it updates." (AuxiliaryViews.swift:1394). That diagnosis is only valid for a confirmed 404 from a legacy relay. `isFetchingMeta` fixes the in-flight window but not the completed-with-error case, so a reachable relay with a transient failure shows a false diagnosis right at rollout time — the moment the section is most useful. CodeRabbit raised this and the loading state only partially addresses it.
**Suggestion:** Track a distinct "unavailable" state (e.g. an enum `metaState: unknown/.unavailable/.loaded`) mapped from a confirmed 404 vs. other failures, and render a neutral message ("Couldn't check compatibility right now") for the latter.

### [severity: minor] Overlapping `refreshMeta()` requests can clobber each other with stale data
**File:** Conduit/Services/PushNotificationService.swift:228
**Issue:** `refreshMeta()` is `@MainActor` but suspends on `URLSession.shared.data`. Two `.task` modifiers (screen-level AuxiliaryViews.swift:1459, section-level :1399) plus the `.onSubmit` handler can run concurrently. Because the earlier request's `await` yields and its completion writes `relayMeta` last, a stale fetch (e.g. against the old relay URL, issued before a URL change) can overwrite the result of the newer fetch, showing the previous relay's compatibility under the new URL. There's no generation guard or cancellation.
**Suggestion:** Cancel the prior fetch (store the `Task` and `.cancel()` before starting a new one) or capture the URL/token at start and only commit `relayMeta`/`isFetchingMeta` if it is still the latest request.

### [severity: minor] Stale relay metadata remains visible while the relay URL is edited but not yet submitted
**File:** Conduit/Views/AuxiliaryViews.swift:1442
**Issue:** `customRelayURL` is an `@AppStorage` bound to the TextField, so it mutates on every keystroke, but `relayMeta` is only refreshed on submit (`.onSubmit`, :1449). Between editing and submitting, the Compatibility section keeps showing the previous relay's support status for the new URL. CodeRabbit's suggestion to clear metadata when the URL changes was only partially implemented.
**Suggestion:** `.onChange(of: customRelayURL)` set `relayMeta = nil` (the section then shows "Checking compatibility…" or unknown) and optionally trigger the refetch, or move the refresh into `onChange` of a debounced value.

### [severity: nit] Redundant duplicate fetch on appear
**File:** Conduit/Views/AuxiliaryViews.swift:1459
**Issue:** With notifications enabled, both the screen-level `.task { refresh(); refreshMeta() }` and the section-level `.task { refreshMeta() }` fire on first appearance, issuing two identical concurrent requests. Harmless (idempotent, and one is cancelled by the race guard if added), but wasteful.
**Suggestion:** Drop the `refreshMeta()` from the screen-level `.task` since the section-level task covers the enabled case; or keep one and delete the other.

### [severity: nit] Duplicate step numbers across sections
**File:** Conduit/Views/AuxiliaryViews.swift:1389
**Issue:** The Compatibility section labels its copy commands "Step 1 Update the notifier / Step 2 Restart the gateway", but the same screen's "Connect a Hermes profile" (steps 1–3) and "Verify pairing" (steps 4–5) sections reuse 1–5, so two different "Step 1"s appear on screen together and the sequence reads ambiguously.
**Suggestion:** Omit step numbers in the Compatibility rows or use distinct labeling (e.g. "1a/1b" or plain bullet commands).

## Verdict
COMMENT

The core design is sound and the review round addressed the major lifecycle issue (section-level `.task` fires exactly when the Compatibility section appears, i.e. the in-place-enable path) plus accessibility and lossy decoding. Concurrency is correct under Swift 6 (`@MainActor` class, no new retain cycles, no force unwraps introduced), and the lossy-decode pattern is covered by a unit test. The remaining items are all minor but worth folding in before merge: the false "predates version reporting" diagnosis for non-404 failures, the overlapping-fetch race, and the stale-metadata-while-editing-URL window.

<a href="https://opencode.ai/s/QVP9cI39">New%20session%20-%202026-08-14T23%3A58%3A10.337Z</a>
[opencode session](https://opencode.ai/s/QVP9cI39)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31852071430)

PR #57, opened 2026-08-13
Fix cross-block Markdown selection and composer session races

kaishi00:
## Summary

- Restore text selection across mixed Markdown responses, including paragraphs separated by tables, code blocks, links, and tool output.
- Preserve selection anchors and lifecycle state for settled responses; a selection holds across re-renders once a reply finishes streaming (it resets while a reply is still streaming).
- Isolate composer submissions, attachment imports, slash-command continuations, and recovery from session/profile handoffs.
- Preserve composer text and attachments when Hermes rotates between runtime and stored session IDs.
- Restore title-tap scroll-to-top ownership across chat transitions and drag activity.

## Root cause

Markdown was rendered as independent selectable blocks, so native selection could not maintain one contiguous range when a response contained non-text blocks. Composer async work also relied on the currently visible session after suspension, allowing stale callbacks and recovery paths to mutate or restore the wrong draft. Runtime/stored session aliases were not consistently treated as the same composer identity.

## Validation

- Focused simulator suite: 100 tests passed, 0 failures.
- Full iOS simulator suite: 578 tests passed, 0 failures.
- Final whole-branch review: clean.

No release metadata or TestFlight upload is included in this PR.

## Summary by CodeRabbit

* **New Features**
  * Select and copy text across Markdown paragraphs, lists, tables, quotes, and code blocks with highlighted ranges.
  * Tap the conversation title to quickly scroll to the top.
  * Composer drafts are preserved separately across profiles and conversations, including attachments.
* **Bug Fixes**
  * Prevented delayed submissions, paste actions, attachment imports, and other asynchronous updates from affecting the wrong conversation.
  * Improved draft restoration and cleanup during conversation changes.
* **Accessibility**
  * Added an accessible label and hint for the conversation title scroll action.

PR #63, opened 2026-08-15
feat(chat): cross-block text selection for responses with tables and code

kaishi00:
## Summary

Selecting text that spans a Markdown table or code block now works end to end: long-press in a paragraph, drag down through the table and code block, lift, adjust either endpoint with handles that travel anywhere in the message, and copy the coordinated text. Within-block selection keeps the fully native behavior.

Follows up on #57/#58, which restored scrolling but deferred cross-block selection as "needs a different architecture."

## How it fits together

- **Observer gesture recognizer** (`MarkdownSelectionObserverGestureRecognizer`) rides every block/cell text view. It never leaves `.possible`, always allows simultaneous recognition, never cancels touches — so UITextView's private selection gesture and the enclosing scroll views keep stock behavior (the #57 failure mode) — yet it receives the full touch stream and forwards finger positions to the retained `MarkdownSelectionCoordinator`, which extends the selection across every registered segment. A nearest-segment fallback keeps the focus tracking through gaps and card chrome at drag speed.
- **Coordinator guards for real gestures:** a re-delivered synthetic `touchesBegan` no longer clears an in-progress drag; the owner's stale range reports can no longer collapse a cross-block selection after lift; a touch landing on an existing selection re-arms the observer (lift, then drag handles).
- **Coordinator-owned chrome for cross-block selections:** the owner resigns first responder at the first cross-block instant (non-FR views render no native selection — note: clearing tint renders iOS 26's selection black, not invisible), and a window-level overlay supplies endpoint handles with native-style hit slop plus a copy pill (Liquid Glass on iOS 26, adaptive thick material below) anchored to the focus end. The overlay highlights every span.
- **Copy paths write the coordinated text directly.** The `menuConfigurationFor:defaultMenu:` interceptor covers system-menu presentation — note the iOS 16 `suggestedActions:` selector is *never called* on iOS 17+ (a silently-dead optional protocol method was today's hardest-won lesson).

## Validation

- 92 unit tests green (selection suites + markdown fallback + session yolo, confirming the branch carries no unrelated drift).
- New UI test target drives real long-press drags across a table fixture (paragraph- and cell-anchored, plus a plain-drag scroll regression check); a DEBUG fixture screen (`-conduitSelectionFixture`) supports manual passes.
- Manual passes on iPhone 17 Pro Max (iOS 26) covered fast/continuous drags, handle adjustment after lift, cross-app paste, and light/dome chrome appearance.

## Known gaps

- No auto-scroll when a drag or handle reaches the screen edge; very large selections are adjusted in stages.
- The full simulator suite has pre-existing infra stalls on this machine (XCTest issue-recording hang, unrelated to this diff — seeFocused suites for what was run).

## Summary by CodeRabbit

* **New Features**
  * Select text across multiple Markdown blocks, including paragraphs, tables, and code blocks.
  * Drag custom selection handles to adjust selections spanning multiple blocks.
  * Copy cross-block selections using the Copy control or system menu.
  * Selection highlights and controls stay aligned while scrolling and dragging.
* **Bug Fixes**
  * Improved selection stability when moving between blocks and touching gaps.
  * Preserved native selection behavior within individual blocks.
  * Improved handling of fast gestures and selections near block boundaries.
  * Cross-block copying now consistently places plain text on the clipboard.


kaishi00:
## Summary
Adds cross-block text selection for Markdown responses containing tables and code blocks: an observer gesture recognizer rides each block/cell text view and feeds a coordinator that extends a native selection across segments, with coordinator-owned handles/copy pill chrome and coordinated copy paths, plus a DEBUG fixture screen and a new UI test target.

## Findings

### [severity: minor] Handle and copy-pill positions go stale while the transcript scrolls
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:67
**Issue:** `MarkdownSelectionHandleContainerView` is mounted at the fixed chat root (ChatView `.overlay`), outside the transcript `ScrollView`. Its `layoutSubviews` runs only on frame change / `updateUIView` (`setNeedsLayout`). Nothing triggers a relayout while the user scrolls, so the endpoint handles and copy pill stay pinned at their pre-scroll window coordinates and detach from the text they annotate; they only snap back once the user touches a handle (which publishes and re-lays-out). The highlight overlay inside the content scrolls correctly, so highlights and controls visually disagree during/after scroll.
**Suggestion:** Drive handle positioning from the scroll view — e.g., observe `scrollViewDidScroll` (or a `UIPanGestureRecognizer`/display-link on the transcript) to `setNeedsLayout()` the container, or host the handles in the scrolling content's coordinate space and convert to window per frame.

### [severity: minor] System-menu Copy is not replaced in non-English locales
**File:** Conduit/Views/Components/SelectableTextView.swift:435
**Issue:** The menu interceptor replaces the Copy action only when its title matches `localizedCaseInsensitiveContains("copy")`. In non-English locales the system "Copy" title doesn't contain the English substring (e.g. French "Copier"), so `replacedCopy` stays false: the coordinated action is inserted at index 0 while the real system Copy remains. Tapping that system Copy triggers the private text-system copy of the owner's native range only — silently copying wrong/incomplete text.
**Suggestion:** Match on the action's selector or `command`/identifier, or compare against `#selector(UIResponderStandardEditActions.copy)` / `UIAction.Identifier`, instead of the localized title.

### [severity: minor] Selection handles are not exposed to VoiceOver
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:331
**Issue:** `MarkdownSelectionHandleView` sets `accessibilityLabel` but never sets `isAccessibilityElement = true`, so the labels are dead — VoiceOver can't reach the custom handles to adjust a cross-block selection (native within-block selection is unaffected). The container also returns `nil` from `hitTest` for itself, so there's no accessibility fallback there.
**Suggestion:** Set `isAccessibilityElement = true` with `.adjustable` trait (and `accessibilityIncrement/Decrement` moving the endpoint by character/word), or expose an `accessibilityCustomAction`.

### [severity: nit] Force-unwrap in nearest-segment scan
**File:** Conduit/Views/Components/MarkdownSelectionCoordinator.swift:860
**Issue:** `distance < nearest!.distance` force-unwraps; it's safe because of the `nearest == nil` short-circuit, but it's fragile under refactors.
**Suggestion:** Bind in the `if let`/switch to avoid the unwrap.

### [severity: nit] Hard-coded tap coordinate in UI test
**File:** ConduitUITests/SelectionObserverUITests.swift:135
**Issue:** `testCopyAfterDragCapturesCrossBlockText` taps the fixture copy button via a hard-coded window offset `(220, 866)`; it breaks on any device/orientation/size-class change.
**Suggestion:** Tap `fixture.copySelection` through its accessibility identifier instead (the other tests already do), or compute the offset from the button's frame.

No retain cycles were found (coordinator references are `weak` on both the text view and the handle container; the observer recognizer holds its view weakly; the pill action and feedback timer capture `self` weakly), force unwraps are otherwise guarded, and the `@MainActor`-isolated coordinator is only touched from UIKit/SwiftUI main-thread contexts.

## Verdict
APPROVE — the architecture is sound and well-tested (unit + UI coverage), and the findings are minor/nits; the scroll-staleness and localized-menu issues are worth follow-up but not blockers.

<a href="https://opencode.ai/s/Sm6BmKvo">New%20session%20-%202026-08-15T02%3A17%3A09.137Z</a>
[opencode session](https://opencode.ai/s/Sm6BmKvo)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31858652462)

kaishi00:
Review round pushed (0a8812a). Triage:

**Fixed**
- **ConduitApp release build (critical, good catch):** the fixture branch referenced the DEBUG-only `SelectionFixtureView` unguarded — every validation run built Debug, so Release would not have compiled. Now `#if DEBUG`-guarded with the root content extracted; verified with a Release simulator build.
- **Suppression flag on collapse:** a handle drag shrinking a cross-block selection back into one segment left it stranded (no native chrome — owner resigned; no coordinator chrome — not cross-segment). The selection now ends, matching a tap. Unit-tested.
- **`nearest!` force unwrap:** replaced with an optional-comparison.
- **Test hygiene:** UI drag tests now assert endpoint handles exist (screenshot capture kept for diagnosis); copy tests tap by identifier instead of hardcoded offsets; assertions read the fixture's pasteboard mirror instead of runner-side `UIPasteboard` (Allow-Paste consent can stall CI); unit tests restore the pasteboard they overwrite.

**Not changed, with rationale**
- **`editMenuForTextIn:suggestedActions:` suggestion:** that selector doesn't exist in UITextViewDelegate; the iOS 16 method was `editMenuForTextRange:suggestedActions:`, replaced in iOS 17 by `menuConfigurationFor:defaultMenu:` — which is what this PR implements, and it was verified empirically: after switching to the correct selector, a manual menu-Copy on the simulator produced the full cross-block text (the iOS 16 selector produced zero delegate callbacks in instrumented logs).
- **Plain-drag scroll test `dy: -12`:** normalized offsets are multiples of the element's size, and the anchor marker is 2pt tall, so this is a 24pt drag — intended. The test's purpose is the #57 regression class (a plain drag must not select/copy), which the pasteboard assertion covers; the fixture's content is shorter than the viewport, so a scroll assertion would require a taller fixture variant and isn't what this test guards.

kaishi00:
## Summary
Adds cross-block text selection for Markdown responses containing tables/code: an observer gesture recognizer rides each block's text view and feeds a coordinator that extends a native selection across segments, with coordinator-owned handles, a copy pill, coordinated copy paths, a DEBUG fixture, and unit + UI test coverage.

## Findings

### [severity: major] `menuConfigurationFor:defaultMenu:` implementation doesn't match the documented delegate method
**File:** Conduit/Views/Components/SelectableTextView.swift:411
**Issue:** Apple's documented method is `textView(_:menuConfigurationFor textItem: UITextItem, defaultMenu: UIMenu) -> UITextItem.MenuConfiguration?` (precise identifier `textView:menuConfigurationForTextItem:defaultMenu:`), and it customizes menus for **text items** (links, attachments), not the selection edit menu. The code implements `textView(_:menuConfigurationFor textRange: UITextRange, defaultMenu: UIMenu) -> UIMenu` — different parameter type, different return type, and a different ObjC selector. UIKit dispatches by selector, so this method is very likely never invoked; the unit test (`testMenuConfigurationReplacesCopyActionForCrossSegmentSelection`, ChatTextSelectionTests.swift:436) calls the method directly, so it cannot catch a dead delegate hook. The code comment's claim that the iOS 16 `suggestedActions:` selector "is never called on iOS 17+" also contradicts Apple's docs, which list `textView(_:editMenuForTextIn:suggestedActions:)` as available iOS 16–27 (not deprecated). If the interceptor is dead, the "system menu Copy" path for cross-block selections silently falls back to the owner's native range — the exact wrong-copy scenario the comment says is the reason for replacing the menu action.
**Suggestion:** Implement `textView(_:editMenuForTextIn:suggestedActions:) -> UIMenu?` (iOS 16+, valid on 17+) and move the Copy-replacement logic there, preserving `suggestedActions`. Confirm the hook is actually invoked (breakpoint/log) on a real iOS 17+/26 run rather than relying on the current claim; the existing test should be reworked to assert against a real menu presentation or kept as a pure utility-function test of the menu-building logic.

### [severity: minor] Handles and copy pill go stale while the transcript scrolls
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:67
**Issue:** `MarkdownSelectionHandleContainerView` is mounted at the fixed chat root (ChatView.swift:468 `.overlay { MarkdownSelectionChromeRoot() }`), outside the transcript `ScrollView`. `layoutSubviews` only runs on frame change/`setNeedsLayout`; nothing triggers a relayout during scroll, so endpoint handles and the copy pill stay pinned at pre-scroll window coordinates and detach from the text they annotate, while the in-content highlight overlay moves correctly — the two disagree visually. Not addressed in the review round.
**Suggestion:** Drive repositioning from the scroll view (e.g. observe `scrollViewDidScroll` or a display-link/pan observer on the transcript to `setNeedsLayout()` the container), or host handles in the scrolling content's coordinate space and convert to window each frame.

### [severity: minor] Selection handles are not exposed to VoiceOver
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:331
**Issue:** `MarkdownSelectionHandleView` sets `accessibilityLabel` but never `isAccessibilityElement = true`, so VoiceOver cannot reach the custom handles to adjust a cross-block selection (native within-block selection is unaffected). The container also returns `nil` from `hitTest` for itself, so there's no accessibility fallback there.
**Suggestion:** Set `isAccessibilityElement = true` with the `.adjustable` trait and implement `accessibilityIncrement/Decrement` (move endpoint by character/word), or add an `accessibilityCustomAction`.

### [severity: minor] UI test still reads the runner-side pasteboard, contradicting the stated fix
**File:** ConduitUITests/SelectionObserverUITests.swift:177
**Issue:** The PR comment claims assertions now "read the fixture's pasteboard mirror instead of runner-side UIPasteboard (Allow-Paste consent can stall CI)", but `testCustomHandlesAndCopyPillAppearAfterCrossBlockDrag` still reads `UIPasteboard.general.string` directly in the runner process. `testCopyAfterDragCapturesCrossBlockText` was converted (uses `fixture.pasteboard` label), but this test wasn't.
**Suggestion:** Assert via the `fixture.pasteboard` label after tapping `copyPill`, matching the other test.

### [severity: minor] System-menu Copy replacement matches on the localized title
**File:** Conduit/Views/Components/SelectableTextView.swift:435
**Issue:** `action.title.localizedCaseInsensitiveContains("copy")` only matches in English. In other locales the system Copy title (e.g. French "Copier") won't match, so `replacedCopy` stays false and the coordinated action is inserted at index 0 while the real system Copy remains — tapping it copies the native range only. (Compounds with the major finding above; if the hook is replaced with the correct `editMenuForTextIn:suggestedActions:` method, match on the selector/identifier instead.)
**Suggestion:** Compare against `#selector(UIResponderStandardEditActions.copy)` or a `UIAction` identifier rather than the localized title.

### [severity: nit] `copyActiveSelection` shadowing, redundant cast, and inset mismatch
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:250
**Issue:** `copiedAttributedTextForActiveSelection() as NSAttributedString?` is a redundant cast (already non-optional); `var copied = pill.configuration` (line 261) shadows the outer `copied` attributed string; install-path insets are top/bottom `8` while the restore path uses `7`, so the pill height changes after every copy. The `DispatchQueue.main.asyncAfter` is never cancelled, so repeated taps schedule overlapping resets.
**Suggestion:** Bind directly without the cast, rename the shadowing variable, match insets, and store/cancel the work item if the last tap should win.

### [severity: nit] `invalidateIntrinsicContentSize()` inside `layoutSubviews`
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:97
**Issue:** Invalidating intrinsic content size during the container's layout pass and immediately reading it marks the pill as needing layout again on every pass; `layoutSubviews` runs on every coordinator update during a drag, scheduling redundant layout work per frame.
**Suggestion:** Use `systemLayoutSizeFitting(...)` or invalidate only on configuration change.

### [severity: nit] Side effect in `SelectionFixtureView.init`
**File:** Conduit/Views/SelectionFixtureView.swift:30
**Issue:** `UIView.setAnimationsEnabled(false)` runs in `init`, which SwiftUI can call multiple times per render pass, and it's never restored.
**Suggestion:** Move to `.task`/`.onAppear`, gated on `CONDUIT_UITEST == "1"`.

No retain cycles found: coordinator references are `weak` on the text views, observer, and handle container; pill action and feedback timer capture `self` weakly; `activeVisibleSelectionCoordinator` is a static `weak`. Force-unwrap in the nearest-segment scan was correctly removed in the review round. The `@MainActor`-isolated coordinator is only touched from main-thread UIKit/SwiftUI contexts, and the release-build `#if DEBUG` guard for the fixture was correctly added.

## Verdict
COMMENT — The architecture and testing are solid and the review-round fixes (release guard, collapse handling, `nearest!` removal, pasteboard hygiene in the copy test) are good. The blocking concern is the `menuConfigurationFor:defaultMenu:` hook: its signature doesn't match Apple's documented delegate method, so the headline "system-menu Copy interceptor" is likely dead code and the fallback copy path may be unreliable. Please verify on-device with a definitive hook check before merge; the scroll-staleness and accessibility findings are worthwhile follow-ups.

<a href="https://opencode.ai/s/rl5rcbmr">New%20session%20-%202026-08-15T03%3A02%3A11.081Z</a>
[opencode session](https://opencode.ai/s/rl5rcbmr)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31860698836)

kaishi00:
Review round 2 pushed (379ff1f).

**OpenCode's major — correct, and conceded with evidence.** Verified against the SDK's own `UITextView.h`: the only `menuConfigurationFor…defaultMenu:` selector is `menuConfigurationForTextItem:` (iOS 17+, for links/attachments); the selection-menu delegates are `editMenuForTextInRange:suggestedActions:` (iOS 16) and `editMenuForTextInRanges:suggestedActions:` (iOS 26). My implementation matched neither selector, so it compiled silently dead — and the manual "verification" I cited earlier was confounded (the fixture's copy button had just been introduced and was used in that pass). The interception is now **deleted** rather than re-pointed: since cross-block selections dismiss the owner's first responder mid-drag, they never present the system menu at all, so there is nothing to intercept — copying goes through the coordinator-owned pill, and within-block selections keep the fully native menu. The directly-invoked unit test went with it. This also moots the localized-title-matching finding.

**Fixed from the same review:**
- Handles now set `isAccessibilityElement`, so they exist in the accessibility tree — this is also why CI's identifier queries couldn't find them (VoiceOver can now reach them too; full adjustable-trait support remains a documented follow-up).
- The chrome container runs a display link while visible, so handles and the pill track the transcript during scrolls instead of detaching at stale window coordinates.
- The handles test no longer reads the runner-side pasteboard (the one leftover OpenCode caught); the plain-drag test now compares the pasteboard mirror before/after instead of assuming a pristine empty pasteboard — earlier tests in the same CI run legitimately dirty it, which is what failed.

**CI failures explained:** both UI-test failures trace to the two items above (handles invisible to AX queries; stale empty-pasteboard assumption). Unit suites 63/63 locally; expecting green CI on this push.

kaishi00:
## Summary
Adds cross-block text selection for Markdown responses containing tables/code: an observer gesture recognizer feeds a coordinator that extends native selection across segments, with coordinator-owned handles/copy pill, coordinated copy paths, a DEBUG fixture, and unit + UI test coverage.

## Findings

### [severity: minor] Two UI drag tests still assert nothing
**File:** ConduitUITests/SelectionObserverUITests.swift:41 (and :105)
**Issue:** `testCellAnchoredLongPressAndDragExtendsAcrossBlocks` and `testFastLongPressAndDragStillExtendsAcrossBlocks` perform the drag, `Thread.sleep`, write a screenshot to `/tmp`, and end — zero assertions. They pass even if the selection never crosses a block, so they cannot detect the regression their names describe. The PR comment ("UI drag tests now assert endpoint handles exist") does not match the code; only `testCustomHandlesAndCopyPillAppearAfterCrossBlockDrag` asserts an outcome. The `/tmp` screenshot is never verifiable on CI.
**Suggestion:** After the drag, assert on the fixture's observable outcomes (`selection.handle.anchor`, `selection.handle.focus`, `selection.copyPill`, or `fixture.pasteboard`), and attach screenshots via `XCTAttachment` instead of writing to `/tmp`.

### [severity: minor] `testPlainDragStillScrolls` never verifies scrolling
**File:** ConduitUITests/SelectionObserverUITests.swift:60-79
**Issue:** The test name claims to guard the #57 scroll regression, but it only asserts the pasteboard mirror is unchanged. No scroll-dependent fact is checked, and `coordinate(withNormalizedOffset:dy: -12)` is a normalized offset applied to a 2pt-tall marker, so the drag start/end resolve to points off the element (and likely outside the scroll content entirely). The fixture content is shorter than the viewport, so no scroll can actually occur regardless.
**Suggestion:** Either rename the test to reflect what it guards ("plain drag must not select/copy"), or add a taller fixture and assert `fixture.top`/`fixture.bottom` frame movement using an in-range offset plus `withOffset`.

### [severity: minor] CADisplayLink retain cycle while the chrome is visible
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:67, 211-235
**Issue:** `CADisplayLink(target: self, ...)` retains its target strongly, and `repositioningDisplayLink` is a strong property on the same view, forming a cycle while the link runs. The link is only invalidated from `layoutSubviews` (hide paths) or `deinit` — but `deinit` is unreachable while the cycle holds. If the chrome is unmounted while a cross-block selection is still active (e.g. the chat screen is dismissed before the coordinator clears), the container and link leak and the link keeps firing `setNeedsLayout` on a windowless view.
**Suggestion:** Invalidate the link in `didMoveToWindow`/`willMove(toWindow:)` when `window == nil`, or use a weak-proxy target.

### [severity: minor] Per-frame layout churn while a selection is displayed
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:96-102, 223-231
**Issue:** The display link runs continuously while a cross-block selection is at rest (not just during scroll), and `repositionFromDisplayLink` calls `setNeedsLayout()` every frame. Each pass runs `layoutSubviews`, which recomputes both caret rects (each calling `textView.layoutIfNeeded()` + glyph queries) and calls `pill.invalidateIntrinsicContentSize()` during the layout pass before reading `intrinsicContentSize` — scheduling redundant layout every frame even when nothing moved. This is main-thread/battery cost while the user merely holds a selection.
**Suggestion:** Gate the link on actual motion (compare the last caret geometry, or run it only while the transcript is scrolling/panning), and size the pill with `systemLayoutSizeFitting(...)` or invalidate only when its configuration changes.

### [severity: minor] Inconsistent pasteboard strategy between the pill and the `copy(_:)` override
**File:** Conduit/Views/Components/SelectableTextView.swift:511-524
**Issue:** `MarkdownSelectionTextView.copy(_:)` writes both `UIPasteboard.general.string` and an RTF item, while the pill's `copyActiveSelection()` deliberately writes string-only with a comment documenting that pairing RTF made external paste targets come up empty. If `copy(_:)` is ever reached for a cross-block selection (e.g. hardware-keyboard Copy while the owner is still first responder mid-drag), it reproduces the known-broken paste. This path is now also completely untested (see below).
**Suggestion:** Make both paths use the same proven string-only write, and cover `copy(_:)` with a unit test.

### [severity: minor] Pill inset mismatch and uncancelled feedback timer
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:179 vs 300, 309
**Issue:** The pill is installed with top/bottom insets of `8` but the copy-feedback and restore paths use `7`, so the pill height changes by 2pt after the first copy and never returns to its original size. The `DispatchQueue.main.asyncAfter` (line 303) is never cancelled, so rapid taps schedule overlapping resets.
**Suggestion:** Match the insets across all three paths; store/cancel the work item if the last tap should win.

### [severity: minor] Side effect in `SelectionFixtureView.init`
**File:** Conduit/Views/SelectionFixtureView.swift:30-34
**Issue:** `UIView.setAnimationsEnabled(false)` runs in `init`, which SwiftUI may invoke multiple times per render pass, and the setting is never restored. The init is also not `@MainActor`-isolated while calling a `@MainActor` UIKit class method (a Swift 6 strict-concurrency issue).
**Suggestion:** Move it to `.task`/`.onAppear`, gated on `CONDUIT_UITEST == "1"`, and restore animations on disappear.

### [severity: minor] Handles reachable by VoiceOver but not operable
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:369-370
**Issue:** The handles now set `isAccessibilityElement = true` (good — this also fixed the UI-test identifier queries), but they carry no `.adjustable` trait and no `accessibilityIncrement`/`Decrement` or custom actions. A VoiceOver user can focus a handle and hear its label but cannot adjust a cross-block selection, which only responds to pans.
**Suggestion:** Add the `.adjustable` trait with increment/decrement moving the endpoint by character/word, or an `accessibilityCustomAction`.

### [severity: nit] Test-coverage regression in round 2
**File:** ConduitTests/ChatTextSelectionTests.swift (removed in 379ff1f)
**Issue:** Round 2 deleted `testCopyPutsCoordinatedCrossSegmentTextOnPasteboard` and `testHandleChromeDragsMoveEachEndpointAndPillCopies` (plus the dead menu test). The pasteboard-writing paths — the pill's `copyActiveSelection` and the `copy(_:)` override — now have no unit coverage; the coordinated copy is only verified through the single UI test, and `copy(_:)` has none at all.
**Suggestion:** Re-add unit tests for the two pasteboard-writing paths (with pasteboard save/restore via `defer`, as round 1 had), keeping them independent of the removed menu hook.

### [severity: nit] `makeSelectionWindow` never calls `makeKeyAndVisible()`
**File:** ConduitTests/MarkdownSelectionCoordinatorTests.swift:918-925
**Issue:** The helper sets `isHidden = false` but never makes the window key; `becomeFirstResponder()` can return `false` for a view in a non-key window, making the `XCTAssertTrue(firstTextView.isFirstResponder)` assertions at lines 645/774 flaky on some CI/simulator configurations.
**Suggestion:** Call `makeKeyAndVisible()` in the helper.

No critical or major issues remain: the release-build `#if DEBUG` guard, the dead `menuConfigurationFor:defaultMenu:` hook (correctly identified as a silently-dead selector and removed), the `nearest!` force-unwrap, and the collapse-suppression flag are all properly addressed. Retain-cycle discipline is otherwise good — coordinator references are `weak` on the text views, observer, and handle container; pill action and feedback timer capture `self` weakly — and the `@MainActor`-isolated coordinator is only touched from main-thread UIKit/SwiftUI contexts.

## Verdict
COMMENT — The architecture and test coverage are solid and the review-round fixes are sound. Remaining items are minor: the two assertion-free UI drag tests, the misleading scroll test, the display-link retain-cycle/per-frame churn, and the copy-path inconsistency are worth follow-up, but none block merge.

<a href="https://opencode.ai/s/A4C1mP0x">New%20session%20-%202026-08-15T03%3A19%3A51.431Z</a>
[opencode session](https://opencode.ai/s/A4C1mP0x)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31861472706)

kaishi00:
## Summary
Adds cross-block text selection for Markdown responses containing tables/code blocks: a never-recognizing observer gesture recognizer rides each block/cell text view and feeds a `@MainActor` coordinator that extends a native selection across segments, with coordinator-owned endpoint handles and a copy pill, coordinated copy paths, a DEBUG fixture screen, and unit + UI test coverage.

## Findings

### [severity: major] CADisplayLink retain cycle leaks the chrome container (and keeps firing forever)
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:67, 211-235
**Issue:** `CADisplayLink(target: self, ...)` retains its target, and `repositioningDisplayLink` is a strong property on the same view, so while the link runs (the `.common` run-loop mode retains the link) the cycle view→link→view is unbreakable. The link is only invalidated from `layoutSubviews` hide paths or `deinit` — but `deinit` is unreachable while the cycle holds, and `layoutSubviews` is not invoked when the overlay is removed. Because `MarkdownSelectionChromeRoot` removes the overlay as soon as `activeCoordinator` is nil (any `clearSelection()`), every dismissal of an active cross-block selection — copying then tapping away, or navigating off the chat — leaves the container leaked with the link running `setNeedsLayout()` on a windowless view indefinitely. This is the common path, not a narrow edge case.
**Suggestion:** Invalidate the link in `didMoveToWindow()`/`willMove(toWindow:)` when `window == nil` (and also when the guard in `repositionFromDisplayLink()` fails), or use a weak-proxy target.

### [severity: minor] Per-frame layout churn while a selection is at rest
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:96-102, 223-231
**Issue:** The display link runs continuously whenever a cross-block selection is displayed — including at rest, not only during scroll — and `repositionFromDisplayLink()` calls `setNeedsLayout()` every frame. Each pass recomputes both caret rects (`textView.layoutIfNeeded()` + glyph queries) and calls `pill.invalidateIntrinsicContentSize()` during the layout pass before reading `intrinsicContentSize`, scheduling redundant layout every frame even when nothing moved. Main-thread/battery cost while the user merely holds a selection.
**Suggestion:** Gate the link on actual motion (compare last caret geometry, or run it only while the transcript is panning/decellerating), and size the pill with `systemLayoutSizeFitting(...)` or invalidate only when its configuration changes.

### [severity: minor] Handles are focusable by VoiceOver but not operable
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:369-370
**Issue:** `isAccessibilityElement = true` was correctly added, so handles now exist in the accessibility tree, but they carry no `.adjustable` trait and no `accessibilityIncrement`/`Decrement` or `accessibilityCustomAction`. A VoiceOver user can focus a handle and hear its label but cannot adjust a cross-block selection, which only responds to pans.
**Suggestion:** Add `.adjustable` with increment/decrement routing to `coordinator.updateSelection`/`updateAnchorSelection` (moving the endpoint by character/word), or expose an `accessibilityCustomAction`; surface the current endpoint in `accessibilityValue`.

### [severity: minor] Two UI drag tests still assert nothing
**File:** ConduitUITests/SelectionObserverUITests.swift:41, 124
**Issue:** `testCellAnchoredLongPressAndDragExtendsAcrossBlocks` and `testFastLongPressAndDragStillExtendsAcrossBlocks` perform the drag, sleep, and write a screenshot to `/tmp` — zero assertions (the PR comment "UI drag tests now assert endpoint handles exist" does not match the code; `performCrossBlockDrag` retries but never asserts). They pass even if the selection never crosses a block. The `/tmp` screenshots are also not verifiable on CI.
**Suggestion:** After the drag, assert on the fixture's observable outcomes (`selection.handle.anchor`, `selection.handle.focus`, `selection.copyPill`, or `fixture.pasteboard`), following `testCustomHandlesAndCopyPillAppearAfterCrossBlockDrag`; attach screenshots via `XCTAttachment` instead of `/tmp`.

### [severity: minor] `testPlainDragStillScrolls` never verifies scrolling
**File:** ConduitUITests/SelectionObserverUITests.swift:64-83
**Issue:** The test only asserts the pasteboard mirror is unchanged. `coordinate(withNormalizedOffset:dy: -12)` is applied to a 2pt-tall marker, so the offset resolves off the element; no scroll-dependent fact is checked, and the fixture content is shorter than the viewport so no scroll can occur regardless. The name overstates what it guards.
**Suggestion:** Rename it to reflect what it actually verifies ("a plain drag must not select/copy"), or add a taller fixture and assert a scroll fact (e.g. `fixture.bottom` becoming hittable) using an in-range offset plus `withOffset`.

### [severity: minor] Copy paths disagree on pasteboard format
**File:** Conduit/Views/Components/SelectableTextView.swift:511-524
**Issue:** `MarkdownSelectionTextView.copy(_:)` writes both `UIPasteboard.general.string` and an RTF item, while the pill's `copyActiveSelection()` deliberately writes string-only with a documented comment that pairing RTF made external paste targets come up empty. If `copy(_:)` is reached for a cross-block selection (e.g. hardware-keyboard Cmd-C while the owner is still first responder mid-drag), it reproduces the known-broken paste. This path is also currently untested.
**Suggestion:** Use the proven string-only write in both paths and cover `copy(_:)` with a unit test.

### [severity: minor] Unit-test first-responder assertions can be flaky
**File:** ConduitTests/MarkdownSelectionCoordinatorTests.swift:918-925
**Issue:** `makeSelectionWindow` sets `isHidden = false` but never calls `makeKeyAndVisible()`; `becomeFirstResponder()` can return `false` for a view in a non-key window, so `XCTAssertTrue(firstTextView.isFirstResponder)` at lines 644/671/773/800 may fail intermittently on CI/simulator.
**Suggestion:** Call `makeKeyAndVisible()` in the helper (and consider `XCTAssertTrue(firstTextView.becomeFirstResponder())` at the call sites).

### [severity: nit] `copyActiveSelection` shadowing, inset mismatch, and uncancelled timer
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:282-313
**Issue:** `var copied = pill.configuration` (line 297) shadows the outer `copied` attributed string; the install path uses top/bottom insets `8` (line 179) while the feedback and restore paths use `7` (lines 300, 309), so the pill height changes by 2pt after each copy and never returns to the original size; the `DispatchQueue.main.asyncAfter` is never cancelled, so rapid taps schedule overlapping resets.
**Suggestion:** Rename the inner variable, match insets across all three paths, and store/cancel the work item if the last tap should win.

### [severity: nit] Side effect in `SelectionFixtureView.init`
**File:** Conduit/Views/SelectionFixtureView.swift:30-34
**Issue:** `UIView.setAnimationsEnabled(false)` runs in `init`, which SwiftUI can invoke multiple times per render pass, and the setting is never restored.
**Suggestion:** Move it to `.task`/`.onAppear`, gated on `CONDUIT_UITEST == "1"`.

No remaining critical issues: the dead `menuConfigurationFor:defaultMenu:` hook (correctly identified and removed), the release-build `#if DEBUG` guard, the `nearest!` force-unwrap, and the collapse-suppression handling are all properly addressed. Retain-cycle discipline is otherwise good (coordinator references are `weak` on text views, observer, and container; the pill action and feedback timer capture `self` weakly), and the `@MainActor`-isolated coordinator is only touched from main-thread UIKit/SwiftUI contexts.

## Verdict
COMMENT — The architecture and test coverage are solid and the review-round fixes are sound. The display-link retain cycle (leaked on every selection dismissal, with per-frame work continuing forever) is worth fixing before merge; the remaining items are minor/nits and do not block.

<a href="https://opencode.ai/s/uutEBG5z">New%20session%20-%202026-08-15T03%3A32%3A56.461Z</a>
[opencode session](https://opencode.ai/s/uutEBG5z)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31862056625)

kaishi00:
## Summary
This PR adds cross-block Markdown text selection for chat responses with tables and code: a never-recognizing observer gesture recognizer rides each block/cell text view and feeds a `@MainActor` coordinator that extends a native selection across segments, with coordinator-owned endpoint handles and a copy pill, coordinated copy paths, a DEBUG fixture, and unit + UI test coverage.

## Findings

### [severity: major] CADisplayLink retain cycle leaks the chrome container on every selection dismissal
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:67, 211-235
**Issue:** `CADisplayLink(target: self, …)` retains its target, and `repositioningDisplayLink` is a strong property on the same view, so while the link runs (retained by the `.common` run-loop mode) the cycle view→link→view is unbreakable. The link is invalidated only from `layoutSubviews` hide paths or `deinit` — but `deinit` is unreachable while the cycle holds, and `layoutSubviews` is not called when SwiftUI removes the overlay from the window. `MarkdownSelectionChromeRoot` removes the overlay as soon as `activeCoordinator` is nil (any `clearSelection()`), so dismissing an active cross-block selection — copying then tapping away, or navigating off the chat — leaves the container leaked with the link firing `setNeedsLayout()` on a windowless view indefinitely. `repositionFromDisplayLink`'s guard-failure branch (line 226) only calls `setNeedsLayout()` and never stops the link, so the failure path can't break the cycle either.
**Suggestion:** Invalidate the link in `willMove(toWindow:)`/`didMove(toWindow:)` when `window == nil`, and also stop it when the guard in `repositionFromDisplayLink()` fails, rather than deferring to a layout pass that may never come.

### [severity: minor] Per-frame layout churn while a selection is at rest
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:96-124, 223-231
**Issue:** The display link runs continuously whenever the chrome is visible — including at rest, not just during scroll — and every frame calls `setNeedsLayout()`. Each pass recomputes both caret rects (`textView.layoutIfNeeded()` + glyph queries in `coordinator.caretRect`) and calls `pill.invalidateIntrinsicContentSize()` during the layout pass before reading `intrinsicContentSize`, re-scheduling layout each frame even when nothing moved. This is continuous main-thread work (60/120Hz) while the user merely holds a selection.
**Suggestion:** Gate the link on actual motion (compare last caret geometry, or run it only while the transcript is panning/decellerating), and size the pill with `systemLayoutSizeFitting(...)` or invalidate only when its configuration changes.

### [severity: minor] Handles are focusable by VoiceOver but not operable
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:369-370
**Issue:** `isAccessibilityElement = true` was correctly added, so handles now exist in the accessibility tree, but they carry no `.adjustable` trait and no `accessibilityIncrement`/`Decrement` or `accessibilityCustomAction`. A VoiceOver user can focus a handle and hear its label but cannot adjust a cross-block selection, which only responds to pans.
**Suggestion:** Add `.adjustable` with `accessibilityIncrement`/`Decrement` routing to `coordinator.updateAnchorSelection`/`updateSelection` (moving the endpoint by character/word), and surface the current endpoint via `accessibilityValue`.

### [severity: minor] Copy paths disagree on pasteboard format
**File:** Conduit/Views/Components/SelectableTextView.swift:511-524
**Issue:** `MarkdownSelectionTextView.copy(_:)` writes both `UIPasteboard.general.string` and an RTF item, while the pill's `copyActiveSelection()` deliberately writes string-only with a comment documenting that pairing RTF made external paste targets come up empty (MarkdownSelectionHandles.swift:290-293). If `copy(_:)` is ever reached for a cross-block selection (e.g. hardware-keyboard Cmd-C while the owner is still first responder mid-drag), it reproduces the known-broken paste. This path is also untested.
**Suggestion:** Use the proven string-only write in both paths and cover `copy(_:)` with a unit test.

### [severity: minor] `testPlainDragStillScrolls` never verifies scrolling
**File:** ConduitUITests/SelectionObserverUITests.swift:61-80
**Issue:** The test only asserts the pasteboard mirror is unchanged. The `dy: -12` "normalized" offset is applied to a 2pt-tall marker, so the end point resolves 24pt above the element (off the scroll content, likely outside it), and no scroll-dependent fact is checked — the fixture content is shorter than the viewport so no scroll can occur regardless. The name overstates what it guards.
**Suggestion:** Rename it to reflect what it verifies ("a plain drag must not select/copy"), or add a taller fixture and assert a scroll fact (e.g. `fixture.bottom` becoming hittable) using an in-range offset plus `withOffset`.

### [severity: minor] `/tmp` screenshot writes remain in UI tests
**File:** ConduitUITests/SelectionObserverUITests.swift:162-163, 197-198, 209-210
**Issue:** Screenshots are still written to absolute `/tmp` paths inside the simulator sandbox; these are not verifiable on CI and silently do nothing useful. Several prior review rounds flagged this.
**Suggestion:** Attach via `XCTAttachment(screenshot:)` so they land in the result bundle.

### [severity: minor] Unit-test first-responder assertions can be flaky
**File:** ConduitTests/MarkdownSelectionCoordinatorTests.swift:918-925
**Issue:** `makeSelectionWindow` sets `isHidden = false` but never calls `makeKeyAndVisible()`; `becomeFirstResponder()` can return `false` for a view in a non-key window, so the `XCTAssertTrue(firstTextView.isFirstResponder)` preconditions at the new tests' `becomeFirstResponder()` call sites may fail intermittently on CI/simulator.
**Suggestion:** Call `makeKeyAndVisible()` in the helper, and prefer `XCTAssertTrue(firstTextView.becomeFirstResponder())` at the call sites.

### [severity: nit] `copyActiveSelection` shadowing, inset mismatch, uncancelled timer
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:282-313
**Issue:** `var copied = pill.configuration` (line 297) shadows the outer `copied` attributed string; the install path uses top/bottom insets `8` (line 179) while the feedback and restore paths use `7` (lines 300, 309), so the pill height changes by 2pt after each copy and never returns to its original size; the `DispatchQueue.main.asyncAfter` is never cancelled, so rapid taps schedule overlapping resets.
**Suggestion:** Rename the inner variable, match insets across all three paths, and store/cancel the work item if the last tap should win.

### [severity: nit] Side effect in `SelectionFixtureView.init`
**File:** Conduit/Views/SelectionFixtureView.swift:30-34
**Issue:** `UIView.setAnimationsEnabled(false)` runs in `init`, which SwiftUI can invoke multiple times per render pass, and the setting is never restored.
**Suggestion:** Move it to `.task`/`.onAppear`, gated on `CONDUIT_UITEST == "1"`.

No critical issues remain: the dead `menuConfigurationFor:defaultMenu:` hook (correctly identified as a silently-dead selector and removed), the release-build `#if DEBUG` guard, the `nearest!` force-unwrap, and the collapse-suppression handling are all properly addressed. Retain-cycle discipline is otherwise good — coordinator references are `weak` on the text views, observer, and container; the pill action and feedback timer capture `self` weakly — and the `@MainActor`-isolated coordinator is only touched from main-thread UIKit/SwiftUI contexts.

## Verdict
COMMENT — The architecture and test coverage are solid and the review-round fixes are sound. The CADisplayLink retain cycle (a leak on the common dismissal path, with per-frame work continuing forever) is worth fixing before merge; the remaining items are minor/nits and do not block.

<a href="https://opencode.ai/s/W8dYQH9P">New%20session%20-%202026-08-15T03%3A47%3A34.768Z</a>
[opencode session](https://opencode.ai/s/W8dYQH9P)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31862692687)

kaishi00:
Review round 3 pushed (17f0f78).

**Major (CADisplayLink leak) — correct, fixed.** The link retained its target while the container retained the link, its only teardown paths were a layout pass or a cycle-blocked deinit, and SwiftUI's overlay removal fires neither — every selection dismissal leaked the container with the link running indefinitely. Teardown is now explicit: `willMove(toWindow: nil)` and the reposition guard both invalidate the link directly.

**Minors fixed:** the reposition pass now compares both endpoints' caret rects and skips `setNeedsLayout` when nothing moved, and the pill no longer invalidates its intrinsic size every layout pass — a resting selection costs no per-frame main-thread work (the caret comparison remains, which is the cheap part of the old pass). `copy:` now uses the same string-only pasteboard write as the pill, closing the reachable-mid-drag ⌘C path that would have reproduced the known broken RTF pairing. Handles carry `.adjustable` with increment/decrement moving their endpoint one character through the coordinator.

**Minor not changed (again, with rationale):** `testPlainDragStillScrolls` verifying scroll — the fixture's content is shorter than the viewport, so no scroll can occur; the test guards the #57 regression class (a plain drag must not select or copy), which its before/after pasteboard assertion does cover. A taller fixture variant would enable a scroll assertion but perturbs the anchor geometry every other drag test depends on; happy to add it as a follow-up if wanted.

kaishi00:
## Summary
This PR adds cross-block text selection for Markdown responses containing tables and code: a never-recognizing observer gesture recognizer feeds a `@MainActor` coordinator that extends native selection across segments, with coordinator-owned endpoint handles and a copy pill, coordinated copy paths, a DEBUG fixture, and unit + UI test coverage. The prior rounds' criticals are all properly addressed at HEAD (release `#if DEBUG` guard, the dead `menuConfigurationFor` hook removed, `nearest!` unwrap replaced, and the `CADisplayLink` leak fixed via `willMove(toWindow:)` + guard-failure teardown in `MarkdownSelectionHandles.swift:254-271`).

## Findings

### [severity: minor] Display link does per-frame caret-geometry work while a selection rests
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:261-283
**Issue:** The link is started whenever cross-block chrome is visible (layoutSubviews:100) and fires at 60/120Hz continuously — not only during scroll. The "quiet rest" optimization (`caretRects unchanged → return`) skips `setNeedsLayout`, but it still computes both caret rects every frame, and each call to `coordinator.caretRect(...)` (MarkdownSelectionCoordinator.swift:716-757) runs `textView.layoutIfNeeded()` plus glyph-range/bounding-rect queries. A user merely holding a selection pays constant main-thread cost.
**Suggestion:** Gate the link on actual motion — e.g., observe the transcript's pan/scroll (or compare only lightweight window frames) — or run the caret comparison only when the coordinator reports a geometry change, instead of every display-link tick.

### [severity: minor] Handles are adjustable but VoiceOver gets no position feedback
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:420-431
**Issue:** `isAccessibilityElement = true`, `.adjustable`, and `accessibilityIncrement/Decrement` (routing through `moveEndpoint`) are correctly in place, but `accessibilityValue` is never set or updated. A VoiceOver user can swipe up/down to move an endpoint by one character but hears no announcement of the resulting position, making the adjustment effectively blind.
**Suggestion:** Set `accessibilityValue` from the endpoint's current offset (or resolved character/word), refreshing it after each `moveEndpoint` call.

### [severity: minor] Pill copy feedback has an inset mismatch, variable shadowing, and an uncancelled timer
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:345-360
**Issue:** The install path sets `contentInsets` top/bottom to `8` (line 182), while the feedback path (line 348) and restore path (line 357) use `7`, so the pill's height changes by 2pt after the first copy and never returns to the installed size. Line 345 also declares `var copied = pill.configuration`, shadowing the outer `copied` attributed string at line 335. The `DispatchQueue.main.asyncAfter` is never cancelled, so rapid taps schedule overlapping 1.2s resets that can race the config restore.
**Suggestion:** Use one inset value in all three paths, rename the shadowing local, and store/cancel the work item so the last tap wins.

### [severity: minor] UI tests write screenshots to `/tmp` and one test overstates what it guards
**File:** ConduitUITests/SelectionObserverUITests.swift:162-163, 197-198, 209-210; 61-80
**Issue:** Screenshots are still written to absolute `/tmp` paths inside the simulator sandbox — unverifiable on CI (the earlier comments even note screenshot requests hang against a busy app). Separately, `testPlainDragStillScrolls` never verifies a scroll fact: its only assertion is the pasteboard mirror, and `coordinate(withNormalizedOffset: dy: -12)` on a 2pt-tall marker resolves 24pt off the element, so the drag doesn't reach scrollable content regardless.
**Suggestion:** Attach screenshots via `XCTAttachment(screenshot:)` so they land in the result bundle. Rename `testPlainDragStillScrolls` to what it actually guards ("a plain drag must not select/copy"), or add a taller fixture and assert a real scroll fact.

### [severity: minor] Unit-test first-responder assertions can flake
**File:** ConduitTests/MarkdownSelectionCoordinatorTests.swift:918-925 (also 644, 671, 773, 800)
**Issue:** `makeSelectionWindow` sets `isHidden = false` but never calls `makeKeyAndVisible()`; `becomeFirstResponder()` can return `false` for a view in a non-key window, and the call sites discard its result (`_ = ...`) before asserting `isFirstResponder`. On some simulator/CI configurations the dismissal/collapse tests could silently pass without exercising the intended first-responder path.
**Suggestion:** Call `makeKeyAndVisible()` in the helper and prefer `XCTAssertTrue(firstTextView.becomeFirstResponder())` at the call sites.

### [severity: minor] `SelectionFixtureView.init` performs a global, non-restored side effect
**File:** Conduit/Views/SelectionFixtureView.swift:30-34
**Issue:** `UIView.setAnimationsEnabled(false)` runs in `init`, which SwiftUI may invoke multiple times per render pass, is never restored, and disables animations app-wide for the session. The struct `init` is also not `@MainActor`-isolated while calling a `@MainActor` UIKit class method — a latent Swift 6 strict-concurrency error if the project later enables complete checking (same pattern exists for the Coordinator's `textViewDidChangeSelection` calling the `@MainActor` coordinator from a nonisolated `NSObject`).
**Suggestion:** Move the toggle to `.onAppear`/`.task` gated on `CONDUIT_UITEST == "1"` (restoring on disappear), and consider annotating the fixture with `@MainActor`.

### [severity: nit] Handle-drag tail of the chrome test has no assertion
**File:** ConduitUITests/SelectionObserverUITests.swift:207-211
**Issue:** The final focus-handle drag back into the table only sleeps and writes a `/tmp` screenshot — nothing asserts the selection shrank or stayed cross-block.
**Suggestion:** Assert the pasteboard mirror or the handle/endpoint state after the drag, matching the earlier assertions in the same test.

No force unwraps or implicitly unwrapped optionals were introduced; coordinator references remain `weak` on the text views, observer, and container, and the pill action/feedback timer capture `self` weakly. The display-link retain cycle and other previously-reported criticals are resolved.

## Verdict
COMMENT — The architecture is sound and well-tested, and the review-round fixes (release guard, dead-hook removal, `nearest!`, collapse handling, display-link lifecycle) are correct. The remaining items are minor/nits — the at-rest per-frame caret work, missing `accessibilityValue`, and the copy-pill inset/timer hygiene are the most worth follow-up, but none block merge.

<a href="https://opencode.ai/s/XBBdAOG2">New%20session%20-%202026-08-15T11%3A39%3A13.532Z</a>
[opencode session](https://opencode.ai/s/XBBdAOG2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31882537526)

kaishi00:
Review round 4 pushed (b707bca) — everything worth fixing from this round, fixed:

- **Quiet rest (their 3rd flag, now fully closed):** the display link compares the transcript's `contentOffset` first — nearly free — and only recomputes caret geometry (layoutIfNeeded + glyph queries) when content has actually moved. A resting selection now does zero per-frame layout/glyph work.
- **VoiceOver feedback:** adjustments announce the endpoint's new position via `accessibilityValue` after every increment/decrement.
- **Pill feedback:** unified insets (the 2pt height jump after first copy is gone), unshadowed local, cancellable reset (rapid taps can no longer race the restore).
- **Fixture:** the animation disable moved out of `init` (global, unrestored side effect; latent Swift 6 error) to `onAppear` with restore-on-disappear.
- **Tests:** selection-window helper calls `makeKeyAndVisible`, first-responder claims are asserted instead of discarded, screenshots use `XCTAttachment` (verifiable in CI's result bundle) instead of `/tmp`, `testPlainDragStillScrolls` renamed to `testPlainDragDoesNotSelectOrCopy` — the name now matches the guard. The handle-drag tail asserts functionally: after dragging the ending handle up into the table and copying, the text still includes the table but no longer reaches the trailing paragraph — CI's first end-to-end verification that handle drags actually move an endpoint.

kaishi00:
## Summary

Adds cross-block text selection for Markdown responses containing tables/code: a never-recognizing observer gesture recognizer feeds a `@MainActor` coordinator that extends native selection across segments, with coordinator-owned endpoint handles/copy pill, coordinated copy paths, a DEBUG fixture, and unit + UI test coverage. Multiple prior review rounds (dead `menuConfigurationFor` hook, release-build `#if DEBUG` guard, `nearest!` unwrap, CADisplayLink leak, collapse-suppression, test hygiene) are all properly addressed at HEAD.

## Findings

### [severity: minor] The PR's namesake UI test never asserts that the selection extends
**File:** ConduitUITests/SelectionObserverUITests.swift:13-33
**Issue:** `testLongPressAndDragExtendsSelectionAcrossTableAndCodeBlocks` performs the long-press drag and then only asserts the pasteboard is *unchanged* ("No copy should happen without user action"). It never checks a cross-block outcome (no `selection.handle.*`, no `selection.copyPill`, no pasteboard content). If the drag produces no selection at all, the test still passes — so the headline behavior of the PR is only verified indirectly by the other, handle-asserting tests. The name overstates what it guards.
**Suggestion:** After the drag, assert an observable cross-block outcome (e.g. `selection.handle.anchor`/`.focus` or `selection.copyPill` exist), matching `testCellAnchoredLongPressAndDragExtendsAcrossBlocks`, and keep the no-auto-copy pasteboard check as a secondary assertion.

### [severity: minor] VoiceOver adjustable handles report no value until adjusted, and the reported offset is unclamped
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:215-239, 447
**Issue:** Handles correctly set `isAccessibilityElement = true` + `.adjustable` and implement increment/decrement, but `accessibilityValue` is only assigned inside `moveEndpoint` (lines 224/232). On first focus VoiceOver reads "Selection start handle, Adjustable" with no current position. Additionally, `moveEndpoint` never clamps: `offset = anchor.offset + step` is stored directly (the coordinator's `updateAnchorSelection`/`updateSelection` don't clamp either), so repeated increments can push the stored offset past the text length (or below 0 on decrement), and `endpointAccessibilityValue` then announces e.g. "character 60" for a 20-char segment even though rendering clamps correctly — the spoken position diverges from the actual handle position.
**Suggestion:** Initialize `accessibilityValue` when the handle is created/positioned, and clamp the adjusted offset to `0...textLength(for: segmentID)` before reporting (and ideally before storing) so VoiceOver announces the true endpoint.

### [severity: minor] Nonisolated UIKit delegate/gesture callbacks invoke the `@MainActor` coordinator (latent Swift 6 breakage)
**File:** Conduit/Views/Components/SelectableTextView.swift:435-467, 601-605, 617-634
**Issue:** `Coordinator.textViewDidChangeSelection` and `MarkdownSelectionObserverGestureRecognizer.touchesMoved/touchesEnded` (nonisolated `NSObject`/`UIGestureRecognizer` contexts) call `@MainActor`-isolated `MarkdownSelectionCoordinator` methods synchronously. This compiles today under Swift 5.9 minimal checking (`SWIFT_VERSION: "5.9"` in project.yml), but enabling strict/complete concurrency or a Swift 6 toolchain makes these errors. It happens to be safe at runtime because UIKit delivers these on the main thread, but nothing in the types enforces it.
**Suggestion:** Annotate the `Coordinator` and the gesture recognizer (or the coordinator-hopping methods) with `@MainActor` so the isolation is enforced statically, or hop via `Task { @MainActor in ... }` if a nonisolated guarantee is required.

### [severity: nit] `testPlainDragDoesNotSelectOrCopy` drags across no selectable text
**File:** ConduitUITests/SelectionObserverUITests.swift:74-80
**Issue:** `coordinate(withNormalizedOffset:dy: -12)` applies a negative normalized offset to a 2pt-tall marker, resolving the end point ~24pt *above* the element (off the scroll content, at/near the screen top). The drag therefore never crosses a text view, so the test never exercises the observer against selectable content — it effectively verifies "a drag on empty space must not copy," which is much weaker than the #57 regression class its comment claims to guard.
**Suggestion:** Anchor the drag on a real text view (e.g. the "Before the table" paragraph) with an in-range normalized offset plus `withOffset` for a small pixel delta, so the observer genuinely sees a plain drag over text and the before/after pasteboard assertion is meaningful.

### [severity: nit] Handle accessibility labels are hardcoded English while the pill is localized
**File:** Conduit/Views/Components/MarkdownSelectionHandles.swift:447
**Issue:** `accessibilityLabel` uses the literal strings "Selection start handle"/"Selection end handle", whereas the copy pill's title goes through `NSLocalizedString`. VoiceOver in non-English locales reads unlocalized labels.
**Suggestion:** Wrap the labels in `NSLocalizedString` for consistency with the rest of the chrome.

No force unwraps (only standard `fatalError("init(coder:)")`), no `unowned`, and no introduced retain cycles: coordinator references are `weak` on the text views, observer, handle container, and `activeVisibleSelectionCoordinator`; the pill action, feedback work item, and accessibility-adjust closures all capture `self` weakly; the CADisplayLink is torn down in both `willMove(toWindow: nil)` and its own guard-failure path.

## Verdict

APPROVE — the architecture is sound, all previously flagged critical/major issues are resolved at HEAD, and the remaining items are minor/nits (weak name-bearing UI test, accessibility-value initialization/clamping, latent strict-concurrency annotations). None block merge.

<a href="https://opencode.ai/s/btStabd4">New%20session%20-%202026-08-15T12%3A22%3A32.265Z</a>
[opencode session](https://opencode.ai/s/btStabd4)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31884389572)

PR #64, opened 2026-08-15
Prompt the notifier update for pre-0.2 plugins

kaishi00:
## Summary

- The Compatibility section now distinguishes three gateway states: current (capabilities ✓), outdated (reported version, missing capabilities), and **legacy pre-0.2** — sent events but never a plugin version, which is by definition an old notifier since 0.2 stamps every event.
- Legacy gateways previously showed "waiting for the first notification" forever; they now read *"This profile's notifier predates decision cards — update it to receive them"* with the copy-paste `hermes plugins update conduit_push` / `hermes gateway restart` commands. This is the dominant rollout state (new app + updated shared relay + not-yet-updated plugin), so the prompt matters.
- `Gateway.lastEventAt` returns (it was dropped as unused in #62 — the legacy derivation is its use); decode + `isLegacyPlugin` unit-tested.

## Validation

Full suite: 661 passed, 0 failed. Companion relay PR stamps `last_event_at` per accepted event (kaishi00/hermes-conduit-notifier `legacy-plugin-prompt`).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved notification compatibility detection for legacy plugins.
  - Update guidance now appears correctly for older plugins, even when capability information is incomplete.
  - Better distinguishes integrations that have sent events without reporting a plugin version from those that have never sent events.
  - Preserved waiting guidance for inactive integrations without version information.

- **Tests**
  - Added coverage for gateway event timestamps and legacy plugin identification.

### DIFF
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -57,10 +57,11 @@ struct RelayMetaInfo: Decodable, Equatable {
         var supportsApprovalCards: Bool { pluginCapabilities.contains("approval-decisions") }
         var supportsClarifyCards: Bool { pluginCapabilities.contains("clarify-loop") }
 
-        /// A gateway that has sent events but never a plugin version runs a
-        /// pre-0.2 notifier (0.2 stamps every event) and should be prompted to
-        /// update, not shown as "waiting".
-        var isLegacyPlugin: Bool { pluginVersion == nil && lastEventAt != nil }
+        /// A gateway that has sent events but never reported a plugin version
+        /// runs a pre-0.2 notifier — every 0.2+ event carries the version, so
+        /// any version-less event is one. Only this evidence justifies the
+        /// update prompt; a gateway that has sent nothing stays "waiting".
+        var hasSentEventsButNeverReported: Bool { pluginVersion == nil && lastEventAt != nil }
 
         enum CodingKeys: String, CodingKey {
             case id
--- a/Conduit/Services/PushNotificationService.swift
+++ b/Conduit/Services/PushNotificationService.swift
@@ -52,15 +52,22 @@ struct RelayMetaInfo: Decodable, Equatable {
         let name: String
         let pluginVersion: String?
         let pluginCapabilities: [String]
+        let lastEventAt: String?
 
         var supportsApprovalCards: Bool { pluginCapabilities.contains("approval-decisions") }
         var supportsClarifyCards: Bool { pluginCapabilities.contains("clarify-loop") }
+
+        /// A gateway that has sent events but never a plugin version runs a
+        /// pre-0.2 notifier (0.2 stamps every event) and should be prompted to
+        /// update, not shown as "waiting".
+        var isLegacyPlugin: Bool { pluginVersion == nil && lastEventAt != nil }
 
         enum CodingKeys: String, CodingKey {
             case id
             case name
             case pluginVersion = "plugin_version"
             case pluginCapabilities = "plugin_capabilities"
+            case lastEventAt = "last_event_at"
         }
     }
 
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -1380,12 +1380,15 @@ private struct NotificationsSettingsDetail: View {
                                 version: gateway.pluginVersion,
                                 isSupported: gateway.supportsApprovalCards && gateway.supportsClarifyCards,
                                 supportedDetail: "Notifier supports approval and clarify cards",
-                                outdatedDetail: gateway.pluginVersion == nil
-                                    ? "Waiting for the first notification from this profile"
-                                    : "Notifier update available — approval and clarify cards need a newer plugin"
+                                outdatedDetail: gateway.isLegacyPlugin
+                                    ? "This profile's notifier predates decision cards — update it to receive them"
+                                    : gateway.pluginVersion == nil
+                                        ? "Waiting for the first notification from this profile"
+                                        : "Notifier update available — approval and clarify cards need a newer plugin"
                             )
-                            if gateway.pluginVersion != nil,
-                               !gateway.supportsApprovalCards || !gateway.supportsClarifyCards {
+                            // Prompt the update both for a reported-but-old
+                            // plugin and for a never-reporting (pre-0.2) one.
+                            if !gateway.supportsApprovalCards || !gateway.supportsClarifyCards || gateway.isLegacyPlugin {
                                 NotificationSetupCommand(step: 1, title: "Update the notifier", command: "hermes plugins update conduit_push")
                                 NotificationSetupCommand(step: 2, title: "Restart the gateway", command: "hermes gateway restart")
                             }
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -1380,15 +1380,19 @@ private struct NotificationsSettingsDetail: View {
                                 version: gateway.pluginVersion,
                                 isSupported: gateway.supportsApprovalCards && gateway.supportsClarifyCards,
                                 supportedDetail: "Notifier supports approval and clarify cards",
-                                outdatedDetail: gateway.isLegacyPlugin
+                                outdatedDetail: gateway.hasSentEventsButNeverReported
                                     ? "This profile's notifier predates decision cards — update it to receive them"
                                     : gateway.pluginVersion == nil
                                         ? "Waiting for the first notification from this profile"
                                         : "Notifier update available — approval and clarify cards need a newer plugin"
                             )
-                            // Prompt the update both for a reported-but-old
-                            // plugin and for a never-reporting (pre-0.2) one.
-                            if !gateway.supportsApprovalCards || !gateway.supportsClarifyCards || gateway.isLegacyPlugin {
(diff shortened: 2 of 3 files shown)


kaishi00:
## Summary
This PR distinguishes pre-0.2 (legacy) gateway notifiers from merely "never-reported" ones by decoding a new `last_event_at` field, so old plugins get an explicit "update the notifier" prompt instead of an eternal "waiting for the first notification" state.

## Findings

### [severity: minor] Update commands now show for never-reported gateways, contradicting the "waiting" message
**File:** Conduit/Views/AuxiliaryViews.swift:1391
**Issue:** The old guard `if gateway.pluginVersion != nil` intentionally suppressed the "Update the notifier" steps for a gateway that had never reported anything. The new condition `if !gateway.supportsApprovalCards || !gateway.supportsClarifyCards || gateway.isLegacyPlugin` is *always* true for the "waiting" state too (`pluginVersion == nil`, `lastEventAt == nil`, empty capabilities → both `supports…` return false), so a freshly paired 0.2 gateway that simply hasn't sent its first event now shows "Waiting for the first notification from this profile" **plus** "Update the notifier / Restart the gateway" steps — contradictory guidance that tells the user to update a notifier that may already be current. The `|| gateway.isLegacyPlugin` clause is also redundant (legacy implies empty capabilities, hence `!supportsApprovalCards`).
**Suggestion:** Tie the commands to actual evidence of oldness, restoring the never-sent case to no-commands:
```swift
if gateway.isLegacyPlugin
    || (gateway.pluginVersion != nil && (!gateway.supportsApprovalCards || !gateway.supportsClarifyCards))
```

### [severity: minor] Partial fix — pre-0.2 plugins with no post-upgrade event still show "waiting" forever
**File:** Conduit/Services/PushNotificationService.swift:63
**Issue:** `lastEventAt` is only stamped on events accepted *after* the relay upgrade. A pre-0.2 plugin whose last notification predates the relay update (or that goes dormant) keeps `lastEventAt == nil`, so it still falls into the "Waiting for the first notification from this profile" branch — the exact bug this PR intends to fix isn't fully closed for that population. The fix only works when the plugin keeps emitting events (the PR's stated dominant rollout state).
**Suggestion:** Acceptable as a stated limitation, but consider a relay-side backfill (stamp the field at meta-report time from any observed event) or document that the legacy prompt requires one fresh event.

### [severity: minor] No test coverage for the non-legacy "waiting" boundary
**File:** ConduitTests/MessageNormalizerTests.swift:362-367
**Issue:** The removed `XCTAssertFalse(neverReported.supportsApprovalCards)` assertion was the only coverage of the never-reported path, and it was dropped without replacement. There is no test asserting that a version-less gateway with `lastEventAt == nil` is *not* flagged `isLegacyPlugin` (i.e., that it keeps the "waiting" state), which is the exact discriminator this PR adds.
**Suggestion:** Add a third gateway (or reuse the fixture) asserting `XCTAssertFalse(neverReported.isLegacyPlugin)` when `last_event_at` is absent.

### [severity: nit] `isLegacyPlugin` conflates "sent an event" with "is pre-0.2"
**File:** Conduit/Services/PushNotificationService.swift:63
**Issue:** The property is derived from `lastEventAt != nil` and only means "legacy" given the invariant that 0.2 always reports a version per event. That coupling is subtle and is precisely what breaks in the scenarios above.
**Suggestion:** Rename to something explicit like `hasSentEventsButNeverReported` or `preZeroTwoNotifier` and encode the invariant in the doc comment.

### [severity: nit] Triple-nested ternary is hard to read
**File:** Conduit/Views/AuxiliaryViews.swift:1383-1387
**Issue:** The nested `? : ? : ?` for `outdatedDetail` is error-prone at a glance and is part of what made the logic change above easy to introduce.
**Suggestion:** Extract to a small `switch`/computed helper on `Gateway` (e.g. `var compatibilityDetail: String`).

### [severity: nit] `last_event_at` decode could drop whole gateway rows
**File:** Conduit/Services/PushNotificationService.swift:55,90
**Issue:** `lastEventAt` is decoded as a strict `String?`. If the relay ever emits a non-string (e.g., an epoch number), the lossy `try? array.decode(Gateway.self)` silently drops the entire gateway row, hiding it from the UI. Low risk since the companion relay controls the format, but the code already goes to lengths to degrade gracefully.
**Suggestion:** Decode via a tolerant wrapper (accept string or number, else nil) to preserve the row.

No retain cycles, force unwraps, or concurrency issues introduced — the change is pure value-type decoding plus a declarative view tweak, and accessibility (combined element + labeled icons) is preserved.

## Verdict
REQUEST_CHANGES — the core `if` condition in AuxiliaryViews.swift is a behavioral regression for the never-reported "waiting" state (contradictory update guidance) and should be fixed, along with closing the test gap for the `isLegacyPlugin` boundary.

<a href="https://opencode.ai/s/1MnVPVKo">New%20session%20-%202026-08-15T13%3A11%3A09.161Z</a>
[opencode session](https://opencode.ai/s/1MnVPVKo)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31886472448)

kaishi00:
Findings addressed in 94ac7df — focused tests green, boundary covered.

**[minor] contradictory update commands for never-reported gateways** — Fixed: the prompt now requires evidence of oldness — `hasSentEventsButNeverReported` **or** a reported-but-old version. A gateway that has sent nothing keeps only the "waiting" copy (the `|| isLegacyPlugin` clause was indeed redundant; the collapsed capability check was the actual regression).

**[minor] dormant pre-0.2 plugin shows "waiting" until its next event** — Accepted limitation (as the review itself allowed): `last_event_at` only exists post-relay-upgrade. Any event flips the state — including `hermes conduit-push test`, which is one command away in the very same settings screen.

**[minor] boundary test** — Added: a third fixture gateway with `last_event_at` null asserts `hasSentEventsButNeverReported == false`. **[nit] rename** — Done, with the invariant encoded in the doc comment (every 0.2+ event carries the version, so any version-less event implies pre-0.2).

kaishi00:
## Summary
This PR decodes a new `last_event_at` field from the relay's gateway metadata to distinguish pre-0.2 (legacy) notifiers — which sent events but never reported a plugin version — from never-reported gateways, so legacy plugins get an explicit "update the notifier" prompt instead of an eternal "waiting for the first notification" state.

## Findings

### [severity: minor] `last_event_at` strict `String?` decode can silently drop a whole gateway row
**File:** Conduit/Services/PushNotificationService.swift:55,91
**Issue:** `lastEventAt` decodes as strict `String?`. If the relay ever emits a non-string value (e.g., an epoch number), `array.decode(Gateway.self)` fails and the lossy path in the custom `init(from:)` drops the entire gateway row from the UI. Low risk since the companion relay controls the format, but the surrounding code deliberately degrades gracefully and a tolerant decode would preserve that property.
**Suggestion:** Decode tolerantly (accept string or number, else `nil`), or leave as-is and document the format contract on the field.

### [severity: minor] Accessibility label contradicts the "waiting" state
**File:** Conduit/Views/AuxiliaryViews.swift:1483,1493
**Issue:** `compatibilityRow` sets the icon's accessibility label to "Update needed" whenever `isSupported == false`, which is true for the never-sent gateway too, so VoiceOver reads "Update needed … Waiting for the first notification from this profile" — contradictory guidance in the exact state this PR is careful to keep silent. Pre-existing, but the PR touches this area and the legacy/never-sent distinction now makes it more prominent.
**Suggestion:** Pass the a11y label in as a parameter derived from the same state machine, or fall back to "Status unknown" when `pluginVersion == nil && lastEventAt == nil`.

### [severity: nit] Test asserts presence, not value
**File:** ConduitTests/MessageNormalizerTests.swift:373
**Issue:** `XCTAssertNotNil(legacy.lastEventAt)` doesn't pin the decoded value, so a decode that yields the wrong timestamp (or any non-nil garbage) still passes. The field's entire meaning rests on the nil/non-nil boundary, so asserting the exact string would catch regressions in the coding-key mapping.
**Suggestion:** `XCTAssertEqual(legacy.lastEventAt, "2026-08-15T01:00:00Z")`.

### [severity: nit] Triple-nested ternary
**File:** Conduit/Views/AuxiliaryViews.swift:1383-1387
**Issue:** The nested `? : ? : ?` for `outdatedDetail` is hard to read at a glance; the earlier regression in this area came from hand-editing this kind of inline logic.
**Suggestion:** Extract a `var compatibilityDetail: String` (or `switch`) on `Gateway`.

The functional regression raised in the prior review (contradictory update commands for never-reported gateways) is correctly fixed: the prompt now requires evidence of oldness (`hasSentEventsButNeverReported` or a reported-but-old version), the redundant `isLegacyPlugin` clause is gone, and the boundary test (gw-3, `last_event_at: null` → `hasSentEventsButNeverReported == false`) closes the test gap. No retain cycles, force unwraps, or concurrency issues introduced — the change is pure value-type decoding plus a declarative view tweak.

## Verdict
APPROVE — remaining items are nits and a pre-existing a11y inconsistency; the behavioral fix and its test coverage are sound.

<a href="https://opencode.ai/s/XO0y1gpF">New%20session%20-%202026-08-15T13%3A21%3A24.642Z</a>
[opencode session](https://opencode.ai/s/XO0y1gpF)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31886938751)

PR #65, opened 2026-08-15
fix(connection): recover from bridge wedge + defer background reconnect churn (0x8BADF00D)

kaishi00:
## Problem

On a flaky connection (phone tethering drifting in and out of range), the app gets stuck "Reconnecting…" forever, and force-closing + relaunching shows iOS's "app had crashed" notice.

An on-device `.ips` from the crashing run settles the mechanism:

```
0x8BADF00D — scene-update watchdog transgression:
exhausted real (wall clock) time allowance of 10.00 seconds
ProcessVisibility: Background
```

The main thread was inside SwiftUI view-body evaluation when the deadline expired, and the report shows ~45% sustained CPU — reconnect churn saturating the main thread until a backgrounded scene update missed its 10s watchdog window.

Two independent problems combine to produce this. Both are fixed here.

## Fix 1: bridge never recovers from a failed initial page load (`99ef160`)

`DashboardTicketBridge` loads `/api/status` in its `WKWebView` exactly once. If that load fails — launch or foreground during an outage, routine on tethering — `isReady` stays `false` forever:

- `mintTicket()`'s `waitUntilReady()` throws `.notReady` *before any HTTP request is attempted*, so the existing `signInRequired` → `reload()` recovery can never fire.
- `prepareDashboardBridge` reuses the dead bridge on every reconnect (same baseURL), so nothing ever reloads the page.

Result: a permanent reconnect loop that survives full network recovery — the "stuck reconnecting" state. `mintTicket()` now treats `.notReady` like an expired session for retry purposes (reload + re-attempt), and surfaces the actual last error on exhaustion instead of a hardcoded `signInRequired` (a cold bridge must not masquerade as an expired session and force a bogus sign-out).

## Fix 2: reconnect churn while backgrounded trips the scene-update watchdog (`a74eee7`)

Every reconnect cycle mints a ticket, reloads the full session catalog, and mutates a series of `@Published` properties — each driving SwiftUI transactions on the main thread. On a flapping link this cycles every few seconds, and backgrounding mid-storm misses the 10s scene-update deadline → `0x8BADF00D` kill → the crash notice users see on relaunch.

`executeReconnect()` now drops the attempt while the scene is not active. No recovery is lost: `handleScenePhase(.active)` already re-establishes the transport (health check + `reconnectForRetry(.automaticReturn)`) on return to foreground.

## Verification

Reproduced against a local chaos gateway (mock Hermes dashboard + WS gateway with outage windows, mid-body kills, resets, hangs, 401s, WS churn; app driven in the iOS 26.5 simulator):

- **Before:** app launched into a 25s outage made **zero** requests after the outage ended — permanent "Failed to refresh the dashboard session: The dashboard session is still loading." banner, spinner forever (screenshots in the linked investigation).
- **After (this branch):** ticket minted and socket reopened **~5s** after the outage ended; session list and transcript fully restored.
- Extended fuzzing (~30 min, Debug + Release, WS churn with live-stream reconciliation, 401 storms, background/foreground cycles) found no other hard crash in the session-loading path.
- Unit suite: **663 tests, 0 failures**, including two new regression tests (`testBackgroundedSceneDropsScheduledReconnectExecution`, `testSceneActivationRestoresReconnectExecution`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection recovery when the app is inactive or running in the background.
  * Reconnect attempts now pause and cancel safely in the background, then resume when the app returns to the foreground.
  * Added retry handling for dashboard sign-in and readiness failures, with clearer final error reporting.

* **Tests**
  * Added coverage for reconnect behavior across app lifecycle transitions.
  * Added coverage for dashboard loading retries and readiness failures.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -3445,7 +3445,9 @@ final class AppState: ObservableObject {
         // A cycle scheduled while the scene is inactive can never run, so
         // don't arm a timer or consume the queued reconnect purpose just to
         // discard them when it fires. handleScenePhase(.active) establishes
-        // the transport on return instead.
+        // the transport on return instead — and intentionally recovers with
+        // .automaticReturn, upgrading the drop-time purpose, since resuming
+        // the saved session on foreground is the expected outcome.
         guard isSceneActive else { return }
         if reconnectTask == nil {
             recoverySequence.clearQueuedReconnect()
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -1202,10 +1202,12 @@ final class AppState: ObservableObject {
         automaticWorkToken: ChatResumeAutomaticWorkToken?,
         automaticReconnectOperationID: UUID?
     ) -> ChatResumeTransportContinuation? {
-        // Checked after every suspension in connect(with:)/reconnectForRetry.
-        // A scene that went inactive/backgrounded mid-flight must not keep
-        // driving reconnect churn (watchdog: 0x8BADF00D); handleScenePhase
-        // (.active) re-establishes the transport on return.
+        // Checked after every suspension in connect(with:),
+        // reconnectForRetry, and the post-connect sync flow — the gate is
+        // transport-wide, not reconnect-only: any chat-resume work that goes
+        // inactive/backgrounded mid-flight must not keep publishing state
+        // (watchdog: 0x8BADF00D). handleScenePhase(.active) re-establishes
+        // the transport and syncs the session catalog on return.
         guard !Task.isCancelled, isSceneActive else { return nil }
         if let automaticReconnectOperationID,
            activeAutomaticReconnectOperation?.id != automaticReconnectOperationID {
@@ -3454,21 +3456,20 @@ final class AppState: ObservableObject {
             break
         }
         let delay = immediately ? 0.1 : min(5.0, pow(2.0, Double(reconnectAttempts)))
-        if !immediately { reconnectAttempts += 1 }
+        // The backoff step is consumed only when the cycle actually runs.
+        // A timer canceled by scene backgrounding — or a cycle dropped for
+        // scene inactivity before execution — never counts as a gateway
+        // failure, so background/foreground cycling cannot ratchet the
+        // retry delay toward its cap without a real failure.
+        let incrementsBackoff = !immediately
 
         reconnectTask = reconnectScheduler(delay) { [weak self] in
             guard let self else { return }
             self.reconnectTask = nil
             let purpose = self.recoverySequence.takeQueuedReconnectPurpose()
                 ?? .preserveCurrent
-            guard self.isSceneActive else {
-                // Dropped for scene inactivity, not for a gateway failure —
-                // undo the backoff step this schedule took so repeated
-                // background/foreground cycling cannot ratchet the retry
-                // delay toward its cap without a real failure.
-                self.reconnectAttempts = max(0, self.reconnectAttempts - 1)
-                return
-            }
+            guard self.isSceneActive else { return }
+            if incrementsBackoff { self.reconnectAttempts += 1 }
             await self.executeReconnect(purpose: purpose)
         }
     }
@@ -3480,11 +3481,13 @@ final class AppState: ObservableObject {
 
     private func executeReconnect(purpose: ChatResumeSyncPurpose) async {
         // A reconnect cycle mints a ticket, reloads the session catalog, and
-        // mutates a handful of @Published properties — each driving SwiftUI
+        // mutates a series of @Published properties — each driving SwiftUI
         // transactions on the main thread. On a flaky link that churn can
         // saturate the main thread, and a backgrounded scene update then
         // misses its 10s watchdog deadline. Drop the attempt here instead;
         // handleScenePhase(.active) re-establishes the transport on return.
+        // This also makes the public reconnect() a no-op while the scene is
+        // inactive/backgrounded — the retry is picked up on the next .active.
         guard isSceneActive else { return }
         if let reconnectExecutor {
             await reconnectExecutor(purpose)
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -646,6 +646,10 @@ final class AppState: ObservableObject {
     private let sessionRenameOperationsOverride: SessionRenameOperation.Operations?
     private let sessionCatalogLoaderOverride: ((Bool) async throws -> [SessionSummary])?
     private var reconnectAttempts = 0
+    /// Whether the UI scene is active. Backgrounded scene updates must
+    /// complete within ~10s of wall clock before the watchdog kills the app
+    /// (0x8BADF00D), so reconnect work is deferred while this is false.
+    private var isSceneActive = true
     private var connectedAt: Date?
     private var sessionCatalogCache = SessionCatalogCache()
     private var projectsRequestGeneration = 0
@@ -3458,6 +3462,13 @@ final class AppState: ObservableObject {
     }
 
     private func executeReconnect(purpose: ChatResumeSyncPurpose) async {
+        // A reconnect cycle mints a ticket, reloads the session catalog, and
+        // mutates a handful of @Published properties — each driving SwiftUI
+        // transactions on the main thread. On a flaky link that churn can
+        // saturate the main thread, and a backgrounded scene update then
+        // misses its 10s watchdog deadline. Drop the attempt here instead;
+        // handleScenePhase(.active) re-establishes the transport on return.
+        guard isSceneActive else { return }
         if let reconnectExecutor {
             await reconnectExecutor(purpose)
         } else {
@@ -3661,6 +3672,7 @@ final class AppState: ObservableObject {
         }
         switch phase {
         case .active:
+            isSceneActive = true
             voiceConversationController.setForegroundActive(true)
             guard connection != nil else { return nil }
             cancelScenePhaseAttempt()
@@ -3711,6 +3723,7 @@ final class AppState: ObservableObject {
             return task
 
         case .background:
+            isSceneActive = false
             voiceConversationController.setForegroundActive(false)
             showVoiceSheet = false
             // Flush any pending coalesced cache writes before the app
@@ -3723,6 +3736,7 @@ final class AppState: ObservableObject {
             return nil
 
         case .inactive:
+            isSceneActive = false
             chatResumeCoordinator.freezeViewport()
             voiceConversationController.setForegroundActive(false)
             return nil
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -3780,6 +3780,10 @@ final class AppState: ObservableObject {
             // center) is recovered by the .active scene task — the same
             // moment the user can see the transcript again.
             cancelScheduledReconnect()
+            // The scene treats .inactive like .background for reconnect
+            // purposes; formally abort the in-flight scene attempt at the
+            // transition too, rather than at its next checkpoint.
+            cancelScenePhaseAttempt()
             chatResumeCoordinator.freezeViewport()
             voiceConversationController.setForegroundActive(false)
             return nil
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -649,6 +649,11 @@ final class AppState: ObservableObject {
     /// Whether the UI scene is active. Backgrounded scene updates must
     /// complete within ~10s of wall clock before the watchdog kills the app
     /// (0x8BADF00D), so reconnect work is deferred while this is false.
+    /// Deliberately true at init: launches head toward active, and blocking
+    /// the cold-start restore on the first scene-phase event would regress
+    /// startup. `.inactive` is treated like `.background` on purpose — it
+    /// immediately precedes backgrounding on home-press, and a socket that
+    /// dies under a system overlay is recovered by the `.active` scene task.
     private var isSceneActive = true
     private var connectedAt: Date?
     private var sessionCatalogCache = SessionCatalogCache()
@@ -3456,6 +3461,14 @@ final class AppState: ObservableObject {
             self.reconnectTask = nil
             let purpose = self.recoverySequence.takeQueuedReconnectPurpose()
                 ?? .preserveCurrent
+            guard self.isSceneActive else {
+                // Dropped for scene inactivity, not for a gateway failure —
+                // undo the backoff step this schedule took so repeated
+                // background/foreground cycling cannot ratchet the retry
+                // delay toward its cap without a real failure.
+                self.reconnectAttempts = max(0, self.reconnectAttempts - 1)
+                return
+            }
             await self.executeReconnect(purpose: purpose)
         }
     }
@@ -3745,6 +3758,12 @@ final class AppState: ObservableObject {
 
         case .inactive:
             isSceneActive = false
+            // Reconnects are suppressed during .inactive too, so an armed
+            // timer would only fire to be discarded. Drop it here; a socket
+            // that dies under a system overlay (incoming call, control
+            // center) is recovered by the .active scene task — the same
+            // moment the user can see the transcript again.
+            cancelScheduledReconnect()
             chatResumeCoordinator.freezeViewport()
             voiceConversationController.setForegroundActive(false)
             return nil
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -1197,7 +1197,11 @@ final class AppState: ObservableObject {
         automaticWorkToken: ChatResumeAutomaticWorkToken?,
         automaticReconnectOperationID: UUID?
     ) -> ChatResumeTransportContinuation? {
-        guard !Task.isCancelled else { return nil }
+        // Checked after every suspension in connect(with:)/reconnectForRetry.
+        // A scene that went inactive/backgrounded mid-flight must not keep
+        // driving reconnect churn (watchdog: 0x8BADF00D); handleScenePhase
+        // (.active) re-establishes the transport on return.
+        guard !Task.isCancelled, isSceneActive else { return nil }
         if let automaticReconnectOperationID,
            activeAutomaticReconnectOperation?.id != automaticReconnectOperationID {
             return nil
@@ -3726,6 +3730,10 @@ final class AppState: ObservableObject {
             isSceneActive = false
             voiceConversationController.setForegroundActive(false)
             showVoiceSheet = false
+            // Drop any armed reconnect timer as well: in-flight cycles abort
+            // at their next transportContinuation checkpoint, and foreground
+            // activation re-establishes the transport.
+            cancelScheduledReconnect()
             // Flush any pending coalesced cache writes before the app
             // suspends — iOS may kill the process before the debounce fires.
             flushPendingPresentationCache()
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -3442,6 +3442,11 @@ final class AppState: ObservableObject {
         purpose: ChatResumeSyncPurpose = .preserveCurrent
     ) {
         guard connection != nil else { return }
+        // A cycle scheduled while the scene is inactive can never run, so
+        // don't arm a timer or consume the queued reconnect purpose just to
+        // discard them when it fires. handleScenePhase(.active) establishes
+        // the transport on return instead.
+        guard isSceneActive else { return }
         if reconnectTask == nil {
             recoverySequence.clearQueuedReconnect()
         }
@@ -3708,6 +3713,14 @@ final class AppState: ObservableObject {
                 defer { self.finishScenePhaseAttempt(id: sceneAttemptID) }
                 guard self.scenePhaseAttemptIsCurrent(sceneAttemptID) else { return }
                 if let client = self.client, client.isConnected {
+                    // A connect that aborted mid-flight (scene went inactive
+                    // before its checkpoint) can leave the UI's transport
+                    // flags unpublished — "connecting…" with sends disabled —
+                    // even though the socket is alive. Publish the healthy
+                    // transport before syncing; the reconnect path manages
+                    // these flags for unhealthy sockets.
+                    self.isConnected = true
+                    self.isConnecting = false
                     do {
                         try await client.healthCheck()
                         guard self.scenePhaseAttemptIsCurrent(sceneAttemptID),
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -249,20 +249,31 @@ final class DashboardTicketBridge: NSObject {
         // A freshly restored session cookie can reach WebKit's cookie store a
         // moment before its network process. A first 401 is therefore not
         // enough evidence to erase the durable session and force a login.
+        var lastError = DashboardTicketBridgeError.signInRequired
         for attempt in 0..<3 {
-            try await waitUntilReady()
             do {
+                try await waitUntilReady()
                 let response = try await requestJSON(path: "/api/auth/ws-ticket", method: "POST")
                 guard let ticket = response["ticket"] as? String, !ticket.isEmpty else {
                     throw DashboardTicketBridgeError.requestFailed("Dashboard did not return a WebSocket ticket.")
                 }
                 return ticket
             } catch DashboardTicketBridgeError.signInRequired where attempt < 2 {
+                lastError = .signInRequired
+                reload()
+                try await Task.sleep(for: .milliseconds(350))
+            } catch DashboardTicketBridgeError.notReady where attempt < 2 {
+                // A failed initial page load (e.g. launch during a network
+                // outage) leaves the bridge cold forever: waitUntilReady()
+                // never sees isReady and no request is ever attempted, so
+                // nothing else triggers a reload. Re-attempt the dashboard
+                // session instead of wedging every future reconnect.
+                lastError = .notReady
                 reload()
                 try await Task.sleep(for: .milliseconds(350))
             }
         }
-        throw DashboardTicketBridgeError.signInRequired
+        throw lastError
     }
 
     private func waitUntilReady() async throws {
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -191,6 +191,9 @@ final class DashboardTicketBridge: NSObject {
     /// session is genuinely absent, which callers must hear as
     /// `.signInRequired` rather than as unreadiness.
     private var didLandOnLogin = false
+    /// Test-only: keep re-asserting the simulated login landing on every
+    /// page load, modeling a dashboard that keeps redirecting to /login.
+    private var simulatedLoginLanding = false
     private var isInvalidated = false
     private var requestID = 0
     private let pendingRequests: DashboardTicketBridgePendingRequests
@@ -440,12 +443,12 @@ final class DashboardTicketBridge: NSObject {
         return literal
     }
 
-    /// Test hook: put the bridge into the state a /login landing produces.
-    /// WKWebView's URL is only settable by real navigation, which the unit
-    /// test host cannot drive deterministically. The simulated landing is
-    /// cleared by the next `loadDashboardSession()` exactly as in
-    /// production.
+    /// Test hook: put the bridge into the state a /login landing produces,
+    /// and keep producing it across reloads — modeling a dashboard whose
+    /// session is genuinely gone. WKWebView's URL is only settable by real
+    /// navigation, which the unit test host cannot drive deterministically.
     func simulateLoginLandingForTesting() {
+        simulatedLoginLanding = true
         isReady = false
         isLoadFailed = false
         didLandOnLogin = true
@@ -458,8 +461,9 @@ final class DashboardTicketBridge: NSObject {
         request = cloudflareAccess?.applying(to: request) ?? request
         isLoadFailed = false
         // The fresh load's landing is unknown until didFinish; a stale
-        // sign-in-required verdict must not leak into its poll window.
-        didLandOnLogin = false
+        // sign-in-required verdict must not leak into its poll window. (The
+        // test simulation re-asserts its landing here on purpose.)
+        didLandOnLogin = simulatedLoginLanding
         webView.load(request)
     }
 
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -181,6 +181,12 @@ final class DashboardTicketBridge: NSObject {
     let cloudflareAccess: CloudflareAccessCredentials?
 
     private var isReady = false
+    /// Whether the current dashboard page load has terminally failed (as
+    /// opposed to still being in flight on a slow link). Retry logic only
+    /// reloads a failed load — restarting an in-progress one would abort a
+    /// load that may be about to finish on a degraded connection.
+    /// Readable for tests; written only by the navigation callbacks.
+    private(set) var isLoadFailed = false
     private var isInvalidated = false
     private var requestID = 0
     private let pendingRequests: DashboardTicketBridgePendingRequests
@@ -279,7 +285,15 @@ final class DashboardTicketBridge: NSObject {
                 return ticket
             } catch DashboardTicketBridgeError.signInRequired where attempt < 2,
                   DashboardTicketBridgeError.notReady where attempt < 2 {
-                reload()
+                // Only restart a page load that terminally failed. On a
+                // degraded link a load can still be in flight when the
+                // readiness poll times out; tearing it down would restart a
+                // navigation that may be about to finish, so keep polling the
+                // same load instead. A silently hung load is eventually
+                // failed by the system request timeout and becomes reloadable.
+                if isLoadFailed {
+                    reload()
+                }
                 try await Task.sleep(for: .milliseconds(350))
             }
         }
@@ -415,6 +429,7 @@ final class DashboardTicketBridge: NSObject {
               let url = URL(string: "\(normalized)/api/status") else { return }
         var request = URLRequest(url: url)
         request = cloudflareAccess?.applying(to: request) ?? request
+        isLoadFailed = false
         webView.load(request)
     }
 
@@ -446,6 +461,7 @@ extension DashboardTicketBridge: WKNavigationDelegate {
     }
 
     func webView(_ webView: WKWebView, didFinish navigation: WKNavigation!) {
+        isLoadFailed = false
         // A redirect back to /login means the HttpOnly dashboard session is no
         // longer valid; never attempt to mint a misleading gateway ticket.
         guard let expectedURL = URL(string: baseURL),
@@ -469,11 +485,13 @@ extension DashboardTicketBridge: WKNavigationDelegate {
 
     func webView(_ webView: WKWebView, didFail navigation: WKNavigation!, withError error: Error) {
         isReady = false
+        isLoadFailed = true
         rejectPending(with: error)
     }
 
     func webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error) {
         isReady = false
+        isLoadFailed = true
         rejectPending(with: error)
     }
 }
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -191,9 +191,15 @@ final class DashboardTicketBridge: NSObject {
     /// session is genuinely absent, which callers must hear as
     /// `.signInRequired` rather than as unreadiness.
     private var didLandOnLogin = false
-    /// Test-only: keep re-asserting the simulated login landing on every
-    /// page load, modeling a dashboard that keeps redirecting to /login.
-    private var simulatedLoginLanding = false
+    /// Test-only landing state, re-asserted by every page load so tests can
+    /// model a dashboard that keeps producing the same landing regardless of
+    /// when WKWebView's nondeterministic callbacks arrive.
+    private var simulatedLanding: SimulatedLanding?
+
+    enum SimulatedLanding {
+        case loginPage
+        case loadFailure
+    }
     private var isInvalidated = false
     private var requestID = 0
     private let pendingRequests: DashboardTicketBridgePendingRequests
@@ -443,15 +449,23 @@ final class DashboardTicketBridge: NSObject {
         return literal
     }
 
-    /// Test hook: put the bridge into the state a /login landing produces,
-    /// and keep producing it across reloads — modeling a dashboard whose
-    /// session is genuinely gone. WKWebView's URL is only settable by real
-    /// navigation, which the unit test host cannot drive deterministically.
-    func simulateLoginLandingForTesting() {
-        simulatedLoginLanding = true
-        isReady = false
-        isLoadFailed = false
-        didLandOnLogin = true
+    /// Test hook: put the bridge into the state a given landing produces
+    /// and keep producing it across reloads (login page or terminally
+    /// failed load). WKWebView's URL and failure callbacks are not
+    /// deterministically drivable in the unit test host, so tests model the
+    /// landing state instead.
+    func simulateLandingForTesting(_ landing: SimulatedLanding) {
+        simulatedLanding = landing
+        switch landing {
+        case .loginPage:
+            isReady = false
+            isLoadFailed = false
+            didLandOnLogin = true
+        case .loadFailure:
+            isReady = false
+            isLoadFailed = true
+            didLandOnLogin = false
+        }
     }
 
     private func loadDashboardSession() {
@@ -459,11 +473,11 @@ final class DashboardTicketBridge: NSObject {
               let url = URL(string: "\(normalized)/api/status") else { return }
         var request = URLRequest(url: url)
         request = cloudflareAccess?.applying(to: request) ?? request
-        isLoadFailed = false
-        // The fresh load's landing is unknown until didFinish; a stale
-        // sign-in-required verdict must not leak into its poll window. (The
-        // test simulation re-asserts its landing here on purpose.)
-        didLandOnLogin = simulatedLoginLanding
+        // The fresh load's landing is unknown until its navigation callback;
+        // a stale verdict must not leak into its poll window. (The test
+        // simulation re-asserts its landing here on purpose.)
+        isLoadFailed = simulatedLanding == .loadFailure
+        didLandOnLogin = simulatedLanding == .loginPage
         webView.load(request)
     }
 
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -184,16 +184,25 @@ final class DashboardTicketBridge: NSObject {
     private var isInvalidated = false
     private var requestID = 0
     private let pendingRequests: DashboardTicketBridgePendingRequests
+    private let readinessPollAttempts: Int
+    private let readinessPollInterval: Duration
+    /// Number of times `reload()` re-attempted the dashboard page load.
+    /// Test visibility for the cold-bridge recovery path.
+    private(set) var reloadCount = 0
 
     init(
         baseURL: String,
         cloudflareAccess: CloudflareAccessCredentials? = nil,
-        pendingRequests: DashboardTicketBridgePendingRequests = DashboardTicketBridgePendingRequests()
+        pendingRequests: DashboardTicketBridgePendingRequests = DashboardTicketBridgePendingRequests(),
+        readinessPollAttempts: Int = 30,
+        readinessPollInterval: Duration = .milliseconds(100)
     ) {
         let normalizedBaseURL = (try? ConnectionURLPolicy.normalizedBaseURL(baseURL)) ?? ""
         self.baseURL = normalizedBaseURL
         self.cloudflareAccess = cloudflareAccess
         self.pendingRequests = pendingRequests
+        self.readinessPollAttempts = readinessPollAttempts
+        self.readinessPollInterval = readinessPollInterval
         let configuration = WKWebViewConfiguration()
         configuration.websiteDataStore = .default()
         if let script = cloudflareAccess?.fetchInjectionUserScript(expectedBaseURL: normalizedBaseURL), !script.isEmpty {
@@ -231,6 +240,7 @@ final class DashboardTicketBridge: NSObject {
 
     func reload() {
         guard !isInvalidated else { return }
+        reloadCount += 1
         isReady = false
         pendingRequests.rejectAll(with: DashboardTicketBridgeError.notReady)
         Task { [weak self] in
@@ -246,11 +256,19 @@ final class DashboardTicketBridge: NSObject {
     }
 
     func mintTicket() async throws -> String {
-        // A freshly restored session cookie can reach WebKit's cookie store a
-        // moment before its network process. A first 401 is therefore not
-        // enough evidence to erase the durable session and force a login.
-        var lastError = DashboardTicketBridgeError.signInRequired
-        for attempt in 0..<3 {
+        // Retry twice on the two recoverable failures, then let the final
+        // attempt's error propagate to the caller unchanged:
+        //  - signInRequired: a freshly restored session cookie can reach
+        //    WebKit's cookie store a moment before its network process, so a
+        //    first 401 is not enough evidence to erase the durable session
+        //    and force a login;
+        //  - notReady: a failed initial page load (e.g. launch during a
+        //    network outage) leaves the bridge cold forever — waitUntilReady()
+        //    never sees isReady and no request is ever attempted, so nothing
+        //    else triggers a reload. Re-attempt the dashboard session instead
+        //    of wedging every future reconnect.
+        var attempt = 0
+        while true {
             do {
                 try await waitUntilReady()
                 let response = try await requestJSON(path: "/api/auth/ws-ticket", method: "POST")
@@ -258,27 +276,18 @@ final class DashboardTicketBridge: NSObject {
                     throw DashboardTicketBridgeError.requestFailed("Dashboard did not return a WebSocket ticket.")
                 }
                 return ticket
-            } catch DashboardTicketBridgeError.signInRequired where attempt < 2 {
-                lastError = .signInRequired
-                reload()
-                try await Task.sleep(for: .milliseconds(350))
-            } catch DashboardTicketBridgeError.notReady where attempt < 2 {
-                // A failed initial page load (e.g. launch during a network
-                // outage) leaves the bridge cold forever: waitUntilReady()
-                // never sees isReady and no request is ever attempted, so
-                // nothing else triggers a reload. Re-attempt the dashboard
-                // session instead of wedging every future reconnect.
-                lastError = .notReady
+            } catch DashboardTicketBridgeError.signInRequired where attempt < 2,
+                  DashboardTicketBridgeError.notReady where attempt < 2 {
+                attempt += 1
                 reload()
                 try await Task.sleep(for: .milliseconds(350))
             }
         }
-        throw lastError
     }
 
     private func waitUntilReady() async throws {
-        for _ in 0..<30 where !isReady && !isInvalidated {
-            try await Task.sleep(for: .milliseconds(100))
+        for _ in 0..<readinessPollAttempts where !isReady && !isInvalidated {
+            try await Task.sleep(for: readinessPollInterval)
         }
         guard !isInvalidated, isReady else { throw DashboardTicketBridgeError.notReady }
     }
@@ -293,8 +302,8 @@ final class DashboardTicketBridge: NSObject {
         timeoutMilliseconds: Int = 12_000,
         maxResponseBytes: Int = DataURLLimits.maxJSONResponseBytes
     ) async throws -> [String: Any] {
-        for _ in 0..<30 where !isReady && !isInvalidated {
-            try await Task.sleep(for: .milliseconds(100))
+        for _ in 0..<readinessPollAttempts where !isReady && !isInvalidated {
+            try await Task.sleep(for: readinessPollInterval)
         }
         guard !isInvalidated, isReady else { throw DashboardTicketBridgeError.notReady }
 
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -333,7 +333,13 @@ final class DashboardTicketBridge: NSObject {
         for _ in 0..<readinessPollAttempts where !isReady && !isInvalidated {
             try await Task.sleep(for: readinessPollInterval)
         }
-        guard !isInvalidated, isReady else { throw DashboardTicketBridgeError.notReady }
+        guard !isInvalidated, isReady else {
+            // Mirror waitUntilReady(): a login-parked bridge is signed out,
+            // not loading, for non-mint callers too.
+            throw didLandOnLogin
+                ? DashboardTicketBridgeError.signInRequired
+                : DashboardTicketBridgeError.notReady
+        }
 
         requestID += 1
         let id = requestID
@@ -434,6 +440,17 @@ final class DashboardTicketBridge: NSObject {
         return literal
     }
 
+    /// Test hook: put the bridge into the state a /login landing produces.
+    /// WKWebView's URL is only settable by real navigation, which the unit
+    /// test host cannot drive deterministically. The simulated landing is
+    /// cleared by the next `loadDashboardSession()` exactly as in
+    /// production.
+    func simulateLoginLandingForTesting() {
+        isReady = false
+        isLoadFailed = false
+        didLandOnLogin = true
+    }
+
     private func loadDashboardSession() {
         guard let normalized = try? ConnectionURLPolicy.normalizedBaseURL(baseURL),
               let url = URL(string: "\(normalized)/api/status") else { return }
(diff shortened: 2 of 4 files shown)


kaishi00:
## Summary
Fixes two connection-recovery bugs: (1) `DashboardTicketBridge.mintTicket` now reloads and retries when the bridge never becomes ready (previously a failed initial page load wedged reconnect forever), and (2) `AppState.executeReconnect` drops reconnect cycles while the scene is backgrounded to avoid watchdog (0x8BADF00D) starvation, relying on `handleScenePhase(.active)` to restore the transport.

## Findings

### [minor] `lastError` is dead code in `mintTicket`
**File:** Conduit/Services/DashboardTicketBridge.swift:252-276
**Issue:** The `throw lastError` at line 276 is unreachable. Each loop iteration either `return`s a ticket or throws. On the final iteration (`attempt == 2`), the `where attempt < 2` guard fails on both catch clauses, so whatever error the `do` block throws (`.notReady`, `.signInRequired`, `.requestFailed`, or `CancellationError`) propagates directly out of `mintTicket` — the loop never falls through to the `throw lastError`. The two `lastError = ...` assignments are dead stores, and the variable misleads readers into thinking there's a fallback error path.
**Suggestion:** Either remove `lastError` and rely on the final iteration's propagated error (adding a comment that the last attempt's error intentionally surfaces), or restructure so all three attempts are caught — e.g. `catch let error as DashboardTicketBridgeError where attempt < 2` and `throw lastError` at the end — so the stated intent ("surface the actual last error on exhaustion") is actually expressed in the control flow.

### [minor] No test coverage for Fix 1
**File:** ConduitTests/AppStateChatResumeTests.swift:2329
**Issue:** Both new regression tests cover only Fix 2 (scene gating in `AppState`). The `mintTicket` change — the subtler of the two, with the new `.notReady` → `reload()` retry path and the `lastError` control flow — has no direct unit test. A future refactor could silently reintroduce the permanent-wedge behavior without any test catching it.
**Suggestion:** Add a `DashboardTicketBridge` test that drives a cold bridge (never-ready) through `mintTicket` and asserts it reloads and returns once readiness arrives, and that exhaustion surfaces `.notReady` rather than `.signInRequired`.

### [minor] Guard only gates *new* reconnect cycles, not in-flight churn
**File:** Conduit/Services/AppState.swift:3471
**Issue:** The `guard isSceneActive` prevents *starting* a reconnect while backgrounded, but a cycle already past the guard (the scene transitions to `.inactive`/`.background` mid-cycle, e.g. a call arrives or home button pressed mid-ticket-mint) continues mutating `@Published` properties and driving SwiftUI transactions. The very churn the fix targets (main-thread saturation missing the scene-update watchdog deadline) can still occur if the phase transition lands mid-cycle. `cancelScenePhaseAttempt()` on `.background` only cancels owned automatic operations, not the `reconnectForRetry` flow already in progress.
**Suggestion:** Check `isSceneActive` at suspension points inside the reconnect flow (e.g. before the catalog reload in `reconnectForRetry`) and bail early, or add a phase-transition observation that aborts in-progress recovery.

### [nit] `isSceneActive` defaults to `true`
**File:** Conduit/Services/AppState.swift:650
**Issue:** On a cold launch into the background (push notification / background fetch / VoIP-style wake), `isSceneActive` stays `true` until SwiftUI first delivers a non-active scene phase. A reconnect scheduled in that window would still churn while backgrounded. The window is small and self-corrects, so this is low risk, but the invariant ("never churn while backgrounded") isn't absolute.
**Suggestion:** If the class can observe the current `ScenePhase` (or a view model that already does), seed `isSceneActive` from it at init rather than hardcoding `true`.

### [nit] `testSceneActivationRestoresReconnectExecution` determinism relies on an implicit failure path
**File:** ConduitTests/AppStateChatResumeTests.swift:2365-2384
**Issue:** After `handleScenePhase(.active)`, the scene task runs an unstructured `reconnectForRetry(.automaticReturn)` concurrently. The test's `XCTAssertEqual(reconnectSpy.purposes, [.preserveCurrent])` only holds because the injected `mintTicket` throws `.notReady`, which sends the scene task down the error branch (`scheduleReconnect`) instead of reaching the executor. That dependency is implicit and easy to break if the harness mintTicket is ever changed.
**Suggestion:** Add a brief comment documenting why the scene task cannot reach the executor spy, or assert on `scheduler.scheduledCount`/cancellations to make the isolation explicit.

## Verdict
COMMENT

The two fixes are sound and target the reported root causes correctly. Main-thread churn is genuinely deferred, the bridge-wedge recovery mechanism is right, and the code is `@MainActor`-consistent (no data races, no new retain cycles — `reload()` and the scheduler closure both capture `self` weakly). No critical or major issues; the findings above are cleanup and hardening items.

<a href="https://opencode.ai/s/5x2useGX">New%20session%20-%202026-08-15T21%3A14%3A40.797Z</a>
[opencode session](https://opencode.ai/s/5x2useGX)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31908840321)

kaishi00:
Addressed all three findings in a04c8a4 — all three were worth taking:

**1. In-flight reconnect churn on background (both bots)** — agreed, the entry guard only covered cycles that hadn't started. `transportContinuation()` is the checkpoint after every suspension in `connect(with:)`/`reconnectForRetry`, so it now also requires `isSceneActive`: a cycle suspended mid-ticket-mint aborts at its next checkpoint when the scene goes inactive/backgrounded, and `handleScenePhase(.background)` additionally drops any armed reconnect timer. This is a real user path (the crashing run was exactly a mid-storm backgrounding — the .ips shows 45% CPU churn with the main thread in SwiftUI body eval when the 10s scene-update deadline expired). Regression test added in the suggested shape: `testBackgroundingMidMintAbortsInFlightReconnect` suspends ticket minting, backgrounds the scene, and asserts the cycle never reaches `connectClient`.

**2. Dead `lastError` in `mintTicket`** — correct, the final iteration's `where attempt < 2` fails and the error propagates directly, so the assignments were dead stores and the trailing throw was unreachable. Restructured to a `while true` with the two retryable catch patterns gated on `attempt < 2`; the control flow now states the contract directly: retry twice, then surface whatever the final attempt threw (exhausted `.notReady` must never masquerade as `.signInRequired` and force a sign-out).

**3. No direct test for the mintTicket fix** — added `testColdBridgeMintTicketRetriesReloadAndSurfacesNotReady`, which drives a real cold bridge (unreachable dashboard origin) through `mintTicket` with injectable short readiness polling and asserts (a) exhaustion surfaces `.notReady` and (b) both retries actually re-attempted the page load (`reloadCount == 2`). The full-recovery variant (readiness arriving mid-retry) isn't unit-testable without a live dashboard; it's covered by the chaos-gateway verification described in the PR description (app launched into a 25s outage re-mints ~5s after recovery).

Unit suite: 665 tests, 0 failures. Note the earlier CI failure (`testCopyAfterDragCapturesCrossBlockText`, pasteboard:empty) is the known-flaky selection/pasteboard UI-test family — the same suite also failed on main's last two runs on different tests in that family (`testCellAnchoredLongPressAndDragExtendsAcrossBlocks` on #64's run); unrelated to this branch's changes.

kaishi00:
## Summary
Fixes two connection-recovery bugs: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady` (previously a failed initial page load wedged reconnect forever) and surfaces the final error instead of a hardcoded `.signInRequired`; `AppState` gates reconnect churn on scene activity to avoid the 0x8BADF00D background watchdog kill, aborting in-flight cycles at suspension checkpoints and cancelling scheduled reconnects on background. Includes 4 new regression tests.

## Findings

### [minor] `isSceneActive` defaults to `true` — cold-launch-into-background window
**File:** Conduit/Services/AppState.swift:652
**Issue:** On a launch into the background (push / background-fetch / VoIP-style wake), `isSceneActive` stays `true` until SwiftUI first delivers a non-active scene phase. A reconnect scheduled in that window still runs the full churn the fix is meant to suppress. The window is small and self-corrects on the next phase event, so risk is low, but the invariant "never churn while backgrounded" isn't absolute.
**Suggestion:** Seed `isSceneActive` from the observed `ScenePhase` at init (or when the scenePhase handler is first wired), or gate `scheduleReconnect` on `isSceneActive` in addition to the execution guard.

### [minor] `.inactive` is gated exactly like `.background`, which may be over-broad
**File:** Conduit/Services/AppState.swift:3746 (and the guard at :3475)
**Issue:** The watchdog report (`0x8BADF00D`, `ProcessVisibility: Background`) is specific to background visibility, yet `executeReconnect`'s guard also drops reconnects during `.inactive` — e.g. control-center pull, app switcher, incoming-call/system overlay, where the app is still doing real work. During a system overlay the scene can stay `.inactive` for a long stretch; a dropped socket there means transport recovery waits for the next `.active` (and the scene task's healthCheck path only partially covers it). Note also the asymmetry: `.background` calls `cancelScheduledReconnect()` but `.inactive` does not, so a timer armed while active fires during `.inactive` only to be discarded by the guard.
**Suggestion:** Consider distinguishing the two states (e.g. gate only on `.background`), or at minimum also `cancelScheduledReconnect()` on `.inactive` so an armed timer isn't fired and thrown away. If `.inactive` suppression is intentional, document the voice/system-overlay tradeoff.

### [minor] Test fragility: scene-task isolation in `testSceneActivationRestoresReconnectExecution` depends on the injected mintTicket throwing
**File:** ConduitTests/AppStateChatResumeTests.swift:2393-2421
**Issue:** `handleScenePhase(.active)` spawns an unstructured scene task that runs `reconnectForRetry(.automaticReturn)`. The assertion `XCTAssertEqual(reconnectSpy.purposes, [.preserveCurrent])` only holds because the harness `mintTicket` throws `.notReady`, routing the scene task onto the controlled scheduler's error branch. If a future change makes the harness mint succeed, the scene task would proceed into `connect(with:)` → live `connectChatResumeClient`, hitting real network inside a unit test. The comment documents the expectation but doesn't enforce it.
**Suggestion:** Inject a `connectClient` into the lifecycle operations that records (or throws) rather than relying on the mint failure, or assert on `scheduler.scheduledCount`/`cancelledCount` to make the isolation explicit.

### [nit] Backoff ratchets on dropped cycles
**File:** Conduit/Services/AppState.swift:3452
**Issue:** `reconnectAttempts += 1` happens at `scheduleReconnect` time, before the scene gate drops the cycle. Repeated background/foreground cycles with no real connection failure therefore drive the backoff toward the 5s cap, and it's only reset on a successful connect. Capped, so harmless, but it makes `reconnectAttempts` a poor signal for actual failure history.
**Suggestion:** Reset or skip the increment when the subsequent `executeReconnect` is dropped for inactivity.

## Verdict
COMMENT

The two fixes are sound, correctly target the reported root causes, and are well-tested: the retry loop in `mintTicket` (DashboardTicketBridge.swift:258-289) now expresses the intended contract directly, the `transportContinuation` checkpoint guard (AppState.swift:1204) closes the in-flight-abort gap from the earlier review, and `cancelScheduledReconnect()` on `.background` drops armed timers. Code is `@MainActor`-consistent (no data races, no new retain cycles — `[weak self]` used throughout, and `WeakScriptMessageHandler` remains correctly weak), and no force unwraps or new UI/accessibility surface were introduced. The findings above are hardening items; none block merge.

Note: `reloadCount` (DashboardTicketBridge.swift:191) is also incremented by `reload()` calls from AppState's sign-in path (AppState.swift:3542), so the new test's `reloadCount == 2` assertion is only meaningful in the isolated bridge context — fine for a diagnostic counter, but worth a doc comment if it's expected to track mint retries specifically.

<a href="https://opencode.ai/s/nCtEqRpx">New%20session%20-%202026-08-15T21%3A43%3A33.227Z</a>
[opencode session](https://opencode.ai/s/nCtEqRpx)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31910170622)

kaishi00:
Round-2 triage in c9b322a — took the hardening items, declined two with rationale:

**Taken:**

- **Negative `readinessPollAttempts` trap** (CodeRabbit) — clamped at 0 in the initializer. Test-injection-only parameter, but a two-line clamp removes the footgun.
- **`.inactive`/`.background` asymmetry** (OpenCode) — the sub-finding was right: suppressing execution during `.inactive` while leaving timers armed meant fire-and-discard. `handleScenePhase(.inactive)` now calls `cancelScheduledReconnect()` too.
- **Backoff ratchets on dropped cycles** (OpenCode) — a scheduled cycle dropped for inactivity now undoes the `reconnectAttempts` increment its scheduling took, so bg/fg cycling without a real gateway failure can't push retry delay to the cap.
- **Test isolation depended on mint throwing** (OpenCode) — `testSceneActivationRestoresReconnectExecution` now injects a `connectClient` recorder and asserts it stays at 0; even a future harness mint that succeeds could never reach a live connection path.
- **`reloadCount` scope** (OpenCode nit) — doc comment now states it counts reloads from any caller, not just mint retries.

**Declined, with rationale:**

- **Seed `isSceneActive` from the observed phase at init** — the cold-launch-into-background window is theoretical for this app: push wakes store their payload and restore on foreground (no reconnect work in the wake path), and the first phase event arrives within milliseconds of launch. Defaulting `true` is load-bearing in the other direction: blocking the cold-start restore on the first phase event would regress startup. The tradeoff is now documented on the property instead.
- **Distinguish `.inactive` from `.background` (gate only on background)** — the watchdog report is specific to background visibility, but `.inactive` immediately *precedes* `.background` on home-press; gating only on `.background` would let a cycle start in the inactive blip and continue into the backgrounded window — which is exactly the crash sequence this PR closes (in-flight cycles do abort at checkpoints now, but the entry gate shouldn't re-open that window). Recovery latency under long system overlays is bounded by the `.active` scene task, which is also the first moment the user can see the transcript again. Documented at the property and in the `.inactive` case.

Note: CodeRabbit's inline "transportContinuation does not check isSceneActive" on this round is a stale re-anchor of the round-1 finding — the guard has been in place since a04c8a4 (AppState.swift:1199-1206) and is covered by `testBackgroundingMidMintAbortsInFlightReconnect`.

Unit suite: 665 tests, 0 failures.

kaishi00:
## Summary

Two connection-recovery fixes: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady`/`.signInRequired` (a failed initial page load no longer wedges the bridge forever) and surfaces the final error instead of a hardcoded `.signInRequired`; `AppState` gates reconnect execution on scene activity to avoid the 0x8BADF00D background watchdog kill, aborting in-flight cycles at transport checkpoints and cancelling armed timers on `.background`/`.inactive`. Four regression tests added.

## Findings

### [minor] Backoff ratchet is only undone for cycles that *fire* while inactive, not for canceled timers
**File:** Conduit/Services/AppState.swift:3464-3471 (and 3749, 3766 vs. 3457)
**Issue:** `reconnectAttempts += 1` is consumed at schedule time (line 3457), but the compensating `- 1` only runs inside the scheduler closure when it fires while inactive. The new `cancelScheduledReconnect()` calls on `.background`/`.inactive` cancel the timer *before* it fires, so that schedule's increment is never undone. Common scenario: socket drops while active → timer armed (attempts++) → user backgrounds → `.background` cancels the timer → attempts stays elevated. The comment at 3466-3468 claims repeated bg/fg cycling "cannot ratchet the retry delay toward its cap," but that only holds when the closure runs, not when the timer is canceled. Impact is bounded (capped at 5s, reset on successful connect at 3598), but the stated invariant isn't fully achieved.
**Suggestion:** Track the pending increment (e.g. decrement inside `cancelScheduledReconnect()` as well, or make `reconnectAttempts += 1` symmetric with cancellation), or narrow the comment to the fired-while-inactive case.

### [minor] `mintTicket`'s `while true` loop is bounded only implicitly
**File:** Conduit/Services/DashboardTicketBridge.swift:272-288
**Issue:** Termination relies on the non-obvious invariant that the multi-pattern catch's `where attempt < 2` stops matching at `attempt == 2`, so the error propagates. The bound is correct today (3 total evaluations), but it is not expressed in the control flow: any future edit that adjusts or drops the guard silently converts this into an unbounded loop that keeps calling `reload()` (each of which rejects pending requests and spawns a page-load task) and sleeping 350ms. The previous `for attempt in 0..<3` form made the bound compile-enforced and self-documenting.
**Suggestion:** Prefer an explicit bounded form, e.g. `for attempt in 0..<3 { ... }` with the retry catch gated on `attempt < 2` and a trailing `throw` for the exhausted path, or a `while attempt < 3` with a terminal `throw`.

### [nit] The scene gate is transport-wide, not reconnect-only
**File:** Conduit/Services/AppState.swift:1205-1209
**Issue:** `transportContinuation()` is the shared checkpoint for the full `connect(with:)`/sync flow, not just reconnect churn. Adding `isSceneActive` there means a foreground-initiated connect or session sync that backgrounds mid-flight (e.g. app-switcher gesture during catalog load) also aborts at its next checkpoint and is left to the `.active` scene task to resume. That recovery path exists (health-check + `syncSession`/`reconnectForRetry`), and the watchdog motivation justifies it, but the doc comment ("reconnect churn", "reconnect work") understates the scope. Worth confirming the `.active` task fully restores the session catalog for a mid-`connect` abort where `connection` is already non-nil and `client.isConnected` is true.
**Suggestion:** Update the comments at 1205-1208 and 3482-3487 to state the gate applies to all chat-resume transport continuations, and add a test asserting catalog restore after a mid-connect background abort.

### [nit] `reconnect()` is a silent no-op while the scene is inactive
**File:** Conduit/Services/AppState.swift:3476-3479
**Issue:** The new `guard isSceneActive else { return }` in `executeReconnect` makes the public `reconnect()` return without action when backgrounded/inactive. A user-initiated retry landing in an `.inactive` blip (e.g. system overlay, control-center pull) silently does nothing and recovery waits for the next `.active`. Low risk since `.active` is guaranteed to arrive, but callers get no signal that the attempt was dropped.
**Suggestion:** Consider returning a `Bool`/result from `reconnect()` indicating whether the attempt ran, or documenting the no-op semantics at the call site.

## Verdict

COMMENT

The two fixes are sound, correctly target the reported root causes, and are well-tested: the retry loop now recovers a cold bridge instead of wedging it, `mintTicket` no longer forces a bogus sign-out, the `transportContinuation` checkpoint (AppState.swift:1209) closes the in-flight-abort gap, and `cancelScheduledReconnect()` on `.background`/`.inactive` drops armed timers. The code is `@MainActor`-consistent (no data races on `isSceneActive`), no new retain cycles (`[weak self]` throughout, `WeakScriptMessageHandler` still weak), and no force unwraps or new UI/accessibility surface were introduced. Findings above are cleanup/hardening items; none block merge.

<a href="https://opencode.ai/s/f7e2j36p">New%20session%20-%202026-08-15T22%3A00%3A36.631Z</a>
[opencode session](https://opencode.ai/s/f7e2j36p)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31910935563)

kaishi00:
Round-3 triage in 4545326 — the first finding was a real bug in my round-2 fix, now corrected; the rest taken as form/doc hardening:

**Backoff ratchet: canceled timers never undid their increment** — correct, and a sharper catch than my round-2 compensation. Rather than decrement inside `cancelScheduledReconnect()` (which fires from several paths where a decrement would be wrong — successful connects, sign-out), the increment now moves from schedule-time into the scheduled closure: the backoff step is consumed only when the cycle actually starts running. Canceled-before-fire and dropped-for-inactivity both now never count, with no compensation logic to keep honest. Failure-sequence arithmetic is unchanged (cycle N runs with `attempts = N-1`, same delays as before). Regression test added: `testCanceledReconnectTimerDoesNotAdvanceBackoff` asserts a timer armed at the first step, canceled by scene backgrounding, still leaves the next reconnect at the first step (the controlled scheduler now records requested delays).

**`while true` bound only implicit in mintTicket** — agreed, the compile-enforced bound is worth more than avoiding the trailing throw. Back to `for attempt in 0..<3` with the two conditional catches; the trailing `throw .notReady` now carries a comment marking it as definite-exit plumbing that is unreachable in practice (on the final attempt neither pattern matches, so the real error already propagated).

**Gate is transport-wide, comments understated it** — taken as documentation: `transportContinuation` and `executeReconnect` comments now say explicitly that the gate covers connects and post-connect syncs, not just reconnects, and why that's intended. On the suggested mid-connect-abort catalog-restore test: declining that one — the `.active` recovery path (health check → `syncSession`/`reconnectForRetry`) is what restores the catalog and is already exercised by the existing scene-phase suite, and a faithful mid-`connect` abort test needs a socket-connected client harness that doesn't exist in the unit target; the end-to-end variant of exactly that scenario (background/foreground during connect churn, catalog restored) is what the chaos-gateway verification in the PR description covered.

**`reconnect()` silent no-op while inactive** — documented at `executeReconnect` (the no-op is deliberate and self-corrects on the next `.active`). Declining the `Bool` return: the only callers are the settings reconnect button and tests, both of which just fire-and-forget; a return value would ripple without a consumer that can act on it. Happy to revisit if a caller ever needs the signal.

Unit suite: 666 tests, 0 failures.

kaishi00:
## Summary

Fixes two connection-recovery bugs: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady` (a failed initial page load no longer wedges reconnect forever) and surfaces the final error instead of a hardcoded `.signInRequired`; `AppState` gates reconnect execution and in-flight transport continuations on scene activity to avoid the 0x8BADF00D background watchdog kill. Four regression tests added; the backoff/`mintTicket` refactors from review rounds 1–3 are correctly folded into HEAD (4545326).

## Findings

### [minor] Transport-wide scene gate can strand `isConnecting`/`isConnected` after an explicit connect aborts mid-flight
**File:** Conduit/Services/AppState.swift:1211 (guard), 1726-1731 (first post-connect checkpoint), 1703/1716 (state set before the first suspension)
**Issue:** The new `isSceneActive` guard is checked after every suspension in the full `connect(with:)` flow, not just reconnects. A user-initiated connect that goes `.inactive`/`.background` between `connectChatResumeClient` and the checkpoint at 1726 aborts with `isConnecting = true`, `isConnected = false`, `turnState = .synchronizing`, but the socket already connected. On return, the `.active` scene task takes the healthy-socket branch (`client.isConnected` is true) → `healthCheck()` → `syncSession(...)`; neither `performSyncSession` nor that scene-task branch resets `isConnecting`/`isConnected` (they're only cleared in the `reconnectForRetry` path at 3599-3600). `isConnected == false` blocks the send-prompt guards (AppState.swift:4373), so the UI can sit on "connecting…" with sends disabled until a future reconnect. This behavior is new to this PR (previously the connect simply ran to completion through the background).
**Suggestion:** Reset `isConnecting`/`isConnected`/`turnState` at the abort points (or in a `defer` that runs when returning due to a nil continuation), or have the `.active` healthy-socket path explicitly re-publish the connected state before `syncSession`.

### [minor] Cold-bridge mint retry performs full page reloads that can restart a slow, recovering load
**File:** Conduit/Services/DashboardTicketBridge.swift:272-289
**Issue:** Each `.notReady` retry calls `reload()`, which cancels the in-flight navigation (`isReady = false`, new `loadDashboardSession()`) and restarts it from zero. With the defaults (3s `waitUntilReady` poll window × 3 attempts ≈ 10s per mint call), a page load that is slowly progressing on a degraded link can be repeatedly aborted and restarted before it ever finishes, and AppState then re-schedules the entire cycle on exhaustion — so recovery latency on marginal links is potentially worse than the old single 3s poll. Strictly better than the permanent wedge, and the chaos-gateway verification covers the recovering case, so this is informational.
**Suggestion:** Consider checking readiness once before reloading (e.g. only `reload()` when the load has actually failed via `didFailProvisionalNavigation`), or lengthening the poll window relative to the retry count so a slow load isn't torn down mid-progress.

### [nit] `scheduleReconnect` arms a timer while the scene is inactive that is guaranteed to be discarded
**File:** Conduit/Services/AppState.swift:3466-3474
**Issue:** A socket drop delivered while `.inactive`/`.background` still arms a timer (up to 5s backoff); when it fires, the closure's `guard self.isSceneActive` discards it, having already consumed the queued reconnect purpose (`takeQueuedReconnectPurpose()`). Correct behavior, but the timer, the queued purpose, and the wakeup are wasted, and nothing re-arms until the next `.active` (which is the intended recovery).
**Suggestion:** Early-return from `scheduleReconnect` when `!isSceneActive` (matching `executeReconnect`'s guard) so no timer/purpose is registered for a cycle that cannot run.

### [nit] `testColdBridgeMintTicketRetriesReloadAndSurfacesNotReady` couples `reloadCount` to caller behavior
**File:** ConduitTests/DashboardTicketBridgeTests.swift:45-68
**Issue:** `reloadCount` is also incremented by AppState's sign-in-recovery `reload()` (AppState.swift:3558), so the `reloadCount == 2` assertion is only meaningful in the isolated bridge harness. This is documented on the property, so it's fine as a diagnostic counter — just noting the coupling for future test authors.
**Suggestion:** None required; consider a `mintReloadCount` if the counter ever needs to be mint-specific.

## Verdict

COMMENT

The two fixes are sound and directly address the reported root causes: `mintTicket` now recovers a cold bridge instead of wedging it and no longer forces a bogus sign-out, and the scene-gated reconnect (entry guard + `transportContinuation` checkpoint + timer cancellation on `.inactive`/`.background` + backoff consumed only on execution) closes the in-flight-abort gap and the canceled-timer ratchet. The retry loop, backoff arithmetic, and trailing-throw-unreachable comment in `mintTicket` are all correct. Code is `@MainActor`-consistent (no data races on `isSceneActive`), `[weak self]` throughout, no new force unwraps, and no new UI/accessibility surface. Findings above are hardening items; none block merge.

<a href="https://opencode.ai/s/InfVnNgT">New%20session%20-%202026-08-15T22%3A23%3A25.276Z</a>
[opencode session](https://opencode.ai/s/InfVnNgT)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31911949707)

kaishi00:
Round-4 triage across 9328fee + 5c7536c — the first finding was real and is fixed; the slow-load refinement is taken; one nit taken, one noted as-is:

**Stranded `isConnecting`/`isConnected` after a mid-connect abort** — verified real: a connect that crosses a scene-inactive transition between socket-open and its first checkpoint returns with `isConnecting=true, isConnected=false`, and the `.active` healthy-socket branch ran health-check + sync without ever re-publishing transport state — leaving sends disabled (`createNewSession` gates on `isConnected`) until some later reconnect. Fixed in the `.active` healthy-socket branch: it now re-publishes `isConnected = true; isConnecting = false` before the health check, so the stranded state self-corrects at the first foreground regardless of what stranded it. Declined the abort-point/`defer` variant: `connect(with:)` has half a dozen checkpoint exits plus the catch path, and a defer can't distinguish abort-from-superseded-connect without clobbering a newer connect's in-flight flags. On unit coverage: a faithful test needs a socket-connected client harness the unit target doesn't have (existing fakes are private to HermesClientTests); the end-to-end variant — background/foreground mid-connect under chaos — is exactly what I re-verified below.

**Cold-bridge reload can abort a slowly-progressing load** — taken, it's the right call for degraded links (the PR's whole audience). The bridge now tracks `isLoadFailed` from the navigation callbacks and `mintTicket` reloads only terminally failed loads; a load still in flight keeps being polled across attempts instead of being torn down at each 3s timeout. A silently hung load becomes reloadable when the system request timeout fails it. Interesting wrinkle found while testing this: in the unit-test host, WKWebView navigation callbacks for a refused port are not delivered deterministically at all, so the test asserts the unconditional contract (exhaustion surfaces `.notReady`, never `.signInRequired`) and the reload path only when a failed load was actually observed — the reload-recovers behavior stays covered by the chaos-gateway end-to-end run.

**scheduleReconnect arming a timer that's guaranteed to be discarded** — taken: early-return while the scene is inactive, so neither the timer nor the queued purpose is consumed. The background-drop test now also asserts nothing is armed.

**reloadCount caller coupling** — no change beyond the round-3 doc note (the review itself said "none required").

**End-to-end re-verification on this head** (chaos gateway): launch into a 20s outage → first mint + WS open as soon as the outage lifted; then 10 rounds of background/foreground torture with mid-round outage windows and WS churn → app alive throughout, final state fully connected with transcript and session list intact. Unit suite: 666/0 across three consecutive runs.

(Note: 9328fee briefly shipped without the AppState half of this round — a local revert intended to drop a repro hook also dropped uncommitted edits; 5c7536c restores them. CI on 9328fee alone will show the background-drop test failing; the PR head is 5c7536c.)

kaishi00:
## Summary
Fixes two connection-recovery bugs: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady` so a failed initial page load no longer wedges reconnect forever, and `AppState` gates reconnect execution and in-flight transport continuations on scene activity to avoid the 0x8BADF00D background watchdog kill. Four regression tests are added.

## Findings

### [major] `signInRequired` retry silently lost its `reload()` recovery
**File:** Conduit/Services/DashboardTicketBridge.swift:286-297
**Issue:** The original code called `reload()` unconditionally when `mintTicket` caught `.signInRequired`. The new combined catch gates `reload()` on `isLoadFailed`, but `isLoadFailed` is *always false* at that point: `requestJSON` only runs after `waitUntilReady()` passes, which requires `isReady == true`, and `isReady` is set only in `didFinish`, which resets `isLoadFailed = false` first. A 401/403 is the only route to `.signInRequired` here, and it requires a completed page load. So the signInRequired path is now a blind 350 ms re-POST with no page reload — the cookie-flush mechanism the original code relied on ("cookie reaches WebKit's cookie store a moment before its network process") is gone. On the slow/flaky links this PR targets, the retry can 401 twice and force a spurious sign-in, which is the exact failure mode the PR's own round-4 note ("cold bridge must not masquerade as an expired session and force a bogus sign-out") is trying to avoid. The `mintTicket` header comment still lists `signInRequired` as a reload-recoverable failure, misrepresenting the new control flow.
**Suggestion:** Split the catches: keep `reload()` unconditional for `.signInRequired` (as before), and keep the `isLoadFailed`-gated reload only for `.notReady`. Or, if the blind retry is intended, update the comment to state the reload no longer applies to the 401 path.

### [minor] `.notReady` retries add ~9s of pure polling for non-reloadable failures
**File:** Conduit/Services/DashboardTicketBridge.swift:278-303
**Issue:** A genuinely logged-out dashboard (page redirects to `/login`, `didFinish` sets `isReady = false`, `isLoadFailed = false`) surfaces as `.notReady`. Because `isLoadFailed` is false, no reload happens, yet the retry loop still re-polls the full `waitUntilReady` window three times: default 30×100 ms × 3 attempts + 2×350 ms sleeps ≈ 9.7 s. Before this PR, `waitUntilReady()` sat *outside* the `do` block, so `.notReady` propagated after one 3 s poll. The reconnect state is now pinned ~6.7 s longer with nothing being attempted.
**Suggestion:** Only retry `.notReady` when the failure is actually reloadable (`isLoadFailed == true`); otherwise propagate immediately. Alternatively, shorten the poll window on later attempts.

### [minor] Cold-launch-into-background can still churn
**File:** Conduit/Services/AppState.swift:652
**Issue:** `isSceneActive` defaults to `true` and only flips on the first scene-phase event. A push/background-fetch/VoIP wake that schedules reconnect work before SwiftUI delivers a non-active phase runs the churn the fix is meant to suppress. The tradeoff is documented on the property and the window is small/self-correcting, so risk is low.
**Suggestion:** Gate `scheduleReconnect`'s new early-return on a `didEnterBackground` notification as a backstop, or accept as documented (stated).

### [nit] Test comment overstates the isolation guarantee
**File:** ConduitTests/AppStateChatResumeTests.swift:2437-2441
**Issue:** The comment claims the injected `connectClient` means "even if the harness mintTicket were changed to succeed, the scene task's recovery attempt could never touch a live connection path." That is not true: if `mintTicket` succeeded, the scene task's `reconnectForRetry(.automaticReturn)` would proceed past the mint to `makeClient` → `connectChatResumeClient` → `connectCount.value += 1`, and the `XCTAssertEqual(connectCount.value, 0)` at line 2459 would fail. The assertion does enforce isolation, but the comment describes a mechanism it doesn't provide.
**Suggestion:** Reword the comment to say the assertion fails if the scene task ever reaches a connection path.

## Verdict
REQUEST_CHANGES

The two fixes are sound and the scene-gating work is well-constructed: `@MainActor`-consistent (no data races on `isSceneActive`), `[weak self]` throughout, `WeakScriptMessageHandler` still weak, no force unwraps or new UI/accessibility surface, and the round-3/4 backoff and stranded-transport fixes are correct. The blocking issue is the silent loss of the `signInRequired` reload recovery — a regression introduced by this PR in a path directly adjacent to its stated goal, with a trivial fix. The remaining findings are minor hardening items.

<a href="https://opencode.ai/s/Wdn9yvXo">New%20session%20-%202026-08-15T23%3A14%3A04.731Z</a>
[opencode session](https://opencode.ai/s/Wdn9yvXo)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31914142412)

kaishi00:
Fair catch on the major — verified real, and it was a regression of my own round-4 change: the `isLoadFailed` gate on the combined catch silently disabled the `signInRequired` path's unconditional `reload()`. Their causal chain is exactly right: a 401/403 can only arrive after `waitUntilReady()` passes → which requires `didFinish` → which resets `isLoadFailed` — so the gate was always false on that path, leaving a blind re-POST where the cookie-race recovery (reload so WebKit's cookie store settles before judging the session expired) used to be. On flaky links that turns recoverable 401s into spurious sign-outs — the failure mode this PR's round-1 commit explicitly guards against.

**Fixed in e88c908:** the catches are split — `signInRequired` reloads unconditionally (pre-round-4 behavior), `.notReady` keeps the terminally-failed-only reload for slow-link protection. The header comment now describes each path's actual behavior.

**Minor #2 taken as well, with a sharper fix than suggested:** rather than only skipping non-reloadable retries, a bridge whose load *landed on the dashboard's /login redirect* is now tracked (`didLandOnLogin`) and `waitUntilReady()` surfaces it as `.signInRequired` instead of `.notReady` — a logged-out dashboard is signed out, not still loading. That removes the ~9.7s of dead polling AND finally delivers the redirect's expiry signal to the reconnect path (previously the redirect only rejected already-pending requests; a fresh mint just saw endless notReady). The flag resets whenever a fresh load starts, so a stale verdict can't leak into a slow reload's poll window — and a stale-or-real signInRequired then flows into the (restored) reload-retry, which is precisely the cookie-race handling that path exists for.

**Minor #3:** accepting as documented, as the review itself allowed (round-2 declined item, rationale unchanged).

**Nit:** comment reworded — the recorder is an isolation tripwire whose assertion fails if a connection path is reached, not a guarantee it can't be.

Unit suite: 666/0 across three consecutive runs.

kaishi00:
## Summary
Two connection-recovery fixes: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady`/`.signInRequired` (a failed initial page load no longer wedges reconnect forever, and a login-landing surfaces `.signInRequired` instead of endless `.notReady`), and `AppState` gates reconnect execution and in-flight transport continuations on scene activity to avoid the 0x8BADF00D background watchdog kill. The round-5 split-catch fix correctly restored the `signInRequired` unconditional reload that round 4 had inadvertently disabled.

## Findings

### [minor] `requestJSON`'s readiness poll doesn't honor `didLandOnLogin` — asymmetry with `waitUntilReady`
**File:** Conduit/Services/DashboardTicketBridge.swift:331
**Issue:** Round 5 made `waitUntilReady()` throw `.signInRequired` when the bridge is parked on the login page, but `requestJSON`'s internal poll still throws `.notReady` for the same condition. There are ~40 `requestJSON` call sites in AppState (catalog, config, skills, health/status, etc.), all of which still report a logged-out dashboard as "not ready" — so the login-recovery that now works for `mintTicket` doesn't apply to the bulk of the bridge's traffic. Not a regression (pre-round-5 behavior), but the fix scope is incomplete.
**Suggestion:** Mirror `waitUntilReady`: `guard !isInvalidated, isReady else { throw didLandOnLogin ? .signInRequired : .notReady }` so non-mint callers route to the same sign-in recovery.

### [minor] `didFinish` origin-mismatch branch never becomes reloadable
**File:** Conduit/Services/DashboardTicketBridge.swift:478-482
**Issue:** On an origin mismatch, `isReady = false` is set but `isLoadFailed` is not. A settled-but-foreign-origin landing (e.g., an SSO/IDP redirect that slips past the navigation-policy checks) is then treated by `mintTicket`'s `.notReady` retry as an in-flight load: no `reload()` happens, and all three attempts re-poll a page that is already settled (~9s guaranteed failure). This is the same "settled-but-unusable page that never gets reloaded" pattern the PR set out to fix, on a defensive branch.
**Suggestion:** Set `isLoadFailed = true` in the origin-mismatch arm of `didFinish` so the retry reloads it, or document why this landing is intentionally left non-reloadable.

### [minor] New `didLandOnLogin` behavior has no direct unit coverage
**File:** ConduitTests/DashboardTicketBridgeTests.swift:45-68
**Issue:** `testColdBridgeMintTicketRetriesReloadAndSurfacesNotReady` only exercises the cold/refused-port path and (as the test's own comment notes) conditionally skips the reload assertion because WKWebView navigation callbacks aren't deterministic in the test host. The round-5 headline behavior — a `/login` landing surfacing `.signInRequired` and flowing into the restored reload-retry — has no unit test at all; the author confirms it's only covered by chaos-gateway e2e. A future refactor of the `didLandOnLogin`/catch logic could silently break the cookie-race recovery without the suite noticing.
**Suggestion:** Add a test that drives the bridge's `didFinish` with a `/login` URL (or injects a navigation-delegate stub) and asserts `mintTicket` surfaces `.signInRequired` and issues a reload.

### [nit] `.inactive` doesn't invalidate the in-flight scene attempt
**File:** Conduit/Services/AppState.swift:3773-3780
**Issue:** `.background` calls `cancelScenePhaseAttempt()`, but `.inactive` only sets `isSceneActive = false` and cancels the timer. An `.active` scene task therefore survives into an `.inactive` window and is not formally invalidated — it continues until its next `transportContinuation` checkpoint. Bounded (`.active` on return calls `cancelScenePhaseAttempt()`), but the two branches are asymmetric for a transition the PR deliberately treats identically for reconnect purposes.
**Suggestion:** Call `cancelScenePhaseAttempt()` in `.inactive` too, matching `.background`, so the scene attempt is aborted at the transition rather than at the next suspension.

## Verdict
COMMENT

The two fixes are sound and directly address the reported root causes, and the round-5 catch split (unconditional reload on `.signInRequired`, `isLoadFailed`-gated reload on `.notReady`) correctly restores the cookie-race recovery that round 4 had regressed. The code is `@MainActor`-consistent (no data races on `isSceneActive`, bridge navigation state), `[weak self]` throughout with no new retain cycles, no force unwraps, and no new UI/accessibility surface. The findings above are hardening and scope-completeness items; none block merge.

<a href="https://opencode.ai/s/Q1FbZtNs">New%20session%20-%202026-08-15T23%3A42%3A59.384Z</a>
[opencode session](https://opencode.ai/s/Q1FbZtNs)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31915394415)

kaishi00:
Round-6 findings all taken in 53181e0 — they were small and each closed a real seam:

- **requestJSON login mirror** — right, the fix scope was incomplete. `requestJSON`'s poll now throws the same `.signInRequired` verdict for a login-parked bridge that `waitUntilReady` does, so all bridge traffic reports a logged-out dashboard consistently.
- **Origin-mismatch arm never reloadable** — taken: a settled foreign-origin landing is now marked load-failed, so mint retries reload back to the dashboard origin instead of polling a finished page for ~9s.
- **Unit coverage for the round-5 recovery** — added `testMintTicketReloadsForSimulatedLoginLanding`. Real WKWebView navigation isn't drivable in the unit host (established in round 4), so the test uses an internal `simulateLoginLandingForTesting` hook that reproduces the exact state a `/login` landing produces — and asserts the deterministic core invariant: the `signInRequired` catch reloads (the round-4 regression would fail this with reloadCount == 0) and never mints against a logged-out session. The hook clears on the next `loadDashboardSession()` exactly as production reloads do.
- **`.inactive` scene-attempt cancellation** — taken; the transition now formally aborts the in-flight scene attempt instead of waiting for its next checkpoint, matching the background branch's handling. (Idempotent with the subsequent `.background` in the home-press sequence.)

Unit suite: 667/0 across three runs.

kaishi00:
## Summary

Fixes two connection-recovery bugs: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady`/`.signInRequired` so a failed initial page load no longer wedges the bridge (with a login-parked bridge surfacing `.signInRequired` via the new `didLandOnLogin` tracking), and `AppState` gates reconnect execution, scheduled timers, and in-flight `transportContinuation` checkpoints on `isSceneActive` to avoid the 0x8BADF00D background watchdog kill. Four regression tests added; the retry-loop and scene-gating logic is `@MainActor`-consistent with `[weak self]` throughout.

## Findings

### [minor] Readiness polling burns the full window on a load already known to be dead
**File:** Conduit/Services/DashboardTicketBridge.swift:310 and 333
**Issue:** `waitUntilReady()`/`requestJSON()` poll `readinessPollAttempts` (default 30 × 100 ms = 3 s) whenever `!isReady`, even after `didFail`/`didFailProvisionalNavigation` has set `isLoadFailed = true`. A terminally failed load can never become ready without a reload, yet each of the three mint attempts sleeps the full 3 s window before the `.notReady` catch finally reloads — roughly 10 s of pure polling per mint call during a persistent outage, with no load even in flight. Round 5 eliminated this exact waste for the login-parked case (fast `.signInRequired`), but the failed-load case still polls a dead page.
**Suggestion:** Break out of the poll loop (or throw immediately) when `isLoadFailed` becomes true mid-poll — the load is dead, and the `.notReady` catch's reload is the correct recovery anyway. The in-flight-load protection is unaffected since `loadDashboardSession()` resets `isLoadFailed = false`.

### [nit] Deferred recovery always runs `.automaticReturn`, dropping the disconnect-time purpose
**File:** Conduit/Services/AppState.swift:3449 (early return) vs. 3742/3751 (scene task `.automaticReturn`)
**Issue:** When `handleDisconnect` runs while the scene is inactive, `scheduleReconnect` early-returns and the drop's purpose (from `chatResumePurposeForDisconnect()`) is never queued. On the next `.active`, the scene task recovers via `reconnectForRetry(.automaticReturn)`, so `ChatResumeSessionResolver.target` uses `savedSessionID` rather than `currentSessionID`. In ordinary use `currentPurpose` is already `.automaticReturn` so there's no observable change; it only diverges after a user-initiated `reconnect()` (which registers `.preserveCurrent`). Low impact, but a real semantic consequence of the early return worth confirming as intended.
**Suggestion:** If preserving the drop-time purpose matters, stash it in `recoverySequence` before the early return so the `.active` recovery can pick it up; otherwise a one-line note that deferred recovery intentionally upgrades to `.automaticReturn` would close the question.

### [nit] `.inactive` now aborts the in-flight scene attempt, forcing a full redo after transient overlays
**File:** Conduit/Services/AppState.swift:3786
**Issue:** Round-6 made `.inactive` symmetric with `.background` by calling `cancelScenePhaseAttempt()`. But `.inactive` is far more transient than `.background`: a control-center pull or notification shade mid-foreground cancels an in-progress health-check + catalog sync, and the next `.active` re-runs the whole recovery from scratch — extra main-thread churn at exactly the boundary this PR targets. Correct and bounded (idempotent, `.active` re-runs), so this is a monitoring note rather than a defect.
**Suggestion:** Consider a lighter `.inactive` treatment that only gates *new* reconnect starts (as the entry guards already do) while letting an already-running scene attempt finish, reserving the full abort for `.background`.

## Verdict

COMMENT

The two fixes are sound and directly target the reported root causes; the retry loop, backoff-consumed-on-execution, `transportContinuation` gate, and login-landing tracking are all correct, and the round-5 catch split correctly restores the `signInRequired` unconditional reload. No critical or major issues: no retain cycles (`[weak self]` throughout, `WeakScriptMessageHandler` still weak), no force unwraps, no data races (`isSceneActive` and bridge navigation state are `@MainActor`-isolated), and no new UI/accessibility surface. The findings above are hardening items.

<a href="https://opencode.ai/s/LBooRF37">New%20session%20-%202026-08-15T23%3A57%3A54.667Z</a>
[opencode session](https://opencode.ai/s/LBooRF37)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31916012123)

kaishi00:
Follow-up to 53181e0: CI caught an ordering race in the new login-landing test — the bridge's init task runs `loadDashboardSession()` asynchronously, and when it lands between the simulated landing and the first poll, the flag reset turns attempt 0 into `.notReady` and no reload ever happens (exactly the `reloadCount == 0` CI reported; my local runs had the init task settle earlier). Fixed in acac621: the simulation now persists across reloads (test-only flag honored by `loadDashboardSession`), modeling a genuinely logged-out dashboard — the flow is order-independent, both retries reload, exhaustion surfaces `.signInRequired`, and the assertions are exact (`reloadCount == 2`). 667/0 ×3 locally.

kaishi00:
I inspected the PR's diff across `AppState.swift`, `DashboardTicketBridge.swift`, and the two test files, and traced the scene-gating, `mintTicket` retry/catch, `requestJSON` poll, and `transportContinuation` checkpoints to verify the fixes against the PR description, though I did not output the final structured review.

<a href="https://opencode.ai/s/zcNovhi7">New%20session%20-%202026-08-16T00%3A13%3A52.108Z</a>
[opencode session](https://opencode.ai/s/zcNovhi7)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31916667627)

kaishi00:
Second follow-up (ec911df): CI then caught the *cold-bridge* test racing the same way from the other direction — the refused-port failure callback landed after `mintTicket` finished, so the assert-time `isLoadFailed` check didn't reflect what the retries saw. Conclusion: real WKWebView callbacks can't anchor either mint test on any host. Both now use one simulated-landing hook (`.loginPage` / `.loadFailure`) that `loadDashboardSession` re-asserts on every load, modeling a dashboard that keeps producing the same landing. Assertions are exact and callback-timing-independent: failed load → reload ×2, exhaust `.notReady`; login landing → reload ×2, exhaust `.signInRequired`. 667/0 ×3 locally.

kaishi00:
## Summary

The PR fixes two connection-recovery bugs on flaky links: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady`/`.signInRequired` (a failed initial page load no longer wedges the bridge forever, and a login-parked bridge surfaces `.signInRequired` rather than endless `.notReady`), and `AppState` gates reconnect execution, scheduled timers, and in-flight `transportContinuation` checkpoints on scene activity to prevent background reconnect churn from tripping the 0x8BADF00D watchdog. Four regression tests are added. The code is `@MainActor`-consistent with `[weak self]` throughout (`WKWebView.navigationDelegate` is weak, `WeakScriptMessageHandler` remains weak — no retain cycles), contains no force unwraps, and adds no UI/accessibility surface.

## Findings

### [minor] Readiness polls burn the full window on a load already known to be failed
**File:** Conduit/Services/DashboardTicketBridge.swift:319-330 and 342-351
**Issue:** `waitUntilReady()` and `requestJSON()` poll `readinessPollAttempts` (default 30 × 100 ms ≈ 3 s) any time `!isReady`, even after `didFail`/`didFailProvisionalNavigation` has set `isLoadFailed = true`. A terminally failed load can never become ready without a reload, yet each of `mintTicket`'s three attempts sleeps the full 3 s window before the `.notReady` catch finally reloads — roughly 9-10 s of dead polling per mint call during a persistent outage (connection-refused fails in milliseconds, then the poll sleeps another ~3 s). The same holds for the ~40 `requestJSON` callers. This was raised in round-7 review and remains unaddressed.
**Suggestion:** Add `!isLoadFailed` to the poll's `where` clause so the loop exits as soon as the navigation terminally fails (only a reload can revive it, and the `.notReady` catch already performs that reload). In-flight slow loads are unaffected since `loadDashboardSession()` resets `isLoadFailed = false` on each fresh load.

### [minor] Deferred recovery silently upgrades the disconnect purpose to `.automaticReturn`
**File:** Conduit/Services/AppState.swift:3449 (early return) vs. 3742/3751 (scene task `.automaticReturn`)
**Issue:** When a socket drop lands while the scene is inactive/backgrounded, `handleDisconnect` computes `chatResumePurposeForDisconnect()` (`recoverySequence.currentPurpose`) but `scheduleReconnect`'s new `guard isSceneActive` returns before the purpose is ever queued. On the next `.active`, the scene task recovers via `reconnectForRetry(.automaticReturn)`, so `ChatResumeSessionResolver` targets `savedSessionID` rather than `currentSessionID`. In ordinary use the purpose is already `.automaticReturn`, so this is invisible; it only diverges after a user-initiated `reconnect()` (`.preserveCurrent`), where the on-foreground recovery loses the "preserve current session" intent.
**Suggestion:** Either stash the drop-time purpose in `recoverySequence` before the early return so the `.active` recovery picks it up, or add a one-line note that deferred recovery intentionally upgrades to `.automaticReturn`.

### [minor] A silently hung navigation is not reloaded until the system navigation timeout
**File:** Conduit/Services/DashboardTicketBridge.swift:303-309
**Issue:** The `.notReady` catch only reloads when `isLoadFailed == true`, which requires a `didFail*` callback. For a load that hangs without any callback (server accepts the connection but never responds — captive portal, stalled tether), `isLoadFailed` stays `false` through all three attempts, so no reload ever fires within `mintTicket`, and the bridge is parked until WKWebView's own navigation timeout (~60 s) surfaces the failure. This is a strict improvement over the pre-PR permanent wedge (which never recovered at all) and is acknowledged in the comment, but recovery latency on this specific failure mode can still be tens of seconds.
**Suggestion:** Consider an explicit in-flight-navigation watchdog inside the bridge (fail the load after it has been in flight longer than the poll window) so a hung load becomes reloadable without waiting on the system timeout.

### [nit] `.inactive` now aborts the full in-flight scene attempt on transient overlays
**File:** Conduit/Services/AppState.swift:3786
**Issue:** Round-6 made `.inactive` symmetric with `.background` by calling `cancelScenePhaseAttempt()`. But `.inactive` is far more transient than `.background`: a control-center pull or notification shade cancels an in-progress health-check + catalog sync, and the next `.active` re-runs the whole recovery from scratch — extra main-thread work at exactly the boundary this PR targets. Correct and bounded (idempotent; `.active` re-runs), so a monitoring note rather than a defect.
**Suggestion:** Consider a lighter `.inactive` treatment that only gates *new* reconnect starts (as the entry guards already do) while letting an already-running scene task finish, reserving the full `cancelScenePhaseAttempt()` for `.background`.

## Verdict

COMMENT

The two fixes are sound and directly target the reported root causes: `mintTicket` recovers a cold bridge instead of wedging it, never forces a bogus sign-out (the `.signInRequired` unconditional reload that round 4 had silently regressed is correctly restored in `e88c908`), and the scene-gated reconnect (entry guard + `transportContinuation` checkpoint at AppState.swift:1211 + timer cancellation on `.inactive`/`.background` + backoff consumed only on execution + stranded-transport re-publish at AppState.swift:3722-3723) is consistent with strong regression coverage. The findings above are hardening items; none block merge.

<a href="https://opencode.ai/s/nYryO3zV">New%20session%20-%202026-08-16T00%3A26%3A29.927Z</a>
[opencode session](https://opencode.ai/s/nYryO3zV)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31917167796)

kaishi00:
Round-7 triage in 2385a75 — one taken, one taken as the suggested note, two declined:

**Taken — stop polling a terminally failed load.** Real latency win for exactly the scenario this PR targets: during a persistent outage each mint attempt slept the full ~3s window after the load had already failed, so ~9-10s of dead polling per cycle before the reload recovery even ran. Both poll loops (`waitUntilReady`, `requestJSON`) now exit as soon as `isLoadFailed` is set — only a reload can revive the load, and the catch performs it. In-flight slow loads are unaffected (`loadDashboardSession` resets the flag per fresh load). Nice side effect: the local suite got ~35% faster because the simulated failed-landing tests exit their polls immediately.

**Taken as note — deferred recovery upgrades to `.automaticReturn`.** As the review itself worked out, this is invisible in ordinary use (the purpose is already `.automaticReturn`) and only diverges after a manual `reconnect()`, where resuming the saved session on foreground is still the sensible outcome. Documented at the early-return guard rather than adding stash-and-restore state for an intent nobody can distinguish in practice.

**Declined — in-flight navigation watchdog.** The review notes this is "a strict improvement over the pre-PR permanent wedge," and the residual exposure (server accepts TCP but never responds, no callback at all) is the rarest failure shape on tethered links — refused/failed dominates by far. An internal watchdog adds timer lifecycle and races with genuinely slow loads (misfiring would reintroduce the abort-a-slow-load problem round 4 fixed). The outer reconnect loop keeps retrying mints, and the system navigation timeout bounds the hang. Worth a follow-up issue if telemetry ever shows this mode mattering.

**Declined — soften `.inactive` scene-attempt cancellation.** This reverses round 6's own request to make `.inactive` symmetric with `.background`. Between the two, symmetry wins: `.inactive` immediately precedes `.background` on home-press, and letting scene work run across that boundary reopens the exact watchdog window this PR closes. The cost of re-running health-check + sync after a control-center pull is bounded and paid at a moment the user is returning anyway. Happy to take the lighter treatment as a follow-up if it shows up in profiling, but I'd rather not oscillate this behavior review-by-review.

Unit suite: 667/0 (two consecutive runs; the suite also got faster from the poll fast-exit).

kaishi00:
## Summary
Fixes two flaky-link connection bugs: `DashboardTicketBridge.mintTicket` now reloads and retries on `.notReady`/`.signInRequired` (so a failed initial page load no longer wedges reconnect forever, and a login-parked bridge surfaces `.signInRequired` instead of endless `.notReady`), and `AppState` gates reconnect scheduling, execution, and in-flight `transportContinuation` checkpoints on `isSceneActive` to prevent background churn from tripping the 0x8BADF00D scene-update watchdog. The code is `@MainActor`-consistent, uses `[weak self]` throughout, and has four targeted regression tests.

## Findings

### [minor] Stale `didLandOnLogin` misclassifies later load failures as `.signInRequired`
**File:** Conduit/Services/DashboardTicketBridge.swift:543-553 (and 518-527)
**Issue:** `didLandOnLogin` is set to `true` when a load lands on `/login`, but neither `webView(_:didFail:)`, `webView(_:didFailProvisionalNavigation:)`, nor the `didFinish` origin-mismatch arm ever resets it. A bridge that once parked on `/login` (sign-in required → CoW native re-auth → `reload()`) retains `didLandOnLogin == true`; if that reload then fails on a flaky link, `waitUntilReady`/`requestJSON` (lines 329-331, 351-353) report `.signInRequired` for what is actually a plain network failure. Because this PR wired `didLandOnLogin` into the ~40 `requestJSON` callers (catalog, config, profiles, skills, status), an outage after an earlier logged-out landing can be misrouted into the sign-in recovery path instead of the `errorMessage` banner path.
**Suggestion:** Reset `didLandOnLogin = false` in `didFail`, `didFailProvisionalNavigation`, and the origin-mismatch arm of `didFinish` (it already gets reset on the success and login-redirect arms), so the flag is scoped to the landing that actually produced it.

### [minor] `reload()` leaves `isLoadFailed`/`didLandOnLogin` stale until the async restore completes
**File:** Conduit/Services/DashboardTicketBridge.swift:262-277 (catch at 303-308)
**Issue:** `reload()` resets `isReady = false` synchronously but not `isLoadFailed`/`didLandOnLogin` — those only reset when the spawned `Task` finishes `DashboardCookiePersistence.restore(...)` and calls `loadDashboardSession()`. Combined with the round-7 fast-exit and the `isLoadFailed`-gated reload in `mintTicket`'s `.notReady` catch, a retry that samples the bridge during that async-restore window still sees the previous load's terminal failure: `waitUntilReady` fast-exits with `.notReady`, and the catch calls `reload()` again — spawning a duplicate restore+load before the first reload's navigation even started, and inflating `reloadCount` beyond the intended one-reload-per-retry. Bounded (3 attempts) and self-correcting, but it adds churn on exactly the degraded links this PR targets.
**Suggestion:** Reset `isLoadFailed = false` (and `didLandOnLogin = false`) synchronously in `reload()`, matching how `isReady` is handled, so the retry loop always clocks against the fresh load's state.

### [minor] Failed/cancelled navigation callbacks can poison a freshly reloaded load
**File:** Conduit/Services/DashboardTicketBridge.swift:543-553
**Issue:** WKWebView delivers `didFail`/`didFailProvisionalNavigation` with `NSURLErrorCancelled` (-999) for a navigation superseded by a newer `load()`. Both handlers now unconditionally set `isLoadFailed = true` and reject pending requests, so a healthy in-flight reload can be marked terminally failed — which, with the fast-exit polling, triggers a spurious extra `reload()` from `mintTicket` or a spurious `.notReady` from non-mint callers. This is amplified by, not introduced in, this PR, but it directly undercuts the "reload only terminally failed loads" distinction the authors built.
**Suggestion:** Early-return from both callbacks when `(error as NSError).code == NSURLErrorCancelled`, so cancellation of a superseded navigation doesn't affect the current load's state or pending requests.

### [nit] Queued reconnect purpose is consumed before the `isSceneActive` re-check
**File:** Conduit/Services/AppState.swift:3476-3478
**Issue:** The scheduler closure calls `recoverySequence.takeQueuedReconnectPurpose()` (line 3476) before the `guard isSceneActive` (line 3478). A timer armed while active that fires concurrently with the scene transition consumes the queued purpose and discards it. Since `.inactive`/`.background` now `cancelScheduledReconnect()` synchronously on the main actor, the practical window is a single fire-during-transition turn and recovery via the `.active` task uses `.automaticReturn` anyway (documented at 3448-3450), so impact is negligible — but ordering the guard before the consume would make the flow self-consistent.
**Suggestion:** Move the `guard self.isSceneActive else { return }` above `takeQueuedReconnectPurpose()` so a dropped cycle leaves the purpose queued rather than silently purging it.

## Verdict
COMMENT

The two fixes are sound and directly address the reported root causes; the round-5 catch split correctly restores the `signInRequired` unconditional reload, and the scene gating (entry guard + `transportContinuation` checkpoint at AppState.swift:1211 + timer cancellation on `.inactive`/`.background` + backoff consumed only on execution + stranded-transport re-publish at AppState.swift:3724-3725) is consistent and well-tested. No critical or major issues: no retain cycles (`[weak self]` throughout, `WeakScriptMessageHandler` remains weak), no force unwraps, no data races (`AppState` and `DashboardTicketBridge` are both `@MainActor`), and no new UI/accessibility surface. The findings above are state-scoping and hard-wiring items in the bridge's new navigation-state handling; none block merge.

<a href="https://opencode.ai/s/69AtZIN2">New%20session%20-%202026-08-16T01%3A18%3A51.174Z</a>
[opencode session](https://opencode.ai/s/69AtZIN2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31919240786)

PR #66, opened 2026-08-16
test(ci): stabilize flaky selection UI tests + one-step CI retry

kaishi00:
## Problem

`SelectionObserverUITests` is the only recurring source of red CI on this repo. Across recent runs, **three different tests from the family failed on three different builds** — including main's own runs:

| Build | Failing test | Failure |
|---|---|---|
| main (#64's run) | `testCellAnchoredLongPressAndDragExtendsAcrossBlocks` | cross-block chrome never appeared |
| PR #65 run 1 | `testCopyAfterDragCapturesCrossBlockText` | `pasteboard:empty` after fixture copy |
| PR #65 run 2 | `testFastLongPressAndDragStillExtendsAcrossBlocks` | no selection after drag + retry |
| PR #65 run 3 | `testCellAnchoredLongPressAndDragExtendsAcrossBlocks` | `Failed to get background assertion … Timed out` (infra) |

Different test each time, same family — the signature of timing flakiness in the drag synthesis and XCUITest infrastructure on CI runners, not of a product regression.

## Fix — two layers

**1. Correct the in-test retry (test-side).** The existing helper (from #63) retried once and only watched `selection.handle.anchor` — but the cell-anchored test *asserts* on `selection.handle.focus`, and the copy test proceeds to tap the fixture copy button regardless. A half-installed chrome (anchor up, focus not) burned the single retry on the wrong condition; a fully failed drag sailed through to `pasteboard:empty`. The helper now takes the drag as a closure (covering the fast-drag velocity variant), runs up to **3 attempts**, waits for **both endpoint handles**, and returns success so each test fails with its own domain message before doing anything that depends on the selection.

**2. Retry the UI CI step once (infra-side).** In-test retries cannot cover failures like the background-assertion acquisition timeout — those happen outside the test. CI now runs unit tests and UI tests as separate steps, and the UI step re-runs once on failure. A genuine double failure still fails the build. Both result bundles upload on failure.

## Verification

- 6/6 UI tests passed twice locally (70s each) and 667/667 unit tests with these changes.
- An initial local run failed all six tests with `Failed to get matching snapshots: Timed out while evaluating UI query` — the known wedged-simulator state after heavy churn on a long-lived simulator (rebooting the sim restored 6/6). CI runners boot a fresh simulator per job, so the step-level retry is the appropriate backstop there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of cross-block drag-and-select interactions, including better handling when selection handles take longer to appear.

* **Tests**
  * Strengthened automated coverage for selection gestures with retry support and fixture relaunches.
  * Improved test execution reliability by retrying UI tests after simulator restart.
  * Test result reports are now retained for both attempts when checks fail.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,6 +85,11 @@ jobs:
             else
               SCOPE="-only-testing:ConduitUITests"
               BUNDLE=TestResults-retry.xcresult
+              # A wedged simulator is the one failure mode in-test retries
+              # cannot clear; restart the device so the retry starts clean.
+              xcrun simctl shutdown "$SIMULATOR" || true
+              sleep 3
+              xcrun simctl boot "$SIMULATOR" || true
             fi
             echo "::group::Test attempt $attempt"
             rm -rf "$BUNDLE"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       - name: List available simulators
         run: xcrun simctl list devices available
 
-      - name: Run tests
+      - name: Pick simulator
         run: |
           set -euo pipefail
           # macOS images change available devices across versions, so avoid
@@ -59,12 +59,18 @@ jobs:
           fi
 
           echo "Using simulator UDID: $SIMULATOR"
+          echo "SIMULATOR=$SIMULATOR" >> "$GITHUB_ENV"
+
+      - name: Run unit tests
+        run: |
+          set -euo pipefail
           xcodebuild test \
             -project Conduit.xcodeproj \
             -scheme Conduit \
             -destination "platform=iOS Simulator,id=$SIMULATOR" \
             -configuration Debug \
-            -resultBundlePath TestResults.xcresult \
+            -only-testing:ConduitTests \
+            -resultBundlePath UnitTestResults.xcresult \
             CODE_SIGN_IDENTITY="" \
             CODE_SIGNING_REQUIRED=NO \
             CODE_SIGNING_ALLOWED=NO \
@@ -74,9 +80,46 @@ jobs:
           # Keep xcodebuild's exit status when output is piped through xcpretty.
           exit ${PIPESTATUS[0]}
 
+      - name: Run UI tests
+        run: |
+          # The synthesized-drag selection tests occasionally lose the race
+          # against slow CI runners (gesture timing, background-assertion
+          # acquisition timeouts). The tests retry drags in-process; this
+          # step-level retry covers the remaining infrastructure blips so a
+          # flaky runner does not paint the build red. A genuine double
+          # failure still fails the build.
+          set -uo pipefail
+          for attempt in 1 2; do
+            echo "::group::UI test attempt $attempt"
+            rm -rf UITestResults.xcresult
+            xcodebuild test \
+              -project Conduit.xcodeproj \
+              -scheme Conduit \
+              -destination "platform=iOS Simulator,id=$SIMULATOR" \
+              -configuration Debug \
+              -only-testing:ConduitUITests \
+              -resultBundlePath UITestResults.xcresult \
+              CODE_SIGN_IDENTITY="" \
+              CODE_SIGNING_REQUIRED=NO \
+              CODE_SIGNING_ALLOWED=NO \
+              DEVELOPMENT_TEAM="" \
+              PROVISIONING_PROFILE_SPECIFIER="" \
+              | xcpretty --color
+            status=${PIPESTATUS[0]}
+            echo "::endgroup::"
+            if [ "$status" -eq 0 ]; then
+              exit 0
+            fi
+            echo "UI tests failed on attempt $attempt (status $status)"
+          done
+          exit 1
+
       - name: Upload test results on failure
         if: failure()
         uses: actions/upload-artifact@v4
         with:
           name: TestResults.xcresult
-          path: TestResults.xcresult
+          path: |
+            UnitTestResults.xcresult
+            UITestResults.xcresult
+          if-no-files-found: ignore
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,44 +61,42 @@ jobs:
           echo "Using simulator UDID: $SIMULATOR"
           echo "SIMULATOR=$SIMULATOR" >> "$GITHUB_ENV"
 
-      - name: Run unit tests
+      - name: Run tests
         run: |
-          set -euo pipefail
-          xcodebuild test \
-            -project Conduit.xcodeproj \
-            -scheme Conduit \
-            -destination "platform=iOS Simulator,id=$SIMULATOR" \
-            -configuration Debug \
-            -only-testing:ConduitTests \
-            -resultBundlePath UnitTestResults.xcresult \
-            CODE_SIGN_IDENTITY="" \
-            CODE_SIGNING_REQUIRED=NO \
-            CODE_SIGNING_ALLOWED=NO \
-            DEVELOPMENT_TEAM="" \
-            PROVISIONING_PROFILE_SPECIFIER="" \
-            | xcpretty --color
-          # Keep xcodebuild's exit status when output is piped through xcpretty.
-          exit ${PIPESTATUS[0]}
-
-      - name: Run UI tests
-        run: |
-          # The synthesized-drag selection tests occasionally lose the race
+          # The synthesized-drag selection UI tests occasionally lose the race
           # against slow CI runners (gesture timing, background-assertion
-          # acquisition timeouts). The tests retry drags in-process; this
-          # step-level retry covers the remaining infrastructure blips so a
-          # flaky runner does not paint the build red. A genuine double
-          # failure still fails the build.
+          # acquisition timeouts). The tests retry drags in-process; if the
+          # full run still fails, retry the UI target once so an
+          # infrastructure blip does not paint the build red. A genuine
+          # double failure still fails the build.
+          #
+          # Attempt 1 always runs the full suite in ONE xcodebuild
+          # invocation (unit + UI). Do not split unit and UI into separate
+          # invocations: on a cold simulator the composer paste unit tests
+          # can stall against the pasteboard XPC service before the UI
+          # runner has warmed the device (observed hanging 35+ minutes on
+          # CI with -only-testing:ConduitTests; reproduced locally on an
+          # erased simulator).
           set -uo pipefail
           for attempt in 1 2; do
-            echo "::group::UI test attempt $attempt"
-            rm -rf UITestResults.xcresult
+            if [ "$attempt" -eq 1 ]; then
+              SCOPE=""
+              BUNDLE=TestResults.xcresult
+            else
+              SCOPE="-only-testing:ConduitUITests"
+              BUNDLE=TestResults-retry.xcresult
+            fi
+            echo "::group::Test attempt $attempt"
+            rm -rf "$BUNDLE"
+            # Unquoted $SCOPE intentionally: empty on attempt 1, one token
+            # on the UI-only retry.
             xcodebuild test \
               -project Conduit.xcodeproj \
               -scheme Conduit \
               -destination "platform=iOS Simulator,id=$SIMULATOR" \
               -configuration Debug \
-              -only-testing:ConduitUITests \
-              -resultBundlePath UITestResults.xcresult \
+              $SCOPE \
+              -resultBundlePath "$BUNDLE" \
               CODE_SIGN_IDENTITY="" \
               CODE_SIGNING_REQUIRED=NO \
               CODE_SIGNING_ALLOWED=NO \
@@ -110,7 +108,7 @@ jobs:
             if [ "$status" -eq 0 ]; then
               exit 0
             fi
-            echo "UI tests failed on attempt $attempt (status $status)"
+            echo "tests failed on attempt $attempt (status $status)"
           done
           exit 1
 
@@ -120,6 +118,6 @@ jobs:
         with:
           name: TestResults.xcresult
           path: |
-            UnitTestResults.xcresult
-            UITestResults.xcresult
+            TestResults.xcresult
+            TestResults-retry.xcresult
           if-no-files-found: ignore
--- a/ConduitUITests/SelectionObserverUITests.swift
+++ b/ConduitUITests/SelectionObserverUITests.swift
@@ -89,6 +89,13 @@ final class SelectionObserverUITests: XCTestCase {
     /// Waiting for both handles matters: the cell-anchored test asserts on
     /// the focus handle, so a retry that only watched the anchor handle
     /// could give up (or stop) while the chrome was half-installed.
+    ///
+    /// A failed attempt can leave a partial selection or the system edit
+    /// menu up, and a synthesized long-press behaves differently against
+    /// that dirty state — so each retry relaunches the fixture for a clean
+    /// slate (captured screen coordinates stay valid; the fixture renders
+    /// identically). Worst case for a genuinely broken feature is roughly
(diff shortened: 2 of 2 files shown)


kaishi00:
## Summary
This PR stabilizes the flaky `SelectionObserverUITests` family: the in-test retry helper now re-runs the drag up to 3 times and waits for both endpoint handles, and the CI workflow splits unit/UI tests into separate steps with a single retry of the UI step.

## Findings

### [minor] Step-level retry runs on the same simulator and cannot recover the documented wedge
**File:** .github/workflows/ci.yml:83-115
**Issue:** The PR's own verification shows the all-six-failures case was caused by a wedged simulator fixed only by rebooting it (`Timed out while evaluating UI query`). But the retry loop re-runs `xcodebuild test` against the *same* simulator, and `xcrun simctl` is never reset. If attempt 1 wedges the simulator, attempt 2 fails identically — the retry does not backstop the root cause it's described as covering. Additionally, `rm -rf UITestResults.xcresult` before attempt 2 discards attempt 1's bundle, so a genuine double failure only uploads the attempt-2 results (the PR text says "both result bundles upload on failure").
**Suggestion:** Reset the simulator between attempts (`xcrun simctl shutdown "$SIMULATOR" && xcrun simctl boot "$SIMULATOR"`, possibly `erase`) before attempt 2, and drop the `rm -rf` so each attempt writes to a distinct path (`UITestResults-attempt1.xcresult`, etc.) uploaded together.

### [minor] Splitting into two `xcodebuild test` invocations doubles the build
**File:** .github/workflows/ci.yml:64-81, 83-115
**Issue:** Previously one `xcodebuild test` built the scheme once and ran both target suites. Now the scheme (app + both test bundles) is built twice per run — once for `-only-testing:ConduitTests`, once for `-only-testing:ConduitUITests` — adding several minutes to every CI run, including green ones.
**Suggestion:** Build once with `xcodebuild build-for-testing`, then run `xcodebuild test-without-building` (with the same `-only-testing` flags) in each step.

### [minor] Retry drags run on partially-installed selection state
**File:** ConduitUITests/SelectionObserverUITests.swift:99-104
**Issue:** If a failed attempt leaves the app with a half-installed selection (e.g., only the anchor handle, or the system edit menu already up from a single-segment selection), the next `press(forDuration:thenDragTo:)` executes against that dirty state. XCTest long-press synthesis on top of an active selection/edit menu can behave differently than on a clean state, so retries may compound rather than recover. The helper also silently waits up to 4+2 seconds per attempt, so a genuinely broken feature costs ~18s before failing.
**Suggestion:** Consider re-launching the app between attempts (or clearing the selection) so each retry starts from a known state, and note the worst-case timing in the comment.

### [nit] `exit ${PIPESTATUS[0]}` in the unit test step is dead code under `set -e`
**File:** .github/workflows/ci.yml:80-81
**Issue:** With `set -euo pipefail`, a failing `xcodebuild | xcpretty` pipeline exits the shell immediately (with the pipeline's non-zero status), so the `exit ${PIPESTATUS[0]}` line is never reached. The outcome is correct, so no behavior is broken — but the "Keep xcodebuild's exit status" comment describes a mechanism that doesn't run.
**Suggestion:** Drop the comment/line, or keep the line but note that `pipefail` + `-e` is what actually preserves the status.

### [nit] Timeout budget imbalance between the two handles
**File:** ConduitUITests/SelectionObserverUITests.swift:101
**Issue:** The focus handle gets only a 2s budget after the anchor's 4s budget. On a slow runner where the anchor appears near the 4s mark, the focus handle (which typically materializes shortly after) may legitimately need more than 2s, causing a spurious retry or false failure.
**Suggestion:** Use symmetric timeouts (e.g., 4s for both) so the second handle isn't starved.

## Verdict
APPROVE

The changes are well-reasoned and address a real flake source. The Swift-side helper is sound (no force unwraps, retain cycles, or concurrency issues — it's non-UI-thread XCUITest code), and failures still surface with domain-specific messages. The CI findings above are refinements, notably that the step-level retry is ineffective against the very simulator-wedge scenario the PR documents as its motivation.

<a href="https://opencode.ai/s/GhUTl8vj">New%20session%20-%202026-08-16T02%3A23%3A22.645Z</a>
[opencode session](https://opencode.ai/s/GhUTl8vj)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31921734047)

kaishi00:
## Summary
This PR hardens the flaky `SelectionObserverUITests` family by (1) making the in-test drag helper retry up to 3× and gate on both selection handles, and (2) adding a one-time step-level retry of CI UI tests via a two-attempt loop in `ci.yml`.

## Findings

### [critical] The CI retry loop is dead code — `set -e` aborts the step before attempt 2 ever runs
**File:** .github/workflows/ci.yml:80-113
**Issue:** GitHub's default `run` shell is `bash -e`. The script only does `set -uo pipefail` (line 80), which leaves errexit enabled. On attempt 1, when `xcodebuild test ... | xcpretty` fails, `pipefail` makes the pipeline fail and `-e` terminates the step immediately — never reaching `status=${PIPESTATUS[0]}` (line 106), the `if` (line 108), or the next loop iteration. The retry — the PR's central feature — is functionally a no-op; any attempt-1 failure still just fails the build, exactly as before this PR.
**Suggestion:** Guard the pipeline so errexit doesn't fire, e.g.:
```bash
if xcodebuild test ... | xcpretty --color; then
  status=0
else
  status=${PIPESTATUS[0]}
fi
```
(or append `|| true` to the pipeline and rely on `PIPESTATUS`).

### [major] Attempt 2 re-runs only the UI target, masking genuine unit-test regressions (false green)
**File:** .github/workflows/ci.yml:86-113
**Issue:** Attempt 1 runs the full suite. If a *unit* test fails while UI tests pass, attempt 2 runs `-only-testing:ConduitUITests` (line 86), passes, and `exit 0` (line 108-109) turns a real regression green. The retry only re-runs the scope that must have failed, so it cannot distinguish "UI-infra flake" from "anything else broke." This is the most dangerous possible failure mode for CI: silently hiding product bugs.
**Suggestion:** Retry the *same* test scope that failed (e.g., capture which target failed from the xcresult and retry that target), or fail the build when attempt 1's failure was not in `ConduitUITests`.

### [minor] Retry re-runs on the same wedged simulator state
**File:** .github/workflows/ci.yml:81-113
**Issue:** Per the PR's own verification, the "all six test failures with `Timed out while evaluating UI query`" case was only recoverable by rebooting the simulator. The attempt-2 loop re-runs `xcodebuild test` against the identical simulator, so a simulator-level wedge fails identically on both attempts — the retry cannot backstop the one failure mode its motivating analysis documents. (The PR text argues CI boots fresh simulators per job, which mitigates this, but the limitation is real for transient infra issues.)
**Suggestion:** Add `xcrun simctl shutdown "$SIMULATOR"` / `boot` (or `erase`) between attempts, or at minimum document that the step retry only covers per-run timing variance, not simulator wedges.

### [minor] In-test retries run against half-installed selection state
**File:** ConduitUITests/SelectionObserverUITests.swift:99-104
**Issue:** If a failed attempt leaves a partial selection (anchor handle up, or the system edit menu active), the next `performDrag()` executes over that dirty state. XCTest long-press synthesis on top of an active selection/edit menu can behave differently than on a clean state, so retries may compound rather than recover. Worst case also costs ~3 × (drag + 4 s + 2 s) ≈ 20 s per call on a genuinely broken feature before failure.
**Suggestion:** Consider re-launching the app between attempts (or at least clearing the selection), and note the bounded worst-case time in the comment so a slow suite isn't a surprise.

### [nit] Asymmetric handle timeouts starve the second handle
**File:** ConduitUITests/SelectionObserverUITests.swift:101
**Issue:** `anchorHandle` gets a 4 s budget, `focusHandle` a 2 s budget evaluated strictly after. On a slow runner where the anchor appears near the 4 s mark, the focus handle (which typically materializes after the anchor) may legitimately need more than 2 s, causing a spurious retry or false failure.
**Suggestion:** Use symmetric timeouts (e.g., 4 s for both).

### [nit] One family test bypasses the retry helper
**File:** ConduitUITests/SelectionObserverUITests.swift:25
**Issue:** `testLongPressAndDragExtendsSelectionAcrossTableAndCodeBlocks` still does a raw, unretried `anchor.press(...→)`. It can't fail from the drag flake (it only asserts the pasteboard is unchanged), so it's safe, but it's inconsistent with the rest of the family now retrying via the helper.
**Suggestion:** `@discardableResult` — no action strictly needed; optionally route it through the helper for consistency.

Positive observations: the helper returns `Bool` so every caller fails with its own domain message and, with `continueAfterFailure = false`, never proceeds to the copy-button tap / handle asserts on a failed drag (the exact bug from the PR-body table). No force unwraps, no retain cycles, and no concurrency concerns in the Swift-side changes — it's plain synchronous XCUITest code.

## Verdict
REQUEST_CHANGES

The Swift-side changes are sound and clearly improve domain-specific failure reporting. But the CI layer — half the PR's stated purpose — is broken: the `set -e`/`pipefail` interaction makes the attempt-2 retry unreachable, and even when fixed, re-running only the UI target on a full-suite failure can turn genuine unit-test regressions green. Both must be addressed before merge.

<a href="https://opencode.ai/s/fUbGTC8z">New%20session%20-%202026-08-16T03%3A23%3A56.993Z</a>
[opencode session](https://opencode.ai/s/fUbGTC8z)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31924098753)

kaishi00:
Thanks — the review ran against the earlier unit/UI-split workflow, which is already gone: CI hung 35+ minutes on it (the composer paste unit tests stall against the pasteboard XPC service on a cold simulator when the run skips UI-runner preparation — reproduced locally on an erased simulator with `-only-testing:ConduitTests`, passes warm). 9294bbe reverted to the single combined invocation CI has run for months; the retry attempt reruns `-only-testing:ConduitUITests` only. That moots the build-doubling finding and the dead `exit ${PIPESTATUS[0]}` nit (the loop deliberately runs under `set -uo pipefail` without `-e` so `PIPESTATUS` is live).

Taken from the review regardless, in a236605:

- **Sim reset between attempts** — right call, and the local evidence is stronger than the PR text suggested: a wedged simulator here is *deterministic* (two consecutive full UI-suite runs wedge at an identical ~963s wall time; only a reboot clears it). The retry attempt now shuts down and boots the device first, so it starts from a clean simulator rather than re-running into the same wedge.
- **Clean-state drag retries** — taken: each retry after the first relaunches the fixture app, so the synthesized long-press never runs against a partial selection or a stale edit menu. Worst case for a genuinely broken feature is ~35s, noted in the doc comment. Result bundles were already distinct per attempt in the rework (the `rm -rf` only clears a bundle's own previous incarnation).
- **Symmetric handle timeouts** — both endpoint handles now get 4s.

kaishi00:
## Summary
This PR hardens the flaky `SelectionObserverUITests` family in two layers: the Swift-side drag helper now runs up to 3 attempts, relaunches the fixture between retries for a clean state, waits for both endpoint handles symmetrically, and returns success; the CI workflow retries the UI test target once via a two-attempt loop that reboots the simulator between attempts.

## Findings

### [critical] The CI retry loop is dead — GitHub's default errexit aborts the step before attempt 2 ever runs
**File:** .github/workflows/ci.yml:80-118
**Issue:** GitHub wraps every `run` script as `bash --noprofile --norc -e -o pipefail {0}`. The `set -uo pipefail` in the script does **not** disable the inherited `-e`/errexit (only `set +e` would). So when attempt 1's pipeline `xcodebuild test | xcpretty --color` fails, `pipefail` propagates a non-zero status and errexit terminates the shell immediately — `status=${PIPESTATUS[0]}` (line 111), the `if` (line 113), and the next loop iteration are never reached. The retry — half the PR's stated purpose — is a no-op on CI; attempt-1 failure still fails the build exactly as before. The PR author's rebuttal claims the loop runs "without `-e`", but that's only true in a local shell; on the runner `-e` is on by default. This is likely why local verification (plain `bash`) showed retries working.
**Suggestion:** Disable errexit explicitly, e.g. `set +e -uo pipefail`, or guard the pipeline: `if xcodebuild test ... | xcpretty --color; then status=0; else status=${PIPESTATUS[0]}; fi`.

### [major] A successful UI-only retry can turn a genuine unit-test regression green
**File:** .github/workflows/ci.yml:82-118
**Issue:** Attempt 1 runs the full suite (unit + UI). If failure came from a *unit* test while UI tests passed, attempt 2 reruns only `-only-testing:ConduitUITests` (line 86, 103) and `exit 0` (line 113-114) — the job is green and, because the step succeeded, the `if: failure()` artifact upload also never fires. This silently masks real product regressions, the most dangerous CI failure mode.
**Suggestion:** Only greenlight the retry when attempt-1 failures were confined to `ConduitUITests` — e.g. classify the failing tests from attempt 1's xcresult (`xcrun xcresulttool get test-results ...`) and `exit 1` if any `ConduitTests` test failed, or rerun the full suite on the retry step.

### [minor] Simulator reset is best-effort and swallowed
**File:** .github/workflows/ci.yml:90-92
**Issue:** Both `xcrun simctl shutdown` and `xcrun simctl boot` are suffixed `|| true`, so a wedged device that refuses to shut down leaves the retry running against the very state it was meant to clear. `simctl boot` also returns when boot *starts*, not when the device is ready; the subsequent `xcodebuild` then waits on the same half-booted device. Forcing the boot on an ephemeral runner VM mostly works, but the failed-shutdown path is exactly the wedge case this is meant to fix.
**Suggestion:** Either fail loudly when reset fails (`xcrun simctl shutdown "$SIMULATOR"` without `|| true`, exit on error) or wait for readiness with `xcrun simctl bootstatus "$SIMULATOR" -b` before attempt 2.

### [minor] Stale XCUIElement handles depend on re-resolution across relaunch
**File:** ConduitUITests/SelectionObserverUITests.swift:104-115
**Issue:** `anchorHandle`/`focusHandle` are created once (lines 104-105) and reused across `app.terminate()` + `app.launch()`. Reusing an element query across a process relaunch relies on XCUITest re-resolving the query against the new session; in practice this usually works, but a snapshot error during the process-switch window can throw (aborting the retry) rather than return false. `dragDestination`/`pressAnchor` coordinates captured pre-relaunch are assumed to stay valid because the fixture renders identically — reasonable but unverified.
**Suggestion:** Re-create the handle queries (or re-derive coordinates) after each relaunch, and/or wrap the post-relaunch wait so a transient snapshot failure is treated as a retry rather than a test crash.

### [nit] One family test bypasses the retry helper
**File:** ConduitUITests/SelectionObserverUITests.swift:25
**Issue:** `testLongPressAndDragExtendsSelectionAcrossTableAndCodeBlocks` still performs a raw, unretried `anchor.press(...)`. It is a negative control (asserts pasteboard unchanged), so the flake risk is bounded, but it's inconsistent with the rest of the family now going through `performCrossBlockDrag`.
**Suggestion:** Optionally route it through the helper for consistency (its pasteboard-unchanged assertion is then simply run after a verified selection).

Positive: the helper returning `Bool` means each caller fails with its own domain message before tapping the copy button or asserting handles (the exact bug in the PR-body table); the symmetric 4s timeouts fix the earlier starvation nit; per-attempt result bundles are distinct so both upload on a double failure; no force unwraps, no `self` captures/retain cycles, and no concurrency concerns in the Swift changes.

## Verdict
REQUEST_CHANGES

The Swift-side changes are sound, but the CI half is not functional: GitHub's default `-e` makes the retry unreachable (the rebuttal premise that errexit is off is incorrect — only a local shell behaves that way), and even once reachable, the UI-only retry can mask unit-test failures as green. Both must be fixed before merge.

<a href="https://opencode.ai/s/oIAd5hVi">New%20session%20-%202026-08-16T04%3A13%3A56.988Z</a>
[opencode session](https://opencode.ai/s/oIAd5hVi)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31926091253)

PR #67, opened 2026-08-16
fix(connection): revive dashboard bridge after web content process death (overnight stuck-reconnecting)

kaishi00:
## Problem

Waking the phone after the app sat suspended overnight: Conduit shows "Reconnecting…" forever; force-closing and reopening works fine. Reported on TestFlight build 130, and — after checking — **still reproducible on `main` with #65 and #66 merged**. It is a third, distinct wedge.

## Root cause

The dashboard bridge hosts a hidden `WKWebView` to mint WebSocket tickets. Under memory pressure, iOS reclaims the web **content** process while the app is suspended overnight. Three facts combine into a permanent silent hang:

1. The bridge's `isReady` flag is in-memory and survives the suspension — it still says "ready" from yesterday.
2. Nothing observed the process death: the bridge never implemented `webViewWebContentProcessDidTerminate`.
3. `requestJSON`'s only timeout lives **inside the injected JavaScript** (`AbortController`), which never runs in a dead web view. The Swift continuation had no deadline.

So the morning reconnect goes: mint ticket → readiness passes instantly (stale flag) → `evaluateJavaScript` into a dead web view → response never arrives → **the await never returns**. No error is thrown, so no retry is scheduled; the reconnect state machine is simply suspended forever. Force-close is the only recovery — a fresh process gets a fresh web view.

Reproduced deterministically in the simulator against a chaos gateway: kill the content process **while the app is backgrounded**, drop the socket, foreground the app → main makes **zero network requests indefinitely** (log shows the socket death, then silence). Killing the process while foregrounded recovers in seconds, which is why this only surfaces after real overnight suspension.

## Fix — two layers

- **`webViewWebContentProcessDidTerminate`** (the proper recovery): marks the bridge cold-but-reloadable (`isReady=false`, `isLoadFailed=true`) and resumes any pending `requestJSON` with `.notReady` — flowing directly into `mintTicket`'s existing reload recovery from #65, which reloads the page and revives the web view.
- **Swift-side deadline on every `requestJSON` continuation** (the belt): request timeout + 3s. If a response is ever silently dropped for any other reason (origin-mismatch on a restored blank page, process-relaunch races), the hang becomes a retryable failure instead of wedging ticket minting forever. All failure-side resume paths funnel through one helper that cancels the deadline first; `rejectPending` cancels all of them.

## Verification

The exact overnight scenario against the fixed build: background → kill content process → drop socket → foreground → **ticket minted, socket reopened, session catalog + transcript restored ~1s after foreground** (vs. never on main). Screenshot in the investigation; unit suite **669/0** including two new regression tests (termination → reloadable state + mint retries the dead page; termination → pending requests resume as `.notReady`).

Fixes the "morning stuck-on-reconnecting" report from build 130 alongside #65 (which fixed the cold-bridge and watchdog wedges); #66 was CI-only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added timeout handling for in-progress dashboard requests.
  * Improved cleanup when requests are canceled, rejected, or completed.
  * Added recovery when the embedded web content process terminates.
  * Pending requests now fail promptly, and the dashboard can recover after reloading.

### DIFF
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -254,6 +254,11 @@ final class DashboardTicketBridge: NSObject {
     }
 
     deinit {
+        // Stored-property access and Task.cancel() are safe from nonisolated
+        // deinit; cancelling here stops deadline tasks from outliving the
+        // bridge they were created to guard.
+        for deadline in requestDeadlines.values { deadline.cancel() }
+        requestDeadlines.removeAll()
         pendingRequests.rejectAll(with: DashboardTicketBridgeError.notReady)
         webView.configuration.userContentController.removeScriptMessageHandler(forName: "dashboard-response")
     }
@@ -262,7 +267,7 @@ final class DashboardTicketBridge: NSObject {
         guard !isInvalidated else { return }
         isInvalidated = true
         isReady = false
-        pendingRequests.rejectAll(with: DashboardTicketBridgeError.notReady)
+        rejectPending(with: DashboardTicketBridgeError.notReady)
     }
 
     func reload() {
@@ -384,9 +389,14 @@ final class DashboardTicketBridge: NSObject {
                 requestDeadlines[id] = Task { [weak self] in
                     try? await Task.sleep(for: .milliseconds(timeout + 3_000))
                     guard !Task.isCancelled else { return }
+                    // .notReady (not .requestFailed) so the failure funnels
+                    // into mintTicket's internal reload-retry instead of
+                    // surfacing past it — a timed-out JavaScript response is
+                    // almost always a dead or stalled web view, i.e. exactly
+                    // the state reloading revives.
                     self?.failPendingRequest(
                         id: id,
-                        with: DashboardTicketBridgeError.requestFailed("Dashboard request timed out.")
+                        with: DashboardTicketBridgeError.notReady
                     )
                 }
                 let script = """
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -203,6 +203,12 @@ final class DashboardTicketBridge: NSObject {
     private var isInvalidated = false
     private var requestID = 0
     private let pendingRequests: DashboardTicketBridgePendingRequests
+    /// Swift-side deadlines for in-flight `requestJSON` continuations. The
+    /// JavaScript timeout only fires inside a live web view; if the content
+    /// process is gone (or a response message is dropped), nothing but these
+    /// deadlines resumes the continuation — without them a single lost
+    /// response hangs ticket minting, and therefore reconnecting, forever.
+    private var requestDeadlines: [Int: Task<Void, Never>] = [:]
     private let readinessPollAttempts: Int
     private let readinessPollInterval: Duration
     /// Number of times `reload()` re-attempted the dashboard page load, from
@@ -369,6 +375,20 @@ final class DashboardTicketBridge: NSObject {
                 }
 
                 pendingRequests.insert(continuation, for: id)
+                // Swift-side deadline: the JavaScript AbortController above
+                // only runs inside a live web view. If the content process
+                // is gone or a response message is dropped, this is the only
+                // thing that resumes the continuation — convert the silent
+                // hang into a failure the reconnect loop can retry through
+                // the reload path.
+                requestDeadlines[id] = Task { [weak self] in
+                    try? await Task.sleep(for: .milliseconds(timeout + 3_000))
+                    guard !Task.isCancelled else { return }
+                    self?.failPendingRequest(
+                        id: id,
+                        with: DashboardTicketBridgeError.requestFailed("Dashboard request timed out.")
+                    )
+                }
                 let script = """
                 (async function() {
                     try {
@@ -428,8 +448,8 @@ final class DashboardTicketBridge: NSObject {
                 true;
                 """
                 webView.evaluateJavaScript(script) { _, error in
-                    guard let error, let pending = self.pendingRequests.removeValue(for: id) else { return }
-                    pending.resume(throwing: error)
+                    guard let error else { return }
+                    self.failPendingRequest(id: id, with: error)
                 }
             }
         }, onCancel: {
@@ -439,9 +459,17 @@ final class DashboardTicketBridge: NSObject {
         })
     }
 
-    private func cancelPendingRequest(id: Int) {
+    /// Cancels the Swift-side deadline and resumes the pending request, if
+    /// any, with `error`. All failure-side resume paths funnel through here
+    /// so the deadline task can never fire on an already-resumed request.
+    private func failPendingRequest(id: Int, with error: Error) {
+        requestDeadlines.removeValue(forKey: id)?.cancel()
         guard let continuation = pendingRequests.removeValue(for: id) else { return }
-        continuation.resume(throwing: CancellationError())
+        continuation.resume(throwing: error)
+    }
+
+    private func cancelPendingRequest(id: Int) {
+        failPendingRequest(id: id, with: CancellationError())
     }
 
     private func javaScriptLiteral(_ value: Any) throws -> String {
@@ -485,6 +513,8 @@ final class DashboardTicketBridge: NSObject {
     }
 
     private func rejectPending(with error: Error) {
+        for deadline in requestDeadlines.values { deadline.cancel() }
+        requestDeadlines.removeAll()
         pendingRequests.rejectAll(with: error)
     }
 }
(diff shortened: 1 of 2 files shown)


kaishi00:
## Summary

This PR fixes a permanent "Reconnecting…" hang where iOS reclaims the hidden `WKWebView`'s content process overnight: it implements `webViewWebContentProcessDidTerminate` to mark the bridge cold-but-reloadable and resume pending requests with `.notReady`, and adds a Swift-side deadline to every `requestJSON` continuation so a lost web-process response can no longer suspend the continuation forever.

## Findings

### [minor] Deadline resumes with `.requestFailed`, bypassing mintTicket's reload/retry recovery
**File:** Conduit/Services/DashboardTicketBridge.swift:389
**Issue:** The belt-timer resumes the continuation with `DashboardTicketBridgeError.requestFailed("Dashboard request timed out.")`. `mintTicket`'s retry loop only catches `.notReady` and `.signInRequired` (lines 306–316), so this error propagates straight out of `mintTicket` into `AppState` (which shows "Failed to refresh the dashboard session: …" and calls `scheduleReconnect`). The termination path, by contrast, resumes with `.notReady` and flows through `mintTicket`'s internal reload-retry. This is an inconsistency with the PR's stated design ("converts the silent hang into a retryable failure… through the reload path"), and in the "callback lost but page dead" corner it skips the reload attint.
**Suggestion:** Resume the deadline with `.notReady` (same as the termination path) so the failure funnels through `mintTicket`'s existing `attempt < 2` retry (which reloads when `isLoadFailed`) instead of surfacing immediately at the top level. If `.requestFailed` is intentional, document that the reload recovery intentionally doesn't apply.

### [minor] `invalidate()` and `deinit` don't cancel `requestDeadlines` tasks (inconsistent cleanup)
**File:** Conduit/Services/DashboardTicketBridge.swift:261-266 and 256-259
**Issue:** `rejectPending` (lines 515–518) cancels and clears `requestDeadlines` before rejecting, but `invalidate()` and `deinit` call only `pendingRequests.rejectAll(...)`, so each in-flight request's deadline task keeps sleeping up to `timeout + 3s` after invalidation. It's memory/CPU-safe (the task holds `weak self`, and `failPendingRequest`'s `removeValue` nil-guard prevents a double resume), but it's dead state kept alive longer than necessary and inconsistent with the new funnel.
**Suggestion:** Route `invalidate()`/`deinit` through the same helper as `rejectPending` (cancel all deadlines, then `rejectAll`), or call `rejectPending(with:)` directly.

### [nit] Deadline task serialized on MainActor
**File:** Conduit/Services/DashboardTicketBridge.swift:384
**Issue:** The `Task { [weak self] in … }` inherits MainActor isolation from `requestJSON`, so deadline bookkeeping and `failPendingRequest` hop through the main actor. This is correct (all `pendingRequests`/`requestDeadlines` access is MainActor-confined), just slightly heavier than a detached nonisolated timer would be. Not a defect — worth a comment explaining the intentional MainActor inheritance since `DashboardTicketBridgePendingRequests` is `nonisolated`.

### [nit] Test models termination callback directly, not the delivery path
**File:** ConduitTests/DashboardTicketBridgeTests.swift:115 and 154
**Issue:** `testWebContentTermination*` call `bridge.webViewWebContentProcessDidTerminate(bridge.webView)` directly, which is a reasonable injection seam given untestable WebKit callbacks, but neither test exercises the deadline fallback (the other half of the fix), and one test inserts a continuation manually that bypasses `requestJSON`'s deadline registration.
**Suggestion:** Consider adding a test that drives a real `requestJSON` to a silent-drop (e.g., never posting the JS response) and asserts the deadline resolves it with the timeout error within ~`timeout + 3s`, otherwise the belt layer ships unverified.

## Verdict

APPROVE — the core fix is sound: no retain cycles (`weak self` in the deadline task, weak proxy already correct), no force unwraps introduced, all `pendingRequests`/`requestDeadlines` access is MainActor-confined with safe single-resume funneling, and the two new regression tests cover the reported wedge. Findings above are minor/nit and don't block the fix.

<a href="https://opencode.ai/s/c5L2AHAj">New%20session%20-%202026-08-16T11%3A30%3A26.401Z</a>
[opencode session](https://opencode.ai/s/c5L2AHAj)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31944451538)

kaishi00:
Review follow-ups in 144d0ea — first two taken, nit declined:

- **Deadline error → **: agreed, the inconsistency was real. A timed-out JavaScript response is almost always a dead or stalled web view — exactly the state the reload recovery exists for — so the belt now funnels into `mintTicket`'s internal retry like the termination path does, instead of bypassing it.
- **Full deadline cleanup**: `invalidate()` now routes through `rejectPending` and `deinit` cancels the deadline tasks inline (stored-property access and `Task.cancel()` are safe from nonisolated deinit), so no deadline outlives the bridge.
- **MainActor deadline task** (nit, declined): the sleep wake is one main-queue hop every ~15s per in-flight request, and keeping the task on the actor means `failPendingRequest` runs without a hop when it fires. A detached task would save nothing measurable and add a hop on the firing path.

669/0.

kaishi00:
## Summary

Fixes the "stuck-on-reconnecting" hang after overnight suspension by implementing `webViewWebContentProcessDidTerminate` (marks the bridge cold-but-reloadable and resumes pending `requestJSON` continuations with `.notReady` so `mintTicket` reload-retries) and adding a Swift-side deadline (`timeout + 3s`) to every `requestJSON` continuation so a dropped JavaScript response can never suspend the await forever. All failure paths funnel through a single `failPendingRequest` helper that cancels the deadline.

## Findings

### [minor] `reload()` still bypasses the deadline-canceling funnel
**File:** Conduit/Services/DashboardTicketBridge.swift:277
**Issue:** The follow-up commit routed `invalidate()` through `rejectPending` and canceled deadlines in `deinit`, but `reload()` still calls `pendingRequests.rejectAll(with: .notReady)` directly. Each resumed continuation's deadline task therefore keeps sleeping up to `timeout + 3s` after a reload. It's memory-safe (weak self, and `failPendingRequest`'s `removeValue` nil-guard prevents a double-resume), but it contradicts the PR's stated design — "All failure-side resume paths funnel through [failPendingRequest]" — and leaves dead task state alive after every reload-triggered rejection.
**Suggestion:** Change `reload()` to `rejectPending(with: DashboardTicketBridgeError.notReady)`, matching `invalidate()`.

### [minor] Belt-only scenario stalls ~45s before reloading because `isLoadFailed` stays `false`
**File:** Conduit/Services/DashboardTicketBridge.swift:314-320, 595-608
**Issue:** Retry only reloads `if isLoadFailed` (line 317). In the exact belt scenario — content process dead but the termination callback missed/undelivered — that flag is still `false`, so each of the up-to-3 retry attempts sleeps through a full `timeout + 3s` deadline and re-runs `evaluateJavaScript` into the dead process without reloading, stalling ~45s before `.notReady` finally propagates to the top level. The callback path recovers quickly (seconds); the belt path only converts a forever-hang into a long, retryable stall.
**Suggestion:** In `mintTicket`'s `.notReady` catch, treat a Swift-deadline-observed timeout as reloadable too (e.g. reload after a deadline-derived `.notReady` even when `isLoadFailed` is false), or accept and document this worst-case latency as intended belt behavior.

### [minor] Belt deadline path ships without a regression test
**File:** ConduitTests/DashboardTicketBridgeTests.swift:99-170
**Issue:** Both new tests drive the termination callback directly, and the pending-request test inserts a continuation manually (id 99) that bypasses `requestJSON`'s deadline registration entirely. Nothing exercises the other half of the fix — the Swift deadline firing on a silently dropped response — so a regression in the deadline/funnel logic would go uncaught.
**Suggestion:** Add a test that calls a real `requestJSON` against a bridge that never posts a `dashboard-response` (small `timeoutMilliseconds`) and asserts the continuation resolves with the expected error within ~`timeout + 3s`. Note the deadline error acknowledged by the author is `.notReady`, which the JS-Abort path can't reproduce, so the test should simulate a silent drop rather than a JS-reported failure.

### [nit] Swift-side deadline error is `.notReady`, which is semantically "still loading", not "timed out"
**File:** Conduit/Services/DashboardTicketBridge.swift:392-400
**Issue:** `DashboardTicketBridgeError.notReady` is described as "The dashboard session is still loading", but here it signals a true timeout. This is intentional (to funnel through the `.notReady` retry), and the comment documents it well, but it overloads `.notReady`'s meaning and means `waitUntilReady`-style polling semantics are borrowed for a timeout condition.
**Suggestion:** No change required; consider a dedicated case (e.g. `.requestTimedOut`) mapped into the retry path if the overloaded semantics ever become confusing. Non-blocking.

No retain cycles: the deadline task uses `[weak self]`, `WeakScriptMessageHandler` already breaks the WebKit cycle, and the new stored deadline/Task properties are all `@MainActor`-confined with single-resume guaranteed by `failPendingRequest`. No force unwraps introduced. No UI/accessibility surface affected (the web view is invisible). Noted positives: termination correctly resets `isLoadFailed = true` and grounds `didLandOnLogin` so a reloaded page re-asserts the real landing, and `deinit`/`invalidate` cleanup is now consistent.

## Verdict

APPROVE

<a href="https://opencode.ai/s/PJwoZLE2">New%20session%20-%202026-08-16T12%3A03%3A07.432Z</a>
[opencode session](https://opencode.ai/s/PJwoZLE2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31945946582)

kaishi00:
Fair hit on merging before your review landed — my process error. Addressed in #68: the deadline-expiry item was the substantive one (the belt resumed `.notReady` but left the bridge looking ready, so mint's retry never reloaded the stalled view); expiry now marks it cold-but-reloadable with a regression test proving the reload happens. reload()'s raw rejectAll (the last uncleaned path — your invalidate/deinit anchors were already fixed in 144d0ea) now routes through rejectPending, and the grace addition is overflow-safe + injectable. 670/0.

PR #68, opened 2026-08-16
fix(connection): deadline expiry marks stalled web view for reload (CodeRabbit follow-up to #67)

kaishi00:
Follow-up to #67 addressing the CodeRabbit review that landed after merge.

**The substantive one — deadline expiry must mark the web view stalled.** The #67 belt resumed timed-out requests with `.notReady`, but left `isReady=true`/`isLoadFailed=false` — so `mintTicket`'s retry saw a "ready" bridge and re-evaluated JavaScript against the *same* stalled view instead of reloading it. In the exact silent-drop corner the belt exists for, that converts "hang forever" into "retry against a dead view forever" rather than recovering. A live page always answers before the deadline (its in-JS `AbortController` posts a response message at the request timeout), so deadline expiry reliably proves the view is stalled — expiry now marks it cold-but-reloadable before resuming, and the next mint attempt reloads it. New regression test: a simulated `.ready` bridge whose response is dropped → deadline expires as `.notReady` → assert mint reloaded the stalled page.

**`reload()` deadline cleanup** — the last raw `rejectAll` path (the review's `invalidate`/`deinit` anchors were already fixed in #67's `144d0ea`); rejections now cancel their deadline tasks immediately everywhere.

**Overflow-safe, injectable grace** — the `timeout + 3s` addition is clamped, and the grace is an init parameter so the regression test runs in ~1s instead of 15.

Unit suite: **670/0**.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of expired requests with safer deadline grace periods.
  - Detects stalled web views, marks them for recovery, and reloads them when necessary.
  - Ensures pending requests are properly cleared during web view recovery.
  - Resets readiness and login state correctly during ready-state simulations.

- **Tests**
  - Added coverage for request expiration, stalled web views, recovery reloads, and related failure scenarios.

### DIFF
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -197,12 +197,14 @@ final class DashboardTicketBridge: NSObject {
     private var simulatedLanding: SimulatedLanding?
 
     enum SimulatedLanding {
+        case ready
         case loginPage
         case loadFailure
     }
     private var isInvalidated = false
     private var requestID = 0
     private let pendingRequests: DashboardTicketBridgePendingRequests
+    private let requestDeadlineGraceMilliseconds: Int
     /// Swift-side deadlines for in-flight `requestJSON` continuations. The
     /// JavaScript timeout only fires inside a live web view; if the content
     /// process is gone (or a response message is dropped), nothing but these
@@ -221,7 +223,8 @@ final class DashboardTicketBridge: NSObject {
         cloudflareAccess: CloudflareAccessCredentials? = nil,
         pendingRequests: DashboardTicketBridgePendingRequests = DashboardTicketBridgePendingRequests(),
         readinessPollAttempts: Int = 30,
-        readinessPollInterval: Duration = .milliseconds(100)
+        readinessPollInterval: Duration = .milliseconds(100),
+        requestDeadlineGraceMilliseconds: Int = 3_000
     ) {
         let normalizedBaseURL = (try? ConnectionURLPolicy.normalizedBaseURL(baseURL)) ?? ""
         self.baseURL = normalizedBaseURL
@@ -230,6 +233,7 @@ final class DashboardTicketBridge: NSObject {
         // A negative count would build an invalid Range in the polling loops.
         self.readinessPollAttempts = max(0, readinessPollAttempts)
         self.readinessPollInterval = readinessPollInterval
+        self.requestDeadlineGraceMilliseconds = max(0, requestDeadlineGraceMilliseconds)
         let configuration = WKWebViewConfiguration()
         configuration.websiteDataStore = .default()
         if let script = cloudflareAccess?.fetchInjectionUserScript(expectedBaseURL: normalizedBaseURL), !script.isEmpty {
@@ -274,7 +278,7 @@ final class DashboardTicketBridge: NSObject {
         guard !isInvalidated else { return }
         reloadCount += 1
         isReady = false
-        pendingRequests.rejectAll(with: DashboardTicketBridgeError.notReady)
+        rejectPending(with: DashboardTicketBridgeError.notReady)
         Task { [weak self] in
             guard let self else { return }
             await DashboardCookiePersistence.restore(into: self.webView.configuration.websiteDataStore.httpCookieStore)
@@ -386,8 +390,12 @@ final class DashboardTicketBridge: NSObject {
                 // thing that resumes the continuation — convert the silent
                 // hang into a failure the reconnect loop can retry through
                 // the reload path.
+                // A live page answers before this fires (its own 12s abort
+                // posts a response message), so expiry means the view is
+                // stalled and the request marks it for reload below.
+                let deadlineDelayMilliseconds = deadlineDelay(afterMilliseconds: timeout)
                 requestDeadlines[id] = Task { [weak self] in
-                    try? await Task.sleep(for: .milliseconds(timeout + 3_000))
+                    try? await Task.sleep(for: .milliseconds(deadlineDelayMilliseconds))
                     guard !Task.isCancelled else { return }
                     // .notReady (not .requestFailed) so the failure funnels
                     // into mintTicket's internal reload-retry instead of
@@ -396,7 +404,8 @@ final class DashboardTicketBridge: NSObject {
                     // the state reloading revives.
                     self?.failPendingRequest(
                         id: id,
-                        with: DashboardTicketBridgeError.notReady
+                        with: DashboardTicketBridgeError.notReady,
+                        markingWebViewStalled: true
                     )
                 }
                 let script = """
@@ -472,10 +481,29 @@ final class DashboardTicketBridge: NSObject {
     /// Cancels the Swift-side deadline and resumes the pending request, if
     /// any, with `error`. All failure-side resume paths funnel through here
     /// so the deadline task can never fire on an already-resumed request.
-    private func failPendingRequest(id: Int, with error: Error) {
+    private func failPendingRequest(
+        id: Int,
+        with error: Error,
+        markingWebViewStalled: Bool = false
+    ) {
         requestDeadlines.removeValue(forKey: id)?.cancel()
         guard let continuation = pendingRequests.removeValue(for: id) else { return }
+        if markingWebViewStalled {
+            // Only reached when the request deadline expired: a live page
+            // answers before that (its in-JS abort posts a response), so
+            // expiry proves the view is stalled. Mark it cold-but-reloadable
+            // so mintTicket's next attempt reloads instead of retrying
+            // evaluateJavaScript against the same dead view.
+            isReady = false
+            isLoadFailed = true
+        }
         continuation.resume(throwing: error)
+    }
+
+    /// Deadline delay with an overflow-safe grace addition.
+    private func deadlineDelay(afterMilliseconds timeout: Int) -> Int {
+        let grace = requestDeadlineGraceMilliseconds
+        return min(timeout, Int.max - grace) + grace
     }
 
     private func cancelPendingRequest(id: Int) {
@@ -498,6 +526,10 @@ final class DashboardTicketBridge: NSObject {
     func simulateLandingForTesting(_ landing: SimulatedLanding) {
         simulatedLanding = landing
         switch landing {
+        case .ready:
+            isReady = true
+            isLoadFailed = false
+            didLandOnLogin = false
         case .loginPage:
             isReady = false
             isLoadFailed = false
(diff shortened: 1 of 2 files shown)


kaishi00:
## Summary
This PR makes Swift-side request deadline expiry mark the web view as stalled (`isReady=false`, `isLoadFailed=true`) so `mintTicket`'s retry reloads a dead view instead of re-evaluating JavaScript against it forever, funnels `reload()`'s rejections through deadline-cancelling `rejectPending`, and makes the deadline grace period clampable/injectable.

## Findings

### [major] Regression test does not actually exercise the expiry belt — and its stated ~1s runtime doesn't hold
**File:** ConduitTests/DashboardTicketBridgeTests.swift:175
**Issue:** The test asserts on `reloadCount` but never proves the deadline fired. `mintTicket()` calls `requestJSON(path:method:)` with the default `timeoutMilliseconds = 12_000` (DashboardTicketBridge.swift:310,357), so the Swift deadline fires at 12,050ms, not ~1s — the injectable `requestDeadlineGraceMilliseconds: 50` only trims the OLD 15s (`timeout + 3s`) add, it does not shrink the 12s base. The test therefore either (a) blocks ~15s waiting for the belt (contradicting the PR's "runs in ~1s" claim), or (b) on hosts where `evaluateJavaScript` returns an immediate error, `mintTicket` throws that error and the broad `catch { }` swallows it — the test passes with **zero** coverage of the stall-marking path it was written to protect. In case (b), `mintTicket`'s retry exhaustion can even throw `.notReady` without the web view ever being marked stalled. The assertion `reloadCount >= 1` cannot distinguish these outcomes.
**Suggestion:** Drive the belt deterministically instead of relying on wall-clock/the 12s default. Thread an injectable `timeoutMilliseconds` through `mintTicket` → `requestJSON` (default 12_000) and pass a small value (e.g. 100–200ms) in the test, then assert the intent directly: `XCTAssertFalse(bridge.isReady)` and `XCTAssertTrue(bridge.isLoadFailed)` after expiry, plus `XCTAssertGreaterThanOrEqual(bridge.reloadCount, 1)`. Also, only swallow the immediate-`evaluateJavaScript`-error case when `reloadCount == 0` was never reached, or remove the catch-all reassignment — otherwise a regressed belt still passes silently.

### [minor] Broad catch-all can mask the very regression this test exists to catch
**File:** ConduitTests/DashboardTicketBridgeTests.swift:193
**Issue:** The `catch { }` branch accepts any non-`.notReady` error and performs no assertion. If the deadline belt is broken so that expiry stops marking the view stalled (the #67 bug class this PR fixes), `mintTicket`'s retry loop still eventually surfaces `.notReady` on most hosts, but on a host that fails `evaluateJavaScript` immediately the test passes untouched — the guard this test adds never fails. Combined with the unasserted 15s path, both branches under-verify.
**Suggestion:** Assert the stalled state unconditionally (`isLoadFailed == true`, `isReady == false`) after `mintTicket()` throws, regardless of which exception was observed, so a host that short-circuits still validates the marking behavior.

### [minor] Stall comment overstates the invariant; deadline firing plus delayed inbound response forces an unnecessary full reload
**File:** Conduit/Services/DashboardTicketBridge.swift:393-395 (and 491-499)
**Issue:** The comment claims "a live page always answers before this fires" and "expiry proves the view is stalled." The deadline is a wall-clock `Task.sleep` racing a main-actor-routed script message. A response message can be inbound (already in `WKScriptMessageHandler`) yet dispatched to the main actor just after the 15s deadline task runs, so the deadline can win on a *live* view: the continuation is resumed with `.notReady`, and `isLoadFailed=true` + `isReady=false` now force a full page reload (aborting all in-flight requests via `rejectPending`) for a request that actually succeeded. Per-request it's a rare, bounded race, but it is a behavior change from the pre-#68 `rejectAll` path where expiry resumed `.notReady` with *no* state mutation, and #67 had already made the same state nullable only via the navigation callbacks. Worth a one-line caveat that the flag is "generally stalled / force-reload" rather than proof.
**Suggestion:** Either broaden the deadline's grace margin (e.g. keep a larger constant add so genuinely-live-but-slow views answer first) or soften the comment to acknowledge expiry fires on a scheduler-delayed-live-page edge case, and note the worst case is one extra reload.

### [minor] `markingWebViewStalled` state set is gated behind the pending-request lookup while a matching deadline ran
**File:** Conduit/Services/DashboardTicketBridge.swift:489-500
**Issue:** When a deadline fires while the response ALSO completed (partial race loss), `pendingRequests.removeValue(for:)` returns nil and the early `guard` skips the `isReady/isLoadFailed` marking entirely — reasonable for a genuinely-successful response (marking would be wrong) — but it also skips it for the cancelled-then-deadline case. Given a reload/cancel funnels `rejectPending`/`cancel` first, this is defensible, but the ordering means any host where the message-handler and deadline contend can leave a truly-dead view unmarked (assertion in test also won't reproduce it reliably).
**Suggestion:** Consider deriving the stall mark from expiry alone (the deadline only reaches this code path on expiry; `requestDeadlines.removeValue(for:)?.cancel()` already ran above), and set the flags before the continuation-lookup guard — the resume is still gated by the guard while the state reflects the expiry independently of whether a continuation remains.

## Verdict
REQUEST_CHANGES

The production fix is correct and meaningful: `reload()` now cancels deadline tasks immediately (`rejectPending` replaces raw `rejectAll`), the overflow-safe `deadlineDelay` addition and `max(0, ...)` clamping are sound, `weak self`/`@MainActor` discipline is clean, and the stall mark genuinely converts the overnight dead-view hang into a reloadable retry. But the headline regression test does not deterministically exercise the belt (it either blocks ~15s on the fixed 12s request timeout or silently passes via the catch-all path), so it does not protect the fix it claims to enforce.

<a href="https://opencode.ai/s/U3QIH8Il">New%20session%20-%202026-08-16T14%3A03%3A22.801Z</a>
[opencode session](https://opencode.ai/s/U3QIH8Il)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31951538695)

kaishi00:
## Summary
Follow-up to #67: Swift-side request deadline expiry now marks the stalled web view cold-but-reloadable (`isReady=false`, `isLoadFailed=true`) so `mintTicket`'s retry reloads it instead of re-evaluating JS against the dead view forever; `reload()` funnels its rejections through deadline-cancelling `rejectPending`; the deadline grace is clamped and injectable; plus a deterministic regression test.

## Findings

### [minor] "Expiry proves the view is stalled" overstates the invariant — a live page can be force-reloaded
**File:** Conduit/Services/DashboardTicketBridge.swift:393-395, 491-499
**Issue:** The deadline is a wall-clock `Task.sleep` racing a main-actor-routed postMessage. A response already dispatched to the main actor can lose to the deadline task, so expiry can win on a *live* view: the continuation is resumed with `.notReady` and the flags now force a full `reload()` (which aborts all in-flight requests via `rejectPending`) for a request that actually succeeded. Post-#68 this is a state mutation where previously expiry resumed `.notReady` with no mutation. It's rare and bounded (one spurious reload), but the "always answers before this fires" comment is stronger than the code guarantees.
**Suggestion:** Soften the comment to acknowledge a scheduler-delayed-response edge case whose worst case is one extra reload, or rely on the 3s grace (already the practical mitigation).

### [minor] Regression test can pass without ever exercising the stall-marking path on some hosts
**File:** ConduitTests/DashboardTicketBridgeTests.swift:204-207
**Issue:** The catch-all accepts any non-`.notReady` error with no assertion. On a host where `evaluateJavaScript` fails immediately, `failPendingRequest(id:with:)` runs *without* `markingWebViewStalled` (AppState will mark nothing), the deadline belt is never reached, and the test still passes. That host is genuinely unable to exercise the fix, but the headline assertion then silently doesn't run — the exact regression class this PR shipped to protect.
**Suggestion:** This is acceptable as a documented host limitation (the deterministic comment at 176-182 covers the navigation-callback flavor), but consider recording which branch ran and asserting the `.notReady` + `isLoadFailed` checks explicitly in that branch, so any future removal of the belt fails loudly rather than vacuously.

### [minor] Stall state persists for non-mint callers and is only cleared by navigation/reload
**File:** Conduit/Services/DashboardTicketBridge.swift:497-498
**Issue:** After expiry, `isLoadFailed=true`/`isReady=false` persist until a navigation event or a mint-driven reload. Many non-mint `requestJSON` callers (AppState.swift:3928, 5700, 6043; HermesVoiceGateway etc.) that hit expiry now inherit a permanently "failed" bridge: their subsequent readiness polls fall out immediately and throw `.notReady` with no auto-reload. It self-heals on the next `mintTicket`, but a single slow-but-alive server response can wedge the whole bridge's dashboard fetches until then.
**Suggestion:** Fine as designed given the overnight-dead-view scenario; worth a comment noting the failure is cleared only by the next successful navigation or mint-triggered reload.

### [nit] Stall-marking correctly gated behind the continuation lookup
**File:** Conduit/Services/DashboardTicketBridge.swift:490-491
**Issue:** Not a defect — flagging for clarity that this is *correct*: if the response already consumed the continuation (deadline task late after the message handler's `requestDeadlines.removeValue(...)?.cancel()`), the page answered, and skipping the stall mark is right. Do not move the marking before the guard.

## Verdict
APPROVE

The production fix is sound and materially fixes the dogged-hang: `reload()` now cancels deadline tasks immediately, `deadlineDelay`'s clamp + overflow-safe addition are correct, all closures use `[weak self]` with deadlines cancelled on every path (message handler, cancellation, `rejectPending`, `deinit`), the `@MainActor`/Swift-6 discipline is clean, and there are no force unwraps or retain cycles. The remaining items are edge-case/documentation minors; the deterministic-test concern from the earlier review was already addressed in `a994874` by driving `requestJSON` directly with an injectable short timeout.

<a href="https://opencode.ai/s/KWuNrXCr">New%20session%20-%202026-08-16T14%3A58%3A36.742Z</a>
[opencode session](https://opencode.ai/s/KWuNrXCr)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31954212851)

kaishi00:
Both findings triaged against how CI's WKWebView actually behaves — declining both with the trace:

**"Require navigation to settle / fail if it doesn't" (minor):** failing the test when `isLoadFailed` never becomes true would break the test host flavor CI has repeatedly exhibited (navigation callbacks never delivered — several passing bridge tests depend on it). The settle loop's purpose isn't to prove settlement; it's to close the window where an in-flight navigation could reject the pending request before the deadline on callback-firing hosts. And that window is closed by timing regardless: the deadline fires at ~1.05s (1s timeout floor + 50ms grace) while CI's observed navigation-failure latency is 1–2s+ (measured in the round-2 CI failure of the earlier conditional test, where the refused-port callback landed after a ~1.1s mint completed). Deadline-first is the only reachable interleaving; a late callback after the deadline finds no pending request and no-ops.

**"Catch-all accepts unrelated errors → vacuous pass" (major):** the vacuous path requires an immediate `evaluateJavaScript` completion error — a flavor never observed on any host in this repo's history (CI shows exactly two flavors: no callbacks at all, or late navigation callbacks; both traced above to exercise the real arm). The fallback arm exists to keep the suite from being hostage to an unobserved third flavor; tightening it to `XCTFail` trades a silent vacuous pass (currently unreachable on CI) for a spurious red build if that flavor ever appears. If you want belt-and-suspenders anyway I'm happy to log which arm executed via `XCTContext` so the xcresult shows it — but failing the build on an unobserved host flavor is the worse trade.

PR #71, opened 2026-08-17
fix(connection): harden bridge recovery state machine (three-model review pass)

kaishi00:
## Summary

Hardening follow-up to the overnight-wedge fixes (#67/#68), produced by a **three-model local review pipeline** (deepseek, step, qwen3 reviewers running as local subagents) over the shipped recovery machinery. Each fix below was found by at least one reviewer and cross-confirmed; the standout was a reviewer-run test reproduction that caught a one-line regression **in this PR's own first draft** before it was ever pushed.

## Fixes

1. **Eval-error funneling** — `evaluateJavaScript` completion errors now always map onto the typed stall path (`.notReady` + stall-marking + flushing sibling pending requests). The injected IIFE already funnels every JS-level failure into the message channel, so any completion error means the engine never ran the script. *(deepseek + qwen independently; consensus fix)*
2. **Login verdict immutability** — `didLandOnLogin` changes only on an origin-matching `didFinish`. The draft cleared it on unknown/foreign landings and failures, but WebKit settles some failed loads as an `about:blank` `didFinish`, flipping a genuine login verdict back to `.notReady` and degrading sign-in recovery into a reconnect loop. *(deepseek, verified by running the failing tests and bisecting to the line)*
3. **Navigation supersession** — `reload()` retires the old navigation into a stale-set immediately; `didFinish` clears identity. Failure callbacks for a replaced load can no longer fail the fresh state — closing both the reload-window race and post-finish suppression. *(all three reviewers, from different angles)*
4. **Stall-marking flushes all pending requests** — recovery latency is one deadline, not one per in-flight request. *(deepseek)*
5. **Synchronous failure-verdict reset in `reload()`** — kills the retry-observes-stale-failure → double-reload cycle. *(step + qwen convergence, deepseek variant)*
6. **Login-park poll fast-exit** — settled login pages never become ready by polling; the `.signInRequired` recovery fires in ~1.1s instead of ~9.7s per exhausted mint. *(qwen)*
7. **Invalidate precedence** — a late navigation callback cannot resurrect `.signInRequired` from a torn-down bridge. *(deepseek)*
8. **Contract documentation + tests** — `requestJSON` error contract documented; new regression tests (invalidate-clears-login, login-park timing), reload-count assertions, deadline-test arm logging.

## Verification

- Bridge suite 8/0 across two consecutive local runs — the login tests that motivated fix 2 flaked on the second run before it, deterministically green after.
- Full-suite validation rides CI (local simulator exhausted for the day — the known churn wedge).
- Reviewer triage: 2 of 14 findings declined with rationale (a data-race claim resting on misreading `@MainActor` isolation; a timing-margin nit superseded by the outcome assertion).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dashboard loading reliability during reloads, navigation changes, and session expiration.
  - Prevented outdated page failures from overriding the current dashboard state.
  - Added faster recovery when page requests stall or browser content fails to load.
  - Improved handling of invalid or incomplete landing pages.
  - Readiness checks now stop promptly when login is required or the dashboard cannot load.
  - Improved session and cookie handling during navigation and reloads.

- **Tests**
  - Added coverage for login redirects, invalidated sessions, and immediate loading failures.

### DIFF
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -203,6 +203,12 @@ final class DashboardTicketBridge: NSObject {
     }
     private var isInvalidated = false
     private var requestID = 0
+    /// Identity of the current dashboard navigation, used to ignore failure
+    /// callbacks for loads a reload has already replaced. `staleNavigations`
+    /// covers the window between a reload and its new load: failures for
+    /// superseded navigations must not fail the fresh state.
+    private var currentNavigation: WKNavigation?
+    private var staleNavigations: Set<WKNavigation> = []
     private let pendingRequests: DashboardTicketBridgePendingRequests
     private let requestDeadlineGraceMilliseconds: Int
     /// Swift-side deadlines for in-flight `requestJSON` continuations. The
@@ -271,6 +277,10 @@ final class DashboardTicketBridge: NSObject {
         guard !isInvalidated else { return }
         isInvalidated = true
         isReady = false
+        // A torn-down bridge must report plain unreadiness: leaving a stale
+        // login verdict would surface .signInRequired (with its session
+        // recovery) from an invalidated instance.
+        didLandOnLogin = false
         rejectPending(with: DashboardTicketBridgeError.notReady)
     }
 
@@ -278,6 +288,21 @@ final class DashboardTicketBridge: NSObject {
         guard !isInvalidated else { return }
         reloadCount += 1
         isReady = false
+        // Supersede the old navigation immediately: its late failure
+        // callbacks must not fail the fresh state during the window before
+        // the new load starts.
+        if let old = currentNavigation {
+            staleNavigations.insert(old)
+        }
+        currentNavigation = nil
+        // Reset the failure verdict synchronously: leaving it set until the
+        // async load task runs lets the next retry observe a stale failure
+        // and reload AGAIN, aborting the load that was just started. (The
+        // login verdict is deliberately NOT reset here — only an
+        // origin-matching landing may change it.) Test simulations re-assert
+        // their landing state so it persists across reloads deterministically.
+        isLoadFailed = false
+        applySimulatedLanding()
         rejectPending(with: DashboardTicketBridgeError.notReady)
         Task { [weak self] in
             guard let self else { return }
@@ -288,6 +313,18 @@ final class DashboardTicketBridge: NSObject {
             )
             guard !self.isInvalidated else { return }
             self.loadDashboardSession()
+        }
+    }
+
+    /// Re-asserts the test-simulated landing state; a no-op in production.
+    private func applySimulatedLanding() {
+        switch simulatedLanding {
+        case .loadFailure:
+            isLoadFailed = true
+        case .loginPage:
+            didLandOnLogin = true
+        case .ready, nil:
+            break
         }
     }
 
@@ -331,25 +368,35 @@ final class DashboardTicketBridge: NSObject {
     }
 
     private func waitUntilReady() async throws {
-        // Stop polling as soon as the navigation terminally fails: a failed
-        // load can only become ready again via reload, so sleeping out the
-        // window would just delay the .notReady recovery.
-        for _ in 0..<readinessPollAttempts where !isReady && !isInvalidated && !isLoadFailed {
+        // Stop polling as soon as the navigation terminally fails or the
+        // page has settled on the login screen: neither can become ready
+        // without a reload or sign-in, so sleeping out the window would
+        // just delay the recovery.
+        for _ in 0..<readinessPollAttempts where !isReady && !isInvalidated && !isLoadFailed && !didLandOnLogin {
             try await Task.sleep(for: readinessPollInterval)
         }
         guard !isInvalidated, isReady else {
             // A bridge parked on the login page is signed out, not loading:
             // surface the expiry so callers run their sign-in recovery
-            // instead of polling an already-settled page.
-            throw didLandOnLogin
-                ? DashboardTicketBridgeError.signInRequired
-                : DashboardTicketBridgeError.notReady
+            // instead of polling an already-settled page. An invalidated
+            // bridge always reports plain unreadiness — a late navigation
+            // callback must not resurrect sign-in state on a torn-down
+            // instance.
+            throw isInvalidated || !didLandOnLogin
+                ? DashboardTicketBridgeError.notReady
+                : DashboardTicketBridgeError.signInRequired
         }
     }
 
     /// Requests authenticated dashboard JSON through the same WebKit cookie
     /// jar that the sign-in flow owns. This intentionally avoids duplicating
     /// HttpOnly session handling in URLSession.
+    ///
+    /// Error contract: unusable page states (foreign-origin landing, stalled
+    /// engine) surface as `.notReady` so callers flow into the reload
+    /// recovery; an origin-matching login landing surfaces as
+    /// `.signInRequired`. Direct callers should not expect `.requestFailed`
+    /// for these states.
     func requestJSON(
         path: String,
         method: String = "GET",
@@ -357,15 +404,16 @@ final class DashboardTicketBridge: NSObject {
         timeoutMilliseconds: Int = 12_000,
         maxResponseBytes: Int = DataURLLimits.maxJSONResponseBytes
     ) async throws -> [String: Any] {
-        for _ in 0..<readinessPollAttempts where !isReady && !isInvalidated && !isLoadFailed {
+        for _ in 0..<readinessPollAttempts where !isReady && !isInvalidated && !isLoadFailed && !didLandOnLogin {
             try await Task.sleep(for: readinessPollInterval)
         }
         guard !isInvalidated, isReady else {
-            // Mirror waitUntilReady(): a login-parked bridge is signed out,
-            // not loading, for non-mint callers too.
-            throw didLandOnLogin
-                ? DashboardTicketBridgeError.signInRequired
-                : DashboardTicketBridgeError.notReady
+            // Mirror waitUntilReady(): a login-parked or terminally failed
+            // bridge is settled, not loading, for non-mint callers too, and
+            // an invalidated bridge always reports plain unreadiness.
+            throw isInvalidated || !didLandOnLogin
+                ? DashboardTicketBridgeError.notReady
+                : DashboardTicketBridgeError.signInRequired
         }
 
         requestID += 1
@@ -468,7 +516,19 @@ final class DashboardTicketBridge: NSObject {
                 """
                 webView.evaluateJavaScript(script) { _, error in
                     guard let error else { return }
-                    self.failPendingRequest(id: id, with: error)
+                    // The injected IIFE funnels every JS-level failure into
+                    // the message channel, so a completion error means the
+                    // engine never ran the script — a dead or stalled view.
+                    // Always route it onto the typed stall path (which
+                    // reloads the page and flushes sibling requests) instead
+                    // of letting an untyped NSError escape past mintTicket's
+                    // retry funnel. A request whose continuation was already
+                    // resumed (e.g. by reload) no-ops here.
+                    self.failPendingRequest(
+                        id: id,
+                        with: DashboardTicketBridgeError.notReady,
+                        markingWebViewStalled: true
+                    )
                 }
             }
         }, onCancel: {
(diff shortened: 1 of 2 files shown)


kaishi00:
## Summary
Hardens the `DashboardTicketBridge` recovery state machine: JS eval-error funneling onto the stall path, login-verdict immutability, navigation-identity supersession on reload, shutdown-fast login/readiness polls, and invalidate precedence, plus regression tests.

## Findings

### [minor] Late `didFail` for an already-finished navigation is not suppressed
**File:** Conduit/Services/DashboardTicketBridge.swift:661-664
**Issue:** `didFinish` assigns `currentNavigation = nil` and removes the navigation from `staleNavigations`. A subsequent `didFail`/`didFailProvisionalNavigation` for that same navigation then passes both guards at lines 705-706 / 713-714 (not stale, `current` is nil) and marks the successfully-loaded bridge `isLoadFailed = true`, `isReady = false`. The code comment itself flags WebKit delivering a cancelled navigation's error after a successful commit, yet the stale-set removal defeats the suppression in exactly the post-finish window the PR claims to close (fix 3). Outcome is recoverable (next mint reloads), but it is a spurious failure of a healthy load.
**Suggestion:** Derive suppression from the navigation that actually *finished* too — e.g. keep a "resolved" set removed only on the next `reload()`, or drop the `staleNavigations.remove` on `didFinish` and instead clear the set only when `loadDashboardSession` starts a fresh load.

### [minor] `evaluateJavaScript` completion handler captures `self` strongly
**File:** Conduit/Services/DashboardTicketBridge.swift:517-532
**Issue:** The script completion closure captures `self` strongly (`self.failPendingRequest(...)` at line 527). WebKit retains the completion block until it fires; if the content process is wedged and never invokes it, the block forms `bridge → webView → completion → bridge`, a permanent cycle — the same class of leak the file explicitly engineers against with `WeakScriptMessageHandler`. The Swift-side deadline task (line 445) guards against dropped responses, so reusing `[weak self]` here is safe: if `self` is gone, `deinit`/`invalidate` already rejected the continuations.
**Suggestion:** Use `[weak self]` with a `guard let self else { return }` in the completion handler.

### [minor] `staleNavigations` grows unbounded
**File:** Conduit/Services/DashboardTicketBridge.swift:210-211, 294-296
**Issue:** Every `reload()` inserts the superseded navigation into `staleNavigations`, and entries are only removed in `didFinish` — which never fires for cancelled navigations, the primary producers of stale entries. A long-lived bridge under repeated recovery cycles accumulates `WKNavigation` objects indefinitely. Small objects, but an unneeded unbounded growth path.
**Suggestion:** Clear `staleNavigations` when `loadDashboardSession` starts the fresh load (the window it protects has closed), or drop entries older than the current reload.

### [nit] Inconsistent login-verdict reset across failure paths
**File:** Conduit/Services/DashboardTicketBridge.swift:707-709, 715-717 vs 560, 731
**Issue:** `failPendingRequest(markingWebViewStalled:)` and `webViewWebContentProcessDidTerminate` reset `didLandOnLogin = false`, but `didFail`/`didFailProvisionalNavigation` and the foreign-origin `didFinish` branch deliberately leave it set. Because the `waitUntilReady`/`requestJSON` guard (lines 385-387, 414-416) prefers `.signInRequired` whenever `didLandOnLogin` is true (regardless of `isLoadFailed`), a bridge parked on login whose subsequent reload genuinely fails (network drop) keeps surfacing `.signInRequired` — with its session-erasing recovery — instead of `.notReady`. Defensible if the last landing was a login page, but the divergence across three failure paths makes the verdict precedence fragile and undocumented in the guard.
**Suggestion:** Either set `didLandOnLogin = false` in `didFail`/`didFailProvisionalNavigation` uniformly, or document the intended precedence (login verdict wins over failure) on the guard itself.

### [nit] Wall-clock timeout assertion in the newest test
**File:** ConduitTests/DashboardTicketBridgeTests.swift:257-260
**Issue:** `testLoginParkedBridgeDoesNotBurnPollWindow` asserts `elapsed < 3.0` from a `Date()` delta. Under overloaded CI, `Task.sleep(for:)` overhead can push a nominally ~1.1 s run over the margin (it is only ~3× headroom). Also, the existing `catch` fallback in `testExpiredRequestDeadlineMarksWebViewStalled` (lines 204-210) is now dead code: the completion-error change means every eval error is mapped to `.notReady`, so that `print`-annotated branch is unreachable; the stale comment it carries should be updated.
**Suggestion:** Drive the timing assertion off a deterministic budget (e.g. assert `reloadCount` and fast-exit via a much larger margin, or inject poll-interval as the independent variable) rather than wall-clock time; update the fallback-arm comment.

### [nit] Reload window where the login verdict reads as stale
**File:** Conduit/Services/DashboardTicketBridge.swift:300-305
**Issue:** `reload()` deliberately does not reset `didLandOnLogin`, and it is only reset later in the async `loadDashboardSession` (line 622). A `requestJSON`/`waitUntilReady` call landing in that window (between the sync `reload()` and the rescheduled load) fast-exits and throws `.signInRequired`, possibly triggering the session-erasing recovery while the bridge is merely reloading. The window is short (cookie-restore latency), but the timing is nondeterministic.
**Suggestion:** Reset `didLandOnLogin = false` synchronously in `reload()` (mirroring `isLoadFailed`) and re-assert via `applySimulatedLanding()` — consistent with the reset already performed in `loadDashboardSession`.

## Verdict
COMMENT

The state-machine hardening is sound on the whole: the eval-error → `.notReady` + stall-marking funnel, invalidate precedence, and stall-flush logic are correct and well-tested, and the navigation-identity guards close the reload-window race for the intended (cancelled-after-reload) case. The residual post-`didFinish` suppression hole, the strong-`self` capture on the newly rewritten completion closure, and the unbounded stale-set growth are worth addressing but are recoverable/recoverable-by-reload and not blockers.

<a href="https://opencode.ai/s/Zj0FFqNB">New%20session%20-%202026-08-17T00%3A33%3A07.783Z</a>
[opencode session](https://opencode.ai/s/Zj0FFqNB)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31982295347)

kaishi00:
## Summary
This PR hardens the `DashboardTicketBridge` recovery state machine: JS eval errors funnel onto the typed `.notReady`/stall path, login verdicts are pinned to origin-matching `didFinish`, `reload()` synchronously supersedes navigation identity and resets landing/failure verdicts, readiness polls fast-exit on login/failure/invalidation, `invalidate()` clears the login verdict, and `staleNavigations`/`currentNavigation` guards suppress stale callbacks — with new regression tests.

## Findings

### [minor] Strong `self` capture in `evaluateJavaScript` completion handler
**File:** Conduit/Services/DashboardTicketBridge.swift:530-545
**Issue:** `webView.evaluateJavaScript(script) { _, error in ... self.failPendingRequest(...) }` captures `self` strongly. WebKit can retain the completion block until the script evaluation resolves; in the exact scenario this PR protects against (wedged/dead content process that never runs the script), the completion may never fire, forming `bridge → webView → retained-block → bridge` — a permanent cycle that prevents `deinit` (which would cancel the deadline tasks and unregister the message handler). The `WeakScriptMessageHandler` proxy exists precisely to engineer against this class of leak. The neighboring Swift-side deadline task correctly uses `[weak self]` (line 458).
**Suggestion:** Use `[weak self]` with a `guard let self else { return }` before calling `failPendingRequest`. The `failPendingRequest` nil-continuation no-op and `deinit`/`invalidate` already make the weak path safe.

### [minor] Post-finish `didFail` for a resolved navigation is still processed
**File:** Conduit/Services/DashboardTicketBridge.swift:683-685, 734-735, 742-743
**Issue:** `didFinish` sets `currentNavigation = nil` and removes the navigation from `staleNavigations`. A subsequent `didFail`/`didFailProvisionalNavigation` for that *same* (already finished) navigation then passes both guards: it is no longer in `staleNavigations`, and `currentNavigation == nil` so the `navigation !== currentNavigation` guard does not trigger. The callback marks the healthy, successfully-loaded bridge `isLoadFailed = true`, `isReady = false`, and rejects pending requests. The identity that would suppress it is discarded at the moment it is needed most. For a single navigation WebKit treats finish and failure as mutually exclusive, so this is a defensive hole rather than an observed race; it is also the exact window the PR claims to close ("post-finish suppression").
**Suggestion:** Track a "resolved" navigation identity (e.g. keep the finished navigation in a `resolvedNavigations` set cleared only on the next `loadDashboardSession`, or compare against a `lastFinishedNavigation`) and suppress failures matching it, or keep the entry in `staleNavigations` until the next fresh load.

### [minor] `.ready` simulated landing is not re-asserted on reload
**File:** Conduit/Services/DashboardTicketBridge.swift:333-342
**Issue:** `reload()` (line 298) and `loadDashboardSession()` (line 636) set `isReady = false`, then call `applySimulatedLanding()`. For `.ready`, the helper is a no-op (`break`), so a bridge whose test seam simulates a persistently-ready landing becomes permanently unready after a reload — contradicting the code's own comments ("Test simulations re-assert their landing so it persists across reloads deterministically", line 313-315; "re-assert any test simulation on purpose", line 633). No current test reloads a `.ready` bridge, so this is latent, but the seam is internally inconsistent.
**Suggestion:** Make the `.ready` case re-assert `isReady = true` (and `isLoadFailed = false`) in `applySimulatedLanding`.

### [minor] `staleNavigations` can grow unboundedly
**File:** Conduit/Services/DashboardTicketBridge.swift:302-304, 219
**Issue:** Every `reload()` inserts the superseded navigation into `staleNavigations`, and entries are only removed when a matching callback arrives (`remove(navigation) != nil`). Superseded/cancelled navigations commonly never produce a `didFinish`/`didFail` (WebKit silently drops them when superseded), so each recovery reload can leak one `WKNavigation` in the set for the bridge's lifetime. Small objects, but an unnecessary unbounded growth path on a long-lived bridge under repeated recovery cycles.
**Suggestion:** Clear `staleNavigations` in `loadDashboardSession()` when the fresh load starts (the window the set protects has closed), since the new `currentNavigation` identity covers suppression from that point on.

### [nit] Login verdict precedence inconsistent across failure paths
**File:** Conduit/Services/DashboardTicketBridge.swift:728-747, 571-573, 760
**Issue:** `failPendingRequest(markingWebViewStalled:)`, `webViewWebContentProcessDidTerminate`, and `invalidate()` clear `didLandOnLogin`, but `didFail`/`didFailProvisionalNavigation` do not (and the foreign-origin `didFinish` branch intentionally leaves it). Because the `waitUntilReady`/`requestJSON` guards (lines 398-400, 427-429) prefer `.signInRequired` whenever `didLandOnLogin` is true regardless of `isLoadFailed`, a bridge parked on login whose next reload genuinely fails (e.g. network drop) surfaces `.signInRequired` — with its session-erasing recovery — instead of `.notReady`. Defensible (the last landing was a login page), but the divergence across three failure paths is undocumented.
**Suggestion:** Either clear `didLandOnLogin` uniformly in both `didFail` methods, or document the intended precedence (login verdict wins over failure) on the guards.

### [nit] Wall-clock timing assertion risks CI flake
**File:** ConduitTests/DashboardTicketBridgeTests.swift:257-260
**Issue:** `testLoginParkedBridgeDoesNotBurnPollWindow` asserts `elapsed < 3.0` on a `Date()` delta for a run whose nominal duration is ~1.1s (three 350ms sleeps plus scheduling). That is only ~2.7× headroom; under an overloaded CI host, task scheduling and sleep overshoot can exceed the bound. The elapsed assertion depends on behavior (reload re-asserting the login sim) already covered deterministically by the `reloadCount == 2` assertion.
**Suggestion:** Drive the assertion off deterministic state (reloadCount, and the error type) rather than wall-clock time, or widen the margin substantially.
**File:** ConduitTests/DashboardTicketBridgeTests.swift:204-210 — the `catch` fallback is now effectively dead (every eval error funnels to `.notReady`), as the PR text acknowledges; the stale "deadline belt is not exercised" comment should be updated to reflect that the immediate-error arm is now the primary path.

## Verdict
COMMENT

The state-machine hardening is well-structured and the dominant logic is correct: the eval-error → `.notReady` + stall-marking funnel, synchronous verdict reset in `reload()`, the `staleNavigations`/`currentNavigation` supersession guards, invalidate precedence, and the fast-exit polls all hold up under the crossing callbacks. The findings above are the residual issues — a strong-capture leak on the newly rewritten completion closure, a post-finish suppression hole, the `.ready` simulation seam gap, and unbounded stale-set growth. None are merge-blocking; all are recoverable by reload and worth a follow-up pass.

<a href="https://opencode.ai/s/UA1maZZr">New%20session%20-%202026-08-17T00%3A52%3A59.233Z</a>
[opencode session](https://opencode.ai/s/UA1maZZr)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31983246966)

kaishi00:
<｜DSML｜tool_calls>
<｜DSML｜invoke name="bash">
<｜DSML｜parameter name="command" string="true">grep -n "@MainActor" Conduit/Services/DashboardTicketBridge.swift | head; awk 'NR>=180 && NR<=200' Conduit/Services/DashboardTicketBridge.swift</｜DSML｜parameter>
</｜DSML｜invoke>
</｜DSML｜tool_calls>

<a href="https://opencode.ai/s/mt6iD7ex">New%20session%20-%202026-08-17T01%3A04%3A03.839Z</a>
[opencode session](https://opencode.ai/s/mt6iD7ex)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/31983775701)

PR #73, opened 2026-08-17
fix(chat): show full markdown table cell content

kaishi00:
## User-visible symptoms

1. Markdown tables truncated cells to a few lines (clean 4-line cuts, sometimes half a line / 1–2 lines).
2. Column dividers shifted horizontally between rows — each row sized its own columns.
3. `:---:` / `---:` alignment was parsed but rendered left-aligned.
4. Even trivial three-column tables forced horizontal scrolling on iPhone.

## Root causes

- `maximumNumberOfLines: 4` on every table cell.
- Non-deterministic measurement width inside the horizontal `ScrollView` (nil proposal + zero bounds on the first pass → intrinsic fallback).
- No shared column widths: each row's cells independently self-sized.
- Alignment only positioned a full-width SwiftUI wrapper; the text itself laid out left.
- A 112pt per-column floor that existed for measurement determinism, not table semantics (~396pt minimum for three columns with padding).

## Fixes

**Full content, deterministic height:** the four-line cap is gone. Table cells self-size at a deterministic width derived from content, independent of layout proposals — first pass and every later pass report the same correct height.

**Shared, responsive column widths:** `MarkdownTableLayout` derives one width per column for the entire table from the header plus all body cells, then resolves against the chat's proposed width (read by a zero-height probe; the first pass falls back to a deterministic cap-and-scroll layout until the width arrives):

- Column width = largest ideal content width in that column, capped at 220pt.
- If the capped columns fit the chat width, the table fits — narrow columns stay genuinely narrow, wide ones keep proportionally more space.
- On overflow, columns shrink proportionally to their excess above a 64pt floor (already-narrow columns don't shrink) and wrap at the resolved width.
- A table still overflowing at the floor keeps horizontal scrolling.
- Cell padding, dividers, and the border subtract from available width exactly once.
- Cell heights measure at the final resolved shared column content width (single-value self-sizing range), preserving the deterministic first-pass/full-height guarantee.

**Alignment reaches the text engine:** `:---`/`:---:`/`---:` flows through `InlineMarkdown → SelectableTextView.textAlignment` into the paragraph style, holding for single-line and wrapped multiline content. `.natural` default leaves every other caller unchanged.

## Scrolling policy

Fit when the content allows; scroll horizontally only when it doesn't. Cells expand vertically without nested scrolling; the outer chat scroll owns vertical scrolling.

## Regression coverage

- **ChatTextSelectionTests (28):** full-chain unlimited-lines render, first-pass height determinism vs known-correct reference, capped-vs-unlimited discriminator, finite-limit callers still truncate.
- **MarkdownTableLayoutTests (13):** pure policy unit tests (fit keeps ideals, cap, proportional shrink, floor-and-scroll, unknown-width fallback, content-metric growth recompute) plus rendered fixtures — tiny three-column table fits without scrolling; narrow status + long description fits with proportional shared widths; wide five-column table scrolls; shared divider widths across header/short/empty cells; long cell wraps fully at its shared width; per-line glyph geometry proves leading/center/trailing for single-line and wrapped cells; selection observers stay attached.
- **MarkdownSelectionCoordinatorTests (32):** cross-block/table selection intact.

## Validation

73/73 across the three markdown suites on the local simulator. Shared-width and content-informed assertions verified to fail against per-cell sizing (re-injected). Known local-sim wedge (`AppStateChatResumeTests`, CoreAudio AURemoteIO) unrelated; CI's fresh runner arbitrates.

### DIFF
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -661,6 +661,7 @@ private struct InlineMarkdown: View {
     var font: UIFont = .preferredFont(forTextStyle: .body)
     var lineSpacing: CGFloat = 0
     var maximumNumberOfLines: Int = 0
+    var selfSizingWidthRange: ClosedRange<CGFloat>? = nil
     var selectionCoordinator: MarkdownSelectionCoordinator?
     var selectionSegment: MarkdownSelectionSegmentDescriptor?
     var trailingCharacterOpacities: [Double] = []
@@ -689,6 +690,7 @@ private struct InlineMarkdown: View {
             lineSpacing: lineSpacing,
             maximumNumberOfLines: maximumNumberOfLines,
             linkColor: usesAccentSurface ? .white : .link,
+            selfSizingWidthRange: selfSizingWidthRange,
             selectionCoordinator: selectionCoordinator,
             selectionSegment: selectionSegment
         )
@@ -923,6 +925,9 @@ private struct MarkdownTable: View {
                     font: isHeader
                         ? UIFont.preferredFont(forTextStyle: .caption1).withTraits(.traitBold)
                         : UIFont.preferredFont(forTextStyle: .footnote),
+                    // Match the frame clamp below so the measured wrapping
+                    // width is deterministic from the first layout pass.
+                    selfSizingWidthRange: 112...220,
                     selectionCoordinator: selectionCoordinator,
                     selectionSegment: selectionSegment(row: rowIndex, column: index)
                 )
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -923,7 +923,6 @@ private struct MarkdownTable: View {
                     font: isHeader
                         ? UIFont.preferredFont(forTextStyle: .caption1).withTraits(.traitBold)
                         : UIFont.preferredFont(forTextStyle: .footnote),
-                    maximumNumberOfLines: 4,
                     selectionCoordinator: selectionCoordinator,
                     selectionSegment: selectionSegment(row: rowIndex, column: index)
                 )
--- a/Conduit/Views/Components/SelectableTextView.swift
+++ b/Conduit/Views/Components/SelectableTextView.swift
@@ -209,16 +209,11 @@ struct SelectableTextView: UIViewRepresentable {
         return CGSize(width: width, height: Self.measuredWrappingHeight(of: textView, at: width))
     }
 
-    /// Shared measurement path for `sizeThatFits` and unit tests. Forces the
-    /// text container to the target width before measuring so the result is
-    /// independent of `textView.bounds` (which may be zero or stale during the
-    /// first layout pass inside a horizontal `ScrollView`).
+    /// Shared measurement path for `sizeThatFits` and unit tests. Delegates
+    /// to `UITextView.sizeThatFits` at the target width. On iOS 17+
+    /// (TextKit 2) the proposed width drives wrapping independent of the
+    /// text container's stored size, so no container mutation is needed.
     static func measuredWrappingHeight(of textView: UITextView, at width: CGFloat) -> CGFloat {
-        let container = textView.textContainer
-        let tracksWidth = container.widthTracksTextView
-        container.widthTracksTextView = false
-        container.size = CGSize(width: width, height: CGFloat.greatestFiniteMagnitude)
-        container.widthTracksTextView = tracksWidth
         let measured = textView.sizeThatFits(CGSize(width: width, height: CGFloat.greatestFiniteMagnitude))
         return ceil(measured.height)
     }
--- a/Conduit/Views/Components/SelectableTextView.swift
+++ b/Conduit/Views/Components/SelectableTextView.swift
@@ -197,27 +197,30 @@ struct SelectableTextView: UIViewRepresentable {
             return measureNonWrapping()
         }
 
-        guard let width = proposal.width, width > 0 else { return nil }
         let textView = uiView.mountedTextView
 
-        // The text container width is set from textView.bounds.width in
-        // configure(), but during the first layout pass (or when the view is
-        // inside a horizontal ScrollView whose content width is unconstrained),
-        // bounds.width may be 0 or stale. Force the container to the proposed
-        // width so UIKit wraps lines at the actual column width, not a
-        // leftover from a previous layout cycle.
-        let previousTracksWidth = textView.textContainer.widthTracksTextView
-        textView.textContainer.widthTracksTextView = false
-        textView.textContainer.size = CGSize(width: width, height: CGFloat.greatestFiniteMagnitude)
-        textView.textContainer.widthTracksTextView = previousTracksWidth
+        // Inside a horizontal ScrollView the width proposal is nil; fall back
+        // to the last laid-out width so SwiftUI measures at the real column
+        // width instead of intrinsicContentSize (which uses a stale container).
+        let width = proposal.width
+            ?? (textView.bounds.width > 0 ? textView.bounds.width : nil)
+        guard let width, width > 0, width.isFinite else { return nil }
 
+        return CGSize(width: width, height: Self.measuredWrappingHeight(of: textView, at: width))
+    }
+
+    /// Shared measurement path for `sizeThatFits` and unit tests. Forces the
+    /// text container to the target width before measuring so the result is
+    /// independent of `textView.bounds` (which may be zero or stale during the
+    /// first layout pass inside a horizontal `ScrollView`).
+    static func measuredWrappingHeight(of textView: UITextView, at width: CGFloat) -> CGFloat {
+        let container = textView.textContainer
+        let tracksWidth = container.widthTracksTextView
+        container.widthTracksTextView = false
+        container.size = CGSize(width: width, height: CGFloat.greatestFiniteMagnitude)
+        container.widthTracksTextView = tracksWidth
         let measured = textView.sizeThatFits(CGSize(width: width, height: CGFloat.greatestFiniteMagnitude))
-
-        if textView.bounds.width != width {
-            textView.bounds.size.width = width
-        }
-
-        return CGSize(width: width, height: ceil(measured.height))
+        return ceil(measured.height)
     }
 
     private func configure(_ textView: UITextView) {
--- a/Conduit/Views/Components/SelectableTextView.swift
+++ b/Conduit/Views/Components/SelectableTextView.swift
@@ -15,6 +15,10 @@ struct SelectableTextView: UIViewRepresentable {
     let maximumNumberOfLines: Int
     let wrapsLines: Bool
     let linkColor: UIColor
+    /// When set, the view self-sizes at a deterministic width derived from
+    /// its content and this range (used by Markdown table cells), ignoring
+    /// layout proposals. See `measuredSize(proposalWidth:textView:)`.
+    let selfSizingWidthRange: ClosedRange<CGFloat>?
     let selectionCoordinator: MarkdownSelectionCoordinator?
     let selectionSegment: MarkdownSelectionSegmentDescriptor?
 
@@ -26,6 +30,7 @@ struct SelectableTextView: UIViewRepresentable {
         maximumNumberOfLines: Int = 0,
         wrapsLines: Bool = true,
         linkColor: UIColor = .link,
+        selfSizingWidthRange: ClosedRange<CGFloat>? = nil,
         selectionCoordinator: MarkdownSelectionCoordinator? = nil,
         selectionSegment: MarkdownSelectionSegmentDescriptor? = nil
     ) {
@@ -36,6 +41,7 @@ struct SelectableTextView: UIViewRepresentable {
         self.maximumNumberOfLines = maximumNumberOfLines
         self.wrapsLines = wrapsLines
         self.linkColor = linkColor
+        self.selfSizingWidthRange = selfSizingWidthRange
         self.selectionCoordinator = selectionCoordinator
         self.selectionSegment = selectionSegment
     }
@@ -48,6 +54,7 @@ struct SelectableTextView: UIViewRepresentable {
         maximumNumberOfLines: Int = 0,
         wrapsLines: Bool = true,
         linkColor: UIColor = .link,
+        selfSizingWidthRange: ClosedRange<CGFloat>? = nil,
         selectionCoordinator: MarkdownSelectionCoordinator? = nil,
         selectionSegment: MarkdownSelectionSegmentDescriptor? = nil
     ) {
@@ -59,6 +66,7 @@ struct SelectableTextView: UIViewRepresentable {
             maximumNumberOfLines: maximumNumberOfLines,
             wrapsLines: wrapsLines,
             linkColor: linkColor,
+            selfSizingWidthRange: selfSizingWidthRange,
             selectionCoordinator: selectionCoordinator,
             selectionSegment: selectionSegment
         )
@@ -72,6 +80,7 @@ struct SelectableTextView: UIViewRepresentable {
         maximumNumberOfLines: Int = 0,
         wrapsLines: Bool = true,
         linkColor: UIColor = .link,
+        selfSizingWidthRange: ClosedRange<CGFloat>? = nil,
         selectionCoordinator: MarkdownSelectionCoordinator? = nil,
         selectionSegment: MarkdownSelectionSegmentDescriptor? = nil
     ) {
@@ -86,6 +95,7 @@ struct SelectableTextView: UIViewRepresentable {
             maximumNumberOfLines: maximumNumberOfLines,
             wrapsLines: wrapsLines,
             linkColor: linkColor,
+            selfSizingWidthRange: selfSizingWidthRange,
             selectionCoordinator: selectionCoordinator,
             selectionSegment: selectionSegment
         )
@@ -193,19 +203,33 @@ struct SelectableTextView: UIViewRepresentable {
     }
 
     func sizeThatFits(_ proposal: ProposedViewSize, uiView: SelectableTextViewHostView, context: Context) -> CGSize? {
+        measuredSize(proposalWidth: proposal.width, textView: uiView.mountedTextView)
+    }
+
+    /// Extracted sizing logic so `sizeThatFits` and unit tests share one
+    /// path (`UIViewRepresentableContext` cannot be constructed outside
+    /// SwiftUI).
+    ///
+    /// Self-sizing range (table cells): Markdown tables lay out inside a
+    /// horizontal ScrollView, where width proposals are nil or transient and
+    /// first-pass UIKit bounds are zero — a proposal- or bounds-derived width
+    /// produces a different wrapped height on the first pass than in steady
+    /// state, which is the intermittent half-line/1–2-line clipping. Instead,
+    /// the measurement width is derived deterministically from the content's
+    /// ideal width clamped to the table's column policy, so the first pass
+    /// and every later pass report the same correct height.
+    func measuredSize(proposalWidth: CGFloat?, textView: UITextView) -> CGSize? {
         if !wrapsLines {
             return measureNonWrapping()
         }
 
-        let textView = uiView.mountedTextView
+        if let range = selfSizingWidthRange {
+            let idealWidth = max(1, measureNonWrapping().width)
+            let width = min(max(idealWidth, range.lowerBound), range.upperBound)
+            return CGSize(width: width, height: Self.measuredWrappingHeight(of: textView, at: width))
+        }
 
-        // Inside a horizontal ScrollView the width proposal is nil; fall back
-        // to the last laid-out width so SwiftUI measures at the real column
-        // width instead of intrinsicContentSize (which uses a stale container).
-        let width = proposal.width
-            ?? (textView.bounds.width > 0 ? textView.bounds.width : nil)
-        guard let width, width > 0, width.isFinite else { return nil }
-
+        guard let width = proposalWidth, width > 0, width.isFinite else { return nil }
         return CGSize(width: width, height: Self.measuredWrappingHeight(of: textView, at: width))
     }
 
--- a/Conduit/Views/Components/SelectableTextView.swift
+++ b/Conduit/Views/Components/SelectableTextView.swift
@@ -198,7 +198,25 @@ struct SelectableTextView: UIViewRepresentable {
         }
 
         guard let width = proposal.width, width > 0 else { return nil }
-        let measured = uiView.mountedTextView.sizeThatFits(CGSize(width: width, height: CGFloat.greatestFiniteMagnitude))
+        let textView = uiView.mountedTextView
+
+        // The text container width is set from textView.bounds.width in
+        // configure(), but during the first layout pass (or when the view is
+        // inside a horizontal ScrollView whose content width is unconstrained),
+        // bounds.width may be 0 or stale. Force the container to the proposed
+        // width so UIKit wraps lines at the actual column width, not a
+        // leftover from a previous layout cycle.
+        let previousTracksWidth = textView.textContainer.widthTracksTextView
+        textView.textContainer.widthTracksTextView = false
+        textView.textContainer.size = CGSize(width: width, height: CGFloat.greatestFiniteMagnitude)
+        textView.textContainer.widthTracksTextView = previousTracksWidth
+
+        let measured = textView.sizeThatFits(CGSize(width: width, height: CGFloat.greatestFiniteMagnitude))
+
+        if textView.bounds.width != width {
+            textView.bounds.size.width = width
+        }
+
         return CGSize(width: width, height: ceil(measured.height))
     }
 
--- a/ConduitTests/ChatTextSelectionTests.swift
+++ b/ConduitTests/ChatTextSelectionTests.swift
@@ -715,4 +715,96 @@ final class ChatTextSelectionTests: XCTestCase {
         XCTAssertLessThanOrEqual(measured, truncationCeiling,
                                  "Long text must be truncated well below full expansion")
     }
+
+    // MARK: - Full table-view-chain regression (MarkdownText → MarkdownTable → cell text views)
+
+    /// Regression: a Markdown table with a long cell must render every cell's
+    /// selectable text view with unlimited lines and enough height for the
+    /// fully wrapped content. This drives the real SwiftUI chain —
+    /// MarkdownText → MarkdownTable.tableRow → InlineMarkdown →
+    /// SelectableTextView — so re-introducing a line cap or under-height
+    /// sizing at any link fails here.
+    @MainActor
+    func testMarkdownTableLongCellRendersFullContentThroughViewChain() throws {
+        let longCellText = (0..<30).map { "Word\($0)" }.joined(separator: " ")
+        let source = """
+        | Item | Details |
+        |---|---|
+        | Long entry | \(longCellText) |
+        """
+        let host = UIHostingController(rootView: MarkdownText(source: source))
+        let window = UIWindow(frame: CGRect(x: 0, y: 0, width: 390, height: 844))
+        window.rootViewController = host
+        window.isHidden = false
+        host.view.setNeedsLayout()
+        host.view.layoutIfNeeded()
+
+        let textViews = allSubviews(of: host.view).compactMap { $0 as? UITextView }
+        XCTAssertFalse(textViews.isEmpty, "Rendered table must mount selectable text views")
+
+        for textView in textViews {
+            XCTAssertEqual(textView.textContainer.maximumNumberOfLines, 0,
+                           "Every table cell must render unlimited-line text; a finite cap here is the truncation bug")
+        }
+
+        let longCell = try XCTUnwrap(
+            textViews.first { $0.attributedText.string.contains("Word29") },
+            "The long table cell's text view must exist in the rendered hierarchy"
+        )
+        XCTAssertGreaterThan(longCell.bounds.width, 0)
+
+        // Reference: a fresh, unlimited text view with the same content must
+        // wrap to the same height the rendered cell was given. A capped or
+        // under-measured cell falls short of the full-wrap reference.
+        let reference = SelectableTextView.makeTextView()
+        reference.attributedText = longCell.attributedText
+        let fullWrapHeight = SelectableTextView.measuredWrappingHeight(of: reference, at: longCell.bounds.width)
+        XCTAssertGreaterThanOrEqual(longCell.bounds.height + 1, fullWrapHeight,
+                                     "Rendered cell height must cover the fully wrapped content")
+    }
+
+    /// Regression: inside the horizontal table ScrollView the first layout
+    /// pass receives a nil width proposal with zero UIKit bounds — the
+    /// conditions that produced intermittent half-line/1–2-line cells.
+    /// Self-sized table cells must return the full-wrap height at their
+    /// deterministic column width on the first pass, and identical sizes on
+    /// every later pass regardless of proposal.
+    @MainActor
+    func testTableCellFirstPassMeasurementMatchesFullWrapReference() throws {
+        let font = UIFont.preferredFont(forTextStyle: .footnote)
+        let longText = (0..<30).map { "Word\($0)" }.joined(separator: " ")
+        let view = SelectableTextView(text: longText, font: font, selfSizingWidthRange: 112...220)
+        let hostView = view.makeUIViewForTests(coordinator: view.makeCoordinator())
+        let textView = hostView.mountedTextView
+
+        // First pass: no width proposal, zero bounds.
+        textView.bounds = .zero
+        let firstPass = try XCTUnwrap(
+            view.measuredSize(proposalWidth: nil, textView: textView),
+            "Self-sized table cell must produce a size even with nil proposal and zero bounds"
+        )
+
+        // Known-correct reference: full wrap at the deterministic column
+        // width (content ideal width clamped to the table column policy).
+        let idealWidth = view.measureNonWrapping().width
+        let expectedWidth = min(max(idealWidth, 112), 220)
+        let referenceHeight = SelectableTextView.measuredWrappingHeight(of: textView, at: expectedWidth)
+
+        XCTAssertEqual(firstPass.width, expectedWidth, accuracy: 1,
+                       "First-pass width must be the clamped ideal column width")
+        XCTAssertEqual(firstPass.height, referenceHeight, accuracy: 1,
+                       "First-pass height must equal the full-wrap height at the column width")
+
+        // A later pass with a wide proposal (or populated bounds) must not
+        // change the answer — that drift was the intermittent clipping.
+        let laterPass = try XCTUnwrap(view.measuredSize(proposalWidth: 390, textView: textView))
+        XCTAssertEqual(laterPass.width, firstPass.width, accuracy: 1,
+                       "Steady-state width must match the first pass")
+        XCTAssertEqual(laterPass.height, firstPass.height, accuracy: 1,
+                       "Steady-state height must match the first pass")
+    }
+
+    private func allSubviews(of view: UIView) -> [UIView] {
+        view.subviews.flatMap { [$0] + allSubviews(of: $0) }
+    }
 }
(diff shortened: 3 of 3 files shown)


kaishi00:
## Summary
Removes the hard-coded 4-line cap on markdown table cell content (allowing cells to expand to full intrinsic height) and adds three tests claiming regression coverage for the truncation fix.

## Findings

### [major] New tests do not actually cover the fix
**File:** ConduitTests/ChatTextSelectionTests.swift:599-657
**Issue:** All three new tests verify behavior that was already true *before* this change:
- `testMarkdownTableParserDeliversLongCellContentUntruncated` tests `MarkdownParser.parse`, which never truncated — truncation was purely a UIKit display-layer effect (`textContainer.maximumNumberOfLines`). This test passes on the old code.
- `testSelectableTextViewDefaultMaximumNumberOfLinesIsUnlimited` and `testSelectableTextViewFiniteLineLimitStillTruncates` test `SelectableTextView`, whose defaults were already `0`/unlimited — unchanged by this PR.

The actual change (removing `maximumNumberOfLines: 4` at `MarkdownTable.tableRow`) has zero direct coverage. A regression that re-adds `maximumNumberOfLines: 4` in MarkdownText.swift:919 would pass all three tests, defeating the stated "regression coverage" purpose.
**Suggestion:** Render the table (e.g. via `MarkdownText`/`MarkdownBlockView` or direct inspection of the view produced by `MarkdownTable.tableRow`) and assert the resulting text container has `maximumNumberOfLines == 0`, or assert `InlineMarkdown` built with default parameters propagates `0`. At minimum, add a test that exercises `MarkdownText` with a long-cell table source and asserts no truncation.

### [minor] Unbounded cell height reintroduces pathological layout cost
**File:** Conduit/Views/Components/MarkdownText.swift:919
**Issue:** The 4-line cap previously bounded the worst case. Now a model-authored cell with hundreds of lines becomes a single tall, non-scrolling `UITextView` laid out eagerly inside a horizontal `ScrollView` within the chat's vertical `ScrollView`. A pathological cell (very large blob of text) forces full text layout in one `UIViewRepresentable` and is measured repeatedly via `sizeThatFits` (SelectableTextView.swift:195-203). This is inherent to the fix's "no nested scrolling" decision, but worth acknowledging.
**Suggestion:** Consider an explicit `maximumNumberOfLines` allowance for tables if model output includes unbounded cells, sized against a documented content cap, or accept the tradeoff and note it in a comment on the cell creation.

### [nit] Test comment cites the wrong finite-limit value
**File:** ConduitTests/ChatTextSelectionTests.swift:647
**Issue:** The comment says the guard targets "RenderCard source preview with `maximumNumberOfLines: 4`", but the actual RenderCard source-preview call passes `5` (MarkdownText.swift:1421). The test hardcodes `4`, so it does not validate the real caller's value.
**Suggestion:** Use `5` to match the actual RenderCard call site, or correct the comment.

### [nit] First test name overstates what it verifies
**File:** ConduitTests/ChatTextSelectionTests.swift:605
**Issue:** `testMarkdownTableParserDeliversLongCellContentUntruncated` (as noted above) tests the parser, not the truncation fix it's documented against. Passing it does not assure the visual truncation is resolved.
**Suggestion:** Rename it to reflect that it validates parser fidelity (e.g. `testMarkdownTableParserPreservesLongCellContent`), and rely on a UI-level test for the actual fix.

## Verdict
COMMENT

The fix itself is correct and minimal — `InlineMarkdown`/`SelectableTextView` already default to unlimited lines with proper `.byWordWrapping`, and removing the cap resolves the symptom exactly as described. No retain cycles, force unwraps, concurrency, or accessibility issues introduced. However, the claimed regression coverage is illusory: none of the three tests exercises the line that was deleted, so the PR does not actually protect against reintroduction.

<a href="https://opencode.ai/s/ad53xjuc">New%20session%20-%202026-08-17T20%3A18%3A32.480Z</a>
[opencode session](https://opencode.ai/s/ad53xjuc)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32064989068)

kaishi00:
## Summary
Removes the 4-line cap on markdown table cells (cells now expand to full intrinsic height) and forces the text container to the proposed width during `SelectableTextView.sizeThatFits`, with four new regression tests.

## Findings

### [major] Regression tests do not exercise either bug being fixed
**File:** ConduitTests/ChatTextSelectionTests.swift:605-636
**Issue:** The `measureTextView` helper builds a bare `UITextView` via `SelectableTextView.makeTextView()`, configures it directly, and calls `UITextView.sizeThatFits` itself. It never invokes `SelectableTextView.sizeThatFits(_:uiView:context:)` (the Bug 2 fix at SelectableTextView.swift:195) nor `MarkdownTable.tableRow` → `InlineMarkdown(maximumNumberOfLines:)` (the Bug 1 fix at MarkdownText.swift:919). The helper even hardcodes the fixed behavior — manually setting `textContainer.size` and `bounds` to the target width (lines 630-631) — which is precisely what the fix was added to do. Consequently every test passes against the pre-change code:
- `testTableCellMeasurementAtRealisticWidthProducesSufficientHeight` measures a container with `maximumNumberOfLines: 0`; the old cap lived upstream at the `InlineMarkdown` call site, so this never sees it.
- `testMeasurementUsesColumnWidthNotStaleBounds` and `testDifferentLengthCellsProportionalHeights` verify generic UIKit wrapping behavior the fix didn't change.
- `testFiniteLineLimitStillTruncates` documents that `maximumNumberOfLines: 4` yields `.byTruncatingTail`, which the default of `0` never produced.

A regression re-adding `maximumNumberOfLines: 4` at MarkdownText.swift:927, or reverting the `sizeThatFits` change, would pass all four tests — the stated "regression coverage" purpose is not achieved.
**Suggestion:** Add a test that renders a real `MarkdownTable`/`MarkdownText` cell (or at least calls `SelectableTextView.sizeThatFits` through a host view with zero/stale bounds) and asserts the resulting container has `maximumNumberOfLines == 0` and sufficient height at the column width.

### [major] `widthTracksTextView` is re-enabled *before* measurement, contradicting the fix's own rationale
**File:** Conduit/Views/Components/SelectableTextView.swift:209-214
**Issue:** The comment says the flag is "temporarily disable[d] ... to prevent the container from reverting to the stale bounds width during measurement," but the code restores it to its previous value (typically `true`, set in `configure()` at line 226) one line *before* `sizeThatFits` runs:
```swift
textView.textContainer.widthTracksTextView = false
textView.textContainer.size = CGSize(width: width, ...)
textView.textContainer.widthTracksTextView = previousTracksWidth   // restored
let measured = textView.sizeThatFits(...)
```
So during the actual measurement the container is free to snap back to the stale `bounds.width` (0 on first pass), reproducing the exact bug on the pass that matters. Only the post-measure `bounds.size.width = width` assignment (line 216-218) mitigates subsequent calls. Either the flag restore is a no-op (making the comment wrong) or it defeats the mechanism (making the fix ineffective on first layout) — the ordering is broken either way.
**Suggestion:** Keep `widthTracksTextView` disabled for the full measurement, set `textView.bounds.size.width = width` *before* measuring, and only restore the flag afterward:
```swift
textView.bounds.size.width = width
let previousTracksWidth = textView.textContainer.widthTracksTextView
textView.textContainer.widthTracksTextView = false
textView.textContainer.size = CGSize(width: width, height: .greatestFiniteMagnitude)
let measured = textView.sizeThatFits(CGSize(width: width, height: .greatestFiniteMagnitude))
textView.textContainer.widthTracksTextView = previousTracksWidth
```

### [minor] Mutating `bounds` inside `sizeThatFits` is a layout-time side effect
**File:** Conduit/Views/Components/SelectableTextView.swift:216-218
**Issue:** Writing `textView.bounds.size.width` from within the measurement callback modifies view state during layout. SwiftUI may call `sizeThatFits` multiple times with different proposals (especially inside the horizontal scroll + vertical chat scroll); each call forces the bounds width and can trigger further layout invalidation, risking extra layout passes or inconsistency if a later proposal is narrower/wider.
**Suggestion:** Prefer setting the width before measuring in `configure()`/the measurement block (see above) and avoid post-measure mutation; if retained, add a comment explaining why the side effect is required.

### [minor] Unbounded cell height reintroduces pathological layout cost
**File:** Conduit/Views/Components/MarkdownText.swift:919
**Issue:** A model-authored cell of hundreds of lines becomes a single tall, non-scrolling `UITextView` laid out eagerly and measured repeatedly via `sizeThatFits` (SelectableTextView.swift:195). The old 4-line cap bounded this worst case. This is the documented "no nested scrolling" tradeoff, but the PR adds no guard against a pathological single-cell blob.
**Suggestion:** Either document the tradeoff in a comment at the cell creation site, or consider an explicit content-cap policy for tables if unbounded cells are a realistic model output.

### [nit] Test comment cites the wrong finite-limit value
**File:** ConduitTests/ChatTextSelectionTests.swift:688,692
**Issue:** The comment claims the guard targets "RenderCard source preview," but the actual RenderCard call passes `maximumNumberOfLines: 5` (MarkdownText.swift:1421). The test hardcodes `4`, so it neither matches the real caller nor would catch a change to that value.
**Suggestion:** Use `5` to match the actual call site, or drop the RenderCard reference from the comment.

## Verdict
REQUEST_CHANGES

The intent is correct and the code change is minimal — `InlineMarkdown`/`SelectableTextView` already default to unlimited lines with `.byWordWrapping`, and removing the cap matches the documented symptom. No retain cycles, force unwraps, or concurrency issues introduced. However, the measurement code re-enables `widthTracksTextView` before measuring (undoing the protection its own comment claims), and none of the four new tests actually executes either changed code path — so neither the fix's correctness nor its regression protection is verified. Both need addressing before merge.

<a href="https://opencode.ai/s/wWzxHB14">New%20session%20-%202026-08-17T21%3A30%3A46.184Z</a>
[opencode session](https://opencode.ai/s/wWzxHB14)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32071378787)

kaishi00:
## Summary

This PR removes the hard-coded `maximumNumberOfLines: 4` cap on markdown table cells (so cells expand to intrinsic height) and adds a shared measurement path that forces the text container to the proposed width before calling `UITextView.sizeThatFits`, plus four regression tests.

## Findings

### [major] `widthTracksTextView` is restored before the measurement, so the fix's own mechanism is nullified on the pass it targets
**File:** Conduit/Views/Components/SelectableTextView.swift:216-224
**Issue:** The new `measuredWrappingHeight` disables `widthTracksTextView`, sets `container.size` to the target width, then restores the flag *one line before* `sizeThatFits` runs:
```swift
container.widthTracksTextView = false
container.size = CGSize(width: width, height: .greatestFiniteMagnitude)
container.widthTracksTextView = tracksWidth      // ← restored here
let measured = textView.sizeThatFits(...)        // ← measurement runs with tracking re-enabled
```
With `widthTracksTextView == true` (the value `configure()` sets at line 229), the container width is derived from `textView.bounds.width`, not from `container.size`, during layout measurement. On the first pass / unconstrained-width pass this PR is meant to fix, `bounds.width` is 0, so the container snaps back to the stale width and the measurement collapses exactly as before. The canonical pattern (Apple's approach) keeps `widthTracksTextView = false` through the `sizeThatFits` call. Either the flag dance is a no-op (making the doc comment at lines 212-215 wrong) or it defeats the mechanism (making the fix ineffective when it matters).
**Suggestion:** Keep the flag disabled for the full measurement and restore after:
```swift
let tracksWidth = container.widthTracksTextView
container.widthTracksTextView = false
container.size = CGSize(width: width, height: .greatestFiniteMagnitude)
let measured = textView.sizeThatFits(CGSize(width: width, height: .greatestFiniteMagnitude))
container.widthTracksTextView = tracksWidth
```

### [major] Regression tests do not protect either fix
**File:** ConduitTests/ChatTextSelectionTests.swift:606-700
**Issue:** None of the tests execute the code that was changed:
- **Bug 1 (removed `maximumNumberOfLines: 4` at MarkdownText.swift:919) has zero coverage.** `measureCell` builds a bare `UITextView` and never renders `MarkdownTable.tableRow` / `InlineMarkdown`. Re-adding the cap at the table-row call site passes every test.
- **The height tests also pass with Bug 2 still present.** At the buggy 1pt/zero-width measurement a long string wraps to far more lines, so the assertions in `testTableCellMeasurementAtRealisticWidthProducesSufficientHeight`, `testMeasurementUsesColumnWidthNotStaleBounds`, and `testDifferentLengthCellsProportionalHeights` ("taller than 4 lines", "taller than one line", "wider ≤ narrower") all hold under the stale-width bug — the buggy result is *huge*, not small. The tests assert magnitudes, not the actual wrapped height at the requested column width.
- The helper hardcodes the intended behavior (sets container width itself) and calls `measuredWrappingHeight` directly, never exercising the hosted `sizeThatFits` path or its nil-width-proposal fallback (SelectableTextView.swift:205-207).
**Suggestion:** Add a test that renders a real `MarkdownTable` cell (or at minimum calls `measuredWrappingHeight` on a text view whose container is left at the stale width) and asserts the measured height equals the height of a correctly-framed reference text view measured at the same column width — i.e., an equality/abs() comparison rather than "taller than X" thresholds.

### [minor] Nil-width fallback can still return nil on the first layout pass inside the horizontal ScrollView
**File:** Conduit/Views/Components/SelectableTextView.swift:205-207
**Issue:** Inside `ScrollView(.horizontal)` the proposal width is nil; the fallback uses `textView.bounds.width`, which is 0 on the first pass, so `sizeThatFits` returns `nil` and SwiftUI falls back to `intrinsicContentSize` (computed from the stale container). Layout self-heals on later passes, but the first frame can still measure at the stale width — the same symptom the PR is fixing, at least transiently.
**Suggestion:** Guarantee a sensible width (e.g. clamp to a minimum column width or the last non-zero `container.size` width) instead of returning nil, or document the intentional transient.

### [minor] Unbounded cell height reintroduces pathological layout cost
**File:** Conduit/Views/Components/MarkdownText.swift:919
**Issue:** A model-authored cell with hundreds of lines becomes a single tall, non-scrolling `UITextView` laid out eagerly inside the horizontal ScrollView and measured repeatedly via `sizeThatFits`. The old 4-line cap bounded this worst case. This is the documented "no nested scrolling" tradeoff, but there is no guard against an unbounded single-cell blob.
**Suggestion:** Add a defensive comment at the cell-creation site, or adopt an explicit content-cap policy for table cells if unbounded model output is realistic.

### [nit] Test comment cites the wrong finite-limit value
**File:** ConduitTests/ChatTextSelectionTests.swift:687-693
**Issue:** The comment says the guard targets "RenderCard source preview," but the actual RenderCard call site passes `maximumNumberOfLines: 5` (MarkdownText.swift:1421), while the test hardcodes `4`. It neither matches the real caller nor would catch a change to that value.
**Suggestion:** Use `5` to match the call site, or drop the RenderCard reference.

### [nit] Test asserts the property re-set instead of the propagation path
**File:** ConduitTests/ChatTextSelectionTests.swift:689-700
**Issue:** `testFiniteLineLimitStillTruncates` manually sets `maximumNumberOfLines = 4` and `lineBreakMode` on a bare text view and asserts those values back — it tests nothing that `configure()` (SelectableTextView.swift:233-236) or the `MarkdownTable` call site do.
**Suggestion:** Call `configure`-equivalent setup or assert on `InlineMarkdown.buildTextView(...)` output so the test actually verifies the finite-limit propagation.

## Verdict

REQUEST_CHANGES

The direction is correct and the code change is minimal — no retain cycles, force unwraps, concurrency, or accessibility issues were introduced. However, `widthTracksTextView` is restored before the measurement it is supposed to protect, which is exactly the stale-width pass the fix targets, and the four new tests pass even with both bugs reintroduced, so neither the fix's correctness nor its claimed regression coverage is verified. Both need addressing before merge.

<a href="https://opencode.ai/s/hnIH2JVr">New%20session%20-%202026-08-17T21%3A47%3A54.419Z</a>
[opencode session](https://opencode.ai/s/hnIH2JVr)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32072843848)

kaishi00:
## Summary
Removes the hard-coded `maximumNumberOfLines: 4` cap on markdown table cells so cells expand to intrinsic height, and routes table-cell measurement through a shared `measuredWrappingHeight` static that measures at the proposed width, with four new regression tests.

## Findings

### [major] No test covers the actual fix — the removed cap can be re-added without any test failing
**File:** ConduitTests/ChatTextSelectionTests.swift:604-708
**Issue:** All four tests build bare `UITextView`s and call `SelectableTextView.measuredWrappingHeight` directly. None renders `MarkdownTable.tableRow` → `InlineMarkdown` (MarkdownText.swift:916-937), which is the exact site where `maximumNumberOfLines: 4` was deleted. The containers in these tests are unlimited by construction, so re-adding `maximumNumberOfLines: 4` at MarkdownText.swift:919 — the regression the PR claims to guard against — passes every test. The stated "regression coverage" for the primary fix is therefore not achieved.
**Suggestion:** Add a test that builds the real `InlineMarkdown`/`MarkdownTable` cell for a long-cell source and asserts the produced `textContainer.maximumNumberOfLines == 0` (or that its measured height exceeds four lines), so the lifted cap is pinned at the call site.

### [minor] The Bug-2 fix path (`sizeThatFits` nil-proposal fallback) is never exercised by tests
**File:** Conduit/Views/Components/SelectableTextView.swift:200-209
**Issue:** The changed code that matters is the nil-width fallback (lines 205-207) and the `isFinite` guard in `SelectableTextView.sizeThatFits(_:uiView:)`. The tests bypass this completely — `measureCell` calls the static `measuredWrappingHeight` directly, never the hosted `sizeThatFits`. `testMeasurementUsesTargetWidthNotStaleBounds` effectively pins plain `UITextView.sizeThatFits` honoring its width argument, which is stock UIKit behavior, unchanged by this PR. A regression that broke the fallback or incorrectly returned `nil` on first pass would not be caught.
**Suggestion:** Add a test that drives measurement through a host view with zero/stale bounds (or at least calls `measuredWrappingHeight` on a text view left in the true degenerate state) and asserts a finite, realistic height is produced.

### [minor] First-pass layout inside the horizontal ScrollView can still collapse to the stale width
**File:** Conduit/Views/Components/SelectableTextView.swift:205-207
**Issue:** In `ScrollView(.horizontal)` the width proposal is nil; if `textView.bounds.width` is 0 on the first measurement, `sizeThatFits` returns `nil` and SwiftUI falls back to `intrinsicContentSize`, which derives from the container width set in `configure()` (`max(bounds.width, 1)` ≈ 1pt, SelectableTextView.swift:226). The cell can be laid out at the enormous/erroneous size for one or more passes before a concrete width lands in `bounds`. The PR's description identifies the 1pt container as a root cause, yet the nil path is exactly where that state is still honored.
**Suggestion:** When the proposal width is nil and the fallback is unusable (0/non-finite), clamp to a sane default (e.g. last known non-zero `textContainer.size` width or a minimum column width) rather than returning nil, or document the transient explicitly.

### [minor] Misleading doc comment about iOS 17+/TextKit 2
**File:** Conduit/Views/Components/SelectableTextView.swift:212-215
**Issue:** The comment asserts "On iOS 17+ (TextKit 2) the proposed width drives wrapping independent of the text container's stored size, so no container mutation is needed." TextKit 2 is the default for selectable text views since iOS 16, and this codebase explicitly supports iOS 16 (see the delegate comment at SelectableTextView.swift:426-430). Asserting the behavior as an iOS-17+-only fact is inaccurate, and "no container mutation is needed" is asserted rather than demonstrated.
**Suggestion:** Correct the version claim and either verify measurement against a container that did not honor the width (the reason the container mutation existed at all) or soften the comment to describe the verified behavior.

### [nit] `testFiniteLineLimitStillTruncates` is circular
**File:** ConduitTests/ChatTextSelectionTests.swift:679-708
**Issue:** The test sets `maximumNumberOfLines = 4` and `.byTruncatingTail` on the container itself, then asserts those same values back plus a height ceiling. It verifies only its own setup; it never exercises `configure()` (SelectableTextView.swift:228-231) or any real caller such as the approval-command preview that actually passes `maximumNumberOfLines: 5` (ChatView.swift:1954 / MarkdownText.swift:1421).
**Suggestion:** Assert on the output of `configure()`-equivalent setup, or build the view with `maximumNumberOfLines: 5` to match a real call site, so the test covers the finite-limit propagation path rather than itself.

### [nit] Test helper avoids the true degenerate state that caused the bug
**File:** ConduitTests/ChatTextSelectionTests.swift:622-626
**Issue:** `measureCell` seeds the container at 90pt and its own comment notes this deliberately avoids the real 1pt state that `configure()` produces on first pass (SelectableTextView.swift:226). The tests therefore never measure in the condition the fix targets — the worst case being the exact scenario a regression would reintroduce.
**Suggestion:** Also exercise the true 1pt container state (and the nil-proposal fallback) in at least one test, asserting e.g. an equality/abs() comparison against a correctly-framed reference measurement rather than magnitude thresholds, which hold under the buggy stale-width output too.

## Verdict
REQUEST_CHANGES

The code change itself is minimal and sound — no retain cycles, force unwraps, concurrency, or accessibility issues are introduced, and removing the cap plus measuring at the proposed width matches the documented symptom. The earlier `widthTracksTextView`-ordering defect from prior review iterations is resolved in the current code (the container is no longer mutated at all). However, the regression suite still does not execute either actual fix path: nothing renders the real `MarkdownTable` cell (the removed cap), and the hosted `sizeThatFits` nil-proposal fallback is never invoked. The tests mostly pin stock UIKit measurement behavior, so the PR's core regression-protection claim is unmet. Adding tests that exercise the real call site and the nil-width path would close the gap.

<a href="https://opencode.ai/s/0qQpXUzC">New%20session%20-%202026-08-17T22%3A10%3A31.391Z</a>
[opencode session](https://opencode.ai/s/0qQpXUzC)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32074685654)

kaishi00:
## Summary
This PR removes the hard-coded `maximumNumberOfLines: 4` cap on markdown table cells (so cells expand to full intrinsic height) and adds a shared `measuredWrappingHeight` measurement path that falls back to the last laid-out bounds width when the width proposal is nil (horizontal ScrollView case), along with four regression tests.

## Findings

### [major] Tests do not exercise either fixed code path
**File:** ConduitTests/ChatTextSelectionTests.swift:599-717
**Issue:** All four tests build bare `UITextView`s via `makeTextView()` and call the static `SelectableTextView.measuredWrappingHeight` directly. None renders `MarkdownTable.tableRow` → `InlineMarkdown` (MarkdownText.swift:919), the site where `maximumNumberOfLines: 4` was deleted, and none invokes the hosted `sizeThatFits(_:uiView:context:)` whose nil-proposal fallback (SelectableTextView.swift:205-207) is the Bug 2 fix. Re-adding `maximumNumberOfLines: 4` at MarkdownText.swift:924, or reverting the `proposal.width ?? bounds.width` fallback, passes every test. The stated regression coverage for both bugs is not achieved.
**Suggestion:** Add (a) a test that renders a real `MarkdownTable`/`InlineMarkdown` cell and asserts the produced `textContainer.maximumNumberOfLines == 0` (or its measured height exceeds four lines), and (b) a test driving measurement through the host view with a nil width proposal and zero bounds, asserting a finite realistic height.

### [minor] First layout pass in the horizontal ScrollView can still fall back to the stale 1pt container
**File:** Conduit/Views/Components/SelectableTextView.swift:205-207
**Issue:** Inside `ScrollView(.horizontal)` the width proposal is nil; if `textView.bounds.width` is 0 on the initial pass, the fallback yields nil and `sizeThatFits` returns nil (line 207), so SwiftUI falls back to `intrinsicContentSize`, which is computed from the container width `configure()` set to `max(bounds.width, 1)` ≈ 1pt (line 226). The cell can be measured at a pathological height for one or more passes before a concrete width lands in `bounds`. The PR description identifies the 1pt container as a root cause, yet the nil path is where that exact state is still honored.
**Suggestion:** When the proposal width is nil and the fallback is unusable (0/non-finite), clamp to a sane default (last non-zero `textContainer.size.width` or a minimum column width) instead of returning nil, or surface the transient measurement in a comment.

### [major] Fix correctness depends on an unverified TextKit assumption that the tests don't pin
**File:** Conduit/Views/Components/SelectableTextView.swift:212-219
**Issue:** `measuredWrappingHeight` performs no container mutation and relies on `UITextView.sizeThatFits` honoring the passed width over the text container's stored width. The test helper deliberately seeds the container at 90pt and its own comment notes this avoids the true 1pt degenerate state that `configure()` produces (SelectableTextView.swift:226) — i.e., the exact worst case a regression would reintroduce. `testMeasurementUsesTargetWidthNotStaleBounds` proves the width parameter wins over a 90pt container but says nothing about the 1pt/bounds-zero production state, which is where the original bug lived.
**Suggestion:** At least one test should measure in the true `configure()`-produced 1pt/bounds-zero state and compare (equality/abs) against a correctly-framed reference text view measured at the same column width, rather than magnitude thresholds that can hold under the stale-width output.

### [minor] Unbounded cell height reintroduces pathological layout cost
**File:** Conduit/Views/Components/MarkdownText.swift:916-937
**Issue:** A model-authored cell with hundreds of lines becomes a single tall, non-scrolling `UITextView` laid out eagerly inside the horizontal ScrollView and measured repeatedly via `sizeThatFits`. The old 4-line cap bounded this worst case. This is the documented "no nested scrolling" tradeoff, but there is no guard or defensive note against an unbounded single-cell blob.
**Suggestion:** Add a comment at the cell-creation site noting the unbounded-height tradeoff, or adopt an explicit content-cap policy for table cells if unbounded model output is realistic.

### [nit] Misleading "iOS 17+ (TextKit 2)" version qualifier
**File:** Conduit/Views/Components/SelectableTextView.swift:213-215
**Issue:** The comment attributes proposal-width-driven wrapping to iOS 17+/TextKit 2. TextKit 2 has been the default for selectable text views since iOS 16; the version qualifier is inaccurate and "the proposed width drives wrapping independent of the stored container size" is asserted, not demonstrated. Since the deployment target is 17.0 (project.yml:15) the practical risk is low, but the claim as written is not a reliable basis for skipping container setup.
**Suggestion:** Soften the comment to state the verified behavior or drop the version qualifier.

### [nit] `testFiniteLineLimitStillTruncates` is circular
**File:** ConduitTests/ChatTextSelectionTests.swift:688-717
**Issue:** The test sets `maximumNumberOfLines = 4` and `.byTruncatingTail` on the container itself, then asserts those same values back plus a height ceiling. It validates its own setup only; it never exercises `configure()` (SelectableTextView.swift:228-231) or a real finite-limit caller such as the approval-command preview that passes `maximumNumberOfLines: 5` (ChatView.swift:1954 / MarkdownText.swift:1421).
**Suggestion:** Build the view through a `configure()`-equivalent path with `5` to match the real call site so the test covers finite-limit propagation rather than itself.

## Verdict
REQUEST_CHANGES

The code change is minimal and removal of the cap plus the shared width-aware measurement matches the documented symptom; no retain cycles, force unwraps, concurrency, or accessibility issues were introduced, and the earlier `widthTracksTextView`-ordering defect from previous iterations is resolved (no container/bounds mutation remains). However, re-adding the deleted `maximumNumberOfLines: 4` at the table call site or breaking the nil-proposal fallback would pass all four tests, and the true 1pt/bounds-zero production measurement state is deliberately avoided in the tests — so neither fix's correctness nor its claimed regression protection is verified. Tests that render the real `MarkdownTable` cell and drive the hosted `sizeThatFits` nil-width path would close the gap.

<a href="https://opencode.ai/s/plyvLsCg">New%20session%20-%202026-08-17T22%3A29%3A28.389Z</a>
[opencode session](https://opencode.ai/s/plyvLsCg)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32076155900)

kaishi00:
## Summary

Removes the hard-coded `maximumNumberOfLines: 4` cap on markdown table cells (cells now expand to full intrinsic height) and makes table-cell measurement deterministic by self-sizing each cell at its content's ideal width clamped to the `112...220` column policy, independent of layout proposals or UIKit bounds.

## Findings

### [minor] Unbounded cell height + double text measurement per layout pass
**File:** Conduit/Views/Components/MarkdownText.swift:928-930, Conduit/Views/Components/SelectableTextView.swift:221-234
**Issue:** A model-authored cell of hundreds of lines becomes a single tall, non-scrolling `UITextView` laid out eagerly inside the horizontal ScrollView. Each cell now also pays two text measurements per `sizeThatFits` call (`measureNonWrapping()` boundingRect plus `measuredWrappingHeight`), re-computed on every layout pass for every cell in every row. The unbounded-height tradeoff is documented in the PR body but there is no comment at the call site and no guard against a pathological blob.
**Suggestion:** Add a short comment at the `tableRow` cell-creation site noting the unbounded-height tradeoff, and consider caching the `measureNonWrapping()`/wrapping measurements per attributed-text change if tables with many cells become a layout bottleneck.

### [minor] Chain-test height reference is self-consistent, not independent ground truth
**File:** ConduitTests/ChatTextSelectionTests.swift:756-765
**Issue:** The full-chain test measures the reference height at `longCell.bounds.width` — a value *produced by the fix under test*. A regression that yields a wrong-but-deterministic width would still make the height assertion pass (the width equality in `testTableCellFirstPassMeasurementMatchesFullWrapReference` is what actually discriminates). The `maximumNumberOfLines == 0` assertion does genuinely pin Bug 1, and the first-pass width/heap equality pins Bug 2, so this is not a coverage hole — just a weaker assertion than intended.
**Suggestion:** Compute the reference at the independently-derived clamped ideal width (rather than `longCell.bounds.width`) so the height assertion is also independent of the rendered cell's own size.

### [nit] Misleading "iOS 17+ (TextKit 2)" qualifier in doc comment
**File:** Conduit/Views/Components/SelectableTextView.swift:236-239
**Issue:** The comment attributes proposal-width-driven wrapping to iOS 17+/TextKit 2, but TextKit 2 has been the default for selectable text views since iOS 16 (and the codebase explicitly accommodates iOS 16 elsewhere). The claim that no container mutation is needed is asserted, not demonstrated — `measureCell` seeds a 90pt container, and only the first-pass test exercises the true zero-width state (via the range path, which skips the container anyway).
**Suggestion:** Drop or soften the version qualifier and describe the verified behavior (e.g. "measured via `UITextView.sizeThatFits` at the target width").

### [nit] `testFiniteLineLimitStillTruncates` is circular and uses the wrong discriminator
**File:** ConduitTests/ChatTextSelectionTests.swift:691-717
**Issue:** The test sets `maximumNumberOfLines = 4` and `.byTruncatingTail` on the container itself, then asserts those same values back plus a height ceiling — it validates its own setup, never `configure()` (SelectableTextView.swift:252-255). The PR description cites "RenderCard source preview" as the guard target, but the real caller passes `maximumNumberOfLines: 5` (MarkdownText.swift:1426), so the test's hardcoded `4` would not catch a change to that call site.
**Suggestion:** Use `5` to match the real caller and assert via a `configure()`-equivalent path, or drop the RenderCard reference from the comment.

## Verdict

APPROVE

The earlier review blockers are resolved in the final state: `maximumNumberOfLines: 4` is removed and now covered directly by `testMarkdownTableLongCellRendersFullContentThroughViewChain` (renders the real `MarkdownText → MarkdownTable → InlineMarkdown → SelectableTextView` chain and asserts unlimited lines), and the Bug-2 nil-proposal/zero-bounds first pass is exercised by `testTableCellFirstPassMeasurementMatchesFullWrapReference`, which pins the width exactly and the first-pass ≡ steady-state invariant. The measurement path no longer mutates the container or bounds, and no retain cycles, force unwraps, concurrency, or accessibility issues were introduced. Remaining items are minor performance/documentation notes and a circular finite-line test; none block merge.

<a href="https://opencode.ai/s/eSoi09c2">New%20session%20-%202026-08-17T22%3A55%3A31.707Z</a>
[opencode session](https://opencode.ai/s/eSoi09c2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32078115814)

kaishi00:
## Summary
Removes the hard-coded `maximumNumberOfLines: 4` cap on markdown table cells (cells now expand to intrinsic height) and makes table-cell self-sizing deterministic: the measurement width is derived from the content's ideal width clamped to the existing `112...220` column policy rather than layout proposals or UIKit bounds, so first-pass equals steady-state height.

## Findings

### [minor] Chain-test height reference is self-referential
**File:** ConduitTests/ChatTextSelectionTests.swift:756-763
**Issue:** `testMarkdownTableLongCellRendersFullContentThroughViewChain` computes `fullWrapHeight` at `longCell.bounds.width` — a value *produced by the fix under test*. If a regression returns a wrong-but-deterministic width, the height assertion still passes because it references that same width. The `maximumNumberOfLines == 0` assertion pins the removal of the cap, and `testTableCellFirstPassMeasurementMatchesFullWrapReference` pins the width exactly, so this is not a coverage hole — just a weaker assertion than the comment implies.
**Suggestion:** Compute the reference at the independently-derived clamped ideal width (as the first-pass test already does) instead of the rendered cell's own bounds width.

### [minor] Two text-layout passes per cell per layout query
**File:** Conduit/Views/Components/SelectableTextView.swift:221-243
**Issue:** Every `sizeThatFits` for a self-sizing cell performs `measureNonWrapping()` (a `boundingRect` layout) plus `measuredWrappingHeight` (a `UITextView.sizeThatFits` layout), and SwiftUI re-invokes `sizeThatFits` across layout passes for every cell in every row. A table with many cells pays this repeatedly; a pathological single-cell blob of hundreds of lines is laid out eagerly on each pass.
**Suggestion:** Not blocking, but cache the `measureNonWrapping()`/wrapping measurements keyed on the attributed-content identity (or at least memoize within a layout pass), and add a short comment at the `tableRow` call site (MarkdownText.swift:930) documenting the unbounded-height tradeoff.

### [nit] Duplicated `112...220` literals across source and tests
**File:** Conduit/Views/Components/MarkdownText.swift:930,934; ConduitTests/ChatTextSelectionTests.swift:776,790
**Issue:** The column-range literals `112...220` (self-sizing) and `minWidth: 112, maxWidth: 220` (frame) appear in three places, plus `112`/`220` hardcoded in the tests. A change to one without the others silently diverges the measurement width from the rendered frame width — exactly the class of bug this PR fixes.
**Suggestion:** Define a shared constant (e.g. `static let tableColumnWidthRange: ClosedRange<CGFloat> = 112...220` on `MarkdownTable`) and reference it from both the `InlineMarkdown` call and the `.frame`, and from the tests.

### [nit] `testFiniteLineLimitStillTruncates` is circular and tests the wrong value
**File:** ConduitTests/ChatTextSelectionTests.swift:688-717
**Issue:** The test sets `maximumNumberOfLines = 4` and `.byTruncatingTail` on a bare container, then asserts those same values back plus a height ceiling — it validates only its own setup and never the `configure()` propagation path. It also hardcodes `4` while the real finite-limit caller passes `5` (MarkdownText.swift:1426), so it wouldn't catch a change at that call site.
**Suggestion:** Match the real caller's `5`, and drive the assertion through a `configure()`-equivalent path (or `InlineMarkdown`/`SelectableTextView` setup) so it verifies propagation rather than its own assignments.

### [nit] Version-qualified doc comment overstates the mechanism
**File:** Conduit/Views/Components/SelectableTextView.swift:236-239
**Issue:** The comment attributes proposal-width-driven wrapping to "iOS 17+ (TextKit 2)". TextKit 2 has been the default for text views since iOS 16, and `UITextView.sizeThatFits` honoring the width argument is demonstrated by `testMeasurementUsesTargetWidthNotStaleBounds` rather than by OS version. Deployment target is 17.0 (project.yml:15), so risk is low, but the phrasing is less precise than the tests it sits next to.
**Suggestion:** Reword to describe the verified behavior (e.g. "measured via `UITextView.sizeThatFits` at the target width, which wins over the stored container size — see tests") and drop the version qualifier.

## Verdict
APPROVE

Both bugs are fixed and now genuinely covered: the full-chain test (MarkdownText → MarkdownTable → InlineMarkdown → SelectableTextView) asserts every mounted cell text view has `maximumNumberOfLines == 0`, and the first-pass test drives the real `measuredSize(proposalWidth: nil)` path with zero bounds and pins first-pass ≡ steady-state width and height. Remaining items are minor test-rigor and maintainability notes. No retain cycles, force unwraps, concurrency, or accessibility regressions introduced; removing the cap improves accessibility by surfacing full cell content.

<a href="https://opencode.ai/s/zNSGQero">New%20session%20-%202026-08-17T23%3A07%3A50.732Z</a>
[opencode session](https://opencode.ai/s/zNSGQero)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32079032044)

PR #75, opened 2026-08-18
fix(chat): resolve Markdown reference-style links message-wide

kaishi00:
## Summary

Reference-style text links now resolve in chat messages:

```markdown
See the [Apple documentation][apple].

[apple]: https://developer.apple.com
```

renders as a tappable link, with the definition line hidden — in paragraphs, headings, lists, quotes, callouts, columns, and table cells. All CommonMark variants Foundation supports work: `[text][id]`, `[text][]` (collapsed), `[id]` (shortcut), case-insensitive labels, `"Title"` / `'Title'` / `(Title)` definition titles, and `<angle>` destinations. Ordinary inline links are untouched, and unresolved references degrade to literal text with no invented destination.

## Root cause

The renderer splits each message into blocks (paragraphs, list items, table cells, …) and parses every fragment **in isolation** with `AttributedString(markdown:, .init(interpretedSyntax: .full))`. A link reference definition elsewhere in the message is block-level Markdown that a fragment parser never sees, so `[text][id]` stayed literal — and a bare `[id]: url` line could end up as visible paragraph content.

## Approach

Keep Foundation as the single resolver — no regex rewriting of `[text][id]` uses:

- **`MarkdownParser.parseDocument`** (new) pre-scans the message and collects definition lines into a `MarkdownReferenceContext` (the raw definition Markdown, retained verbatim). Extracted lines leave a blank-line boundary so neighboring paragraphs don't merge. Collection skips fenced code and math blocks, requires CommonMark's ≤3-space indent, and — critically — only collects a line Foundation itself consumes invisibly (a regex match with, say, an unbalanced `)` in the destination stays visible in place instead of being duplicated after every block). A title-continuation line folds into its definition when Foundation consumes the pair. `MarkdownParser.parse` remains as a compatibility wrapper.
- **Both render paths** re-append the definitions when parsing each fragment (`fragment + "\n\n" + definitions`): the selectable-flow fast path via a `references` parameter on `MarkdownSelectionFormatter`, and the SwiftUI block path via a private, `MarkdownText`-scoped environment key read by `InlineMarkdown`. Fragments without references parse exactly as before (the context is appended only when non-empty), and a parse failure falls back to the bare fragment so definitions can never surface as text.
- **`MarkdownRenderCache`** retains the context alongside blocks; the existing cache key already contains the full source, so definition edits invalidate naturally.

Out of scope by design: footnotes (`[^1]:` stays visible text) and reference-style images.

## Tests

30 new tests in `MarkdownReferenceLinkTests.swift`:
- Foundation characterization pinning that `.full` resolves every reference form and hides definitions (the design's load-bearing assumption, verified first);
- extraction: before/after use, fence/math/unclosed-fence/indent/list-item/directive boundaries, footnote exclusion, blank-line boundary, Foundation-rejected and balanced-paren destinations, title folding, definitions-only messages;
- the selectable-flow path and the block-view path end-to-end (a table in `UIHostingController` forces `MarkdownText → MarkdownTable → InlineMarkdown → SelectableTextView`), plus `.link`-attribute assertions for ordinary links and no-invented-link assertions for unresolved ones.

All Markdown/selection suites green locally (99 tests). A local full-suite run hit the known simulator audio-subsystem crash (`AURemoteIO` RPC timeout) in unrelated `AppStateChatResumeTests` — 0 assertion failures; the same classes pass in isolation — so CI is the authoritative full-suite check for this PR.

Reviewed locally by three independent reviewers before push; findings triaged in the commit messages (taken: indent guard, Foundation-gated collection, title folding, test hardening; declined findings verified invalid with probe evidence).

## Summary by CodeRabbit

* **New Features**
  * Added support for Markdown reference-style links across messages, including rich text and selectable text.
  * Reference definitions can span paragraphs and lists while remaining hidden from rendered output.
  * Added support for titles and continuations in valid link definitions.
  * Unresolved references remain visible as plain text, while ordinary inline links continue to work.

* **Bug Fixes**
  * Prevented link definitions inside code blocks, math regions, directives, and footnotes from being incorrectly interpreted.

### DIFF
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -1632,15 +1632,20 @@ enum MarkdownParser {
         let lines = source.replacingOccurrences(of: "\r\n", with: "\n").components(separatedBy: "\n")
         var visibleLines: [String] = []
         var definitions: [String] = []
-        // Fenced code and math blocks are raw content — the same regions the
-        // block walker below consumes wholesale — so a definition-looking line
-        // inside them must stay put.
         var openFence: String?
         var mathClose: String?
 
         for line in lines {
             let trimmed = line.trimmingCharacters(in: .whitespaces)
 
+            // Fenced code and math blocks are raw content — the same regions
+            // the block walker below consumes wholesale — so a
+            // definition-looking line inside them must stay put. One known
+            // divergence: a fence opener inside a blockquote ("> ```") reads
+            // as a quote line to the walker but opens a fence here, so a
+            // definition-looking line in that region is collected while the
+            // walker renders the region as prose. Quote-nested fences are not
+            // a construct this renderer supports; accepted as an edge case.
             if let close = mathClose {
                 visibleLines.append(line)
                 if trimmed == close { mathClose = nil }
@@ -1671,9 +1676,11 @@ enum MarkdownParser {
 
         let visibleSource = visibleLines.joined(separator: "\n")
         let blocks = parseBlocks(visibleLines, recognizesGatewayMedia: recognizesGatewayMedia)
-        // Preserve the original fallback (non-empty source must render), but
-        // against the stripped body so a definitions-only message doesn't
-        // resurface its definitions as a visible paragraph.
+        // Defensive safety net carried over from the original parse(): if the
+        // walker ever produces nothing for non-whitespace input, surface the
+        // stripped body rather than rendering an empty message. Definitions
+        // were already removed from visibleSource, so they cannot resurface
+        // here even in that fallback.
         let finalBlocks = blocks.isEmpty && !visibleSource.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
             ? [.paragraph(visibleSource)]
             : blocks
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -1637,7 +1637,9 @@ enum MarkdownParser {
         var openFence: String?
         var mathClose: String?
 
-        for line in lines {
+        var index = 0
+        while index < lines.count {
+            let line = lines[index]
             let trimmed = line.trimmingCharacters(in: .whitespaces)
 
             // Fenced code and math blocks are raw content — the same regions
@@ -1651,29 +1653,36 @@ enum MarkdownParser {
             if let close = mathClose {
                 visibleLines.append(line)
                 if trimmed == close { mathClose = nil }
+                index += 1
                 continue
             }
             if let fence = openFence {
                 visibleLines.append(line)
                 if trimmed.hasPrefix(fence) { openFence = nil }
+                index += 1
                 continue
             }
             if let fence = fenceStart(trimmed) {
                 openFence = fence
                 visibleLines.append(line)
+                index += 1
                 continue
             }
             if let close = mathBlockOpening(trimmed) {
                 mathClose = close
                 visibleLines.append(line)
+                index += 1
                 continue
             }
-            if isDefinitionIndent(line), let definition = referenceDefinition(trimmed) {
-                definitions.append(definition)
-                visibleLines.append("")
+            if isDefinitionIndent(line), let definition = referenceDefinition(trimmed),
+               let span = consumedDefinitionSpan(definition, following: lines, at: index) {
+                definitions.append(span.markdown)
+                for _ in 0..<span.lineCount { visibleLines.append("") }
+                index += span.lineCount
                 continue
             }
             visibleLines.append(line)
+            index += 1
         }
 
         let visibleSource = visibleLines.joined(separator: "\n")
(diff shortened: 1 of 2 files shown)


kaishi00:
## Summary
This PR makes CommonMark reference-style links (`[text][id]` + `[id]: url`) resolve across an entire chat message by pre-collecting definition lines into a `MarkdownReferenceContext`, stripping them from the visible body, and re-appending them to every fragment parsed by Foundation in both render paths.

## Findings

### [minor] Per-frame parse amplification of the definitions suffix on the streaming path
**File:** Conduit/Views/Components/MarkdownText.swift:740-743
**Issue:** `InlineMarkdown.attributed` and `MarkdownSelectionFormatter.inline` re-parse `fragment + "\n\n" + definitionsMarkdown` for every fragment. On the uncached streaming path each frame re-runs `parseDocument`, and every fragment then re-processes the full definitions block. A message with M fragments and L definition lines is O(M×L) extra line-processing per streamed delta, and each definition-shaped line also triggers up to two whole-document Foundation parses inside `consumedDefinitionSpan` (1855-1868, 1871-1877) just to verify invisibility. For large messages with many definitions this parsing tax is paid on the main actor every frame.
**Suggestion:** Guard with an early-out for the common no-definition case (already done), but consider caching the parsed definition suffix (it is identical for every fragment in a message) or at least skipping the `consumedDefinitionSpan` Foundation probe for lines that cannot be a definition. At minimum, note the streaming cost for very large definition sets.

### [minor] Two sources of truth for "is this a valid definition"
**File:** Conduit/Views/Components/MarkdownText.swift:1833-1838
**Issue:** The regex `referenceDefinition` pre-filters, and Foundation is the actual gate in `consumedDefinitionSpan`. This is a redundant, independently-maintained grammar that can drift from Foundation's behavior (or from a future OS Foundation update). A line the regex rejects but Foundation would consume stays visible where a definition would have been consumed; a line the regex accepts but Foundation rejects is correctly gated out. The divergence risk is real but contained.
**Suggestion:** Acceptable given the Foundation gate, but consider documenting the regex as a pure perf pre-filter (it already is) and adding a characterization test that pins a couple of known regex/Founded mismatch cases so a Foundation behavior change is caught by CI.

### [nit] Title-continuation fold silently removes content-adjacent quoted lines
**File:** Conduit/Views/Components/MarkdownText.swift:1865-1876
**Issue:** A definition immediately followed by a line shaped like `"…"`, `'…'`, or `(…)` is consumed as a title pair and removed from the visible body whenever Foundation consumes the pair invisibly. This matches CommonMark (that line genuinely is a title continuation), so it is spec-correct, but it means a user who wrote a definition line followed by a standalone quoted line will see that line vanish.
**Suggestion:** No change required — behavior is CommonMark-correct and tested (`testParseDocumentFoldsTitleContinuationLineIntoDefinition`). Flagging only so the trade-off is explicit.

## Verdict
APPROVE

No retain cycles, force unwraps, concurrency/data-race violations, SwiftUI lifecycle bugs, or accessibility regressions were found. The definition removal actually improves VoiceOver output by keeping `[id]: url` metadata out of the accessibility tree. The two minor findings are performance/robustness observations, not correctness defects.

<a href="https://opencode.ai/s/JrTOxQx7">New%20session%20-%202026-08-18T01%3A28%3A45.299Z</a>
[opencode session](https://opencode.ai/s/JrTOxQx7)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32088285849)

kaishi00:
## Summary
Adds message-wide CommonMark reference-style link resolution by pre-collecting `[id]: url` definition lines in `MarkdownParser.parseDocument`, stripping them from the visible body, and re-appending them to every block fragment parsed by Foundation in both the fast selectable path and the SwiftUI block path.

## Findings

### [minor] Glued definition lines silently delete user-visible text
**File:** Conduit/Views/Components/MarkdownText.swift:1677-1681
**Issue:** A definition-looking line that directly follows paragraph text with no blank line (e.g. `Alpha beta\n[def]: https://x`) is stripped and replaced with a blank line, then re-appended after the shortened fragment. Your own directive-body comment and `testParseDocumentLeavesFoundationRejectedDefinitionVisibleInPlace` establish Foundation's semantic that definitions cannot interrupt a paragraph — so in that position Foundation/CommonMark render the line as visible paragraph text. After this change it disappears entirely, and a `[def]` use on the preceding line becomes a link it wasn't before. This is a deliberate, test-pinned superset (`testParseDocumentCollectsDefinitionAdjacentToParagraphWithoutBlankLine`), but it does change rendered content for input that strictly was not a definition, and it contradicts the PR's stated principle that ordinary text is untouched.
**Suggestion:** Consider keeping collection to lines preceded by a blank line or start-of-message (matching Foundation's own consumption context) so glued continuation text stays visible, or document this trade-off in the API docs since it is currently only discoverable in a commit message.

### [minor] Per-fragment re-parse amplification plus Foundation probes on the main actor
**File:** Conduit/Views/Components/MarkdownText.swift:282-288, 1850-1869
**Issue:** Every streaming frame re-runs `parseDocument` (MarkdownText.swift:237), and for each definition-shaped line `consumedDefinitionSpan` runs up to two whole-document `AttributedString(markdown:)` parses (lines 1871-1877) just to verify invisibility. At render time each of the M fragments then re-parses the full definitions suffix (`markdownForParsing`), so a message with L definitions pays O(M×L) extra line-processing per frame on the main actor.
**Suggestion:** The no-definition early-out is already in place. For the rarer message with definitions, memoize the parsed definition suffix once per document (it is identical for every fragment) rather than re-parsing it M times, and skip the Foundation probe for lines the plain regex already pins (e.g. no need to probe both the pair and the single line when the next line is not title-shaped).

### [minor] Regex pre-filter vs Foundation gate drift; valid definitions can be left uncollected
**File:** Conduit/Views/Components/MarkdownText.swift:1832-1848
**Issue:** `referenceDefinition` is an independently-maintained grammar that must stay in sync with Foundation. Forms the regex rejects but Foundation accepts — e.g. an empty angle destination `[id]: <>`, labels containing escaped brackets, or destinations with balanced-but-empty content — are never collected, so their uses stay literal *and* the line stays visible in a position where Foundation would have consumed it invisibly. Rejections are safe (gated by `consumedDefinitionSpan`) but acceptances are silently missing features, and the two grammars can drift across OS releases.
**Suggestion:** Acceptable as-is given the Foundation gate, but a characterization test pinning a few regex-reject/Foundation-accept mismatches would catch drift, as the previous review noted.

### [nit] Title-continuation fold can remove a standalone quoted line
**File:** Conduit/Views/Components/MarkdownText.swift:1865-1869
**Issue:** A definition immediately followed by a `"…"` / `'…'` / `(…)`-shaped line is folded into the definition whenever Foundation consumes the pair invisibly, even when the author clearly intended the second line as ordinary prose. This matches CommonMark (the shape filter correctly keeps block openers out), and it is tested, but a user who writes a definition followed by a quoted sentence will see that sentence vanish.
**Suggestion:** None required — flagging only so the trade-off stays explicit.

### [nit] Whitespace-only fallback semantics changed
**File:** Conduit/Views/Components/MarkdownText.swift:1692-1697
**Issue:** The defensive fallback changed from `!source.isEmpty` to `!visibleSource.trimmingCharacters(...).isEmpty`, so a whitespace-only message (e.g. `"   "`) now yields zero blocks instead of one paragraph. Practically irrelevant for chat content.
**Suggestion:** No action needed.

## Verdict

APPROVE

No retain cycles, force unwraps, concurrency/data-race violations, or SwiftUI lifecycle bugs introduced; the environment-scoped `MarkdownReferenceContext` is unambiguously scoped to the `MarkdownText` subtree (all `InlineMarkdown` call sites are inside it), the default `.empty` keeps both existing call paths and existing tests unchanged, and stripping definitions from the accessibility tree is a VoiceOver improvement. All findings are minor robustness/perf observations about an otherwise clean, well-tested implementation.

<a href="https://opencode.ai/s/jgHUDGs1">New%20session%20-%202026-08-18T02%3A28%3A37.439Z</a>
[opencode session](https://opencode.ai/s/jgHUDGs1)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32091899183)

Issue #74, opened 2026-08-18
Allow HTTP connections to local/VPN Hermes instances with explicit user confirmation


CypherNaught-0x:
After the latest Conduit update, I can no longer connect to my local Hermes instance because Conduit now blocks HTTP connections unless the host is localhost or a Tailscale tailnet host.

My Hermes dashboard is running on my LAN, for example:

http://192.168.178.xx:9119

This worked before the update. I also access the same network remotely through WireGuard, so even when I’m away from home the Hermes instance remains reachable over a private VPN without needing HTTPS.

Currently Conduit shows:

Remote dashboards must use HTTPS; HTTP is allowed only for localhost and Tailscale tailnet hosts.

I understand the motivation behind preventing credentials from being sent over unencrypted connections on arbitrary networks, but outright blocking HTTP makes Conduit difficult to use with otherwise secure private networks such as:

* Local LANs
* WireGuard VPNs
* Other private/self-hosted VPN setups
* Potentially other trusted RFC1918/private network environments

Suggested behavior

Instead of preventing the connection entirely, Conduit could show a prominent warning when HTTP is used and require explicit confirmation, for example:

This connection does not use HTTPS.

Credentials and other traffic may be visible to devices on the network. Only continue if you trust this network or are connecting through a secure VPN.

[Cancel] [I understand the risk — Connect]

Optionally, this acknowledgement could be remembered per host so the warning does not appear on every connection.

It might also make sense to require an additional confirmation before allowing Save credentials over HTTP, while still permitting the connection itself.

Why this matters

HTTPS is obviously preferable when traffic crosses an untrusted network, but it is not necessarily adding meaningful protection when the transport is already inside an encrypted WireGuard tunnel. Requiring users to deploy TLS purely to satisfy the client adds configuration and certificate-management overhead without materially improving security in that setup.

I think warning users strongly and requiring an explicit opt-in would preserve the safer default while still allowing advanced/self-hosted configurations.

Conduit version: latest version as of 2026-08-18
Hermes endpoint: self-hosted, HTTP on private LAN/WireGuard network

kaishi00:
Sorry, didn't catch the regression (not a setup I use or encourage people to use). I'll get it back into the next build. In the meantime though, if it's available to you, use one of the other HTTPS methods.

kaishi00:
[...]

PR #76, opened 2026-08-18
Allow private-LAN HTTP dashboards (restore RFC1918 policy)

kaishi00:
## Summary

Restores private-LAN HTTP dashboard support that was previously implemented on the old PR #9 integration branch (commit `609fef2`) and lost when main was rebuilt.

`http://192.168.x.x` dashboards were rejected with the insecure-transport error after the rebuild — this ports the previously-supported RFC1918 semantics forward as the smallest possible patch against current main.

## Changes

**`Conduit/Services/ConnectionURLPolicy.swift`**
- `isInsecureTransportAllowed` now also permits RFC1918 private IPv4 via a restored `isPrivateLANHost` helper:
  - `10.0.0.0/8`
  - `172.16.0.0/12`
  - `192.168.0.0/16`
- Extracted a shared strict `ipv4Octets` parser (exactly 4 dot-separated components, each an integer 0–255) used by both the private-LAN check and the existing Tailscale CGNAT check, so malformed hosts are rejected: `192.168.1.999`, `192.168.attacker.1`, partial IPv4 addresses. This also slightly tightens the CGNAT check (the previous inline split accepted trailing/leading-dot forms).
- Updated the insecure-transport error copy to mention private LAN again.

**`ConduitTests/SecurityBoundaryTests.swift`**
- Restored `testPrivateLANHTTPIsAllowed` (equivalent to the old test): pins `http://192.168.1.42:9119` — the user-visible regression — plus positive/negative RFC1918 boundaries (`172.15.255.255` / `172.32.0.1` / `192.167.0.1` rejected, `172.31.255.254` / `192.168.255.254` / `10.0.0.1` allowed) and malformed-host rejections.
- Restored `testInsecureTransportErrorExplainsAllowedLocalNetworks`.

## What this is NOT

- No revert/cherry-pick of PR #9: all later security/origin-matching and connection work on main is untouched (`originMatches`, `effectivePort`, URL normalization, WebSocket construction, Cloudflare/Tailscale handling, `DashboardTicketBridge`, `HermesClient`, reconnect/recovery state).
- Not an "allow arbitrary insecure HTTP host" preference — WireGuard/VPN host opt-in is separate scope (#74).

## Test plan

- [x] TDD: new tests fail against unmodified main (7 assertion failures), all pre-existing tests pass
- [x] Focused `SecurityBoundaryTests`: 11/11 passed
- [x] Full suite: 710 unit + 6 UI tests, 0 failures
- Historical reference only: PR #9's `isPrivateLANHost`/`ipv4Octets` semantics; the historical file was not restored wholesale.

## Summary by CodeRabbit

* **New Features**
  * HTTP connections are now supported for private local-network addresses, alongside localhost and Tailscale hosts.
  * Private-network URLs are normalized for consistent handling.
  * Insecure-connection messaging now explains which network types support HTTP.

* **Bug Fixes**
  * Improved validation prevents invalid or out-of-range private IP addresses from being accepted.

### DIFF
--- a/Conduit/Services/ConnectionURLPolicy.swift
+++ b/Conduit/Services/ConnectionURLPolicy.swift
@@ -9,7 +9,7 @@ enum ConnectionURLPolicyError: LocalizedError, Equatable {
         case .invalidURL:
             return "Enter a valid dashboard URL."
         case .insecureTransport:
-            return "Remote dashboards must use HTTPS; HTTP is allowed only for localhost and Tailscale tailnet hosts."
+            return "Remote dashboards must use HTTPS; HTTP is allowed only for local networks (localhost, private LAN, and Tailscale)."
         }
     }
 }
@@ -143,7 +143,7 @@ enum ConnectionURLPolicy {
 
     private static func isInsecureTransportAllowed(_ value: String) -> Bool {
         let host = value.lowercased().trimmingCharacters(in: CharacterSet(charactersIn: "[]"))
-        return isLoopbackHost(host) || isTailscaleHost(host)
+        return isLoopbackHost(host) || isPrivateLANHost(host) || isTailscaleHost(host)
     }
 
     private static func isLoopbackHost(_ value: String) -> Bool {
@@ -158,9 +158,31 @@ enum ConnectionURLPolicy {
         // Tailscale CGNAT range: 100.64.0.0/10 (100.64.0.0 – 100.127.255.255)
         // Require a well-formed IPv4 address to prevent label-based bypasses
         // like 100.64.attacker.example being accepted.
-        let octets = host.split(separator: ".").compactMap { Int($0) }
-        guard octets.count == 4,
-              octets.allSatisfy({ (0...255).contains($0) }) else { return false }
+        guard let octets = ipv4Octets(host) else { return false }
         return octets[0] == 100 && octets[1] >= 64 && octets[1] <= 127
+    }
+
+    private static func isPrivateLANHost(_ value: String) -> Bool {
+        // RFC1918 private ranges: 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16
+        guard let octets = ipv4Octets(value) else { return false }
+        switch octets[0] {
+        case 10:
+            return true
+        case 172:
+            return (16...31).contains(octets[1])
+        case 192:
+            return octets[1] == 168
+        default:
+            return false
+        }
+    }
+
+    private static func ipv4Octets(_ value: String) -> [Int]? {
+        let components = value.split(separator: ".", omittingEmptySubsequences: false)
+        guard components.count == 4 else { return nil }
+        let octets = components.compactMap { Int($0) }
+        guard octets.count == 4,
+              octets.allSatisfy({ (0...255).contains($0) }) else { return nil }
+        return octets
     }
 }
--- a/ConduitTests/SecurityBoundaryTests.swift
+++ b/ConduitTests/SecurityBoundaryTests.swift
@@ -51,6 +51,30 @@ final class SecurityBoundaryTests: XCTestCase {
         XCTAssertFalse(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://100.64.1")))
     }
 
+    func testPrivateLANHTTPIsAllowed() {
+        XCTAssertTrue(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://192.168.1.42:9119")))
+        XCTAssertEqual(
+            try? ConnectionURLPolicy.normalizedBaseURL("http://192.168.1.42:9119/"),
+            "http://192.168.1.42:9119"
+        )
+        XCTAssertTrue(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://10.0.0.1")))
+        XCTAssertTrue(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://172.16.0.1")))
+        XCTAssertTrue(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://172.31.255.254")))
+        XCTAssertTrue(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://192.168.255.254")))
+        XCTAssertFalse(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://172.15.255.255")))
+        XCTAssertFalse(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://172.32.0.1")))
+        XCTAssertFalse(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://192.167.0.1")))
+        XCTAssertFalse(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://192.168.1.999")))
+        XCTAssertFalse(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://192.168.attacker.1")))
+    }
+
+    func testInsecureTransportErrorExplainsAllowedLocalNetworks() {
+        XCTAssertEqual(
+            ConnectionURLPolicyError.insecureTransport.errorDescription,
+            "Remote dashboards must use HTTPS; HTTP is allowed only for local networks (localhost, private LAN, and Tailscale)."
+        )
+    }
+
     func testCGNATOctetBoundaryRejectsInvalidValues() {
         // Octet > 255 is not a valid IPv4 octet
         XCTAssertFalse(ConnectionURLPolicy.isAllowedTransport(URL(string: "http://100.64.256.1")))


kaishi00:
## Summary
Restores http:// access to RFC1918 private-LAN IPs in `ConnectionURLPolicy.isInsecureTransportAllowed` via a new `isPrivateLANHost` helper backed by a shared, tightened `ipv4Octets` parser, plus restored regression tests.

## Findings

### [minor] ATS runtime behavior isn't verified — policy layer only
**File:** ConduitTests/SecurityBoundaryTests.swift:54 and Conduit/Info.plist:63
**Issue:** The tests exercise only the policy logic against `URL(string:)`. Nothing verifies that a real `URLSession` HTTP load to a private-LAN IP is actually permitted by App Transport Security. ATS's treatment of numeric IPs, combined with `NSAllowsLocalNetworking` / `NSExceptionDomains`, is nuanced: the existing `100.64.0.0/10` exception implies numeric IPs *are* subject to ATS in this app, so whether `http://192.168.x.x` installs visibly depends on `NSAllowsLocalNetworking` covering RFC1918 ranges at the OS layer. A unit test cannot catch a mismatch here.
**Suggestion:** Add a runtime integration check (or at least a manual verification step in the PR description) that an actual insecure load to a private-LAN address in the simulator succeeds, since the policy now advertises it as allowed in the error copy.

### [minor] Security surface expansion: cleartext HTTP over private LAN
**File:** Conduit/Services/ConnectionURLPolicy.swift:165
**Issue:** This intentionally restores plaintext HTTP for the entire RFC1918 space with no opt-out, no per-host allowlist, and no confirmation. Any process on the LAN (or a website exploiting DNS rebinding/CSRF against the connected dashboard) can now reach these dashboards, and credentials flow in cleartext. The PR is explicit that this is restoring prior parity, so it's a deliberate product decision — but it's worth flagging that the blast radius is larger than the old Tailscale-only allowance.
**Suggestion:** Consider surfacing an explicit in-app disclosure when saving a plaintext LAN dashboard (the error copy already claims these are "allowed"), and revisit as a candidate for the separate opt-in scope referenced in #74 rather than an unconditional allow.

### [nit] `Int` token parsing accepts sign prefixes in `ipv4Octets`
**File:** Conduit/Services/ConnectionURLPolicy.swift:180
**Issue:** `Int($0)` tolerates an ASCII `+`/`-` sign, so a host like `192.168.+1.1` parses to octets `[192, 168, 1, 1]` and passes the private-LAN check (URL may retain the `+` in the host). Also, any future callers relying on octet bounds inherit this looseness. Low practical risk since `+` in a host won't resolve, but the function is described as "strict."
**Suggestion:** Validate tokens explicitly, e.g. only accept if the token is all ASCII digits with no sign (reject a leading `+`/`-`), or map via a `UInt8`-style parse with a length/canonicality check.

### [nit] Redundant normalization in host helpers
**File:** Conduit/Services/ConnectionURLPolicy.swift:149
**Issue:** `isInsecureTransportAllowed` already lowercases and trims brackets before dispatching, yet `isLoopbackHost` re-trims brackets and `isTailscaleHost` re-lowercases. Harmless, but the public-safety intent of the central normalization is obscured by the duplicated work.
**Suggestion:** Drop the redundant normalization in the per-checks or centralize it once, since `isPrivateLANHost`/`ipv4Octets` are already case/symbol-sensitive by design.

## Verdict
APPROVE

The range logic and parser are correct for all the stated boundaries (`172.16.0.0/12` via `16...31`, `192.168.0.0/16`, `10/8`), malformed-input rejection is airtight for realistic bypasses (labels, partial octets, trailing dots), no force unwraps, no state/concurrency issues, and the tightening of the CGNAT check is a genuine improvement. Findings above are non-blocking.

<a href="https://opencode.ai/s/mL5K7c2n">New%20session%20-%202026-08-18T03%3A09%3A28.435Z</a>
[opencode session](https://opencode.ai/s/mL5K7c2n)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32094381469)

kaishi00:
## Summary
Restores plaintext HTTP access for RFC1918 private-LAN dashboards (`ConnectionURLPolicy.isInsecureTransportAllowed`) via a new `isPrivateLANHost` helper backed by a shared `ipv4Octets` parser, plus regression tests pinning valid/invalid boundaries and updated error copy.

## Findings

### [minor] `ipv4Octets` accepts non-canonical IPv4 forms, contradicting the "strict" parser claim
**File:** Conduit/Services/ConnectionURLPolicy.swift:180
**Issue:** `Int($0)` tolerates a leading `+` and leading zeros. `http://192.168.+1.1` parses to `[192,168,1,1]` and passes the private-LAN gate; `http://010.0.0.1` / `http://172.16.01.1` also parse and pass. RFC 3986 disallows both (`dec-octet` has no leading zeros or signs), and resolvers may interpret `010` as octal (8.0.0.1 — not private), so the app can end up claiming an "allowed private-LAN" URL while the resolver targets a different, possibly non-private host. The `0...255` range check does not catch these because the numeric value is already in range.
**Suggestion:** Validate canonical grammar before converting — reject any component with a leading `+`/`-` or a multi-digit component with a leading `0` — then apply the existing 4-component and range checks. Add regression cases (`192.168.+1.1`, `010.0.0.1`, `192.168.01.1`).

### [minor] Policy layer and ATS config are never verified together at runtime
**File:** Conduit/Services/ConnectionURLPolicy.swift:144 (and Conduit/Info.plist:63)
**Issue:** Tests exercise only the policy against `URL(string:)`. Nothing confirms a real `URLSession`/`WKWebView` cleartext load to a numeric private-LAN IP is accepted by ATS, which the error copy now advertises as allowed. Notably, the `NSExceptionDomains` key `100.64.0.0/10` in Info.plist:76 is a literal that never matches an actual host (CIDR keys aren't supported), so ATS coverage for both the CGNAT and RFC1918 ranges silently relies on `NSAllowsLocalNetworking`/IP-literal behavior. A unit test can't catch this.
**Suggestion:** Add an integration/manual verification step (a real HTTP load to `192.168.x.x` in the simulator) and confirm whether ATS coverage should be made explicit, e.g. a functional exception mechanism or documented reliance on local-networking defaults.

### [minor] Security surface expansion without disclosure
**File:** Conduit/Services/ConnectionURLPolicy.swift:165
**Issue:** Restores unconditional cleartext HTTP for the entire RFC1918 space (10/8, 172.16/12, 192.168/16) — a much larger blast radius than the prior localhost/Tailscale-only allowance. Any process on the LAN can reach these dashboards, and the error copy now asserts they're "allowed." This matches the stated parity-restore intent, but the broader host-matching gate in DashboardTicketBridge.swift:663 still permits subframe loads to any allowed-transport host.
**Suggestion:** Surface an in-app disclosure when a plaintext LAN dashboard is saved, and track this as a candidate for the (#74) opt-in scope rather than an unconditional allow.

### [nit] Redundant normalization in host checks
**File:** Conduit/Services/ConnectionURLPolicy.swift:149
**Issue:** `isInsecureTransportAllowed` already lowercases and strips brackets, yet `isLoopbackHost` re-trims brackets (line 150) and `isTailscaleHost` re-lowercases (line 155). Harmless, but it obscures that `isPrivateLANHost`/`ipv4Octets` are deliberately case/symbol-sensitive.
**Suggestion:** Drop the duplicated normalization inside the per-checks since callers guarantee a normalized input.

### [nit] Rejection cases lack `normalizedBaseURL` coverage
**File:** ConduitTests/SecurityBoundaryTests.swift:67
**Issue:** Production entry points (`AppState.swift`, `DashboardTicketBridge.swift:629`, `LoginView.swift`) use `normalizedBaseURL`, but the invalid-host assertions only run through `isAllowedTransport(URL(string:))`. `URL(string:)` can fail or normalize before the validator runs, so the guard in `normalizedBaseURL` (line 47) is untested for these inputs.
**Suggestion:** Add `XCTAssertThrowsError(try ConnectionURLPolicy.normalizedBaseURL("http://\(host)"))` for the malformed hosts (`192.168.1.999`, `192.168.attacker.1`, partial/out-of-range cases).

## Verdict
APPROVE

The range math is correct for all stated boundaries (10/8, 172.16/12 via `16...31`, 192.168/16), the CGNAT tightening is genuine (`omittingEmptySubsequences: false` catches mid-string empty octets like `100.64.1..2` that the old split accepted), no force unwraps or concurrency issues, and the tests pin the user-visible regression. Findings are non-blocking; the canonical-grammar gap is the one worth hardening before broad adoption.

<a href="https://opencode.ai/s/b5ZZ6oMG">New%20session%20-%202026-08-18T03%3A35%3A19.368Z</a>
[opencode session](https://opencode.ai/s/b5ZZ6oMG)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32095955777)

PR #72, opened 2026-08-17
fix(ci): retry full test suite after simulator reset

kaishi00:
## Problem

CI run 316 failed because `ComposerPasteTextViewTests.testPasteItemProvidersFallsBackToUIImageWhenImageDataFails` timed out waiting 5 seconds for a "fallback image callback". The `DashboardTicketBridgeTests` all passed, confirming this is unrelated CI/test-runner flakiness.

## Root cause: retry never reliably executes

GitHub Actions runs each `run:` step under `bash -eo pipefail {0}`. When the `xcodebuild test | xcpretty` pipeline fails, the shell exits immediately on the pipeline result. The line `status=${PIPESTATUS[0]}` never executes, and the retry loop never reaches attempt 2.

## Root cause: UI-only retry masks unit failures

Attempt 2 was scoped to `-only-testing:ConduitUITests`. If attempt 1 failed in a unit test, a subsequent UI-only retry could pass and make CI green while the original unit-test failure was never retried.

## Fix

1. **Wrap the pipeline in `if`/`else`** — this is an exception context; errexit sees success (the condition evaluated) and does not kill the step. Both `PIPESTATUS` values are preserved: if xcodebuild succeeds but xcpretty fails (or vice versa), CI still fails and retries.

2. **Remove `-only-testing:ConduitUITests`** — both attempts run the full suite. A genuine double failure still fails CI.

3. **Preserve separate result bundles** (`TestResults.xcresult` / `TestResults-retry.xcresult`) uploaded on final failure.

## Validation

- Local shell reproduction confirmed: `bash -eo pipefail -c "if true | false; then ...; else ps=(\"${PIPESTATUS[@]}\"); status=$(( ps[0] != 0 ? ps[0] : ps[1] )); fi"` correctly captures status=1 when xcpretty fails.
- Attempt 2 has no `-only-testing:ConduitUITests`.
- A failed second attempt returns nonzero (`exit 1` after the loop).
- Successful attempt 1 exits successfully without retrying (`exit 0` inside the loop).
- No other files modified; no changes to test code or assertions.

## Summary by CodeRabbit

* **Tests**
  * Improved automated test reliability by fully resetting and restarting the simulator before retries.
  * The complete test suite now retries after failures, with clearer reporting of each attempt.
  * Test runs correctly fail when either build or test execution is unsuccessful.

PR #77, opened 2026-08-18
ci+test: bound hung test attempts with per-attempt wall-clock timeout

kaishi00:
## Problem

CI run #328 (PR #75) wedged for 37 minutes: a HermesClient test suspended forever on an unbounded wait (XCTest has **no per-test timeout for async tests**), xcodebuild emitted nothing further and never exited, and the #72 retry loop could never advance — the 45-minute job timeout killed the job and attempt 2 never ran. The job cap was doing wedge *detection*; it should only be an emergency limit.

## Fix — two layers

**1. Tests fail boundedly (`ConduitTests/HermesClientTests.swift`)**
- `Gate.wait()` is now deadline-bounded (NSLock + detached timer racing the signal; exactly-once continuation handoff under the lock — same shape as `HermesClient.waitForSocketOpen`'s open-timeout task). A missed signal records `XCTFail` naming the phase ("Timed out after 10.0s waiting for socket A's receive loop to suspend in socket.receive()") and throws to abort the test.
- New `awaitCompletion(of:_:)` helper bounds every `Task.value` wait the same way.
- All tests in the file use the helpers, not just the observed `testSupersededReceiveLoopDoesNotClobberNewConnection`.

**2. CI bounds each attempt (`scripts/ci-run-tests.sh`, new; retry loop moved out of `ci.yml`)**
- Per-attempt wall-clock budget: **16 min** (healthy runs are 7–13 min; sized so worst case — setup + killed attempt 1 + bootstatus-bounded reboots + full attempt 2 — fits the 45-min cap).
- xcodebuild runs backgrounded under `set -m` in its own process group; on deadline the whole group gets TERM → 10s grace → KILL, so xcodebuild's children (xctest, sim agents) die with it.
- A timed-out attempt exits **124** and flows through the retry loop exactly like a test failure → simulator shutdown/boot/bootstatus → **attempt 2 runs the full suite**.
- Simulator is explicitly booted before **every attempt, including the first**.
- Diagnostics survive: raw per-attempt logs stream into the step log and upload via `ci-logs/`, plus `simctl list devices` state and any partial `.xcresult` (upload step now includes `ci-logs/`).
- The 45-min job timeout stays as the outer emergency limit only.

## Validation

| Scenario | Result |
|---|---|
| Deliberately hung attempt killed within budget | ✅ harness: status 124, killed ~1s past deadline, group-kill verified against child pids |
| Attempt 2 executes after a timeout | ✅ harness: full reboot + attempt 2 runs |
| Normal test failure still retries | ✅ harness: exit 65 → attempt 2 → pass |
| Two failed/timed-out attempts fail CI | ✅ harness: exit 1 |
| Healthy run succeeds on attempt 1 | ✅ harness + this PR's own CI run |
| HermesClientTests cannot hang | ✅ injected never-signalling gate + never-completing task each fail boundedly in ~2s with the phase named; all 6 tests green on simulator |

Harness: mock-xcodebuild/xcrun via PATH driving the real script under `/bin/bash` 3.2 (GitHub's macOS default) — 13/13 checks.

## Review

Three local reviewer passes pre-push. Taken: a real bash dynamic-scoping bug where `local printed` shadowed the pass-by-name counter (log would re-print fully every poll), honest worst-case budget math vs the job cap, log-`ls` diagnostics, and Swift-6-friendly lock/sleep cleanups. Declined: perl-alarm wrappers around `simctl shutdown/boot` (job cap is the documented backstop), and a cancellation-handler rework of `Gate.wait` (XCTest never cancels a suspended async test; checked continuations are documented as cancellation-inert).

PR #75 touches only `MarkdownText.swift` + its own test file, so no conflict; it picks this up on merge to main.

## Summary by CodeRabbit

* **Tests**
  * Improved automated test reliability with bounded waits and clearer timeout reporting.
  * Added simulator reset and boot handling before each test attempt.
  * Test runs now retry automatically after failures or timeouts.
  * Live output, diagnostics, and partial results are preserved for failed attempts.

* **Chores**
  * Extended CI failure artifacts to include additional test-run logs.
  * Added an emergency time limit to prevent indefinitely stalled CI jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,9 +18,12 @@ jobs:
   test:
     name: Build & Test
     runs-on: macos-26
-    # A wedged simulator can hang a test invocation for hours (the default
-    # step timeout is 360 minutes). Healthy runs finish well under 15
-    # minutes; cap the job so a wedge surfaces as a bounded failure.
+    # Outer emergency limit only — wedge detection lives in
+    # scripts/ci-run-tests.sh (per-attempt wall-clock budget of 16 min).
+    # Worst case: ~4 min setup + 2 × (≤3 min bootstatus-bounded reset +
+    # 16 min budget + ~25 s kill/diagnostics) ≈ 43 min, so a bounded run
+    # fits; this cap exists for the residual cases the budget cannot reach
+    # (e.g. a wedged CoreSimulatorService blocking simctl itself).
     timeout-minutes: 45
 
     steps:
@@ -66,65 +69,15 @@ jobs:
           echo "SIMULATOR=$SIMULATOR" >> "$GITHUB_ENV"
 
       - name: Run tests
-        run: |
-          # If the full suite fails, reset the simulator and retry the full
-          # suite once. A genuine double failure still fails CI. Attempt 2
-          # runs the entire suite (not just UI tests) because attempt 1 can
-          # fail in a unit test; a UI-only retry would mask that failure.
-          #
-          # NOTE: GitHub Actions runs `bash -eo pipefail {0}` by default.
-          # We wrap the piped xcodebuild|xcpretty invocation in an `if`
-          # block so that an expected nonzero exit code is intentionally
-          # handled by the loop instead of killed by errexit.  (Running
-          # `set +e` would also work but the `if` scopes the failure
-          # cleanly and makes the intent obvious.)
-          set -uo pipefail
-          for attempt in 1 2; do
-            if [ "$attempt" -eq 1 ]; then
-              BUNDLE=TestResults.xcresult
-            else
-              BUNDLE=TestResults-retry.xcresult
-              # A wedged simulator is the one failure mode in-test retries
-              # cannot clear; restart the device so the retry starts clean.
-              xcrun simctl shutdown "$SIMULATOR" || true
-              sleep 3
-              xcrun simctl boot "$SIMULATOR" || true
-              # Wait for a complete boot before handing the device to
-              # xcodebuild; a half-booted simulator wedges UI tests.
-              xcrun simctl bootstatus "$SIMULATOR" -b -t 180 || true
-            fi
-            echo "::group::Test attempt $attempt"
-            rm -rf "$BUNDLE"
-            status=0
-            if xcodebuild test \
-              -project Conduit.xcodeproj \
-              -scheme Conduit \
-              -destination "platform=iOS Simulator,id=$SIMULATOR" \
-              -configuration Debug \
-              -resultBundlePath "$BUNDLE" \
-              CODE_SIGN_IDENTITY="" \
-              CODE_SIGNING_REQUIRED=NO \
-              CODE_SIGNING_ALLOWED=NO \
-              DEVELOPMENT_TEAM="" \
-              PROVISIONING_PROFILE_SPECIFIER="" \
-              | xcpretty --color; then
-              status=0
-            else
-              # Preserve both pipeline statuses: xcodebuild may succeed
-              # while xcpretty fails (or vice versa). Fail/retry if
-              # either command fails.
-              ps=("${PIPESTATUS[@]}")
-              status=$(( ps[0] != 0 ? ps[0] : ps[1] ))
-            fi
-            echo "::endgroup::"
-            if [ "$status" -eq 0 ]; then
-              echo "tests passed on attempt $attempt"
-              exit 0
-            fi
-            echo "tests failed on attempt $attempt (exit status $status)"
-          done
-          echo "both test attempts failed"
-          exit 1
+        # scripts/ci-run-tests.sh owns the retry loop. Each attempt gets its
+        # own wall-clock budget (default 16 min; healthy runs take 7-13) so a
+        # hung test or wedged simulator is killed and retried instead of
+        # stalling until the 45-min job timeout — which remains the outer
+        # emergency limit, not the wedge-detection mechanism. The script also
+        # boots the simulator before every attempt, streams raw xcodebuild
+        # output into this step's log, and leaves per-attempt logs + any
+        # partial .xcresult bundles for the upload step below.
+        run: bash scripts/ci-run-tests.sh
 
       - name: Upload test results on failure
         if: failure()
@@ -134,4 +87,5 @@ jobs:
           path: |
             TestResults.xcresult
             TestResults-retry.xcresult
+            ci-logs/
           if-no-files-found: ignore
--- a/ConduitTests/HermesClientTests.swift
+++ b/ConduitTests/HermesClientTests.swift
@@ -6,7 +6,7 @@ final class HermesClientTests: XCTestCase {
 
     // MARK: - rpc() guard
 
-    func testRPCGuardRejectsBeforeHandshake() async {
+    func testRPCGuardRejectsBeforeHandshake() async throws {
         // The handshake is held open (we never call transport.open), so the
         // socket is installed but `isConnected` is still false. `rpc()`'s
         // `isConnected` requirement must reject the call — the pre-fix
@@ -19,7 +19,7 @@ final class HermesClientTests: XCTestCase {
         let client = makeClient(transport: transport)
 
         let connectTask = Task { try? await client.connect() }
-        await reachedResume.wait()  // connect() has installed + resumed the socket and is now suspended in the handshake
+        try await reachedResume.wait("connect() to resume the socket (handshake pending)")  // connect() has installed + resumed the socket and is now suspended in the handshake
 
         do {
             try await client.healthCheck()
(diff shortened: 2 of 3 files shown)


kaishi00:
## Summary
Adds two layers of wedge-bounding so a hung async test can never wedge CI again: (1) all `HermesClientTests` wait/task-completion points are now deadline-bounded via a lock-protected `Gate` + detached deadline timer so a missed signal fails boundedly naming the phase, and (2) a new `scripts/ci-run-tests.sh` gives each xcodebuild attempt its own wall-clock budget enforced by a process-group kill, with the retry loop moved out of `ci.yml` and the 45-min job timeout demoted to an emergency backstop.

## Findings

### [minor] Timeout-default documentation diverges from the actual default
**File:** scripts/ci-run-tests.sh:27 / .github/workflows/ci.yml:73
**Issue:** The script header says the default budget is `1080` (18 min) and the ci.yml step comment says "default 18 min", but the actual code default is `960` (16 min) at scripts/ci-run-tests.sh:35, matching the 16-min math used for the 45-min cap. Three places state the default; two of them are stale. Anyone tuning `ATTEMPT_TIMEOUT_SECS` from the documented contract will get different behavior than expected.
**Suggestion:** Point both comments at 960 s / 16 min (or bump the actual default to 1080 and re-verify the worst-case math against the 45-min cap).

### [minor] Liveness log streaming can drop the wedge point on kill
**File:** scripts/ci-run-tests.sh:65-76, 148-149
**Issue:** `stream_new_lines` uses `wc -l` + `tail -n +N`, which only counts newline-terminated lines. When the deadline fires, the group is TERM'd/KILL'd while xcodebuild may still have an unflushed partial line in its stdio buffer — the very line describing what it was wedged on. The "last 200 log lines" diagnostic then risks omitting exactly the useful part.
**Suggestion:** Tolerable for diagnostics; minor improvement would be appending any trailing non-newline tail (`tail -c` residue) or grepping the live log for test-phase markers in the kill diagnostics. Not blocking.

### [minor] Success-path of `Gate.wait` leaks a detached watchdog task per wait
**File:** ConduitTests/HermesClientTests.swift:268-271
**Issue:** The deadline task is created for every `Gate.wait`, including the many that succeed long before the timeout. It is never retained or cancelled on the happy path, so each successful wait leaves a detached task sleeping the full `timeout` (10 s) before `fireDeadline` no-ops. Harmless at this scale (a handful of gates per run) but it grows with every helper call.
**Suggestion:** Store the watchdog task in a property and `cancel()` it inside `signal()` (under the lock, before resuming) so the race stays exactly-once and no timer outlives its wait.

### [minor] On timeout, `awaitCompletion` cancels only the watcher, not the awaited task
**File:** ConduitTests/HermesClientTests.swift:206-212
**Issue:** When `finished.wait` times out, the deferred `watcher.cancel()` runs but the original `task` (e.g. a `connect()` still mid-flight) is left running and can keep mutating the client's state after the test has already failed and is unwinding. Documented as intentional, and checked continuations are cancellation-inert, but the `connect()` itself never observes cancellation.
**Suggestion:** Optionally add `task.cancel()` alongside `watcher.cancel()` in the defer so a failed test at least attempts to stop the orphaned work. Low impact given the test process is already considered failed.

### [nit] Heartbeat cadence assumes a fixed 15 s poll
**File:** scripts/ci-run-tests.sh:122-126
**Issue:** Progress echoes on `heartbeat % 4`, which works out to ~60 s only while the poll stays at 15 s; once `remaining < 15` the cadence shortens and the elapsed/remaining arithmetic at line 123 remains correct but the heartbeat interval is no longer uniform. Cosmetic.
**Suggestion:** Compute cadence from wall-clock elapsed (`now - start`) instead of iteration count, or leave as-is — output cadence is not correctness-critical.

## Verdict
APPROVE

The core logic is sound and carefully reviewed: the `Gate` continuation hand-off is exactly-once under a single NSLock across the signal/timeout race, closures that outlive locals use `[weak self]`, no force unwraps or unsafe optionals were introduced, the mock-types stay MainActor-isolated while the deadline timers run detached behind a lock, and the bash retry/kill plumbing (dynamic-scoping counter fix, `set -m` process-group kill with TERM→grace→KILL, 124 flowing through the retry loop) was verified with a real harness including group-kill of child PIDs. The budget math vs. the 45-min cap is documented honestly, including the residual cases the budget cannot reach. Remaining items are documentation drift and minor resource-hygiene nits — none block merging.

<a href="https://opencode.ai/s/GQ81zjDJ">New%20session%20-%202026-08-18T03%3A13%3A19.281Z</a>
[opencode session](https://opencode.ai/s/GQ81zjDJ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32094618501)

kaishi00:
Noted — the budget was retimed 1080s→960s (16 min) during local review to keep the worst case (setup + killed attempt 1 + bootstatus-bounded reboots + full attempt 2) under the 45-min job cap; two comment references still said 1080/18 min. Fixed in 428ac6d (comment-only change).

kaishi00:
## Summary
Adds two wedge-bounding layers so a hung async test or wedged simulator can no longer stall CI: all `HermesClientTests` waits and task completions are deadline-bounded via a lock-protected `Gate` + detached timer, and a new `scripts/ci-run-tests.sh` gives each xcodebuild attempt a 16-min wall-clock budget enforced by a process-group kill, with the retry loop moved out of `ci.yml`.

## Findings

### [minor] Watchdog task outlives every successful `Gate.wait`
**File:** ConduitTests/HermesClientTests.swift:268-271
**Issue:** Each `wait` spawns a `.detached` task that sleeps the full `timeout` (10s) even on the happy path where `signal()` wins in milliseconds. The task is never retained or cancelled, so every gate that succeeds promptly still keeps a detached sleeper alive until it no-ops in `fireDeadline`. Harmless at this scale, but it grows linearly with helper calls.
**Suggestion:** Store the watchdog task in a property and `cancel()` it inside `signal()` (under the lock, before resuming) so no timer outlives its wait.

### [minor] On timeout, `awaitCompletion` leaves the awaited task running
**File:** ConduitTests/HermesClientTests.swift:206-212
**Issue:** When `finished.wait` times out, the deferred `watcher.cancel()` does not cancel the original `task` (e.g. a `connect()` still mid-flight), which can keep mutating client state while the test unwinds. The continued watcher also stays permanently suspended on `task.value` (documented as intentional). Correctness of the checks survives because continuations are cancellation-inert, but orphaned work persists.
**Suggestion:** Add `task.cancel()` alongside `watcher.cancel()` in the defer so a failed test at least attempts to stop the orphaned work. Low impact regardless.

### [minor] No validation of `ATTEMPT_TIMEOUT_SECS` / `ATTEMPTS`
**File:** scripts/ci-run-tests.sh:35-36, 108, 157
**Issue:** A non-numeric or zero value is silently evaluated as 0 in the deadline arithmetic at line 108 and passed to `seq` at line 157. Every attempt would then "time out" immediately (status 124) or `seq` would error, making a configuration mistake look like a hung test.
**Suggestion:** Validate both are positive integers after the defaults are applied, and emit a `::error::` with the offending value before the loop.

### [minor] No cleanup trap for workflow cancellation / job timeout
**File:** scripts/ci-run-tests.sh:90-104, 135-141
**Issue:** `set -m` places xcodebuild in its own process group, which is exactly right for the deadline kill — but it also means a signal sent to the script's own process group (workflow cancellation, the 45-min job cap) never reaches xcodebuild/xctest/simulator agents, which can be orphaned. Ephemeral GitHub-hosted runners limit the blast radius.
**Suggestion:** Install a trap on EXIT/INT/TERM that TERM→KILLs the recorded child process-group id, and clear it after normal `wait` completion.

### [minor] `ATTEMPTS > 2` reuses the second result bundle
**File:** scripts/ci-run-tests.sh:85, 157-162 and .github/workflows/ci.yml:88-89
**Issue:** Attempts 3+ delete and rewrite `TestResults-retry.xcresult`, and the upload step only knows the two fixed paths, so earlier attempt bundles are lost and later ones overwrite them.
**Suggestion:** Either clamp `ATTEMPTS` to a max of two or use attempt-unique bundle names and extend the upload `path` list.

### [nit] `wc -l`-based streaming drops unterminated trailing lines
**File:** scripts/ci-run-tests.sh:65-76, 148-149
**Issue:** `wc -l` counts only newline-terminated lines; when the deadline kills an unflushed partial line (potentially the line describing the wedge), it is omitted from both the live stream and the "last 200 log lines" diagnostic.
**Suggestion:** Append any trailing non-newline tail, or grep the live log for test-phase markers in the kill diagnostics. Cosmetic.

### [nit] Heartbeat cadence assumes a fixed 15s poll
**File:** scripts/ci-run-tests.sh:121-126
**Issue:** The `heartbeat % 4` cadence is only ~60s while the poll stays at 15s; once `remaining < 15` the interval shortens to a few seconds. Cosmetic, output cadence not correctness-critical.
**Suggestion:** Compute cadence from wall-clock elapsed time, or leave as-is.

## Verdict
APPROVE

The core logic is sound: the `Gate` signal/timeout hand-off is exactly-once under a single NSLock with resumption outside the lock; the detached deadline timers use `[weak self]` (no retain cycles); no force unwraps or implicitly unwrapped optionals were introduced; mock types stay MainActor-isolated while the timers run detached; and the bash retry/kill plumbing (process-group kill with TERM→grace→KILL, 124 flowing through the retry loop, dynamic-scoping counter fix) was verified against a harness, including child-PID group kills. The 960s/16-min budget is consistently documented across the script and `ci.yml` and the worst-case math against the 45-min cap is stated honestly. Remaining findings are resource-hygiene and diagnostics nits only, none blocking.

<a href="https://opencode.ai/s/MGkFXmlb">New%20session%20-%202026-08-18T03%3A32%3A33.320Z</a>
[opencode session](https://opencode.ai/s/MGkFXmlb)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32095768807)

PR #79, opened 2026-08-18
fix(chat): shared responsive table column widths and text-level alignment

kaishi00:
Continues the Markdown-table work from #73 (merged at `5845ddc`): the remaining QA layout bugs whose commits landed on the old branch after it was merged. Rebased cleanly onto current main (incl. #75 reference links); all markdown-suite tests pass on the combined state.

## User-visible symptoms (from QA)

1. **Column dividers shifted horizontally between rows** — each row sized its own columns, so vertical boundaries didn't line up down the table.
2. **`:---:` / `---:` alignment was parsed but rendered left-aligned everywhere.**
3. **Even trivial three-column tables forced horizontal scrolling on iPhone** — the 112pt per-column minimum (a measurement artifact, not a table semantic) made three columns need ~396pt with padding.

## Fixes

**Shared column widths:** `MarkdownTableLayout` derives one width per column for the entire table — the largest ideal (unwrapped) text width among that column's header and body cells — and every cell in the column renders at exactly that width (`.frame(width:)` + single-value self-sizing range), so dividers align. Wider content wraps within the shared column.

**Alignment reaches the text engine:** the parsed alignment flows `MarkdownTable → InlineMarkdown → SelectableTextView.textAlignment` into the paragraph style, so center/trailing hold for single-line *and* wrapped multiline content. `.natural` default leaves every other caller unchanged.

**Responsive width policy:** a zero-height probe reads the chat's proposed width; the pure `resolveColumnContentWidths(ideals:availableWidth:)` step then decides:
- capped columns (≤220pt) that fit the chat width → table fits, narrow columns stay genuinely narrow, wide columns keep proportionally more space;
- on overflow, columns shrink proportionally to their excess above a 64pt floor (already-narrow columns don't shrink) and wrap at the resolved width;
- still overflowing at the floor → horizontal scrolling, as before;
- cell padding, dividers, and border subtract from available width exactly once;
- cell heights measure at the final resolved shared column content width, preserving #73's deterministic first-pass/full-height guarantee.

**Reference-aware measurement:** table cells can contain message-level reference-style links (#75). The exact inline-attributed-string construction `InlineMarkdown` renders is extracted into `InlineMarkdownContent.attributed(source:references:)` and used for both rendering and column-width measurement, with `MarkdownTable` threading the injected `\.markdownReferences` environment into `MarkdownTableLayout` (cache keys on the definitions). A `[Handbook][1]` cell now measures as `Handbook`, so widths and rendered content cannot diverge.

## Scrolling policy

Fit when the content allows; scroll horizontally only when it doesn't. Cells expand vertically without nested scrolling; the outer chat scroll owns vertical scrolling.

## Regression coverage

New `MarkdownTableLayoutTests` (15): pure policy unit tests (fit keeps ideals, cap, proportional shrink, floor-and-scroll, unknown-width fallback, content-metric growth recompute, reference-link measured at its resolved label) plus rendered fixtures through the real `MarkdownText` chain — tiny three-column table fits without scrolling; narrow status + long description fits with proportional shared widths; wide five-column table still scrolls; identical column widths/divider edges across header/short/empty cells; long cell wraps fully at its shared width; per-line glyph-rect geometry proves leading/center/trailing for single-line and wrapped cells; selection observers stay attached; reference-link column sized identically to a literal-label table with the URL attribute intact. The existing 28 `ChatTextSelectionTests` and 32 `MarkdownSelectionCoordinatorTests` pass unchanged.

## Validation

75/75 across the three markdown suites on the rebased branch (includes #75's reference-link changes).

## Summary by CodeRabbit

* **New Features**
  * Improved Markdown table layout with content-aware column sizing and consistent alignment.
  * Wide tables now shrink proportionally when possible and retain horizontal scrolling when needed.
  * Selectable text supports consistent alignment across rendered content.

* **Bug Fixes**
  * Improved text wrapping, divider alignment, link rendering, and selection behavior in Markdown tables.
  * Improved table display across different screen sizes and Dynamic Type settings.

### DIFF
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -721,21 +721,18 @@ private extension EnvironmentValues {
     }
 }
 
-private struct InlineMarkdown: View {
-    let source: String
-    let foregroundStyle: Color
-    let usesAccentSurface: Bool
-    var font: UIFont = .preferredFont(forTextStyle: .body)
-    var lineSpacing: CGFloat = 0
-    var maximumNumberOfLines: Int = 0
-    var textAlignment: NSTextAlignment = .natural
-    var selfSizingWidthRange: ClosedRange<CGFloat>? = nil
-    var selectionCoordinator: MarkdownSelectionCoordinator?
-    var selectionSegment: MarkdownSelectionSegmentDescriptor?
-    var trailingCharacterOpacities: [Double] = []
-    @Environment(\.markdownReferences) private var references
-
-    private var attributed: AttributedString {
+/// The single inline-attributed-string construction shared by
+/// `InlineMarkdown` rendering and `MarkdownTableLayout` width measurement,
+/// so a table column is always measured at exactly the content its cells
+/// render — reference-style links resolve to their labels in both paths,
+/// and any future inline-syntax change updates measurement with it.
+enum InlineMarkdownContent {
+    static func attributed(
+        source: String,
+        references: MarkdownReferenceContext,
+        foregroundStyle: Color = .primary,
+        trailingCharacterOpacities: [Double] = []
+    ) -> AttributedString {
         // Re-append the message's definitions so Foundation resolves
         // reference-style links for this fragment too. The definition block
         // is invisible under `.full`; if parsing fails, fall back to the bare
@@ -753,6 +750,30 @@ private struct InlineMarkdown: View {
             endIndex = startIndex
         }
         return attributed
+    }
+}
+
+private struct InlineMarkdown: View {
+    let source: String
+    let foregroundStyle: Color
+    let usesAccentSurface: Bool
+    var font: UIFont = .preferredFont(forTextStyle: .body)
+    var lineSpacing: CGFloat = 0
+    var maximumNumberOfLines: Int = 0
+    var textAlignment: NSTextAlignment = .natural
+    var selfSizingWidthRange: ClosedRange<CGFloat>? = nil
+    var selectionCoordinator: MarkdownSelectionCoordinator?
+    var selectionSegment: MarkdownSelectionSegmentDescriptor?
+    var trailingCharacterOpacities: [Double] = []
+    @Environment(\.markdownReferences) private var references
+
+    private var attributed: AttributedString {
+        InlineMarkdownContent.attributed(
+            source: source,
+            references: references,
+            foregroundStyle: foregroundStyle,
+            trailingCharacterOpacities: trailingCharacterOpacities
+        )
     }
 
     var body: some View {
@@ -985,17 +1006,24 @@ enum MarkdownTableLayout {
     }()
 
     @MainActor
-    static func columnWidths(headers: [String], rows: [[String]], availableWidth: CGFloat) -> [CGFloat] {
+    static func columnWidths(
+        headers: [String],
+        rows: [[String]],
+        availableWidth: CGFloat,
+        references: MarkdownReferenceContext = .empty
+    ) -> [CGFloat] {
         let columnCount = max(headers.count, rows.map(\.count).max() ?? 0)
         guard columnCount > 0 else { return [] }
 
         // Fonts resolve against the current Dynamic Type size, so the widths
-        // key on the content category just like MarkdownRenderCache, and on
-        // the viewport width that feeds the fit-or-shrink decision.
+        // key on the content category just like MarkdownRenderCache, on the
+        // viewport width that feeds the fit-or-shrink decision, and on the
+        // reference definitions (they change how cell source measures).
         let key = (
             [
                 UIApplication.shared.preferredContentSizeCategory.rawValue,
-                String(format: "%.1f", availableWidth)
+                String(format: "%.1f", availableWidth),
+                references.definitionsMarkdown
             ]
                 + headers
                 + rows.flatMap { $0.isEmpty ? ["<empty-row>"] : $0 }
@@ -1008,11 +1036,11 @@ enum MarkdownTableLayout {
 
         var ideals = [CGFloat](repeating: 0, count: columnCount)
         for (index, header) in headers.enumerated() {
-            ideals[index] = max(ideals[index], idealWidth(of: header, font: headerFont))
+            ideals[index] = max(ideals[index], idealWidth(of: header, font: headerFont, references: references))
         }
         for row in rows {
             for (index, cell) in row.enumerated() where index < columnCount {
-                ideals[index] = max(ideals[index], idealWidth(of: cell, font: bodyFont))
+                ideals[index] = max(ideals[index], idealWidth(of: cell, font: bodyFont, references: references))
             }
         }
 
@@ -1057,18 +1085,18 @@ enum MarkdownTableLayout {
         widths.map { ($0 * 2).rounded() / 2 }
     }
 
-    /// Measures the ideal (single-fragment) width of a cell the same way the
-    /// cell's own text view measures itself — InlineMarkdown's attributed
-    /// string bridged with the cell font — so the shared column widths and
-    /// the rendered wrapping agree.
-    private static func idealWidth(of source: String, font: UIFont) -> CGFloat {
-        let attributed = (try? AttributedString(
-            markdown: source,
-            options: .init(interpretedSyntax: .full)
-        )) ?? AttributedString(source)
+    /// Measures the ideal (single-fragment) width of a cell from the exact
+    /// attributed content InlineMarkdown renders (`InlineMarkdownContent`
+    /// bridged with the cell font), so the shared column widths and the
+    /// rendered wrapping agree — including message-level reference links,
+    /// which measure as their resolved labels.
+    private static func idealWidth(of source: String, font: UIFont, references: MarkdownReferenceContext) -> CGFloat {
         let bridged = NSMutableAttributedString(
             attributedString: SelectableTextView.bridge(
-                attributed, defaultFont: font, defaultColor: .label, linkColor: .link
+                InlineMarkdownContent.attributed(source: source, references: references),
+                defaultFont: font,
+                defaultColor: .label,
+                linkColor: .link
             )
         )
         let fullRange = NSRange(location: 0, length: bridged.length)
@@ -1103,9 +1131,15 @@ private struct MarkdownTable: View {
     /// the layout then falls back to cap-and-scroll and re-resolves a frame
     /// later, once the width is known.
     @State private var availableWidth: CGFloat = 0
+    @Environment(\.markdownReferences) private var references
 
     private var columnWidths: [CGFloat] {
-        MarkdownTableLayout.columnWidths(headers: headers, rows: rows, availableWidth: availableWidth)
+        MarkdownTableLayout.columnWidths(
+            headers: headers,
+            rows: rows,
+            availableWidth: availableWidth,
+            references: references
+        )
     }
 
     var body: some View {
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -1151,12 +1151,13 @@ private struct MarkdownTable: View {
             }
             .frame(height: 0)
 
+            let widths = columnWidths
             ScrollView(.horizontal, showsIndicators: false) {
                 VStack(alignment: .leading, spacing: 0) {
-                    tableRow(headers, rowIndex: 0, isHeader: true)
+                    tableRow(headers, rowIndex: 0, isHeader: true, widths: widths)
                     ForEach(Array(rows.enumerated()), id: \.offset) { rowOffset, row in
                         Divider().overlay(usesAccentSurface ? Color.white.opacity(0.22) : Color.secondary.opacity(0.18))
-                        tableRow(row, rowIndex: rowOffset + 1, isHeader: false)
+                        tableRow(row, rowIndex: rowOffset + 1, isHeader: false, widths: widths)
                     }
                 }
                 .background(
@@ -1171,9 +1172,8 @@ private struct MarkdownTable: View {
         }
     }
 
-    private func tableRow(_ cells: [String], rowIndex: Int, isHeader: Bool) -> some View {
-        let widths = columnWidths
-        return HStack(spacing: 0) {
+    private func tableRow(_ cells: [String], rowIndex: Int, isHeader: Bool, widths: [CGFloat]) -> some View {
+        HStack(spacing: 0) {
             ForEach(Array(cells.enumerated()), id: \.offset) { index, cell in
                 let width = widths.indices.contains(index)
                     ? widths[index]
--- a/Conduit/Views/Components/MarkdownText.swift
+++ b/Conduit/Views/Components/MarkdownText.swift
@@ -960,12 +960,22 @@ private struct MarkdownColumns: View {
 }
 
 /// Table-wide column layout: one width per column shared by the header and
-/// every row, so vertical dividers align across rows. Each column width is
-/// the largest ideal (unwrapped) text width among its cells — header
-/// included — clamped to the table's 112–220pt column policy; wider content
-/// wraps within the shared width instead of widening one row's cell.
+/// every row, so vertical dividers align across rows. Each column's width
+/// is the largest ideal (unwrapped) text width among its cells — header
+/// included — capped at `maxColumnContentWidth`. When the capped columns
+/// fit the chat width the table fits too (narrow columns stay narrow, wide
+/// ones keep their earned space); when they overflow, columns above the
+/// shrink floor give up width proportionally to their excess and wrap, and
+/// a table that still overflows at the floor keeps horizontal scrolling.
 enum MarkdownTableLayout {
-    static let columnWidthRange: ClosedRange<CGFloat> = 112...220
+    /// Widest a column's text may be before it wraps.
+    static let maxColumnContentWidth: CGFloat = 220
+    /// Narrowest a column shrinks to when the table must compress; columns
+    /// whose ideal is already below this keep their ideal width.
+    static let shrinkFloorContentWidth: CGFloat = 64
+    static let cellHorizontalPadding: CGFloat = 10
+    static let dividerWidth: CGFloat = 1
+    static let borderAllowance: CGFloat = 2
 
     private static let cache: NSCache<NSString, NSArray> = {
         let cache = NSCache<NSString, NSArray>()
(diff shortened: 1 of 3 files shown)


kaishi00:
## Summary
Rebased table-layout QA fix: Markdown tables now derive one shared, content-informed width per column across all rows (aligned dividers), thread parsed `:---:`/`:---`/`---:` alignment into the text engine's paragraph style, and resolve a fit-or-shrink-or-scroll width policy against the chat's proposed viewport width instead of the old fixed 112pt floor. Includes 13 new unit/rendered-fixture tests. All three markdown suites reportedly pass.

## Findings

### [minor] First-pass table renders at capped widths, then re-lays out one frame later
**File:** Conduit/Views/Components/MarkdownText.swift:113 (probe), MarkdownTableLayout.swift:1031 (fallback)
**Issue:** `availableWidth` starts at 0, so the first layout pass renders every column at its uncapped/ideal width (`resolveColumnContentWidths` returns `halfPointRounded(capped)` for `availableWidth <= 0`). The zero-height GeometryReader then mutates `@State` from `.onAppear`, triggering a second layout pass that shrinks columns to the shared fit/shrink widths. Every newly-appeared table will visibly reflow/jump by a frame — noticeable in a streaming chat where tables appear mid-stream. Setting `@State` from `onAppear`/`onChange` inside a GeometryReader is also a known fragile pattern (it doesn't cover rotation-on-arrival edge cases the way continuous measurement would).
**Suggestion:** Prefer observing the proposal continuously (e.g. `onGeometryChange(for:of:action:)` on iOS 16+, or mirroring the width via a `GeometryReader`-driven preference) so the first pass already has the true width, or accept the cap-and-settle behavior and explicitly document the one-frame reflow.

### [minor] `trailing → .right` and `leading → .natural` are LTR-hardcoded
**File:** Conduit/Views/Components/MarkdownText.swift:540-544
**Issue:** `.trailing` maps to `NSTextAlignment.right` unconditionally. In a RTL locale, Markdown `---:` (right-aligned relative to the table) will render on the wrong side, and `.leading → .natural` will silently become right-aligned for RTL text. Markdown table alignment is usually relative to the table, not the user's base direction.
**Suggestion:** Map `.leading → .left` and `.trailing → .right` explicitly for LTR, or align relative to the view's effective layout direction (e.g. use `UIView.userInterfaceLayoutDirection` to pick `.left`/`.right`), keeping tests asserting the explicit value so this is a deliberate choice.

### [minor] Per-cell `boundingRect` measurement runs on the main thread on every cache miss
**File:** Conduit/Views/Components/MarkdownText.swift:1051-1054, 1064-1088
**Issue:** `MarkdownTableLayout.columnWidths` measures the ideal width of every header and body cell synchronously on the main thread inside the view body. For a large table (e.g. the wide five-column case) this is presumably cheap, but during streaming the cache key changes every frame, so all cells are re-measured per delta on the main actor. The existing `MarkdownRenderCache` deliberately skips caching while streaming to avoid dead entries — the table layout cache does not, so streaming wide tables pay N `boundingRect` calls per frame.
**Suggestion:** Consider caching the ideal widths (keyed by content + content-size category, without availableWidth) separately from the cheap pure distribution step, so streaming frames only pay the O(columns) distribution, not the per-cell font measurement.

### [nit] `halfPointRounded` can push the sum back above the available width
**File:** Conduit/Views/Components/MarkdownText.swift:1056-1058
**Issue:** After proportional shrink, rounding each column independently to half-points can make the total exceed `availableWidth`, resurrecting a small horizontal scroll in "fits" cases (the tests account for `+0.5`/`+1.5` slack). Also, a shrink result already at the floor can be rounded to `63.5`, dipping below `shrinkFloorContentWidth`.
**Suggestion:** Clamp shrunk columns to `>= shrinkFloorContentWidth` after rounding, and optionally scale the last column down to guarantee `total <= available` when the deficit fits.

### [nit] Cache key uses a single `\u{1F}` delimiter that cell content could theoretically contain
**File:** Conduit/Views/Components/MarkdownText.swift:1002
**Issue:** `joined(separator: "\u{1F}")` concatenates headers + rows + availableWidth into one key. Markdown cell content containing U+001F (unit separator) would collide elements of the key, producing a wrong cached width. Unlikely from agent output, but a cheap robustness fix exists.
**Suggestion:** Use a struct key or a longer, less collision-prone delimiter (e.g. an array of components hashed/joined with a rare multi-char separator).

### [nit] Dead fallback branch in `tableRow`
**File:** Conduit/Views/Components/MarkdownText.swift:1144-1146
**Issue:** `widths.indices.contains(index) ? widths[index] : shrinkFloorContentWidth` — `columnWidths` always returns `columnCount = max(headers.count, max row count)` entries, so `widths` covers every cell index by construction and the `else` branch is unreachable.
**Suggestion:** Drop the fallback and use `widths[index]` directly, or keep it and add a defensive comment explaining when it could trigger.

## Assessment notes
- No new force unwraps or implicitly unwrapped optionals; `SelectableTextView` closures use `[weak selectionCoordinator]` consistently.
- No new retain cycles: the GeometryReader closures capture only value-type views; NSCache entries are cheap and evictable; the stored `cache` on `MarkdownTableLayout` matches the pre-existing `MarkdownRenderCache` pattern.
- Concurrency is well-handled: `resolveColumnContentWidths` is a pure, nonisolated, unit-testable step; UIKit/`UIApplication` access is `@MainActor`; `onChange` covered by the iOS 17 deployment target.
- Tests are strong: the pure-policy/isolation split, empty-cell and shared-width geometry assertions, per-line glyph-rect alignment checks, and the settle+relayout pump in `renderedTableHost` are exactly the right coverage for these bugs.
- The alignment test asserting `rect.maxX == bounds.width` and the empty-cell band check are slightly coupled to internal layout details, but they verify the actual user-visible symptoms, which is good.

## Verdict
APPROVE

<a href="https://opencode.ai/s/lNEXlX6Z">New%20session%20-%202026-08-18T09%3A00%3A18.846Z</a>
[opencode session](https://opencode.ai/s/lNEXlX6Z)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32119196266)

kaishi00:
## Summary
PR #79 reworks Markdown table layout to derive one shared, content-aware width per column, thread `:---:`/`---:` alignment into the text engine's paragraph style, and resolve a fit-or-shrink-or-scroll width policy against the chat's proposed width instead of the old fixed floor. Includes 15 new unit/rendered-fixture tests.

## Findings

### [minor] First layout pass renders at capped widths, then reflows a frame later
**File:** Conduit/Views/Components/MarkdownText.swift:1133 (state), 1147-1152 (probe), 1059 (fallback)
**Issue:** `availableWidth` starts at 0, so pass one resolves every column to its uncapped/capped ideal (`resolveColumnContentWidths` returns `halfPointRounded(capped)` for `availableWidth <= 0`). The zero-height GeometryReader then mutates `@State` from `.onAppear`, forcing a second pass that shrinks columns to the fit/shrink widths. Every newly-appearing table visibly jumps wide→fit for a frame — including tables that stream into view, and the mock-probe is exactly the "settle and re-layout" dance in `renderedTableHost` (MarkdownTableLayoutTests.swift:452-457). Setting `@State` from `onAppear` inside a GeometryReader is also fragile for mid-layout width changes (rotation/compact split arriving during the first pass).
**Suggestion:** Feed the width via a continuous observation (`onGeometryChange(for:of:action:)` on iOS 16+, or deliver the proxy width through a `PreferenceKey`/`onChange` on an always-measured background) so the first pass already has the true viewport width, or accept and explicitly document the one-frame settle.

### [minor] Alignment mapping is LTR-hardcoded
**File:** Conduit/Views/Components/MarkdownText.swift:541-544
**Issue:** `.trailing` maps unconditionally to `.right` and `.leading` to `.natural`. Under an RTL base direction `.natural` resolves to right-alignment, so a `:---` (left) column renders on the wrong side and `---:` (right) also lands wrong. Markdown table alignment is conventionally relative to the table, not the user's base direction.
**Suggestion:** Either map `.leading → .left`, `.trailing → .right` explicitly for LTR (and derive from `UIView.userInterfaceLayoutDirection` if RTL support is intended), and keep the tests asserting the explicit value so the choice is deliberate.

### [minor] All cell `boundingRect` measurements run on the main thread on every cache miss, including streaming frames
**File:** Conduit/Views/Components/MarkdownText.swift:1037-1049, 1093-1116
**Issue:** `columnWidths` measures the ideal width of every header and body cell synchronously in `@MainActor` inside the view body. The cache key includes headers, rows, content-size category, reference definitions, and availableWidth, so while a reply streams the source grows each frame and the key changes — every cell is re-measured per delta. `MarkdownRenderCache` deliberately skips insertions while streaming (MarkdownText.swift:254) to avoid dead entries; the table cache does not, so streaming wide tables pay N `boundingRect` calls per frame on the main actor and churn the 64-entry cache, evicting settled tables' entries.
**Suggestion:** Cache the ideal widths (keyed on content + content-size category only, no availableWidth) separately from the cheap pure `resolveColumnContentWidths` distribution step, so streaming frames pay O(columns) distribution rather than per-cell font measurement, and skip insertion while `isStreaming` like the render cache does.

### [nit] `halfPointRounded` can push the total above the available width
**File:** Conduit/Views/Components/MarkdownText.swift:1084-1086, used at 1076/1081
**Issue:** Each column is rounded to the nearest half-point independently after proportional shrink; for large column counts the aggregate can exceed `available`, resurrecting a small horizontal scroll in "fits" cases (the tests budget `+0.5`/`+1.5` slack for this, e.g. MarkdownTableLayoutTests.swift:44). A shrink result in `[63.5, 63.75)` also rounds to 63.5, dipping below the documented `shrinkFloorContentWidth`.
**Suggestion:** Clamp shrunk columns back up to `shrinkFloorContentWidth` after rounding and, when the total exceeds `available`, scale the last column down enough to guarantee `total <= available`.

### [nit] Cache key uses a single U+001F delimiter cell content could contain
**File:** Conduit/Views/Components/MarkdownText.swift:1022-1030
**Issue:** Headers, rows, `availableWidth`, content-size category and reference definitions are joined with `"\u{1F}"` into one cache key. Cell content containing U+001F (unit separator) would merge key elements and produce a wrong cached width. The `<empty-row>` placeholder marker is likewise collidable with a literal cell string.
**Suggestion:** Use a struct key (or a multi-char separator that markdown cell content cannot plausibly produce), or hash the array of components.

### [nit] Dead fallback branch in `tableRow`
**File:** Conduit/Views/Components/MarkdownText.swift:1178-1180
**Issue:** `widths.indices.contains(index) ? widths[index] : shrinkFloorContentWidth` — `columnWidths` always returns `max(headers.count, max row count)` entries, and header/row cell counts never exceed that, so the `else` branch is unreachable by construction.
**Suggestion:** Use `widths[index]` directly, or keep it with a comment explaining the defensive intent.

## Assessment notes
- No new force unwraps/IOUs; closures capturing the coordinator use `[weak selectionCoordinator]` and `weak var selectionCoordinator` is used consistently.
- No new retain cycles: probe closures capture only value-type views; both NSCaches hold cheap, evictable values matching the existing `MarkdownRenderCache` pattern.
- Concurrency is well handled: `resolveColumnContentWidths` is pure/nonisolated and unit-tested; UIKit/`UIApplication` access is `@MainActor`; measurement and rendering share the exact `InlineMarkdownContent.attributed` + `bridge` construction so widths can't diverge from what's rendered (including reference links).
- Tests target the real user-visible symptoms (divider alignment, fit-no-scroll, proportional widths, per-line glyph geometry, empty cells) through the full `MarkdownText` chain; the pure-policy unit tests keep fallbacks deterministic.
- Accessibility holds up: cells are real `UITextView`s (VoiceOver-readable), Dynamic Type flows through `preferredFont` and is keyed in the caches.

## Verdict
APPROVE

<a href="https://opencode.ai/s/rYcFOEM8">New%20session%20-%202026-08-18T09%3A13%3A10.037Z</a>
[opencode session](https://opencode.ai/s/rYcFOEM8)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32120297874)

kaishi00:
Review triage (79c1575):

- **shortA marker bug — taken.** `cell(containing: "aaa")` is a substring match that also hits `aaaa-column-one-driver`, so the shared-width assertion could bind to the wrong cell. Now matched by exact string, like the `"b"` cell beside it.
- **Width resolution per row — taken.** `columnWidths` resolved inside `tableRow` (cache lookup per row); now hoisted to once per `MarkdownTable.body` evaluation and passed down.
- **TextKit-2 geometry comment — evaluated, kept as smoke check.** The `layoutManager` access does flip the test view into TextKit-1 compat mode, so the glyph-rect assertions measure compat layout rather than production TextKit-2. The paragraph-style alignment assertions above them are the production contract; the geometry stays as a zero-cost end-to-end sanity signal with a comment stating exactly that, and the block is explicitly not to grow.

Remaining nits left as-is per triage.

kaishi00:
## Summary
Rebases and continues #73's Markdown-table work: tables now derive one shared, content-aware width per column (aligned dividers), thread `:---:`/`:---`/`---:` alignment into the cell text's paragraph style, and resolve a fit-or-shrink-or-scroll width policy against the chat's proposed width instead of the old fixed 112pt floor — with shared reference-aware measurement between rendering and width computation, plus 15 new tests.

## Findings

### [minor] Every cache miss re-measures every cell on the main actor, once per streaming frame
**File:** Conduit/Views/Components/MarkdownText.swift:1008-1050, 1093-1116
**Issue:** `columnWidths`'s cache key embeds the full headers+rows text, `availableWidth`, and reference definitions. While a reply streams, `source` changes every frame, so the key changes every frame → all N header/body cells are re-measured via `boundingRect` inside the view body on the main actor, and the 64-entry NSCache is churned with dead entries that can evict settled tables' widths. `MarkdownRenderCache` deliberately skips insertion while `isStreaming` (line 254) for exactly this reason; the table cache does not.
**Suggestion:** Cache the content-only ideal widths (key on content + content-size category + references, no `availableWidth`) separately from the cheap pure `resolveColumnContentWidths` distribution, so streaming frames pay O(columns) distribution rather than per-cell font measurement; optionally skip insertion while streaming.

### [minor] First frame renders at the capped fallback, then reflows once the width probe lands
**File:** Conduit/Views/Components/MarkdownText.swift:1147-1152 (probe), 1059 (fallback)
**Issue:** `availableWidth` starts at 0, so the first pass resolves every column to its capped ideal (`resolveColumnContentWidths` returns `halfPointRounded(capped)` for `availableWidth <= 0` — and that width does not subtract the overhead, so the first frame can also flash a horizontal overflow). The zero-height GeometryReader then mutates `@State` from `.onAppear`, forcing a second pass at the fit/shrink widths. Tables that must shrink visibly jump wide→fit for one frame, including tables that appear mid-stream. The tests even codify this settle-and-relayout pump (MarkdownTableLayoutTests.swift:457-462).
**Suggestion:** Feed the width continuously (e.g. `onGeometryChange(for:of:action:)`, iOS 16+) or deliver it through a `PreferenceKey` on an always-present background view so the first pass already has the true width; otherwise document the one-frame settle as accepted behavior.

### [minor] Alignment mapping is LTR-hardcoded (`.leading → .natural`, `.trailing → .right`)
**File:** Conduit/Views/Components/MarkdownText.swift:539-545
**Issue:** Under an RTL base direction, `.natural` resolves to right-alignment, so a Markdown `:---` (leading) cell renders on the wrong side and `---:`/`:---:` behave inconsistently. Markdown table alignment is conventionally relative to the table, not the user's base direction. The test asserts `.natural` at MarkdownTableLayoutTests.swift:391, so the choice is currently deliberate.
**Suggestion:** Map `.leading → .left` / `.trailing → .right` explicitly (deriving from `UIView.userInterfaceLayoutDirection` if true RTL support is intended), and update the tests to assert the explicit value.

### [minor] Half-point rounding can exceed the available width and dip below the shrink floor
**File:** Conduit/Views/Components/MarkdownText.swift:1076, 1081, 1084-1086
**Issue:** Independent per-column rounding after proportional shrink can push the total up to ~0.5×N above `available`, resurrecting a small horizontal scroll in tables that were meant to fit (the tests budget `+1.5` slack for exactly this, e.g. line 44). A shrunken column landing in `[63.5, 64)` also rounds to 63.5, below the documented `shrinkFloorContentWidth` (64).
**Suggestion:** Clamp shrunk columns back up to `shrinkFloorContentWidth` after rounding, and scale the last column down when the rounded total exceeds `available` so a "fits" table never scrolls.

### [nit] Unconditional 220pt column cap wraps content even when the chat is far wider
**File:** Conduit/Views/Components/MarkdownText.swift:993, 1058, 1068-1069
**Issue:** Every column is clamped to `maxColumnContentWidth` (220) before the fit check, so on a wide iPad/landscape viewport a column whose ideal is, say, 300pt is forced to wrap at 220 even though the table has room to show it on one line.
**Suggestion:** Cap only when needed (e.g. cap after establishing that the uncapped total overflows), or document the cap as an intentional bound on table max width.

### [nit] Cache key uses a single U+001F separator and an `<empty-row>` marker
**File:** Conduit/Views/Components/MarkdownText.swift:1022-1030
**Issue:** Headers, rows, `availableWidth`, content-size category, and reference definitions are joined with `"\u{1F}"`; cell content containing U+001F (unit separator) merges key elements and could yield a wrong cached width, and a literal `<empty-row>` cell collides with the placeholder marker.
**Suggestion:** Use a struct cache key (or hash the component array).

### [nit] Dead fallback branch in `tableRow`
**File:** Conduit/Views/Components/MarkdownText.swift:1178-1180
**Issue:** `widths.indices.contains(index) ? widths[index] : shrinkFloorContentWidth` — `columnWidths` returns `max(headers.count, max row count)` entries, which covers every cell index a header or row can produce, so the `else` branch is unreachable.
**Suggestion:** Use `widths[index]` directly, or keep the fallback with a comment documenting the defensive intent.

### [nit] Duplicated and misnamed test helpers
**File:** ConduitTests/MarkdownTableLayoutTests.swift:290-295, 385-387, 469-471
**Issue:** `cell(containing:)` is defined twice (with different failure semantics — one `try XCTUnwrap(nil, ...)`, the other `XCTUnwrap(cells.first...)`), and `allTextViewsDeep` is poorly named since it returns every descendant `UIView` (it's used to find scroll views, not text views).
**Suggestion:** Collapse into one shared helper and rename to something like `allViewsDeep`.

## Assessment
- No new force unwraps or implicitly unwrapped optionals.
- No new retain cycles: the probe closures capture only value types; `NSCache` holds cheap evictable `[CGFloat]` values. `[weak selectionCoordinator]` usage is consistent.
- Concurrency is sound: `resolveColumnContentWidths` is pure/nonisolated and unit-tested; the `@MainActor` measurement work is confined to the main thread as it must be, and the two-parameter `onChange` matches the iOS 17 deployment target.
- Accessibility/dynamic type is preserved: cells are real `UITextView`s, font metrics key the caches, and wrapping (not clipping) absorbs size changes.
- Tests are strong — the pure-policy unit tests, empty-cell shared-width band checks, per-line glyph geometry, and reference-link width-vs-label assertions all target the actual user-visible symptoms.

## Verdict
APPROVE

<a href="https://opencode.ai/s/W9dtewyT">New%20session%20-%202026-08-18T09%3A25%3A34.903Z</a>
[opencode session](https://opencode.ai/s/W9dtewyT)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32121423871)

PR #82, opened 2026-08-19
Add preferred return surface: open to sessions

### LINKED ISSUE #52: App restores an older conversation after backgrounding (swipe home / app switch)
## Bug Description

After backgrounding the app (swipe home or switch to another app — NOT force-quitting), Conduit returns to a different, older conversation instead of the one that was open.

## Repro

1. Open an active conversation and exchange messages (agent turn can be running or completed).
2. Swipe home or switch to another app. Leave Conduit backgrounded for a short while (a few minutes is enough).
3. Return to Conduit.

## Expected

The app reopens exactly where it was: same conversation, same scroll position/state.

## Actual

The app lands in an earlier conversation from the session list instead of the active one. The conversation that was open is not lost (still in the list), but the app consistently restores a different, older thread. If you don't notice and type, the message goes into the wrong conversation.

## Environment

- Device: iPhone, App Store build (current version as of 2026-08-12)
- Hermes: self-hosted gateway v0.20.0 (2026.8.3) on Linux, dashboard WebSocket connection
- No profile switching involved; single profile

## Notes

- The app is backgrounded, not killed — iOS state restoration path, not cold launch.
- Reported while the Hermes core was ~98 commits behind origin/main (incl. session-continuity work). Will re-test after updating the server; will report back here whether the behavior changes, since the dashboard API may be a factor.

kaishi00:
Closes #52.

Adds the requested "open Conduit into the sessions drawer" option as a small, presentation-layer preference. This is **not** a third `ChatResumeBehavior` — the existing resume setting still answers *which conversation is active*; the new setting answers *which surface you see first*:

- `ChatReturnSurface` (`.conversation` / `.sessions`), default `.conversation`, persisted separately under `conduit.chatReturnSurface.v1` (resume store schema untouched, verified by test)
- Settings ▸ Chat: "When returning to Conduit" now has an "Open to [Conversation|Sessions]" picker above the existing behavior picker, with helper text when Sessions is selected. Changing the setting affects the next qualifying return only.

## Semantics

- **Sessions + Continue where I left off** → saved conversation and viewport restore normally underneath, then the drawer presents; dismissing it without choosing another conversation reveals the restored chat
- **Sessions + Jump to latest activity** → latest conversation resolves underneath, then the drawer presents
- No "no active session" state is introduced

## Triggers

- Cold launch into the authenticated app (once per process — disconnect → re-login cycles don't re-trigger)
- A real background → active transition, and only while authenticated (a background cycle on the login screen consumes the arming without issuing, so it can't leak into the next sign-in)
- Ordinary inactive → active (Control Center, call banner) never triggers

## Precedence

notification / deep-link > existing modal presentation > preferred return surface > normal chat surface. Suppression is enforced at issuance (`requestPreferredReturnSurface`) and re-checked at presentation time in MainView. Single scenePhase ownership stays in `RootView`; MainView consumes a one-shot `preferredReturnSurfaceRequest` token and presents the existing `showSidebar` sheet.

Untouched per spec: ChatResumeCoordinator, ChatResumeSessionResolver, ChatResumeStore schema, viewport restoration, notification routing, reconnect/recovery, connection state machine, streaming, markdown/selection/composer.

## Tests

16 new tests in `ChatReturnSurfaceTests` covering all ten spec scenarios plus persistence, once-per-process cold launch, and the signed-out-cycle leak guard. Full local run: 783 pass / 1 fail in `SelectionObserverUITests` selection UI test — known local-sim flake, green on CI for this exact base commit (37fd7c3); CI here is the authoritative check.

Review triage: local reviewer pass (code-reviewer, second-reviewer, third) produced 7 findings — 6 taken (re-login over-trigger, stale-request leak, presentation-time preference re-check, misplaced accessibility hint, two test-strength fixes, one rename), 1 declined with evidence (claimed second settings-presentation path doesn't exist — `settingsPresentation` has a single assignment site).

## Summary by CodeRabbit

* **New Features**
  * Added separate chat return preferences for opening the conversation or sessions view.
  * Added independent options for resuming the previous position or showing the latest activity.
  * Preferences are saved and restored across launches and available in chat settings.
  * Sessions view now opens automatically on qualifying returns when no navigation or modal takes precedence.
* **Bug Fixes**
  * Improved return behavior across cold launches, authentication changes, notifications, voice navigation, and restored sessions.
  * Prevented duplicate or premature sessions-view openings during pending navigation.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -1091,6 +1091,16 @@ final class AppState: ObservableObject {
         requestPreferredReturnSurface()
     }
 
+    /// Auth teardown (sign-out) is not a qualifying return. Retire any
+    /// outstanding preferred-return request at the boundary — claimed,
+    /// unclaimed, or deferred — so a MainView recreated on the next sign-in
+    /// can never present a stale request. The monotonic counter is left
+    /// untouched; only the consumed watermark advances.
+    private func retireOutstandingPreferredReturnSurfaceRequests() {
+        consumedReturnSurfaceRequest = max(consumedReturnSurfaceRequest, preferredReturnSurfaceRequest)
+        deferredReturnSurfaceRequest = nil
+    }
+
     /// Claims the pending preferred-return-surface request for presentation.
     /// Returns true exactly once per issued request — the watermark lives in
     /// AppState, so a MainView recreated after sign-out/sign-in can never
@@ -1937,6 +1947,7 @@ final class AppState: ObservableObject {
         isVoiceEnabled = false
         voiceTranscriptionMode = .hermes
         appleSpeechAvailability = AppleOnDeviceSpeechTranscriber.currentAvailability()
+        retireOutstandingPreferredReturnSurfaceRequests()
         showLogin = true
         sessions = []
         archivedSessions = []
@@ -2057,6 +2068,7 @@ final class AppState: ObservableObject {
         projectsLoading = false
         KeychainHelper.clearConnection()
         turnState = .idle
+        retireOutstandingPreferredReturnSurfaceRequests()
         showLogin = true
         errorMessage = message
     }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -372,6 +372,13 @@ final class AppState: ObservableObject {
     @Published private(set) var activeSessionTitle = "New conversation"
     @Published private(set) var isChatRefreshing = false
     @Published private(set) var chatResumeBehavior: ChatResumeBehavior = .continueWhereLeftOff
+    @Published private(set) var chatReturnSurface: ChatReturnSurface = .conversation
+    /// One-shot request for MainView to present the sessions drawer as the
+    /// preferred return surface. MainView consumes the latest value once per
+    /// increment; the request is only issued for qualifying returns (cold
+    /// launch or a real background → active transition) when the preference
+    /// is `.sessions` and no explicit navigation or modal owns the surface.
+    @Published private(set) var preferredReturnSurfaceRequest: UInt64 = 0
     @Published private(set) var chatResumeRestorationRequest: ChatResumeRestorationRequest?
     @Published private(set) var chatViewportTransitionGeneration: UInt64 = 0
     /// Keeps the current transcript visible while a notification destination is
@@ -474,6 +481,9 @@ final class AppState: ObservableObject {
     @Published var showGatewaySheet = false
     @Published var showAgentsSheet = false
     @Published var showVoiceSheet = false
+    /// Mirrors MainView's Settings sheet item so return-surface decisions can
+    /// tell whether Settings owns the surface across a background/foreground cycle.
+    @Published var isSettingsSheetPresented = false
     @Published var errorMessage: String?
     @Published var showLogin = true
     @Published private(set) var composerPrefillText = ""
@@ -581,6 +591,10 @@ final class AppState: ObservableObject {
     private var responseHaptics = ResponseHapticState()
     private var scenePhaseTask: Task<Void, Never>?
     private var scenePhaseAttemptID: UUID?
+    /// Armed only by a real .background phase. Ordinary inactive → active
+    /// transitions (Control Center, incoming-call banner) must not be treated
+    /// as reopening the app, so they never arm the preferred return surface.
+    private var hasEnteredBackgroundScenePhase = false
     private var explicitSessionOpenTask: Task<Bool, Never>?
     private var explicitSessionOpenRequestID: UUID?
     private var activeAutomaticChatResumeWork: ChatResumeAutomaticWorkToken?
@@ -784,6 +798,9 @@ final class AppState: ObservableObject {
     // MARK: - Persistence
 
     private let defaults: UserDefaults
+    /// Kept outside ChatResumeStore on purpose: the return surface is a
+    /// presentation preference, not part of the resume schema.
+    static let chatReturnSurfaceKey = "conduit.chatReturnSurface.v1"
     private let activeSessionTitlesByProfileKey = "conduit.activeSessionTitlesByProfile.v1"
     private let pinnedSessionIDsByProfileKey = "conduit.pinnedSessionIdsByProfile.v1"
     private let activeProfileKey = "conduit.activeProfile"
@@ -874,6 +891,8 @@ final class AppState: ObservableObject {
             ?? defaults.string(forKey: "conduit.dashboardURL")
                 .flatMap(Self.normalizedChatResumeServerIdentity)
         chatResumeBehavior = self.chatResumeCoordinator.behavior
+        chatReturnSurface = defaults.string(forKey: Self.chatReturnSurfaceKey)
+            .flatMap(ChatReturnSurface.init(rawValue:)) ?? .conversation
         defaultProfileName = ProfileAppearanceStore.loadDefaultName()
         profileAvatarURLs = ProfileAppearanceStore.loadAvatarURLs()
         appIconChoice = UIApplication.shared.alternateIconName == AppIconChoice.light.alternateIconName ? .light : .dark
@@ -1015,6 +1034,33 @@ final class AppState: ObservableObject {
         recoverySequence.preserveTransportAfterAutomaticIntentCancellation()
         chatResumeBehavior = chatResumeCoordinator.behavior
         chatResumeRestorationRequest = nil
+    }
+
+    /// True when any presented sheet owns the surface other than the sessions
+    /// drawer itself. Explicit navigation and existing modals take precedence
+    /// over the preferred return surface.
+    var isModalSheetPresented: Bool {
+        showModelPicker || showContextSheet || showWorkspaceSheet || showGatewaySheet
+            || showAgentsSheet || showVoiceSheet || isSettingsSheetPresented
+    }
+
+    /// Persisted separately from the resume store: this only changes which
+    /// surface is presented first on a qualifying return, and deliberately
+    /// issues no presentation request — the next qualifying return picks it up.
+    func setChatReturnSurface(_ surface: ChatReturnSurface) {
+        guard surface != chatReturnSurface else { return }
+        chatReturnSurface = surface
+        defaults.set(surface.rawValue, forKey: Self.chatReturnSurfaceKey)
+    }
+
+    /// Issues a one-shot preferred-return-surface request. Suppressed when a
+    /// notification destination is being opened or a modal sheet already owns
+    /// the surface; MainView re-checks at presentation time as a backstop.
+    func requestPreferredReturnSurface() {
+        guard chatReturnSurface == .sessions else { return }
+        guard !isOpeningNotificationSession else { return }
+        guard !isModalSheetPresented else { return }
+        preferredReturnSurfaceRequest &+= 1
     }
 
     func recordChatViewport(_ snapshot: ChatScrollSnapshot, for key: ChatScrollSessionKey) {
@@ -1523,6 +1569,7 @@ final class AppState: ObservableObject {
             theme: themePreference,
             busyInputMode: busyInputMode,
             chatResumeBehavior: chatResumeBehavior,
+            chatReturnSurface: chatReturnSurface,
             displayPreferences: displayPreferences,
             cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: connection?.baseUrl)
         )
@@ -3701,6 +3748,14 @@ final class AppState: ObservableObject {
         case .active:
             isSceneActive = true
             voiceConversationController.setForegroundActive(true)
+            // Issue before the transport work below: MainView presents the
+            // drawer while the automatic resume sync restores the chat
+            // underneath, and explicit navigation still wins via the
+            // suppression guards on both sides.
+            if hasEnteredBackgroundScenePhase {
+                hasEnteredBackgroundScenePhase = false
+                requestPreferredReturnSurface()
+            }
             guard connection != nil else { return nil }
             cancelScenePhaseAttempt()
             // Publish the foreground reconciliation boundary synchronously.
@@ -3759,6 +3814,7 @@ final class AppState: ObservableObject {
 
         case .background:
             isSceneActive = false
+            hasEnteredBackgroundScenePhase = true
             voiceConversationController.setForegroundActive(false)
             showVoiceSheet = false
             // Drop any armed reconnect timer as well: in-flight cycles abort
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -595,6 +595,7 @@ final class AppState: ObservableObject {
     /// transitions (Control Center, incoming-call banner) must not be treated
     /// as reopening the app, so they never arm the preferred return surface.
     private var hasEnteredBackgroundScenePhase = false
+    private var hasRequestedColdLaunchReturnSurface = false
     private var explicitSessionOpenTask: Task<Bool, Never>?
     private var explicitSessionOpenRequestID: UUID?
     private var activeAutomaticChatResumeWork: ChatResumeAutomaticWorkToken?
@@ -1061,6 +1062,15 @@ final class AppState: ObservableObject {
         guard !isOpeningNotificationSession else { return }
         guard !isModalSheetPresented else { return }
         preferredReturnSurfaceRequest &+= 1
+    }
+
+    /// MainView's first authenticated appearance in this process is the
+    /// cold-launch return. Mid-process re-entries (disconnect → sign back
+    /// in) are not cold launches; they rely on the scene-phase path.
+    func requestPreferredReturnSurfaceForColdLaunch() {
+        guard !hasRequestedColdLaunchReturnSurface else { return }
+        hasRequestedColdLaunchReturnSurface = true
+        requestPreferredReturnSurface()
     }
 
     func recordChatViewport(_ snapshot: ChatScrollSnapshot, for key: ChatScrollSessionKey) {
@@ -3748,15 +3758,19 @@ final class AppState: ObservableObject {
         case .active:
             isSceneActive = true
             voiceConversationController.setForegroundActive(true)
-            // Issue before the transport work below: MainView presents the
-            // drawer while the automatic resume sync restores the chat
-            // underneath, and explicit navigation still wins via the
-            // suppression guards on both sides.
-            if hasEnteredBackgroundScenePhase {
-                hasEnteredBackgroundScenePhase = false
+            // Consume the background arming even while signed out, so a
+            // background → active cycle on the login screen doesn't surface
+            // a stale request after the user signs back in.
+            let didReturnFromBackground = hasEnteredBackgroundScenePhase
+            hasEnteredBackgroundScenePhase = false
+            guard connection != nil else { return nil }
+            // The preferred return surface belongs to the authenticated
+            // app: MainView presents the drawer while the automatic resume
+            // sync restores the chat underneath, and explicit navigation
+            // still wins via the suppression guards on both sides.
+            if didReturnFromBackground {
                 requestPreferredReturnSurface()
             }
-            guard connection != nil else { return nil }
             cancelScenePhaseAttempt()
             // Publish the foreground reconciliation boundary synchronously.
             // ChatView may receive the same scene transition before the health
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -596,6 +596,11 @@ final class AppState: ObservableObject {
     /// as reopening the app, so they never arm the preferred return surface.
     private var hasEnteredBackgroundScenePhase = false
     private var hasRequestedColdLaunchReturnSurface = false
+    /// Presentation watermark for the preferred return surface: how far
+    /// through `preferredReturnSurfaceRequest` MainView has consumed, kept
+    /// here (not in view state) so it survives MainView teardown on sign-out.
+    private var consumedReturnSurfaceRequest: UInt64 = 0
+    private var deferredReturnSurfaceRequest: UInt64?
     private var explicitSessionOpenTask: Task<Bool, Never>?
     private var explicitSessionOpenRequestID: UUID?
     private var activeAutomaticChatResumeWork: ChatResumeAutomaticWorkToken?
@@ -1084,6 +1089,35 @@ final class AppState: ObservableObject {
         guard !hasRequestedColdLaunchReturnSurface else { return }
         hasRequestedColdLaunchReturnSurface = true
         requestPreferredReturnSurface()
+    }
+
+    /// Claims the pending preferred-return-surface request for presentation.
+    /// Returns true exactly once per issued request — the watermark lives in
+    /// AppState, so a MainView recreated after sign-out/sign-in can never
+    /// re-present an already-claimed request.
+    ///
+    /// Consumption semantics: while explicit navigation is pending the claim
+    /// defers without consuming; the next unblocked claim of the same request
+    /// then drops it, because the explicit route fully won that qualifying
+    /// return. In-flight notification opens, modals, and preference changes
+    /// consume-and-drop immediately (established precedence losers).
+    func claimPreferredReturnSurfacePresentation() -> Bool {
+        let current = preferredReturnSurfaceRequest
+        guard current > consumedReturnSurfaceRequest else { return false }
+        if hasPendingExplicitNavigation {
+            deferredReturnSurfaceRequest = current
+            return false
+        }
+        defer { deferredReturnSurfaceRequest = nil }
+        if deferredReturnSurfaceRequest == current {
+            consumedReturnSurfaceRequest = current
+            return false
+        }
+        consumedReturnSurfaceRequest = current
+        guard chatReturnSurface == .sessions,
+              !isOpeningNotificationSession,
+              !isModalSheetPresented else { return false }
+        return true
     }
 
     func recordChatViewport(_ snapshot: ChatScrollSnapshot, for key: ChatScrollSessionKey) {
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -1045,6 +1045,17 @@ final class AppState: ObservableObject {
             || showAgentsSheet || showVoiceSheet || isSettingsSheetPresented
     }
 
+    /// True when an explicit destination exists but has not been routed yet
+    /// (a notification tap or voice intent recorded before the app was
+    /// connected). Routing services own this fact; only the read lives here,
+    /// mirroring how syncSession already defers to a pending notification
+    /// target. Explicit navigation outranks the preferred return surface from
+    /// the moment the destination exists, not just once routing starts.
+    var hasPendingExplicitNavigation: Bool {
+        PushNotificationService.shared.pendingTarget != nil
+            || PendingVoiceIntentStore.shared.hasPendingIntent
+    }
+
     /// Persisted separately from the resume store: this only changes which
     /// surface is presented first on a qualifying return, and deliberately
     /// issues no presentation request — the next qualifying return picks it up.
@@ -1054,11 +1065,13 @@ final class AppState: ObservableObject {
         defaults.set(surface.rawValue, forKey: Self.chatReturnSurfaceKey)
     }
 
-    /// Issues a one-shot preferred-return-surface request. Suppressed when a
-    /// notification destination is being opened or a modal sheet already owns
-    /// the surface; MainView re-checks at presentation time as a backstop.
+    /// Issues a one-shot preferred-return-surface request. Suppressed while
+    /// an explicit destination is pending or being opened, or when a modal
+    /// sheet already owns the surface; MainView re-checks at presentation
+    /// time as a backstop.
     func requestPreferredReturnSurface() {
         guard chatReturnSurface == .sessions else { return }
+        guard !hasPendingExplicitNavigation else { return }
         guard !isOpeningNotificationSession else { return }
         guard !isModalSheetPresented else { return }
         preferredReturnSurfaceRequest &+= 1
--- a/Conduit/Services/ChatReturnSurface.swift
+++ b/Conduit/Services/ChatReturnSurface.swift
@@ -7,7 +7,7 @@ import Foundation
 /// conversation is active underneath. Choosing `.sessions` still performs
 /// the normal chat resume behind the drawer; dismissing the drawer without
 /// picking another conversation reveals the restored chat.
-enum ChatReturnSurface: String, CaseIterable {
+enum ChatReturnSurface: String, CaseIterable, Hashable {
     case conversation
     case sessions
 
--- a/Conduit/Services/ChatReturnSurface.swift
+++ b/Conduit/Services/ChatReturnSurface.swift
@@ -1,0 +1,20 @@
+import Foundation
+
+/// The surface Conduit presents first when returning to the app.
+///
+/// This is a presentation-layer preference only: it decides what the user
+/// sees on top, while `ChatResumeBehavior` independently decides which
+/// conversation is active underneath. Choosing `.sessions` still performs
+/// the normal chat resume behind the drawer; dismissing the drawer without
+/// picking another conversation reveals the restored chat.
+enum ChatReturnSurface: String, CaseIterable {
+    case conversation
+    case sessions
+
+    var title: String {
+        switch self {
+        case .conversation: "Conversation"
+        case .sessions: "Sessions"
+        }
+    }
+}
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -809,7 +809,11 @@ private struct ChatReturnBehaviorSettings: View {
                     }
                 }
                 .pickerStyle(.segmented)
-                .accessibilityHint("Used if you dismiss the session list without choosing another conversation.")
+                .accessibilityHint(
+                    surface == .sessions
+                        ? "Conduit opens to the session list. Used if you dismiss it without choosing another conversation."
+                        : "Conduit opens to your conversation."
+                )
                 if surface == .sessions {
                     Text("Used if you dismiss the session list without choosing another conversation.")
                         .font(.footnote)
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -111,6 +111,7 @@ struct SettingsSnapshot: Identifiable {
     let theme: ThemePreference
     let busyInputMode: BusyInputMode
     let chatResumeBehavior: ChatResumeBehavior
+    let chatReturnSurface: ChatReturnSurface
     let displayPreferences: ProfileDisplayPreferences
     let cloudflareAccess: CloudflareAccessCredentials?
 }
@@ -456,6 +457,7 @@ struct SettingsView: View {
     let saveTheme: (ThemePreference) -> Void
     let persistBusyInputMode: (BusyInputMode) async -> Bool
     let persistChatResumeBehavior: (ChatResumeBehavior) -> Void
+    let persistChatReturnSurface: (ChatReturnSurface) -> Void
     let loadProfileSettings: ([String]) async -> [String: ProfileSettingValue]
     let persistProfileSetting: (String, ProfileSettingValue) async -> Bool
     let loadProfileConfigOptions: () async -> ProfileConfigOptions
@@ -499,6 +501,8 @@ struct SettingsView: View {
                 persistBusyInputMode: persistBusyInputMode,
                 chatResumeBehavior: snapshot.chatResumeBehavior,
                 persistChatResumeBehavior: persistChatResumeBehavior,
+                chatReturnSurface: snapshot.chatReturnSurface,
+                persistChatReturnSurface: persistChatReturnSurface,
                 load: loadProfileSettings,
                 save: persistProfileSetting,
                 loadOptions: loadProfileConfigOptions
@@ -703,6 +707,8 @@ private struct ChatSettingsDetail: View {
     let persistBusyInputMode: (BusyInputMode) async -> Bool
     let chatResumeBehavior: ChatResumeBehavior
     let persistChatResumeBehavior: (ChatResumeBehavior) -> Void
+    let chatReturnSurface: ChatReturnSurface
+    let persistChatReturnSurface: (ChatReturnSurface) -> Void
     let load: ([String]) async -> [String: ProfileSettingValue]
     let save: (String, ProfileSettingValue) async -> Bool
     let loadOptions: () async -> ProfileConfigOptions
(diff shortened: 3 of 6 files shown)


kaishi00:
Now I have a thorough understanding. Let me also check `ChatReturnSurface: String, CaseIterable` — `ForEach(ChatReturnSurface.allCases, id: \.self)` requires `Hashable`. `String`-backed enum conforms to Hashable automatically, and `CaseIterable` + Hashable works. `id: \.self` requires Hashable — enum with String raw value is implicitly Hashable. Fine. But `ChatReturnSurface` isn't declared `Hashable` explicitly — however enums with raw-value conform implicitly? No. Actually, enums get `Hashable` synthesized only if they declare conformance. `id: \.self` in `ForEach` requires Hashable. Let me check — does `id: \.self` compile without explicit Hashable conformance? `ForEach(_, id: \.self)` requires the Element to be Hashable. `ChatReturnSurface` does NOT declare `Hashable`. This would be a compile error... unless the codebase compiles. Hmm, enums without associated values DO get automatic Hashable/Equatable synthesis only when you declare conformance. Without declaring `Hashable`, `\.self` as id would fail to compile.

Wait, actually — `Swift` provides implicit `Equatable`/`Hashable`/`Comparable` etc. only via declaration. But there's a subtlety: enums with no associated values automatically conform to `Equatable` and `Hashable` WITHOUT explicit declaration? No — that's not right either for the general case. Actually, yes: enums with no associated values (raw-value or simple enums) get `Equatable`, `Hashable`, and `Comparable` automatically via compiler synthesis even without explicit conformance? Let me recall. In Swift, enums with no associated values automatically conform to `Equatable` and `Hashable` without explicit declaration. Yes — simple enums (no payload) get automatic synthesis of `Equatable`, `Hashable`, and `Comparable` (for `Comparable` only raw-value with Comparable raw type in newer Swift). For Hashable/Equatable, no-raw-value and raw-value enums without associated values do get automatic synthesis. Since this is a shipping codebase with passing CI, I'll assume it compiles. This is a non-issue.

Let me now finalize findings. Let me re-examine the key state issue and check `handleScenePhase(.active)` once more and whether `requestPreferredReturnSurface` is called on .active after the connection guard reorders. The guard order:

```swift
case .active:
    isSceneActive = true
    voiceConversationController.setForegroundActive(true)
    let didReturnFromBackground = hasEnteredBackgroundScenePhase
    hasEnteredBackgroundScenePhase = false
    guard connection != nil else { return nil }
    if didReturnFromBackground {
        requestPreferredReturnSurface()
    }
    ...
```

Good — arming consumed before the guard. And request only when authenticated. This handles the leak.

Now the main issue remains the consume-in-`presentPreferredReturnSurfaceIfNeeded` before guards. Let me finalize.

Let me also double-check there's no retain cycle: `.task {}` and `.onChange` are view modifiers; the closures reference `appState` (EnvironmentObject property wrapper) — these are captured briefly for view lifetime, standard SwiftUI, no cycle. The `persistChatReturnSurface: { appState.setChatReturnSurface($0) }` in the sheet captures appState — fine.

Let me consider whether `showSidebar`'s `didSet` streaming side effects could interact. When presenting the drawer on return, `showSidebar.didSet` cancels streaming publish. That's expected behavior.

Let me finalize the review with findings. The strongest:

1. Consume-before-guards in `presentPreferredReturnSurfaceIfNeeded` — request lost if suppressed at presentation time; one-shot request is discarded, drawer never appears even though the user selected `.sessions`. (major/minor)

2. `@State` in `ChatReturnBehaviorSettings` seeded from a snapshot; if settings sheet stays open and app returns, the `surface` state is stale relative to `chatReturnSurface`. Minor — actually the same pattern existed for behavior already. Non-new.

Let me reflect on severity. The consume-before-guards: In the scene-phase path, when the `.active` sets request and `.onChange` fires — at that point guards likely pass in the common case. The suppression scenarios (notification open, modal present, sidebar already open) are precisely the "precedence" cases where we WANT the drawer suppressed in favor of the higher-priority surface. In those cases discarding the request is arguably intended (notification reset). But consider a subtle one: `showSidebar` already true at return — then the drawer is ALREADY showing, so no need to re-present. And `isModalSheetPresented` — if another modal is up, suppressed. If the user closes that modal, the return drawer won't come. Is that intended? The PR says "Changing the setting affects the next qualifying return only" — implying one-shot per qualifying return. If suppressed at presentation, arguably it should NOT be consumed so it could still fire... but then it'd fire on next `.onChange`, which won't happen since request won't increment. So the consumed request is just dropped. This is a behavioral edge worth flagging as minor.

Actually, I realize the more clearly problematic aspect: the request is consumed *before* the guards, so if the surface preference changed away from `.sessions` after issuance (user changed it in Settings between issue and presentation — quite plausible if the user changes it right before a return), it's consumed. But that's intended ("affects the next qualifying return").

I'll present it as a minor-to-major design concern with concrete fix. Given the precedence semantics explicitly state "re-checked at presentation time in MainView ... as a backstop", the fact that the backstop *discards* rather than defers is worth a minor finding.

Let me also note: `background→active` with `.sessions` while a modal (`isModalSheetPresented`) — can a modal be presented while app is backgrounded? Sheets get dismissed on background typically (SwiftUI auto-dismisses). So `isModalSheetPresented` would be false. But `isSettingsSheetPresented` might remain true if the settings sheet doesn't get auto-dismissed... Actually SwiftUI does dismiss sheets on background by default? No, it doesn't necessarily. The onDismiss sets it false. Hmm, but if the sheet is still up, the `.active` still might run... 

I'll keep the review focused. Let me write it.

Let me reconsider severity for the consume-before-guard. There's a genuinely concrete failure: cold launch `.task` runs `requestPreferredReturnSurfaceForColdLaunch()` then immediately `presentPreferredReturnSurfaceIfNeeded()`. At first MainView appearance, is it possible `isOpeningNotificationSession` is true? Yes — if the app was launched via a notification/deep link, `isOpeningNotificationSession` could be true during the initial session-open flight, which consumes the cold-launch request and prevents the drawer. That's a realistic scenario (launch from notification while `.sessions`). However, notification precedence says notification should win anyway, so dropping the drawer is correct there.

Overall the consume-before-guard is a defensible-but-smelly design. I'll rate it minor and clearly describe the trade-off, and recommend moving consume after successful presentation.

Let me write the review.## Summary
This PR adds a presentation-layer preference ("Open to Conversation/Sessions") controlled by a persisted `ChatReturnSurface` enum, arming a one-shot `preferredReturnSurfaceRequest` that MainView consumes to present the sessions drawer on cold launch or a real background→active return. It keeps the ChatResumeBehavior/schema untouched and adds 16 tests.

## Findings

### [minor] Request is consumed even when presentation-time suppression drops it
**File:** Conduit/Views/RootView.swift:205
**Issue:** `presentPreferredReturnSurfaceIfNeeded()` advances `consumedReturnSurfaceRequest` (line 206) *before* the precedence guards on lines 207-210. Because the request is a one-shot counter that will never increment again this cycle, any suppression at presentation time (notification open in flight, another modal owning the surface, or `showSidebar` already true) permanently discards the pending request — the drawer never appears, and the user's `.sessions` choice is silently ignored for that return. The PR descends this as a "backstop" re-check, but the backstop currently defeats itself by burning the token before it validates.
**Suggestion:** Move `consumedReturnSurfaceRequest = appState.preferredReturnSurfaceRequest` to *after* the guards (only advance once the drawer has actually been requested to present), so a momentarily-occupied surface defers rather than drops the request. If dropping is genuinely intended for some cases (e.g. notification precedence), document that explicitly and consume only those.

### [minor] `ChatReturnSurface` used with `id: \.self` relies on implicit Hashable
**File:** Conduit/Views/AuxiliaryViews.swift:806
**Issue:** `ForEach(ChatReturnSurface.allCases, id: \.self)` requires `Hashable`. The enum in ChatReturnSurface.swift:10 declares only `String, CaseIterable`, no explicit `Hashable`. This compiles only because simple raw-value enums get synthesized conformance — which is fragile and unstated. It's a nit, but declaring the conformance makes the contract explicit and robust against a future associated-value addition.
**Suggestion:** Add `Hashable` (or `Identifiable` using the raw value) to `ChatReturnSurface` explicitly.

### [nit] Snapshot-seeded `@State` can go stale if the settings sheet outlives a surface change
**File:** Conduit/Views/AuxiliaryViews.swift:787
**Issue:** `ChatReturnBehaviorSettings` seeds `@State surface` from the `SettingsSnapshot` captured at presentation. `chatReturnSurface` is also read live at presentation time in MainView (RootView.swift:207). If the settings sheet stays presented across a return and the user changes the surface elsewhere, the picker and the effective decision can disagree. This mirrors the pre-existing `behavior` pattern, so it's not a regression, but it's worth noting.
**Suggestion:** Acceptable as-is given precedent; if desired, drive the picker selection from `@State` but reconcile on the snapshot's `appear`/`onChange`.

## Verdict
COMMENT

The core logic (arming only on real `.background`, consuming arming before the auth guard to prevent sign-in leak, persistence kept out of the resume schema) is sound and well-tested. The one substantive concern is the consume-before-guard ordering in `presentPreferredReturnSurfaceIfNeeded`, which can drop a legitimate request; that's the item I'd want resolved or explicitly justified before this is fully finished.

<a href="https://opencode.ai/s/X5ciEmwB">New%20session%20-%202026-08-19T00%3A33%3A33.849Z</a>
[opencode session](https://opencode.ai/s/X5ciEmwB)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32201628013)

kaishi00:
Triage of the opencode review — thanks, two of three declined with rationale:

**1. Consume-before-guards drops a suppressed request — declined (intended).** The drop is the specified semantics for precedence losers. Issue #52 requires: *"if Settings, voice, model picker, or another existing sheet is already active across background/foreground, do not bulldoze it with the sessions drawer"* and *"The new preference is only for the normal app-return surface."* Deferring the token until the surface frees would pop the drawer later, after the return moment has passed — exactly the surprise the spec prohibits. Each qualifying return gets exactly one decision, made at return time. The notification case is likewise correct precedence (and `openNotificationTarget` closes the drawer anyway).

**2. Implicit `Hashable` under `id: \.self` — taken** in ae9af70: explicit `Hashable` declaration on `ChatReturnSurface`.

**3. Snapshot-seeded `@State` staleness — declined.** Mirrors the pre-existing `behavior` picker pattern in the same view; the review itself notes it's not a regression and is acceptable as-is.

kaishi00:
Re: CodeRabbit pre-merge warnings — both stem from reading #52 as a conversation-restoration bug. #52 is the feature request for this preference: it asks for an "open Conduit into the sessions drawer" option implemented as a presentation-layer setting. The restoration-underneath behavior the warnings ask about is the existing `ChatResumeBehavior` machinery, which this PR deliberately leaves untouched (per the issue's own do-not-modify list) and covers with tests (`testSessionsWithContinueBehaviorStillRestoresSavedSessionUnderneath`, `testSessionsWithLatestActivityStillResolvesLatestChatUnderneath`, plus the untouched-by-test `testReturnSurfacePreferenceDoesNotTouchResumeStoreSchema` payload check). No scope mismatch.

kaishi00:
Confirmed. The re-login stale-request finding is valid. Let me also verify the re-login teardown point with how the `if appState.showLogin || appState.connection == nil` branch switches — MainView is destroyed when login is shown, and re-created on re-login, resetting its `@State`.

Let me check whether there's any additional strong finding about the `.onChange` on `preferredReturnSurfaceRequest` being called during the `.task` cold-launch presentation (double call). Both call `presentPreferredReturnSurfaceIfNeeded` but the second (onChange) sees `consumedReturnSurfaceRequest` already updated, so no double. Fine.

Let me finalize my review. I'll write up the two findings plus a couple of minor ones.

Let me reconsider severity of finding 1. When exactly does re-login happen with a stale count? 
- App opened, `.sessions`, cold launch → request=1 → drawer presented, consumedReturnSurfaceRequest=1 (MainView alive).
- User navigates to something, logs out → showLogin=true → MainView destroyed. `@State consumedReturnSurfaceRequest` discarded.
- User logs back in → MainView recreated, consumedReturnSurfaceRequest=0. `.task` → cold-launch guarded (no increment, request stays 1) → `presentPreferredReturnSurfaceIfNeeded()`: 1 > 0 → presents drawer.

Yes, this presents the drawer on re-login with no qualifying background return. Major.

Let me also double check whether, at the moment `presentPreferredReturnSurfaceIfNeeded` runs on re-login, `isOpeningNotificationSession` or `isModalSheetPresented` could be false (returns true) allowing presentation. After sign-in, likely no modal and not opening notification → presents. Confirmed.

I'm confident. Let me write the review.

Also one more check: is there a scenario where the drawer would re-present on re-login even without `.sessions`? No — needs `.sessions`. The finding is specific to `.sessions` users.

Let me write it up.## Summary
This PR adds a persisted `ChatReturnSurface` preference (`.conversation`/`.sessions`) and a one-shot `preferredReturnSurfaceRequest` token that MainView consumes to present the sessions drawer on cold launch or a genuine background→active return, while leaving the ChatResumeBehavior/store schema untouched. It includes 16 tests.

## Findings

### [major] Stale `preferredReturnSurfaceRequest` re-presents the drawer on re-login due to `@State` reset vs process-global counter
**File:** Conduit/Views/RootView.swift:204 and Conduit/Views/RootView.swift:184
**Issue:** `consumedReturnSurfaceRequest` is a `@State` on `MainView`, which is torn down whenever the login screen shows (`RootView.swift:16`) and recreated on re-login — resetting it to `0`. Meanwhile `appState.preferredReturnSurfaceRequest` is a process-global counter that is never reset. On re-login, MainView's `.task` calls `presentPreferredReturnSurfaceIfNeeded()` (line 186), which sees the stale `1 > 0` from a *previous* process, consumes it, and presents the drawer immediately after sign-in — with no qualifying background→active return. The `hasRequestedColdLaunchReturnSurface` guard prevents *re-issuing* but does not guard the *presentation* re-consuming the stale nonzero counter. The PR's "disconnect → re-login cycles don't re-trigger" claim covers issuance (`testBackgroundToActiveWhileSignedOutDoesNotLeakIntoNextSignIn`) but not this presentation-side leak.
**Suggestion:** Track consumption in AppState (e.g. a persisted/last-seen request value alongside `preferredReturnSurfaceRequest`), or have the cold-launch `.task` only present the drawer when it corresponds to the value it just armed (capture the pre-call counter), rather than presenting any leftover nonzero counter on a fresh MainView. Add a test covering "arm once, log out, log back in → drawer must not present."

### [minor] Cold-launch request is swallowed when suppressed, with no retry this process
**File:** Conduit/Services/AppState.swift:1053
**Issue:** `requestPreferredReturnSurfaceForColdLaunch()` sets `hasRequestedColdLaunchReturnSurface = true` *before* calling `requestPreferredReturnSurface()`. If the cold-launch issuance is suppressed (e.g. a notification open or a modal owns the surface at first appearance), the once-per-process budget is consumed without ever issuing, so the cold-launch `.sessions` drawer never appears for the rest of the process.
**Suggestion:** Only mark `hasRequestedColdLaunchReturnSurface` when the request is actually issued, or re-arm on a later scene activation.

### [nit] Two independent lifecycle flags with overlapping concerns add non-obvious coupling
**File:** Conduit/Views/RootView.swift:184, Conduit/Services/AppState.swift:1053
**Issue:** The once-per-process cold-launch flag lives in `AppState`, while the consume-tracking lives in `MainView` `@State`, and the scene-phase trigger uses a separate `hasEnteredBackgroundScenePhase`. The non-obvious interaction across these three locations (findings above) is easy to regress. Not a bug by itself, but worth consolidating the request/consume state onto a single owner.
**Suggestion:** Consider moving both the counter and its "consumed up to" watermark into AppState so consumption survives view teardown.

## Verdict
REQUEST_CHANGES

The design (real-`.background` arming, consume-before-auth-guard to prevent sign-in leak, schema isolation, notification/modal precedence) is sound and well-tested. However, the `@State`-vs-process-global counter reset in `presentPreferredReturnSurfaceIfNeeded` can present the sessions drawer on re-login without a qualifying return, contradicting the stated "affects the next qualifying return only" semantics; this should be fixed (with a test) before merge.

<a href="https://opencode.ai/s/onTGnPAs">New%20session%20-%202026-08-19T00%3A47%3A26.830Z</a>
[opencode session](https://opencode.ai/s/onTGnPAs)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32202490342)

kaishi00:
I have a complete picture. Let me write the review.

## Summary
This PR adds a presentation-layer `ChatReturnSurface` preference (`.conversation`/`.sessions`), persisted separately from the resume store, that arms a one-shot `preferredReturnSurfaceRequest` token on cold launch or a genuine background→active return. MainView consumes the token to present the existing `showSidebar` drawer, with notification/deep-link and modal precedence, plus 16 tests. It deliberately leaves `ChatResumeBehavior`/store schema untouched.

## Findings

### [major] Stale return-surface request re-presents the drawer on re-login with no qualifying return
**File:** Conduit/Views/RootView.swift:43,207,209 and Conduit/Views/RootView.swift:184
**Issue:** `preferredReturnSurfaceRequest` is an `AppState` counter that is only ever incremented (`AppState.swift:1077`) and never reset, while `consumedReturnSurfaceRequest` is a `@State` on `MainView` (RootView.swift:43). `MainView` is torn down whenever the login screen shows (`showLogin || connection == nil`, RootView.swift:16) and recreated on sign-in, which resets `consumedReturnSurfaceRequest` to `0`. On re-login, MainView's cold-launch `.task` (RootView.swift:184) calls `presentPreferredReturnSurfaceIfNeeded()`, which sees the leftover nonzero request (e.g. `1`) `> 0`, consumes it, and presents the sessions drawer immediately — even though the user merely signed back in while foregrounded and never triggered a background→active return. This directly contradicts the stated "affects the next qualifying return only" / "disconnect → re-login cycles don't re-trigger" semantics. The existing test `testBackgroundToActiveWhileSignedOutDoesNotLeakIntoNextSignIn` (ChatReturnSurfaceTests.swift:59) only covers the *issuance* side (arming consumed before the auth guard); it does not cover this *presentation* side where a previously issued-and-consumed counter is re-consumed by a fresh MainView. Your own review triage at 00:48 confirmed this finding as valid, but the fix is not present in the current HEAD.
**Suggestion:** Track the consumed watermark in `AppState` (e.g. a `consumedReturnSurfaceRequest` alongside the counter, or reset/arm the counter during login teardown) so a recreated MainView does not re-present a stale value. Have the cold-launch `.task` only present a request that it itself armed (capture the pre-call counter), and add a test: "arm once, log out, log back in → drawer must not present."

### [minor] Inconsistent defer-vs-drop handling of explicit navigation states
**File:** Conduit/Views/RootView.swift:208,211
**Issue:** In `presentPreferredReturnSurfaceIfNeeded`, `hasPendingExplicitNavigation` (line 208) *defers* without consuming, so the request survives if the route is dropped; but `isOpeningNotificationSession` (line 211) *consumes and drops* the request. Both are explicit-navigation states, yet get opposite treatment — and `isOpeningNotificationSession` can become true after issuance in exactly the scenario where the pending-target branch was meant to preserve the request (e.g. a request issued, then a notification open kicks off before presentation). The asymmetry is that the not-yet-routed case preserves the request but the in-flight case discards it. Your review declined this as intended precedence, but the dual handling is easy to regress and unclear to readers.
**Suggestion:** Make the handling explicit and uniform — either defer (don't consume) in both cases, or document precisely why one survives and the other is dropped. If deferral is intended for pending-not-yet-routed only, state that contract on the method.

### [minor] Cold-launch once-per-process budget is consumed even when issuance is suppressed
**File:** Conduit/Services/AppState.swift:1084-1086
**Issue:** `requestPreferredReturnSurfaceForColdLaunch()` sets `hasRequestedColdLaunchReturnSurface = true` *before* calling `requestPreferredReturnSurface()`. If issuance is suppressed at cold launch (e.g. `hasPendingExplicitNavigation` true because the app launched from a notification tap — a common case), the once-per-process flag is consumed without ever issuing. In practice the scene-phase `.background`→`.active` path can still recover the drawer later, so impact is limited, but the flag no longer reliably means "cold launch was honored."
**Suggestion:** Only set `hasRequestedColdLaunchReturnSurface = true` when the request is actually issued, or allow a re-arm on a later real background→active return.

### [nit] Redundant presentation attempt between `.task` and `.onChange`
**File:** Conduit/Views/RootView.swift:184-190
**Issue:** The cold-launch `.task` calls `presentPreferredReturnSurfaceIfNeeded()` unconditionally, and the `.onChange(of: preferredReturnSurfaceRequest)` fires for the same request value shortly after. This is harmless because the second call sees the already-updated `consumedReturnSurfaceRequest`, but it couples correctness to the guard ordering.
**Suggestion:** Consider having the `.task` call only `requestPreferredReturnSurfaceForColdLaunch()` and rely solely on `.onChange` to consume, so there is a single consumption path.

## Other checks (no issues)
- No retain cycles: `.task`/`.onChange` closures hold only `EnvironmentObject` access; the settings `persistChatReturnSurface` closure captures `appState` for the sheet's lifetime only.
- Concurrency is clean: `AppState`, `PendingVoiceIntentStore`, and `PushNotificationService` are all `@MainActor`, and the new `hasPendingIntent`/`pendingTarget` reads and `preferredReturnSurfaceRequest` mutation all occur on the main actor — no Swift 6 / data-race concerns.
- Accessibility is good: the "Open to" segmented picker (AuxiliaryViews.swift:806) has a label and a conditionally-updated `.accessibilityHint`; no missing labels or traits.
- The explicit `Hashable` declaration on `ChatReturnSurface` (ChatReturnSurface.swift:10) addresses the `id: \.self` requirement cleanly.

## Verdict
REQUEST_CHANGES

The design is thoughtful, well-documented, and tested, and the schema isolation, precedence rules, and main-actor safety are sound. However, the `@State`-reset vs process-global-counter mismatch causes the sessions drawer to present on re-login without a qualifying return — contradicting the PR's own semantics and confirmed valid in your triage but not yet fixed. That should be resolved (with a regression test) before merge.

Note: the author's own review triage confirmed this re-login finding ("The re-login stale-request finding is valid") yet propose/state it remains in the current HEAD; I'm flagging it as the blocking item rather than the two declined findings, which are acceptable design trade-offs.

<a href="https://opencode.ai/s/nqGiXJdE">New%20session%20-%202026-08-19T00%3A54%3A12.551Z</a>
[opencode session](https://opencode.ai/s/nqGiXJdE)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32202910070)

kaishi00:
Triage of the re-review on ad077ff:

**[major] Stale request re-presents on re-login — taken, fixed in 031de48.** Correct catch: the watermark was MainView `@State`, reset by sign-out teardown. Consumption now lives in AppState as `claimPreferredReturnSurfacePresentation()`, which claims exactly once per issued request with the watermark surviving view recreation. Regression tests: `testClaimedRequestCannotBeReclaimedAfterMainViewRecreation` (issue → claim → re-claim returns false) and the claim-once assertions in the defer test.

**[minor] Inconsistent defer-vs-drop — taken via redesign.** The asymmetry is now an explicit, documented contract on the claim method: pending navigation defers without consuming, and the *next unblocked claim of the same request drops it* — the explicit route fully won that qualifying return (per the issue's precedence requirement). In-flight opens, modals, and preference changes keep the consume-and-drop semantics from the original PR's triage. `testRequestDeferredByPendingNavigationIsDroppedOnceRouteWins` locks the full lifecycle: issue → payload lands → defer → route completes → drop → next qualifying return claims fresh.

**[minor] Cold-launch budget spent while suppressed — declined.** The flag records that the cold-launch hook ran, not that a request issued; per the hardening spec, a launch with a pending notification is fully won by the explicit route — no drawer for that launch under any timing. Re-arming would only re-introduce a surprise presentation. The scene-phase path already serves later returns.

**[nit] Redundant .task + .onChange consumption — declined.** `.onChange` cannot observe changes that happened before the view attached, and scene activation at cold launch can issue before MainView exists. The `.task` claim is the only path that consumes those; `.onChange` handles subsequent increments. Two entry points, one consumer (the claim), no double-presentation possible.

Local: 21/21 `ChatReturnSurfaceTests` + 82 resume + 43 notification-payload suites green (146/146). CI running on 031de48.

kaishi00:
`Hashable` is now declared explicitly. Good — my earlier finding and the re-login finding are both fixed.

Now let me assess the deferred-request-never-re-evaluated issue once more. This is the only substantive remaining finding. Let me verify there's truly no production trigger for the deferred drop by checking the full picture once more. The claim's deferral path (`hasPendingExplicitNavigation`) sets `deferredReturnSurfaceRequest = current` and returns false. Then "the next unblocked claim of the same request drops it" — but `presentPreferredReturnSurfaceIfNeeded` is only invoked from `.task` and `.onChange(of: preferredReturnSurfaceRequest)`. Neither fires when pending navigation clears. So in production, the deferral is never resolved/dropped through the documented path; the deferred value lingers until superseded by a later request.

This is a real (if low-impact) gap — the coderabbit finding is accurate, and the author's triage claims the drop logic is "locked" by the test, but the test exercises it by manually calling the claim a second time in isolation, which no production code path does. I'll report this as a minor finding.

Let me finalize. Given the re-login major is fixed, and the remaining deferred-re-evaluation issue is minor (bounded impact, no user-visible bug), plus the snapshot-staleness nit, I'll go with COMMENT or APPROVE. Since there's a legitimate but minor finding worth the author consciously addressing, COMMENT is appropriate.

Let me write the review.

## Summary
This PR adds a persisted `ChatReturnSurface` preference (`.conversation`/`.sessions`) armed by `preferredReturnSurfaceRequest` on cold launch or a real background→active return, consumed by `claimPreferredReturnSurfacePresentation()` in AppState so MainView can present the sessions drawer, with notification/modal precedence and 21 tests. The prior major re-login re-presentation bug was fixed in 031de48 by moving the consumption watermark into AppState.

## Findings

### [minor] Deferred return-surface request is never re-evaluated when pending navigation clears
**File:** Conduit/Services/AppState.swift:1104-1121 and Conduit/Views/RootView.swift:203-207
**Issue:** `claimPreferredReturnSurfacePresentation()` implements a documented defer-and-drop path: while `hasPendingExplicitNavigation` is true it stores `deferredReturnSurfaceRequest = current` and returns false (without consuming), promising "the next unblocked claim of the same request then drops it." But the only production callers of `presentPreferredReturnSurfaceIfNeeded()` are MainView's cold-launch `.task` (RootView.swift:185) and `.onChange(of: appState.preferredReturnSurfaceRequest)` (RootView.swift:188). Neither fires when `PushNotificationService.shared.pendingTarget` clears or `PendingVoiceIntentStore` is consumed (e.g. ConduitApp.swift:91 `clearPendingTarget` after `openNotificationTarget`), because clearing explicit navigation never increments `preferredReturnSurfaceRequest`. Consequently the deferred request is left lingering: `deferredReturnSurfaceRequest` stays set to the old `current` until a later qualifying return issues a higher request that supersedes it. The test `testRequestDeferredByPendingNavigationIsDroppedOnceRouteWins` (ChatReturnSurfaceTests.swift:172) validates the claim only by calling it a second time manually in isolation — a path no production code exercises.
**Suggestion:** Wire the resolution into a real trigger: either observe pending-navigation clearing in MainView (e.g. publish a low-volume revision through AppState and invoke `presentPreferredReturnSurfaceIfNeeded()` when the pending state clears), or explicitly reset `deferredReturnSurfaceRequest` when `openNotificationTarget` completes. Add an end-to-end test covering request → pending route → route cleared → drawer drop/present.

### [minor] Snapshot-seeded `@State` in the settings picker can report a stale surface
**File:** Conduit/Views/AuxiliaryViews.swift:776-792
**Issue:** `ChatReturnBehaviorSettings` seeds `@State surface` from the `SettingsSnapshot` captured at sheet-presentation time (matching the pre-existing `behavior` pattern). If the settings sheet remains presented across a return while `chatReturnSurface` changes elsewhere, the picker's selection diverges from the effective `chatReturnSurface` read at presentation time in `claimPreferredReturnSurfacePresentation` (AppState.swift:1117). Not a regression — it mirrors the established behavior picker — but worth noting.
**Suggestion:** Acceptable as-is given precedent; if desired, reconcile the picker selection with the live value on `onChange`.

## Other checks (no issues)
- Retain cycles: `.task`/`.onChange` closures in MainView call view-struct methods on `EnvironmentObject` access only; settings persist closures capture `appState` for the view's lifetime. No cycle.
- Concurrency: AppState, PendingVoiceIntentStore, and PushNotificationService are `@MainActor`; `preferredReturnSurfaceRequest`/`consumedReturnSurfaceRequest`/`deferredReturnSurfaceRequest` are only mutated/read on the main actor. No Swift 6 / data-race concerns. `&+=` overflow-safe wrapping increment is appropriate for a monotonic counter.
- Accessibility: the "Open to" segmented picker (AuxiliaryViews.swift:806) has a label and a conditionally-updated `.accessibilityHint` for the Sessions state; no missing labels/traits.
- `ChatReturnSurface` now explicitly declares `Hashable` (ChatReturnSurface.swift:10), satisfying `ForEach(..., id: \.self)` robustly.
- The re-login re-presentation bug (previously flagged major) is correctly resolved: consumption now lives in AppState via `claimPreferredReturnSurfacePresentation`, so a MainView recreated after sign-out cannot re-claim a stale request; covered by `testClaimedRequestCannotBeReclaimedAfterMainViewRecreation`.

## Verdict
COMMENT

The major re-login issue is resolved and the design (real-`.background` arming, consume-before-auth-guard to prevent sign-in leak, schema isolation, main-actor safety, mobile accessibility) is sound. The remaining deferred-request issue is low-impact (bounded to a single return and correctly subsumed by later returns at runtime, despite the documented drop path not being wired to a production trigger); I'd recommend wiring the deferred-drop resolution to the pending-navigation-clears event and backing it with an end-to-end test, but it does not block merge.

<a href="https://opencode.ai/s/NEhQGUeA">New%20session%20-%202026-08-19T00%3A59%3A32.639Z</a>
[opencode session](https://opencode.ai/s/NEhQGUeA)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32203234558)

kaishi00:
I now have a complete picture. Let me write the review.

Key analysis:
- The prior **major** re-login finding is resolved: the consumed watermark lives in AppState (`consumedReturnSurfaceRequest`), and `retireOutstandingPreferredReturnSurfaceRequests()` at logout (lines 1950, 2071) advances the watermark to retire any outstanding request, so a recreated MainView can't re-present a stale request.
- Concurrency: AppState is `@MainActor`; all request/claim/mutation on main actor. No races.
- No retain cycles: `[weak self]` in the scene task; view `.task`/`.onChange` closures only touch EnvironmentObject.
- Accessibility: picker has label and conditional hint.
- `Hashable` explicitly declared.
- The consume-before-guards issue is no longer present (claim logic moved fully to AppState, consumption watermark in AppState).
- The deferred-request-never-re-evaluated issue: remains as a genuinely valid but low-impact concern — the "deferred drop" path (AppState.swift:1122-1125) requires a second call to `claimPreferredReturnSurfacePresentation()` with the same `current`, which no production trigger provides (`.task`/`.onChange` only). But as traced, the deferred request is harmless: it never re-presents (consumed watermark not advanced), and a later qualifying return issues a higher `current` that supersedes it. Bounded and consistent with the "notification won that launch" precedence. Minor.
- The cold-launch budget: flag set before issuance, declined as intended. Reasonable.
- One new consideration: in `claimPreferredReturnSurfacePresentation`, the drop branch (1122-1125) advances `consumedReturnSurfaceRequest` = current but the `defer { deferred = nil }` clears the deferral. This is only reachable via a second call with same current — which requires `current > consumed` still true and `!hasPendingExplicitNavigation` now. Since deferred is only set while pending-nav, that means pending-nav cleared between the two calls but current didn't change. Since no production trigger re-calls with same current, it's dead code in practice but documents intent.

Let me also double check the `.task` cold launch and `.onChange` dual-entry concern. `.task` calls `requestPreferredReturnSurfaceForColdLaunch()` then `presentPreferredReturnSurfaceIfNeeded()`. `.onChange(of: preferredReturnSurfaceRequest)` fires when the counter increments (scene-phase returns). Both share the same consumer `claimPreferredReturnSurfacePresentation()` which is exactly-once via the watermark. No double present. Good.

One thing to verify: the scene-phase `.onChange` in MainView fires on `preferredReturnSurfaceRequest` increment, which happens on `.active` after real background. But does `.onChange` in a `.sheet`-holding MainView reliably fire while backgrounded→active? The MainView is in the window; `.onChange` should fire. Fine.

Let me check one subtle thing: `presentPreferredReturnSurfaceIfNeeded` has `guard !appState.showSidebar else { return }`. If showSidebar is already true at return, the drawer is already showing — fine, skip. But note that when a claim is consumed (returns true) and this guard fails, the request is already consumed by the claim (consumedReturnSurfaceRequest advanced) — so it won't retry. That's fine since the drawer is already present.

Actually wait — there's a subtle issue. `claimPreferredReturnSurfacePresentation()` returns true but advances `consumedReturnSurfaceRequest = current` **before** the guards (lines 1126-1129 the guard is AFTER `consumedReturnSurfaceRequest = current`). So even if the guard in `claim` fails (e.g. modal presented), the request is consumed and dropped. That's the intended "consume-and-drop on precedence losers" semantics. And in MainView, the `guard !showSidebar` after a successful claim drops presentation but request already consumed — fine because showSidebar already true means drawer present.

Now, is there any scenario where claim returns true, consumed advanced, then MainView's `showSidebar` guard drops it AND the drawer isn't already showing? `showSidebar` true means drawer is showing (it's bound to the sheet). So no. Good.

Overall the code is sound. The only remaining substantive finding is the dead-code deferred-drop path — minor. I'll also note the snapshot staleness nit.

Let me write it up as VERDICT COMMENT.

## Summary
Adds a persisted `ChatReturnSurface` preference (`.conversation`/`.sessions`) whose one-shot `preferredReturnSurfaceRequest` token is armed on cold launch or a genuine background→active return and claimed via `claimPreferredReturnSurfacePresentation()` (watermark in AppState) so MainView presents the existing sessions drawer, with notification/deep-link and modal precedence. 21 tests cover all spec scenarios.

## Findings

### [minor] Deferred-request "drop" path is only exercised by tests, not production code
**File:** Conduit/Services/AppState.swift:1117-1125
**Issue:** The claim documents a defer-and-drop contract: while `hasPendingExplicitNavigation` is true it stores `deferredReturnSurfaceRequest = current` and returns false; the "next unblocked claim of the same request then drops it" via the `deferredReturnSurfaceRequest == current` branch (1122-1125). But the only production callers of `presentPreferredReturnSurfaceIfNeeded()` are MainView's cold-launch `.task` (RootView.swift:185) and `.onChange(of: preferredReturnSurfaceRequest)` (RootView.swift:188). Neither fires when `PushNotificationService.shared.pendingTarget` clears or `PendingVoiceIntentStore` is consumed, because clearing explicit navigation never increments the counter. So the drop branch is effectively dead in production — `testRequestDeferredByPendingNavigationIsDroppedOnceRouteWins` exercises it only by invoking the claim a second time manually, a path with no real trigger.
**Suggestion:** The impact is bounded and benign — the deferred request never re-presents (watermark is not advanced while deferred), and any later qualifying return issues a higher `current` that supersedes it, so the "notification won that launch" precedence still holds. Consider either wiring the deferred-drop resolution to a pending-navigation-cleared event (publish through AppState and re-run the claim) or removing the misleading defer/drop branch and documenting the simpler reality. Not blocking.

### [minor] `@State` in `ChatReturnBehaviorSettings` is seeded from a snapshot and can diverge from the live preference
**File:** Conduit/Views/AuxiliaryViews.swift:780,791
**Issue:** `surface` (and `behavior`) are `@State` seeded from the `SettingsSnapshot` captured when the settings sheet is presented, while the authoritative `chatReturnSurface` is also read live at presentation time inside `claimPreferredReturnSurfacePresentation` (AppState.swift:1127). If the settings sheet stays presented across a qualifying return that changes `chatReturnSurface` elsewhere, the picker selection can disagree with the decision the return actually uses. This mirrors the pre-existing `behavior` pattern in the same view, so it's not a regression.
**Suggestion:** Acceptable given precedent. If desired, reconcile the picker selection with the live value on `onChange(of:)`.

## Other checks (no issues found)
- Retain cycles: scene-phase task uses `[weak self]` (AppState.swift:3841); MainView `.task`/`.onChange` closures only touch the `@EnvironmentObject`, captured for view lifetime. No cycle.
- Concurrency: `AppState` is `@MainActor`; `preferredReturnSurfaceRequest`, `consumedReturnSurfaceRequest`, `deferredReturnSurfaceRequest`, and `hasEnteredBackgroundScenePhase` are only mutated/read on the main actor. No data races / Swift 6 violations. `&+=` wrapping increment on a monotonic counter is appropriate.
- Force unwraps / IUOs: none introduced.
- The prior re-login re-presentation defect is correctly resolved — the consumed watermark now lives in AppState and `retireOutstandingPreferredReturnSurfaceRequests()` (AppState.swift:1099, called at 1950/2071) advances it on sign-out, and `MainView` teardown no longer resets the watermark; covered by `testClaimedRequestCannotBeReclaimedAfterMainViewRecreation`.
- `ChatReturnSurface` explicitly declares `Hashable` (ChatReturnSurface.swift:10), satisfying `ForEach(..., id: \.self)`.
- Lifecycle: real-`.background` arming consumed before auth guard (AppState.swift:3823-3825) so login-screen cycles can't leak; `handleScenePhase(.inactive)` never arms, matching "only background/active triggers" spec.
- Accessibility: the "Open to" segmented picker has a label plus a conditionally-updated `.accessibilityHint` for the Sessions state (AuxiliaryViews.swift:812-816); no missing labels/traits.

## Verdict
COMMENT

The design is sound and well-tested: main-actor safety, the AppState-owned claim watermark (resolving the earlier re-login re-presentation bug), notification/modal precedence, separate persistence, and accessibility are all correct. The sole lingering concern is the defer-and-drop path in `claimPreferredReturnSurfacePresentation` being unreachable from production triggers (only the tests drive it); it's low-impact and behaves safely, so it doesn't block merge, but wiring it to a pending-navigation-cleared event (or documenting the simpler behavior) would make the code match its stated contract.

<a href="https://opencode.ai/s/Hi1XV3qL">New%20session%20-%202026-08-19T01%3A04%3A57.996Z</a>
[opencode session](https://opencode.ai/s/Hi1XV3qL)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32203571459)

kaishi00:
Final triage on 5513efc — both minors declined, both consistent with the bot's own non-blocking assessment:

**Deferred-drop path "only exercised by tests" — declined; it has a production trigger the review missed.** MainView's cold-launch `.task` (RootView.swift) re-runs every time MainView enters the hierarchy. The realistic production sequence: scene activation at cold launch issues a request before MainView exists (login still dismissing), a notification payload defers the first claim, and the route clears before MainView's first `.task` — that claim runs the drop branch with the unchanged counter value. Beyond that, the branch is deliberately defensive: it guarantees the deferred request can never present after its route won, under any future trigger ordering, which is the invariant the hardening pass asked for. Removing it to satisfy a reachability argument would trade a documented guarantee for a narrower one.

**Snapshot-seeded `@State` staleness — declined (third pass).** Mirrors the pre-existing `behavior` picker pattern in the same view; not a regression, and the sheet re-snapshots on every presentation via `makeSettingsSnapshot()`.

Verification-pass summary for reviewers: Case 1 (claimed request) held via the AppState watermark; Case 2 (issued-unclaimed/deferred at teardown) was reachable through the deferred-claim path and is closed by `retireOutstandingPreferredReturnSurfaceRequests()` in `disconnect()`/`requireSignIn()` — watermark advances to the counter, monotonic counter untouched, sign-in neither issues nor re-arms. Regression tests drive the real `disconnect()` API, cover the deferred variant, and prove the next genuine background→active still issues and claims a fresh request. Local: 787 unit + 6 UI tests green; CI green on 5513efc.

PR #83, opened 2026-08-19
Manual per-message read aloud for assistant responses

### LINKED ISSUE #81: request: support for per-message TTS / Read aloud
Apologies if I've just missed something, and this is already supported...

Far i can tell, TTS & STT is supported, but only as part of Voice Conversation mode. I have TTS setup & working in the settings, but no message-level option to read-aloud.

Be good to allow support for per-message TTS / Read Aloud similar to Hermes Desktop for when you just one the occasional response read out to you.

kaishi00:
Adds the manual Read Aloud action from #81: a speaker control on completed assistant messages that streams the response through the existing TTS pipeline. No automatic TTS — playback only starts on tap.

## What changed

**`MessageReadAloudController`** (new, `Conduit/Voice/`) — UI-independent, `@MainActor`, mirrors the `VoiceConversationController` pattern:
- Owns the active message ID, `idle / preparing / playing / failed` state, the active `VoiceSpeechStream`, the playback task, and an operation generation token. Every post-`await` mutation of shared state sits behind an `isCurrent(generation)` guard, so a superseded operation (tap-to-stop, tap-to-switch, background, disconnect) can never clear or overwrite the newer operation's stream or state — including in the error path.
- Reuses `VoiceGatewayService.openSpeechStream` (streaming PCM + encoded fallback) and the shared `SpeechPlaybackService`; no second transport, no new endpoints, no `HermesVoiceGateway` changes.
- After stream completion it finishes and drains the playback engine, returning to idle only if it still owns the operation. Errors stop any in-flight audio and surface via the existing `errorMessage` convention.

**AppState** — TTS-only availability (`isConnected` + `isVoiceEnabled` + `snapshot.supportsSpeech`; deliberately no mic/STT/Apple Speech/transcription requirements). Lifecycle stops on scene background, disconnect, forced sign-out, and profile switch; mutual exclusion with the voice conversation sheet and the live TTS test. The capability refresh keeps the live gateway across refreshes so playback isn't churned — but a kept gateway must still match the active profile **and** the live `DashboardTicketBridge` (a gateway outliving its invalidated bridge is rebuilt, which is what a same-profile re-login would otherwise leak).

**ChatView** — 34×34 SF Symbol control in `AssistantBubble` beside Copy/Branch: `speaker.wave.2` idle, spinner while preparing, `stop.fill` playing (accent-tinted), existing haptic conventions, "Read response aloud" / "Stop reading response" labels. Only the owning message shows the active state; the button is absent on in-progress (partial) messages and disabled when read aloud is unavailable.

## Tests

18 focused tests in `MessageReadAloudControllerTests` (patterned after `VoiceConversationControllerTests`, checked-continuation mocks for deterministic interleavings), covering: full-content open, PCM and encoded fallback paths, finish→drain→idle, tap-to-stop, cross-message takeover, superseded-operation isolation (success, late-open, late-finish, and the deferred-catch path where a stale operation's error lands after the newer operation owns its stream), mid-stream failure settling the audio engine, empty-content no-op, repeated stops, TTS-only availability, background stop, and gateway replacement. Existing voice suites unchanged.

## Verification

- `MessageReadAloudControllerTests` 18/18 and `VoiceConversationControllerTests` 28/28 green locally; broader suites (AppState chat resume, chat title scroll, voice config, session cache/rename/yolo, audio session) green locally.
- Full suite + build via CI on this PR.
- UI flows reasoned through: idle start, tap-active stop, tap-different takeover, background stop, disconnect/logout stop, profile-switch stop, and voice-conversation/TTS-test mutual exclusion.

## Summary by CodeRabbit

* **New Features**
  * Added an option to read assistant messages aloud.
  * Added playback controls with preparation status, accessibility labels, and availability feedback.
  * Added persisted chat return-surface preferences.
  * Improved sessions-drawer presentation when returning to chat.
* **Bug Fixes**
  * Improved recovery from interruptions, cancellations, and failed speech generation.
  * Read-aloud playback now responds reliably to connectivity, voice settings, sign-out, backgrounding, and profile changes.
  * Prevented read-aloud playback from conflicting with voice conversations.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -7838,7 +7838,13 @@ final class AppState: ObservableObject {
     /// transcription, mic permission, or Apple Speech — a profile with TTS
     /// but no STT must still be able to read responses aloud.
     var readAloudUnavailableReason: String? {
-        MessageReadAloudController.unavailableReason(
+        // Opening a stream needs the dashboard bridge; while it is absent
+        // (mid sign-out, before the connection lands) disable the button
+        // instead of failing at tap time. Mirrors makeVoiceGateway().
+        guard dashboardTicketBridge != nil, connection != nil else {
+            return "Read aloud needs a connected Hermes gateway."
+        }
+        return MessageReadAloudController.unavailableReason(
             isConnected: isConnected,
             isVoiceEnabled: isVoiceEnabled,
             snapshot: voiceCapabilitySnapshot
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -500,6 +500,18 @@ final class AppState: ObservableObject {
             await self?.interruptForVoice()
         }
     )
+    /// Manual per-message read aloud for completed assistant responses.
+    /// TTS-only: independent of the voice conversation and of STT.
+    lazy var messageReadAloudController = MessageReadAloudController(
+        reportError: { [weak self] message in
+            self?.errorMessage = message
+        }
+    )
+    /// The bridge instance the read aloud gateway was built against. The
+    /// gateway captures its bridge at init, so a gateway outliving its bridge
+    /// (disconnect, re-login, bridge rotation) is dead and must be rebuilt
+    /// rather than kept.
+    private var readAloudGatewayBridge: DashboardTicketBridge?
 
     // MARK: - Cron
 
@@ -1828,6 +1840,11 @@ final class AppState: ObservableObject {
         dashboardTicketBridge?.invalidate()
         dashboardTicketBridge = nil
         voiceConversationController.stop()
+        messageReadAloudController.stop()
+        // The bridge is invalidated above; a gateway built against it can
+        // never open a stream again, so it must not survive the re-login.
+        messageReadAloudController.setGateway(nil)
+        readAloudGatewayBridge = nil
         showVoiceSheet = false
         voiceCapabilitySnapshot = .unavailable
         isVoiceEnabled = false
@@ -1948,6 +1965,11 @@ final class AppState: ObservableObject {
         connection = nil
         dashboardTicketBridge?.invalidate()
         dashboardTicketBridge = nil
+        // A forced sign-out kills the bridge mid-playback; stop the read
+        // aloud and drop its gateway the same way Disconnect does.
+        messageReadAloudController.stop()
+        messageReadAloudController.setGateway(nil)
+        readAloudGatewayBridge = nil
         projects = []
         supportsProjects = false
         projectsLoading = false
@@ -3701,6 +3723,7 @@ final class AppState: ObservableObject {
         case .active:
             isSceneActive = true
             voiceConversationController.setForegroundActive(true)
+            messageReadAloudController.setForegroundActive(true)
             guard connection != nil else { return nil }
             cancelScenePhaseAttempt()
             // Publish the foreground reconciliation boundary synchronously.
@@ -3760,6 +3783,7 @@ final class AppState: ObservableObject {
         case .background:
             isSceneActive = false
             voiceConversationController.setForegroundActive(false)
+            messageReadAloudController.setForegroundActive(false)
             showVoiceSheet = false
             // Drop any armed reconnect timer as well: in-flight cycles abort
             // at their next transportContinuation checkpoint, and foreground
@@ -3788,6 +3812,7 @@ final class AppState: ObservableObject {
             cancelScenePhaseAttempt()
             chatResumeCoordinator.freezeViewport()
             voiceConversationController.setForegroundActive(false)
+            messageReadAloudController.setForegroundActive(false)
             return nil
 
         @unknown default:
@@ -5910,6 +5935,9 @@ final class AppState: ObservableObject {
             return false
         }
         guard !isProfileSwitching else { return false }
+        // A profile change replaces the voice gateway; any in-flight read
+        // aloud belongs to the outgoing profile.
+        messageReadAloudController.stop()
         let transitionGeneration: UInt64
         if let viewportTransitionGeneration {
             guard chatViewportTransitionIsCurrent(
@@ -7676,12 +7704,47 @@ final class AppState: ObservableObject {
 
     var canStartVoiceConversation: Bool { voiceUnavailableReason == nil }
 
+    /// TTS-only availability for read aloud: a connected gateway with voice
+    /// enabled and a ready speech provider. Deliberately does not require
+    /// transcription, mic permission, or Apple Speech — a profile with TTS
+    /// but no STT must still be able to read responses aloud.
+    var readAloudUnavailableReason: String? {
+        MessageReadAloudController.unavailableReason(
+            isConnected: isConnected,
+            isVoiceEnabled: isVoiceEnabled,
+            snapshot: voiceCapabilitySnapshot
+        )
+    }
+
+    /// Chat bubble entry point for manual read aloud. Toggling the active
+    /// message stops it without touching the gateway; starting a different
+    /// message takes over from whatever is playing.
+    func toggleReadAloud(message: ChatMessage) {
+        if messageReadAloudController.isActiveMessage(message.id) {
+            messageReadAloudController.stop()
+            return
+        }
+        guard !message.content.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else { return }
+        if let reason = readAloudUnavailableReason {
+            errorMessage = reason
+            return
+        }
+        // The capability refresh keeps the gateway current; this covers the
+        // gap before the next refresh has run — including right after a
+        // profile switch or re-login, when a stale gateway may still be set.
+        if !readAloudGatewayIsCurrent {
+            assignReadAloudGateway()
+        }
+        messageReadAloudController.toggle(messageID: message.id, content: message.content)
+    }
+
     func refreshVoiceCapabilities() async {
         let profile = activeProfile
         guard isConnected, let bridge = dashboardTicketBridge else {
             voiceCapabilitySnapshot = .unavailable
             isVoiceEnabled = false
             voiceConversationController.setGateway(nil)
+            refreshReadAloudGateway()
             return
         }
         installVoiceAssistantObserverIfNeeded()
@@ -7695,6 +7758,7 @@ final class AppState: ObservableObject {
         voiceTranscriptionMode = preferences.resolvedTranscriptionMode
         voiceConversationController.setProfilePreferences(preferences)
         refreshVoiceControllerGateway()
+        refreshReadAloudGateway()
     }
 
     @discardableResult
@@ -7703,6 +7767,7 @@ final class AppState: ObservableObject {
         defaults.set(enabled, forKey: voiceEnabledPreferenceKey(profile: activeProfile))
         isVoiceEnabled = enabled
         refreshVoiceControllerGateway()
+        refreshReadAloudGateway()
         if !enabled {
             voiceConversationController.stop()
             showVoiceSheet = false
@@ -7739,12 +7804,16 @@ final class AppState: ObservableObject {
         voiceTranscriptionMode = mode
         voiceConversationController.setProfilePreferences(preferences)
         refreshVoiceControllerGateway()
+        refreshReadAloudGateway()
         return true
     }
 
     @discardableResult
     func openVoiceConversation(_ intent: PendingVoiceIntent) async -> Bool {
         guard isConnected else { return false }
+        // Mutual exclusion: the voice conversation owns playback while its
+        // sheet is open, so a read aloud started before must not continue.
+        messageReadAloudController.stop()
         if showVoiceSheet { closeVoiceConversation() }
         if let rawProfile = intent.profile {
             let requestedProfile = rawProfile.trimmingCharacters(in: .whitespacesAndNewlines)
@@ -7825,6 +7894,9 @@ final class AppState: ObservableObject {
         guard let gateway = makeVoiceGateway() else {
             return .failure("Conduit could not connect this test to the selected profile.")
         }
+        // The live TTS test shares the playback engine with read aloud; the
+        // test must not fight a message that is still being read.
+        messageReadAloudController.stop()
         voiceConversationController.setGateway(gateway)
         return await voiceConversationController.runSpeechTest(
             text: "Conduit voice is ready for this profile."
@@ -7852,6 +7924,31 @@ final class AppState: ObservableObject {
         voiceConversationController.setGateway(
             isVoiceEnabled && selectedTranscriptionIsAvailable ? makeVoiceGateway() : nil
         )
+    }
+
+    /// Keeps the read aloud gateway in step with capability refreshes without
+    /// churning a live instance: swapping the gateway mid-playback would stop
+    /// the message. A kept gateway must still match the active profile AND
+    /// the live dashboard bridge — a gateway built against an invalidated or
+    /// rotated bridge (re-login, profile switch) is rebuilt instead.
+    private var readAloudGatewayIsCurrent: Bool {
+        guard let gateway = messageReadAloudController.gateway else { return false }
+        return gateway.profile == activeProfile && readAloudGatewayBridge === dashboardTicketBridge
+    }
+
+    private func assignReadAloudGateway() {
+        messageReadAloudController.setGateway(makeVoiceGateway())
+        readAloudGatewayBridge = dashboardTicketBridge
+    }
+
+    private func refreshReadAloudGateway() {
+        if readAloudUnavailableReason != nil {
+            messageReadAloudController.setGateway(nil)
+            readAloudGatewayBridge = nil
+            return
+        }
+        guard !readAloudGatewayIsCurrent else { return }
+        assignReadAloudGateway()
     }
 
     private func loadVoiceProfilePreferences(profile: String) -> VoiceProfilePreferences {
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -7932,8 +7932,9 @@ final class AppState: ObservableObject {
     /// the live dashboard bridge — a gateway built against an invalidated or
     /// rotated bridge (re-login, profile switch) is rebuilt instead.
     private var readAloudGatewayIsCurrent: Bool {
-        guard let gateway = messageReadAloudController.gateway else { return false }
-        return gateway.profile == activeProfile && readAloudGatewayBridge === dashboardTicketBridge
+        guard let gateway = messageReadAloudController.gateway,
+              let bridge = dashboardTicketBridge else { return false }
+        return gateway.profile == activeProfile && readAloudGatewayBridge === bridge
     }
 
     private func assignReadAloudGateway() {
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -7870,6 +7870,11 @@ final class AppState: ObservableObject {
         if !readAloudGatewayIsCurrent {
             assignReadAloudGateway()
         }
+        // Mutual exclusion (reverse of openVoiceConversation / runVoiceTTSTest):
+        // the voice conversation and read aloud each own their own playback
+        // instance, so a speech test or conversation still speaking must stop
+        // before the manual stream opens — otherwise both engines play at once.
+        voiceConversationController.stop()
         messageReadAloudController.toggle(messageID: message.id, content: message.content)
     }
 
@@ -8029,8 +8034,11 @@ final class AppState: ObservableObject {
         guard let gateway = makeVoiceGateway() else {
             return .failure("Conduit could not connect this test to the selected profile.")
         }
-        // The live TTS test shares the playback engine with read aloud; the
-        // test must not fight a message that is still being read.
+        // The speech test speaks through the voice controller's own playback
+        // instance (the same AVSpeech infrastructure read aloud uses, but a
+        // separate instance), so the test and a still-playing message would
+        // otherwise both produce audio. AppState enforces mutual exclusion in
+        // both directions; toggleReadAloud stops this controller in turn.
         messageReadAloudController.stop()
         voiceConversationController.setGateway(gateway)
         return await voiceConversationController.runSpeechTest(
@@ -8085,6 +8093,23 @@ final class AppState: ObservableObject {
         }
         guard !readAloudGatewayIsCurrent else { return }
         assignReadAloudGateway()
+    }
+
+    /// Test-only: installs the post-connect voice capability state (bridge,
+    /// snapshot, preference) without a dashboard round trip, so the read
+    /// aloud / voice conversation mutual exclusion can be exercised with
+    /// mock controllers. `readAloudGatewayBridge` is set to the same bridge
+    /// so a mock read aloud gateway installed on the controller counts as
+    /// current and is not replaced by a real one at tap time.
+    func installVoiceCapabilityStateForTesting(
+        bridge: DashboardTicketBridge,
+        snapshot: VoiceCapabilitySnapshot,
+        isVoiceEnabled: Bool
+    ) {
+        dashboardTicketBridge = bridge
+        readAloudGatewayBridge = bridge
+        voiceCapabilitySnapshot = snapshot
+        self.isVoiceEnabled = isVoiceEnabled
     }
 
     private func loadVoiceProfilePreferences(profile: String) -> VoiceProfilePreferences {
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -773,7 +773,10 @@ struct MessageBubble: View {
         case .user:
             UserBubble(message: message)
         case .assistant:
-            AssistantBubble(message: message)
+            AssistantBubble(
+                message: message,
+                readAloudController: appState.messageReadAloudController
+            )
         case .reasoning:
             ThinkingCard(message: message)
         case .tool:
@@ -797,7 +800,10 @@ struct MessageBubble: View {
                 SystemBubble(message: message)
             }
         case .partial:
-            AssistantBubble(message: message)
+            AssistantBubble(
+                message: message,
+                readAloudController: appState.messageReadAloudController
+            )
         }
     }
 }
@@ -1055,7 +1061,13 @@ private struct UserDocumentAttachmentChip: View {
 struct AssistantBubble: View {
     let message: ChatMessage
     @EnvironmentObject var appState: AppState
+    @ObservedObject private var readAloudController: MessageReadAloudController
     @State private var copied = false
+
+    init(message: ChatMessage, readAloudController: MessageReadAloudController) {
+        self.message = message
+        self.readAloudController = readAloudController
+    }
 
     var body: some View {
         VStack(alignment: .leading, spacing: 6) {
@@ -1117,10 +1129,45 @@ struct AssistantBubble: View {
                 .disabled(appState.isBusy || appState.isBranchingChat)
                 .opacity(appState.isBusy || appState.isBranchingChat ? 0.45 : 1)
                 .accessibilityLabel("Branch from this response")
+
+                if message.role == .assistant {
+                    Button {
+                        if readAloudIsActive {
+                            Haptics.medium()
+                        } else {
+                            Haptics.light()
+                        }
+                        appState.toggleReadAloud(message: message)
+                    } label: {
+                        Group {
+                            if case .preparing(let id) = readAloudController.state, id == message.id {
+                                ProgressView()
+                                    .controlSize(.small)
+                            } else {
+                                Image(systemName: readAloudIsActive ? "stop.fill" : "speaker.wave.2")
+                            }
+                        }
+                        .font(.subheadline.weight(.semibold))
+                        .frame(width: 34, height: 34)
+                    }
+                    .buttonStyle(.plain)
+                    .foregroundStyle(readAloudIsActive ? Color.conduitAccent : Color.secondary)
+                    .disabled(readAloudUnavailable && !readAloudIsActive)
+                    .opacity(readAloudUnavailable && !readAloudIsActive ? 0.45 : 1)
+                    .accessibilityLabel(readAloudIsActive ? "Stop reading response" : "Read response aloud")
+                }
             }
             .padding(.top, 2)
         }
         .frame(maxWidth: .infinity, alignment: .leading)
+    }
+
+    private var readAloudIsActive: Bool {
+        readAloudController.isActiveMessage(message.id)
+    }
+
+    private var readAloudUnavailable: Bool {
+        appState.readAloudUnavailableReason != nil
     }
 
     private func copyResponse() {
--- a/Conduit/Voice/MessageReadAloudController.swift
+++ b/Conduit/Voice/MessageReadAloudController.swift
@@ -1,0 +1,210 @@
+//
+//  MessageReadAloudController.swift
+//  Conduit
+//
+//  Manual per-message read aloud. Deliberately a single-message playback
+//  operation rather than a voice conversation: it opens the same streaming
+//  TTS transport and plays through the same playback service, but never
+//  touches microphone, STT, or voice-conversation state, so it stays usable
+//  on profiles that configure TTS without transcription.
+//
+
+import Combine
+import Foundation
+
+@MainActor
+final class MessageReadAloudController: ObservableObject {
+    enum State: Equatable {
+        case idle
+        case preparing(messageID: String)
+        case playing(messageID: String)
+        case failed(messageID: String, message: String)
+    }
+
+    @Published private(set) var state: State = .idle
+
+    private let playback: SpeechPlaybackService
+    private let reportError: @MainActor (String) -> Void
+    private var activeGateway: VoiceGatewayService?
+    private var speechStream: VoiceSpeechStream?
+    private var playbackTask: Task<Void, Never>?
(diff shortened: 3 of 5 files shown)


kaishi00:
I now have a thorough understanding of the implementation and tests. Let me write the review.

Let me reconsider the most material findings once more:

The design is robust on concurrency (all `@MainActor`, generation guard token, `[weak self]` everywhere). No retain cycles found. The main issues are test determinism, all-bubbles re-render performance, and a couple of minor correctness/edge items. Nothing rises to critical/major.

Actually, let me reconsider one thing more carefully as a potential real (if minor) correctness concern: In the controller, `toggle()` calls `stop()` then `startPlayback()`. `startPlayback` increments generation. But `stop()` also sets `state = .idle`, then `startPlayback` sets `.preparing`. Fine.

Also — the `@MainActor` `MessageReadAloudController` is stored as `@ObservedObject` in `AssistantBubble`, but `MessageReadAloudController` is only used in views via `@ObservedObject` which requires it to be `ObservableObject` — it is. Good.

One more: `AssistantBubble` now has a custom `init`. Does `MessageBubble` (the parent) use `AssistantBubble` anywhere else without the new parameter? The diff showed both call sites updated. Good. And `AssistantBubble` takes `readAloudController` param explicitly. Good.

Let me write up the review.

## Summary
Adds a manual per-message "Read Aloud" button to completed assistant chat bubbles, streaming the response through the existing TTS pipeline via a new `@MainActor MessageReadAloudController`, with AppState lifecycle/availability wiring and 18 tests.

## Findings

### [minor] Fixed-sleep timing makes tests nondeterministic/flaky
**File:** ConduitTests/MessageReadAloudControllerTests.swift (multiple, e.g. lines 25, 42, 59, 73, 92)
**Issue:** The PR description claims "checked-continuation mocks for deterministic interleavings," but most assertions are driven by `try? await Task.sleep(nanoseconds: 50_000_000)` followed by a state check. These are timing-dependent: on a slow or heavily loaded CI runner, 50ms may not be enough for the task to reach a given state, and on a fast runner the interleaving the mock was meant to pin may already have resolved. This makes the suite flaky and reduces confidence in the "deterministic" guarantee for the interleaving tests (e.g. `testSupersededCatchCannotReleaseNewerStream`, `testStopDuringStreamCreationCannotRevivePlayback`).
**Suggestion:** Use the checked-continuation / explicit release points (`releaseAppendForTest`, `resumeOpenForTest`, etc.) that already exist to drive state transitions deterministically instead of relying on wall-clock sleeps for the assertions that depend on ordering.

### [minor] Shared `@ObservedObject` re-evaluates every assistant bubble
**File:** Conduit/Views/ChatView.swift:1064
**Issue:** Every `AssistantBubble` holds `@ObservedObject` on the single shared `messageReadAloudController`. Each state transition (idle → preparing → playing → idle/failed) observed through `readAloudController.state`, `readAloudIsActive`, and `readAloudUnavailable` recomputes the body of every assistant message in the list. For long chats this is a whole-list re-render on each playback transition. The `unavailableReason` branch calls into `appState.readAloudUnavailableReason` per bubble per body evaluation as well.
**Suggestion:** Consider gating observation (e.g. only observing `.preparing(id)`/`.playing(id)` for the owning message, or scoping state observations so non-owning bubbles don't re-render), or note this as an accepted tradeoff given chat scale.

### [minor] `deinit` doesn't explicitly tear down the active speech stream
**File:** Conduit/Voice/MessageReadAloudController.swift:44
**Issue:** `deinit` only cancels `playbackTask`. The active `speechStream` is left to ARC release, which happens to trigger `HermesSpeechStream.deinit` (cancelling the websocket/tasks), so it works — but only indirectly. If the stream was opened but never assigned or if cleanup order matters, audio could linger briefly on controller dealloc.
**Suggestion:** Cancel the stream in `deinit` (e.g. `speechStream?.cancel()`) for symmetry with `stop()` and to make teardown explicit rather than relying on the stream's own deinit.

### [minor] Availability check can be decoupled from gateway availability
**File:** Conduit/Services/AppState.swift:7711 (`readAloudUnavailableReason`) vs line 7674 (`makeVoiceGateway`)
**Issue:** `readAloudUnavailableReason` gates only on `isConnected`, `isVoiceEnabled`, and `snapshot.supportsSpeech` — not on the bridge/connection actually being present to build a gateway. In the narrow window where `isConnected` is true but `dashboardTicketBridge`/`connection` is nil (e.g. mid-reconnect after a credential refresh), `makeVoiceGateway()` returns nil; `toggleReadAloud` then calls `toggle` with a nil gateway and the user sees a confusing generated error ("Read aloud needs a connected Hermes gateway.") from the controller rather than the button simply being unavailable. (The second commit correctly hardened `readAloudGatewayIsCurrent` for this nil-bridge case but not the availability gate.) Also note `toggle()` in the controller itself doesn't guard `isBackground` for the case where `readAloudUnavailableReason` is nil.
**Suggestion:** Have `readAloudUnavailableReason` (or `toggleReadAloud`) also require a non-nil `dashboardTicketBridge`/gateway-constructibility so the button is disabled rather than failing at tap time.

### [nit] Fixed 34×34 tap target and `.plain` button rely on image-only label
**File:** Conduit/Views/ChatView.swift:1150-1157
**Issue:** The control uses a fixed 34×34 frame with `.buttonStyle(.plain)` and only an image (no text), with the dynamic-type context inherited from `.subheadline`. Accessibility label is present (good), but the 34pt target is below the 44pt recommended minimum, and while preparing there is no accessibility state exposing "loading."
**Suggestion:** Optionally enlarge the hit target to ≥44pt and add `.accessibilityValue`/state to convey the preparing/playing state for VoiceOver.

## Verdict
APPROVE

The core logic is strong: `@MainActor`, a generation token guarding every post-await shared-state mutation, all closures use `[weak self]` (no retain cycles), and error/teardown paths settle the playback engine. Findings are minor/nit — no critical or major issues. The test suite's reliance on fixed `Task.sleep` durations is the most worthwhile follow-up for CI reliability.

<a href="https://opencode.ai/s/xQg6k8Pc">New%20session%20-%202026-08-19T01%3A31%3A25.986Z</a>
[opencode session](https://opencode.ai/s/xQg6k8Pc)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32205142175)

kaishi00:
Triage per finding (lens: will a user actually hit this?):

**1. Fixed-sleep tests — accepted as written (convention match).** The interleaving itself is pinned by checked continuations (blocking open, `pauseAfterEmission`, deferred-cancel delivery); the 50ms sleeps only yield control so the task can run to its next pinned suspension point. This is the same pattern the existing `VoiceConversationControllerTests` uses throughout (e.g. lines 207, 237, 258, 263, 294, 299), so the suite follows codebase convention rather than introducing a new flakiness surface — and the full suite passed on the fresh CI runner. Converting every sleep to an explicit release point would be a fair follow-up if a flake ever shows up in CI; I don't want to rewrite both suites' convention in this PR.

**2. Shared `@ObservedObject` re-renders — accepted tradeoff.** `AssistantBubble` already holds `@EnvironmentObject appState` and re-evaluates on every one of its published changes (streaming message appends, turn state, etc.) — far more frequently than read-aloud's transitions, which are at most three per playback (tap → preparing → playing → idle/failed) and always user-initiated. Gating observation with per-bubble derived state would add real complexity for a cost dwarfed by what the existing environment-object pattern already does. Noted as accepted given chat scale.

**3. `deinit` stream teardown — declined.** The controller is a `lazy var` on `AppState` (the app's root object) and is never deallocated in production; its `deinit` only runs at process exit or test teardown. `HermesSpeechStream`'s own `deinit` already cancels the websocket/tasks, and `VoiceSpeechStream.cancel()` is `@MainActor`-isolated — calling it from a nonisolated `deinit` would need an isolation escape for a path users can't reach.

**4. Availability decoupling — stale premise.** Verified against the code: every `isConnected = true` assignment happens only after the connection *and* the dashboard bridge are established, and both teardown paths (`disconnect()`, `requireSignIn()`) set `isConnected = false` synchronously in the same MainActor turn as the bridge teardown. So `isConnected == true && dashboardTicketBridge == nil` is not a reachable state — there is no mid-reconnect window where the button could be enabled without a constructible gateway (reconnect paths publish `isConnected = false` first). The sub-claim that `toggle()` doesn't guard backgrounding is also incorrect: it checks `isForegroundActive` before starting, pinned by `testSceneBackgroundCancelsActivePlayback`. Even in a hypothetical gap the outcome is a recoverable error banner (playback invariant 10), not corruption.

**5. 34×34 target + preparing a11y state — declined.** The 34×34 frame is mandated by the issue spec and matches the adjacent Copy/Branch controls exactly (ChatView.swift:1125). And the preparing state *is* exposed to VoiceOver: `isActiveMessage` covers `.preparing`, so the label switches to "Stop reading response" — alongside the ProgressView icon and accent tint — the moment preparation begins.

CI: full suite green on the fresh runner for `e02a1fe`.

kaishi00:
@coderabbitai review

kaishi00:
**Nitpick (record stream only when open succeeds): fixed in `c5e8170`.**

Agreed — the mock was recording the stream before the `openError` throw, so a failed open left an orphan in `streams` that the controller never received. Reordered so `streams.append(stream)` runs only after the open succeeds, and added `XCTAssertEqual(gateway.streams.count, 0, ...)` to `testGatewayFailureLeavesRecoverableState` to pin the invariant. The blocked-open path is unaffected (it records, then suspends), and all 18 tests pass locally.

kaishi00:
Confirmed: `makeVoiceGateway()` returns nil if `dashboardTicketBridge` or `connection` is nil, independent of `isConnected`. So the availability gate (`isConnected && isVoiceEnabled && supportsSpeech`) is genuinely decoupled from gateway constructibility. However, the prior reviewer's argument — that `isConnected == true` implies bridge and connection non-nil via assignment ordering — is plausible. But `connection` (baseURL) could theoretically be present while bridge is nil in some transition. The `readAloudGatewayIsCurrent` hardening added in e02a1fe treats nil bridge as stale, confirming the codebase considers "isConnected but bridge nil" a real transitional state worth guarding. Yet `readAloudUnavailableReason` doesn't. So there's an inconsistency: the button's *enabled* state is gated only on the snapshot, and `toggleReadAloud` will call `assignReadAloudGateway()` → `setGateway(nil)` and then `toggle(...)` which hits `runPlayback`'s nil-gateway `fail(...)`, showing "Read aloud needs a connected Hermes gateway." — despite the button appearing enabled. This is a genuine minor finding, worth noting as the PR itself hardened the sibling path for the same nil-bridge condition.

I have enough to write a defensible, accurate review. Findings are all minor/nit. No critical/major, no retain cycles, no force unwraps, no concurrency races.

## Summary
Adds a manual, per-message "Read Aloud" control on completed assistant bubbles that streams the response through the existing TTS pipeline, backed by a new `@MainActor MessageReadAloudController` with generation-guarded supersession handling, AppState lifecycle/availability wiring, and 18 tests.

## Findings

### [minor] Read-aloud availability gate is decoupled from gateway constructibility
**File:** Conduit/Services/AppState.swift:7707 (`readAloudUnavailableReason`) vs 7674 (`makeVoiceGateway`)
**Issue:** `readAloudUnavailableReason` enables the button based only on `isConnected + isVoiceEnabled + snapshot.supportsSpeech`, but `toggleReadAloud` (AppState.swift:7720-7731) can still reach `assignReadAloudGateway()` → `makeVoiceGateway()` which returns nil when `dashboardTicketBridge`/`connection` are nil even with `isConnected == true`. In that window the controller's `runPlayback` hits the `activeGateway == nil` branch and surfaces the awkward "Read aloud needs a connected Hermes gateway." error at tap time, while the button appeared enabled. This is inconsistent with the PR's own `readAloudGatewayIsCurrent` hardening (e02a1fe), which explicitly treats a nil bridge as a real transitional state.
**Suggestion:** Include a non-nil `dashboardTicketBridge`/gateway-constructibility check in `readAloudUnavailableReason` (or in `toggleReadAloud`) so the button is disabled rather than failing on the tap, matching how the nil-bridge case is already treated as stale elsewhere.

### [minor] Tests assert intermediate state via fixed 50 ms sleeps
**File:** ConduitTests/MessageReadAloudControllerTests.swift (34 occurrences, e.g. lines 25, 96, 238)
**Issue:** The suite heavily relies on `try? await Task.sleep(nanoseconds: 50_000_000)` followed by a state assertion. The checked-continuation mocks pin the interleavings, but the sleeps are wall-clock timing that can be too short on a loaded CI runner, or allow the pinned suspension to already resolve on a fast one — potential flakiness for the supersession/error-ordering tests that the PR claims are "deterministic."
**Suggestion:** For order-sensitive assertions, drive transitions through the existing explicit release points (`resumeOpenForTest`, `pauseAfterEmission`, block flags) instead of sleeps; or accept and document this as matching the `VoiceConversationControllerTests` convention.

### [nit] Playback state transitions re-render every assistant bubble
**File:** Conduit/Views/ChatView.swift:1133-1166
**Issue:** Every `AssistantBubble` holds `@ObservedObject` on the single shared controller, so each `idle → preparing → playing → idle/failed` transition recomputes the body of every assistant cell in the list (and each evaluation calls `appState.readAloudUnavailableReason`). For long chats this is a whole-list re-render per playback transition.
**Suggestion:** Accept as a stated tradeoff (consistent with the existing per-message `@EnvironmentObject` re-renders), or scope observation so only the owning bubble reacts.

### [nit] `toggle()` transitions through `.idle` between stop and start
**File:** Conduit/Voice/MessageReadAloudController.swift:81-82
**Issue:** `toggle()` calls `stop()` (which bumps generation and sets `.idle`) then `startPlayback()` (bumps again, sets `.preparing`). Each tap that starts a new message publishes two state transitions and two generation increments. Harmless (supersession still correct) but slightly wasteful.
**Suggestion:** Optionally keep `stop()`'s generator bump but avoid the intermediate `.idle` publish when immediately starting a new playback.

## Verdict
APPROVE

The core design is strong and correct: `@MainActor` isolation, `[weak self]` everywhere (no retain cycles), a generation token guarding every post-`await` shared-state mutation (including the error path), no new force unwraps, and consistent teardown of the speech stream and playback engine across stop, backgrounding, disconnect, profile switch, and gateway replacement. Findings are minor/nit only — the availability-gate coupling and test-sleep reliance are the most worthwhile follow-ups for CI robustness.

<a href="https://opencode.ai/s/2jDfq9zb">New%20session%20-%202026-08-19T01%3A51%3A06.860Z</a>
[opencode session](https://opencode.ai/s/2jDfq9zb)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32206336416)

kaishi00:
I have all the information needed. Let me note one more thing to verify — the AppState toggle double-stop and whether there's actually a subtle bug in `toggleReadAloud` when it stops a non-active then reassigns gateway. Let me trace: `toggleReadAloud` for starting a new message B while A is active. `isActiveMessage(message.id)` is false (B), so proceeds. Gateway reassign check, then `controller.toggle(B)`. Inside toggle: not active, non-empty, foreground, `stop()` (stops A), `startPlayback(B)`. Correct.

Now writing the review.

## Summary
Adds a manual per-message "Read Aloud" button to completed assistant chat bubbles that streams the response through the existing TTS pipeline, backed by a new `@MainActor MessageReadAloudController` with generation-guarded supersession handling, AppState lifecycle/availability wiring, and 18 tests.

## Findings

### [minor] Availability gate is decoupled from gateway constructibility
**File:** Conduit/Services/AppState.swift:7840 (`readAloudUnavailableReason`) vs 7803 (`makeVoiceGateway`) / 3744, 3875
**Issue:** `readAloudUnavailableReason` enables the button based only on `isConnected + isVoiceEnabled + snapshot.supportsSpeech`, but `toggleReadAloud` can still reach `assignReadAloudGateway()` → `makeVoiceGateway()` which returns nil when `dashboardTicketBridge`/`connection` are nil. `isConnected = true` is set in the reconnect path (3744) and scene-foreground path (3875) in turns where the dashboard bridge isn't explicitly re-established, so "connected but stale/nil bridge" is a real (narrow) transitional state — the very case the PR itself treats as stale in `readAloudGatewayIsCurrent` (8057) but not in the availability gate. In that window the button appears enabled and tapping surfaces the awkward "Read aloud needs a connected Hermes gateway." error from `runPlayback`'s nil-gateway branch (MessageReadAloudController.swift:126-128).
**Suggestion:** Also require a non-nil gateway/`dashboardTicketBridge` in `readAloudUnavailableReason` (or in `toggleReadAloud`) so the button is disabled at tap time instead of failing after the fact, matching the nil-bridge staleness handling already in `readAloudGatewayIsCurrent`.

### [minor] Fixed 50 ms sleeps drive the ordering-sensitive assertions
**File:** ConduitTests/MessageReadAloudControllerTests.swift (e.g. lines 25, 42, 59, 96, 238; 30+ occurrences)
**Issue:** The checked-continuation mocks pin the interleavings, but the tests then assert intermediate states after `try? await Task.sleep(nanoseconds: 50_000_000)`. On a loaded CI runner 50 ms may be too short for the task to reach the pinned suspension; on a fast runner the target suspension may already have resolved. This adds wall-clock dependence to the supersession/error-ordering tests the PR describes as deterministic.
**Suggestion:** Drive order-sensitive transitions through the existing explicit release points (`releaseAppendForTest`, `resumeOpenForTest`, block flags) instead of sleeps, or accept/document this as matching the `VoiceConversationControllerTests` convention. Low urgency since it's convention-consistent and CI is green.

### [minor] `deinit` only cancels the task, not the active speech stream
**File:** Conduit/Voice/MessageReadAloudController.swift:44
**Issue:** `deinit` cancels `playbackTask` but not `speechStream`. Teardown relies on `HermesSpeechStream.deinit` (HermesVoiceGateway.swift:125-130) to cancel the websocket/tasks when ARC releases the stream, rather than on explicit controller teardown. Works today, but it's indirect and asymmetric with `stop()`.
**Suggestion:** Cancel the stream in `deinit` (`speechStream?.cancel()`) for symmetry and explicit teardown. Low impact since the controller is a `lazy var` on AppState and effectively never deallocates in production.

### [nit] Shared `@ObservedObject` recomputes every assistant bubble
**File:** Conduit/Views/ChatView.swift:1064, 1143-1166
**Issue:** Every `AssistantBubble` holds `@ObservedObject` on the single shared controller, so each `idle → preparing → playing → idle/failed` transition recomputes the body of every assistant cell (and each evaluation calls `appState.readAloudUnavailableReason`, and reading `readAloudController.state` for all bubbles).
**Suggestion:** Accept as a stated tradeoff (consistent with the existing per-message `@EnvironmentObject` re-renders), or scope observation so only the owning bubble reacts to `state`.

### [nit] PCM enqueue failure path is untested
**File:** ConduitTests/MessageReadAloudControllerTests.swift (MockReadAloudPlayback.enqueueError, line 419)
**Issue:** `enqueueError` is declared but never exercised. `onPCM16` (MessageReadAloudController.swift:137-141) propagates `enqueuePCM16` throws up through the stream into `runPlayback`'s catch, but that specific path isn't covered by a test.
**Suggestion:** Add a test that sets `enqueueError`, delivers PCM, and asserts `.failed`, playback stopped, stream cancelled.

## Verdict
APPROVE

The design is strong and correct: `@MainActor` isolation throughout, `[weak self]` on every closure (no retain cycles), a generation token guarding every post-`await` shared-state mutation including the error path, no new force unwraps, and consistent stream/playback teardown across stop, background, disconnect, profile switch, and gateway replacement. All findings are minor/nit; the availability-gate consistency and fixed-sleep test reliance are the most worthwhile follow-ups.

<a href="https://opencode.ai/s/qDpBslPc">New%20session%20-%202026-08-19T01%3A57%3A11.917Z</a>
[opencode session](https://opencode.ai/s/qDpBslPc)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32206714625)

kaishi00:
I have enough understanding of the code to write my review. Let me verify the availability gate logic once more against the actual current code.

In the diff, `readAloudUnavailableReason` now includes the `dashboardTicketBridge != nil, connection != nil` guard (the earlier reviewer's finding #4 was addressed with the second commit `e02a1fe`). So the availability decoupling concern has been fixed. Let me verify the final state of that function is as shown in the diff — yes, lines 7841-7847 show the bridge/connection guard added.

Now let me write the review.

## Summary
Adds a manual per-message "Read Aloud" control on completed assistant chat bubbles that streams the response through the existing TTS pipeline, backed by a new `@MainActor MessageReadAloudController` with generation-guarded supersession handling, plus AppState lifecycle/availability wiring and 18 tests.

## Findings

### [minor] Ordering-sensitive tests rely on fixed 50 ms sleeps
**File:** ConduitTests/MessageReadAloudControllerTests.swift (e.g. lines 25, 42, 59, 73, 92, 238; 30+ occurrences)
**Issue:** The checked-continuation mocks (blocked open, `pauseAfterEmission`, `releaseAppendForTest`) pin interleavings, but assertions on intermediate state follow `try? await Task.sleep(nanoseconds: 50_000_000)`. On a loaded CI runner 50 ms can be too short for the task to reach its pinned suspension; on a fast runner the target suspension can already have resolved. This adds wall-clock dependence to the supersession/error-ordering tests the PR describes as deterministic.
**Suggestion:** For order-sensitive transitions, drive state forward through the existing explicit release points (`releaseAppendForTest`, `resumeOpenForTest`, block flags) rather than sleeps, or accept/document this as matching the `VoiceConversationControllerTests` convention. Low urgency; it is convention-consistent and CI is green.

### [minor] `deinit` cancels the task but not the active speech stream
**File:** Conduit/Voice/MessageReadAloudController.swift:44
**Issue:** `deinit` cancels only `playbackTask`; the active `speechStream` is left to ARC release, relying on `HermesSpeechStream.deinit` to cancel the websocket/tasks. This works but is indirect and asymmetric with `stop()` (which also calls `stream?.cancel()` plus `playback.stop()`). It also leaves `playback` undisposed when the controller deallocates mid-playback.
**Suggestion:** In `deinit`, also `speechStream?.cancel()` and `playback.stop()` for explicit, symmetric teardown. Low impact in production (the controller is a `lazy var` on `AppState` and effectively never deallocs), but it matters for the `defer { controller.stop() }` idiom and test isolation.

### [minor] Empty-content test leaves a suspended stream unsettled (test-isolation leak)
**File:** ConduitTests/MessageReadAloudControllerTests.swift (~289-308)
**Issue:** `testEmptyContentDoesNotStartOrStopPlayback` (and the empty-content path generally) relies on `guard` early-returns in `toggle()` — but the sibling `testStopDuringStreamCreation...`/blocked-open flows leave `MockReadAloudStream.append` suspended on a checked continuation that is only resumed from `cancel()`. If any test reaches the empty-content branch while a prior suspend is outstanding, or a test exits with a suspended continuation, the continuation may never be resumed and the test can hang or leak state. The production `runPlayback`'s guard on `isCurrent` after `openSpeechStream` covers supersession, but this empty-content guard happens *before* any operation/generation exists — so it can't stop an in-flight one by design, yet the test should still stop the controller (e.g. `defer { controller.stop() }`) to release its suspended continuation.
**Suggestion:** Add `defer { controller.stop() }` after controller creation in tests that leave streams suspended, so teardown cancels the stream and resumes any pending continuation.

### [minor] Shared `@ObservedObject` recomputes every assistant bubble on each transition
**File:** Conduit/Views/ChatView.swift:1064, 1133-1166
**Issue:** Every `AssistantBubble` holds `@ObservedObject` on the single shared controller, so each `idle → preparing → playing → idle/failed` transition recomputes the body of every assistant cell, and each evaluation calls `appState.readAloudUnavailableReason` (which reads `isConnected`, `isVoiceEnabled`, `voiceCapabilitySnapshot`) for all bubbles. Long chats re-render entirely per playback transition.
**Suggestion:** Accept as a stated tradeoff (consistent with existing per-message `@EnvironmentObject` re-renders), or scope observation so only the owning bubble reacts to `state`.

### [nit] 34×34 tap target below the 44 pt accessibility minimum
**File:** Conduit/Views/ChatView.swift:1151-1152
**Issue:** The button uses a fixed 34×34 frame with `.buttonStyle(.plain)` and only an image. The accessibility label is present and switches correctly between "Read response aloud" / "Stop reading response" (`.preparing` is covered by `isActiveMessage`, so state is somewhat conveyed via the label). However, the 34 pt usable target is under the 44 pt recommended hit area, and dynamic-type scaling does not grow the frame.
**Suggestion:** Optionally enlarge the hit target to ≥44 pt (e.g. `.contentShape`/padding) and/or expose the playing/preparing state via `.accessibilityValue` for clearer VoiceOver feedback. Matches the adjacent Copy/Branch controls, so note as a broader convention question rather than this PR alone.

## Verdict
APPROVE

The core design is strong and correct: `@MainActor` isolation throughout, `[weak self]` on all closures (no retain cycles), a generation token guarding every post-`await` shared-state mutation including the error path, no new force unwraps, and consistent stream/playback teardown across stop, backgrounding, disconnect, profile switch, and gateway replacement. The earlier availability-gate concern (decoupled from gateway constructibility) was already addressed in `e02a1fe` by adding the `dashboardTicketBridge != nil && connection != nil` guard to `readAloudUnavailableReason`. All remaining findings are minor/nit; the test-sleep reliance and the suspended-continuation teardown in tests are the most worthwhile follow-ups for CI robustness.

<a href="https://opencode.ai/s/5s7WbnHv">New%20session%20-%202026-08-19T02%3A16%3A45.763Z</a>
[opencode session](https://opencode.ai/s/5s7WbnHv)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32207913782)

kaishi00:
**Second-round triage (OpenCode re-reviews of c5e8170/aafe722 + CodeRabbit incrementals) — addressed in `8225a5a`**

**Fixed (3):**

1. **Read-aloud availability gate** (OpenCode, both re-reviews, with line cites) — `readAloudUnavailableReason` now also requires gateway constructibility (`dashboardTicketBridge` + `connection`, mirroring `makeVoiceGateway()`), returning "Read aloud needs a connected Hermes gateway." while the bridge/connection is torn down. The button disables during that window instead of failing at tap time, and `refreshReadAloudGateway()` now clears the gateway for it as well. The pure static `MessageReadAloudController.unavailableReason` is unchanged, so the TTS-only (no STT/mic/Apple Speech) contract stays testable in isolation.
2. **PCM enqueue failure** (flagged independently by both bots) — the mock's `enqueueError` was declared but never exercised. Added `testPCMEnqueueFailureStopsAndFails`: a mid-stream enqueue failure settles the playback engine, cancels the stream, reports the error, and leaves a tappable `.failed` state. Controller tests are now 19.
3. **Empty-content test cleanup** (CodeRabbit) — `testEmptyContentDoesNotStartOrStopPlayback` now calls `controller.stop()` before exiting, releasing the paused mock stream instead of ending with a suspended append continuation.

**Declined, with rationale:**

- **Fixed-sleep tests** — interleavings are pinned by checked continuations (the mock suspends until the test explicitly releases it); the 50 ms sleep only yields control so the MainActor task can advance. Same pattern as the existing `VoiceConversationControllerTests`, and OpenCode itself noted "low urgency since it's convention-consistent."
- **Stream cancel in `deinit`** — `VoiceSpeechStream.cancel()` is `@MainActor`-isolated and can't be called from a nonisolated `deinit` without an isolation escape. The controller is an app-lifetime lazy var in `AppState` (never deallocates in production), and `HermesSpeechStream` already cancels its own stream in `deinit`.
- **`@ObservedObject` for the controller in `AssistantBubble`** — the bubble's body already re-evaluates on every `AppState` change (it observes the environment object), and read-aloud state changes are ≤ a handful per user-initiated playback. OpenCode: "Accept as a stated tradeoff."
- **Toggle passing through `.idle`** — `stop()` and `startPlayback()` run in the same MainActor turn, so both publishes coalesce before SwiftUI renders; `.idle` is never observable and no render is wasted.
- **Return-surface test cohort** (CodeRabbit, AppState 1057–1143) — that points at PR #82's return-surface feature, not read-aloud. That state machine already ships with a focused test file (`ConduitTests/ChatReturnSurfaceTests.swift`, 617 lines, landed with #82); out of scope for this PR.

**Follow-up since this triage:** `878b5c4` closes the one remaining audio-ownership gap — starting a new manual Read Aloud now stops an in-flight VoiceConversationController operation (e.g. a Settings TTS test still speaking) before opening the manual stream, making playback mutual exclusion symmetric with `openVoiceConversation`/`runVoiceTTSTest`. Adds `AppStateReadAloudTests` (stop-before-start ordering, no voice reset on the stop-only path, A→B takeover, TTS-only availability independence) and corrects the `runVoiceTTSTest` comment that claimed a shared playback engine.

kaishi00:
The `readAloudUnavailableReason` gate now mirrors `makeVoiceGateway` exactly (both require `dashboardTicketBridge` and `connection` non-nil). Good — the earlier availability-coupling concern was addressed.

I have enough to write the review. Let me put together the final review.

## Summary

Adds a manual per-message "Read Aloud" control on completed assistant bubbles that streams the response through the existing TTS pipeline, backed by a new `@MainActor MessageReadAloudController` with generation-guarded supersession handling, AppState lifecycle/availability/mutual-exclusion wiring, and 19 controller + 4 AppState tests.

## Findings

### [minor] `deinit` cancels only the playback task, not the active speech stream
**File:** Conduit/Voice/MessageReadAloudController.swift:44
**Issue:** `deinit` does `playbackTask?.cancel()` but leaves `speechStream`/`playback` teardown to ARC, relying on `VoiceSpeechStream`'s own `deinit` to cancel the websocket. `stop()` (lines 88-98) is explicit and symmetric, but deinit isn't.
**Suggestion:** Cancel the stream in `deinit` for symmetric teardown. Note the constraint: `cancel()` is `@MainActor`-isolated so it can't be called from a nonisolated `deinit` without an isolation escape — the author's decline is defensible. Low impact in production since the controller is an app-lifetime `lazy var` on `AppState`, but it matters for clean test isolation (a suspended mock continuation on a deallocated controller is never released).

### [minor] Ordering-sensitive tests assert state after fixed 50 ms sleeps
**File:** ConduitTests/MessageReadAloudControllerTests.swift (e.g. lines 25, 42, 73, 96, 114; 30+ occurrences)
**Issue:** Checked continuations pin the interleavings, but intermediate-state assertions follow `try? await Task.sleep(nanoseconds: 50_000_000)`. On a loaded CI runner 50 ms may be too short for the task to reach its pinned suspension; on a fast runner the target suspension may already have resolved — adding wall-clock dependence to the supersession/error-ordering tests.
**Suggestion:** For order-sensitive transitions, drive state forward through the explicit release points (`releaseAppendForTest`, `resumeOpenForTest`, block flags) rather than sleeps, or accept as matching the `VoiceConversationControllerTests` convention. Low urgency since it's convention-consistent and locally green.

### [minor] Shared `@ObservedObject` recomputes every assistant bubble on each transition
**File:** Conduit/Views/ChatView.swift:1064, 1136-1168
**Issue:** Every `AssistantBubble` holds `@ObservedObject` on the single shared controller, so each `idle → preparing → playing → idle/failed` transition recomputes the body of every assistant cell, and each evaluation calls `appState.readAloudUnavailableReason` (reading `isConnected`, `isVoiceEnabled`, snapshot) for all bubbles. Long chats re-render entirely per playback transition.
**Suggestion:** Accept as a stated tradeoff (consistent with the existing per-message `@EnvironmentObject appState` re-renders, which occur far more frequently than read-aloud's user-initiated transitions), or scope observation so only the owning bubble reacts to `state`.

### [nit] Preparing state exposes no "loading" accessibility value
**File:** Conduit/Views/ChatView.swift:1140-1166
**Issue:** While preparing, the label switches to "Stop reading response" and the icon is a `ProgressView`, but VoiceOver gets no explicit state/value signaling that loading is in progress (only the action label + accent tint). The 34×34 target with `.buttonStyle(.plain)` is also below the 44 pt recommended hit area, and the frame doesn't scale with dynamic type.
**Suggestion:** Optionally add `.accessibilityValue` (e.g. "Loading"/"Playing") and enlarge the hit target via padding/`contentShape`. Matches the adjacent Copy/Branch controls, so this is a codebase-wide convention question rather than a blocker for this PR.

### [nit] `toggle()` publishes through `.idle` between stop and start
**File:** Conduit/Voice/MessageReadAloudController.swift:81-82
**Issue:** Starting a new message calls `stop()` (sets `.idle`) then `startPlayback()` (sets `.preparing`), two `@Published` changes and two generation increments per start. Harmless — supersession stays correct, and both publishes coalesce in the same MainActor turn before SwiftUI renders.
**Suggestion:** Optionally avoid the intermediate `.idle` publish when immediately starting a new playback, though this is purely cosmetic.

## Verdict
APPROVE

The core design is strong and correct: `@MainActor` isolation throughout; `[weak self]` on every closure (no retain cycles found); a generation token guarding every post-`await` shared-state mutation including the error path and the deferred-catch supersession case; no new force unwraps; consistent stream/playback teardown across stop, backgrounding, disconnect, profile switch, and gateway replacement. The availability gate now mirrors `makeVoiceGateway` (both require non-nil bridge and connection), closing the earlier decoupling gap. Mutual exclusion with the voice conversation and TTS test is enforced in both directions and pinned by AppState-level tests. All remaining findings are minor/nit; the fixed-sleep test reliance is the most worthwhile follow-up for CI robustness.

<a href="https://opencode.ai/s/Vs3MUCPr">New%20session%20-%202026-08-19T03%3A04%3A21.905Z</a>
[opencode session](https://opencode.ai/s/Vs3MUCPr)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32210745998)

PR #86, opened 2026-08-19
Omit Hermes compaction summaries from resumed transcripts

kaishi00:
## Problem

When reopening/resuming a Hermes session, persisted context-compaction summaries come back in history with a normal conversational role (usually `user`). Conduit normalized them into visible chat bubbles — and because a summary can be megabytes large, the Markdown/UI rendering pipeline stalled on it.

## Fix

Filtering lives entirely at the history-normalization boundary in `MessageNormalizer` (`Conduit/Services/HermesClient.swift`) — the single point both resume paths already use (`session.resume` and the dashboard `/api/sessions/{id}/messages` prefetch). A new `visibleContentRemovingCompaction(from:content:isToolResult:)` runs right after raw-content extraction, before role/system/runtime normalization, so summary text can never reach a `ChatMessage` or the renderer. No view-layer changes.

Detection (role-independent):

- `_compressed_summary == true` — top-level or inside the record's `metadata` object
- content starts with `[CONTEXT COMPACTION` (current) or `[CONTEXT SUMMARY]:` (legacy), case-insensitive and start-anchored, checked against a bounded 32-char head so a summary-sized payload is never copied

Merged records containing `[END OF PRIOR CONTEXT — COMPACTION SUMMARY BELOW]`:

- keep only the genuine prompt before the delimiter
- strip the `[PRIOR CONTEXT — for reference only; not a new message]` wrapper only when it leads the retained text (a copy the user quoted mid-message survives)
- drop the record if the remainder is empty or is itself an older summary (double-compaction shape)
- the summary suffix is never passed into `ChatMessage`

Tool-result rows bypass the filter entirely: compaction artifacts only ride conversational roles, while a tool's own output may legitimately print the delimiter text.

## Tests

14 new tests in `MessageNormalizerTests`: all eight shapes from the issue (metadata-flagged user summary, current/legacy prefix without metadata, assistant-role summary, merged real-content + summary, merged with no genuine content, false-positive protection, large-summary regression) plus reviewer-driven edges (double-compaction prefix, flagged merged record keeps genuine content, bare delimiter, wrapper-absent merge, quoted-wrapper preservation, delimiter-printing tool output). Local runs green: 194/194 across `MessageNormalizerTests`, `AppStateChatResumeTests`, `ChatResumePolicyTests`, `SessionPresentationCacheTests`, `DashboardTicketBridgeTests`.

## Out of scope

- Legitimate extremely large user/assistant messages still render through synchronous Markdown parsing — separate performance concern.
- Dash/case-tolerant delimiter matching deliberately declined: Hermes emits exact programmatic constants, and tolerance widens the false-positive surface.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conversation transcript cleanup by removing internal context-compaction summaries.
  * Preserved genuine user messages, tool results, and ordinary discussion content.
  * Correctly handles legacy, merged, malformed, and summary-only transcript entries.
  * Omits records that contain no visible conversational content.
  * Improved handling of large summaries and varied formatting without affecting legitimate prompts.

### DIFF
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -1438,9 +1438,19 @@ enum MessageNormalizer {
                 ?? obj["message_id"]?.descriptiveStringValue
                 ?? String(index)
 
-            let rawContent = isToolResult
+            let extractedContent = isToolResult
                 ? extractToolOutput(obj)
                 : extractContent(obj["content"] ?? obj["text"] ?? .null)
+            // Compaction bookkeeping is removed before role/system/runtime
+            // normalization so no summary text — standalone or merged onto a
+            // real prompt — can reach a ChatMessage and the renderer.
+            guard let rawContent = visibleContentRemovingCompaction(
+                from: obj,
+                content: extractedContent,
+                isToolResult: isToolResult
+            ) else {
+                continue
+            }
             let modelChange = modelChangeActivity(fromText: rawContent)
             let systemNotice = systemNoticeText(fromText: rawContent)
             if modelChange != nil || systemNotice != nil { role = .system }
@@ -1598,6 +1608,68 @@ enum MessageNormalizer {
     static func isUserCorrectionInterruptionNotice(_ text: String) -> Bool {
         text.trimmingCharacters(in: .whitespacesAndNewlines)
             .caseInsensitiveCompare("[This response was interrupted by a user correction.]") == .orderedSame
+    }
+
+    /// Hermes persists context-compaction summaries as ordinary transcript
+    /// records — usually under a conversational role, and sometimes appended
+    /// to a row that still holds the genuine user prompt. They are
+    /// model-context bookkeeping, not chat messages, and can be large enough
+    /// to stall the Markdown pipeline when mistaken for a visible bubble.
+    /// Returns the content that should remain visible, or nil when the whole
+    /// record is compaction bookkeeping and must not produce a ChatMessage.
+    /// Tool-result rows bypass the filter: compaction artifacts only ride
+    /// conversational roles, while a tool's own output may legitimately print
+    /// the delimiter text.
+    private static func visibleContentRemovingCompaction(
+        from obj: [String: AnyCodable],
+        content: String,
+        isToolResult: Bool
+    ) -> String? {
+        if isToolResult { return content }
+
+        if let range = content.range(of: compactionSummaryDelimiter) {
+            // Merged row: the genuine prompt sits before the delimiter behind
+            // an optional wrapper header; the summary suffix is never shown.
+            var visible = String(content[..<range.lowerBound])
+            // Strip the wrapper only as a leading header — a copy the user
+            // quoted mid-message belongs to their message.
+            if visible.trimmingCharacters(in: .whitespacesAndNewlines)
+                .hasPrefix(priorContextWrapperHeader),
+                let wrapperRange = visible.range(of: priorContextWrapperHeader) {
+                visible.removeSubrange(wrapperRange)
+            }
+            let trimmed = visible.trimmingCharacters(in: .whitespacesAndNewlines)
+            // After a second compaction the retained prefix can itself be an
+            // older summary rather than a genuine prompt.
+            guard !trimmed.isEmpty, !hasCompactionSummaryPrefix(trimmed) else { return nil }
+            return trimmed
+        }
+
+        let flaggedAsSummary = obj["_compressed_summary"]?.boolValue == true
+            || obj["metadata"]?.objectValue?["_compressed_summary"]?.boolValue == true
+        if flaggedAsSummary || hasCompactionSummaryPrefix(content) {
+            return nil
+        }
+        return content
+    }
+
+    private static let compactionSummaryDelimiter =
+        "[END OF PRIOR CONTEXT — COMPACTION SUMMARY BELOW]"
+
+    private static let priorContextWrapperHeader =
+        "[PRIOR CONTEXT — for reference only; not a new message]"
+
+    /// Hermes emits two generations of compaction headers, and the reload
+    /// path can drop the `_compressed_summary` flag while keeping the
+    /// recognizable text. Match case-insensitively like the other Hermes
+    /// bracketed notices, anchored at the start of the content so ordinary
+    /// discussion of "context compaction" is never mistaken for a summary.
+    /// Only a short head is lowercased — the payload being tested can be a
+    /// multi-megabyte summary that must not be copied before being dropped.
+    private static func hasCompactionSummaryPrefix(_ content: String) -> Bool {
+        let head = content.trimmingCharacters(in: .whitespacesAndNewlines).prefix(32).lowercased()
+        return head.hasPrefix("[context compaction")
+            || head.hasPrefix("[context summary]:")
     }
 
     /// Normalizes the gateway's native clarification event. Current Hermes
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -1627,9 +1627,16 @@ enum MessageNormalizer {
     ) -> String? {
         if isToolResult { return content }
 
+        // Standalone summaries announce themselves within the first bytes.
+        // Dropping them here skips the merged-delimiter search, which would
+        // otherwise walk a potentially multi-megabyte payload end-to-end.
+        if hasCompactionSummaryPrefix(content) { return nil }
+
         if let range = content.range(of: compactionSummaryDelimiter) {
             // Merged row: the genuine prompt sits before the delimiter behind
             // an optional wrapper header; the summary suffix is never shown.
+            // The flag does not preempt this branch — it marks the record as
+            // carrying a summary, not as lacking a genuine prompt.
             var visible = String(content[..<range.lowerBound])
             // Strip the wrapper only as a leading header — a copy the user
             // quoted mid-message belongs to their message.
@@ -1647,10 +1654,8 @@ enum MessageNormalizer {
 
         let flaggedAsSummary = obj["_compressed_summary"]?.boolValue == true
             || obj["metadata"]?.objectValue?["_compressed_summary"]?.boolValue == true
-        if flaggedAsSummary || hasCompactionSummaryPrefix(content) {
-            return nil
-        }
-        return content
+        // Flagged with no merged form is pure bookkeeping.
+        return flaggedAsSummary ? nil : content
     }
 
     private static let compactionSummaryDelimiter =
@@ -1664,10 +1669,15 @@ enum MessageNormalizer {
     /// recognizable text. Match case-insensitively like the other Hermes
     /// bracketed notices, anchored at the start of the content so ordinary
     /// discussion of "context compaction" is never mistaken for a summary.
-    /// Only a short head is lowercased — the payload being tested can be a
-    /// multi-megabyte summary that must not be copied before being dropped.
-    private static func hasCompactionSummaryPrefix(_ content: String) -> Bool {
-        let head = content.trimmingCharacters(in: .whitespacesAndNewlines).prefix(32).lowercased()
+    /// Classification reads only a bounded head: leading whitespace is
+    /// skipped by index and only the next 32 characters are compared, so a
+    /// multi-megabyte summary is never copied or tail-scanned to classify it.
+    static func hasCompactionSummaryPrefix(_ content: String) -> Bool {
+        var headStart = content.startIndex
+        while headStart < content.endIndex, content[headStart].isWhitespace {
+            headStart = content.index(after: headStart)
+        }
+        let head = content[headStart...].prefix(32).lowercased()
         return head.hasPrefix("[context compaction")
             || head.hasPrefix("[context summary]:")
     }
--- a/ConduitTests/MessageNormalizerTests.swift
+++ b/ConduitTests/MessageNormalizerTests.swift
@@ -1138,6 +1138,45 @@ final class MessageNormalizerTests: XCTestCase {
         XCTAssertEqual(messages.first?.content, "Still here after the summary was dropped.")
     }
 
+    func testLargeStandalonePrefixSummaryIsDroppedByPrefixAlone() {
+        // No flag and no merged delimiter anywhere: classification must rest
+        // on the prefix alone, without the full-payload delimiter search.
+        let hugeSummary = "[CONTEXT COMPACTION 12:04 — 48% of window used]\n"
+            + String(repeating: "Turn summary with no merged marker in the body. ", count: 100_000)
+        let messages = MessageNormalizer.normalizeMessages([
+            .object([
+                "id": .number(97),
+                "role": .string("user"),
+                "content": .string(hugeSummary)
+            ]),
+            .object([
+                "id": .number(98),
+                "role": .string("assistant"),
+                "content": .string("Neighbor row survives.")
+            ])
+        ])
+
+        XCTAssertEqual(messages.map(\.content), ["Neighbor row survives."])
+    }
+
+    func testCompactionPrefixHelperClassifiesFromABoundedHead() {
+        // Multi-megabyte payloads must be classifiable from their head
+        // without trimming or copying the whole string.
+        let hugeTail = String(repeating: "Summary detail line. ", count: 250_000)
+
+        XCTAssertTrue(MessageNormalizer.hasCompactionSummaryPrefix(
+            "[CONTEXT COMPACTION 12:04]\n" + hugeTail + "\n  \n"
+        ))
+        XCTAssertTrue(MessageNormalizer.hasCompactionSummaryPrefix(
+            "\n\t  [context summary]: legacy opening\n" + hugeTail
+        ))
+        XCTAssertFalse(MessageNormalizer.hasCompactionSummaryPrefix(
+            "Can you explain how context compaction works?\n" + hugeTail
+        ))
+        XCTAssertFalse(MessageNormalizer.hasCompactionSummaryPrefix("   \n\t"))
+        XCTAssertFalse(MessageNormalizer.hasCompactionSummaryPrefix(""))
+    }
+
     func testInterruptedIdenticalCorrectionDoesNotCreateASecondUserBubble() {
         let messages = MessageNormalizer.normalizeMessages([
             .object([
--- a/ConduitTests/MessageNormalizerTests.swift
+++ b/ConduitTests/MessageNormalizerTests.swift
@@ -889,6 +889,255 @@ final class MessageNormalizerTests: XCTestCase {
         XCTAssertEqual(messages[0].rawContent, original)
     }
 
+    // MARK: - Persisted context-compaction summaries
+
+    func testPersistedCompactionSummaryFlaggedByMetadataIsOmitted() {
+        let messages = MessageNormalizer.normalizeMessages([
+            .object([
+                "id": .number(81),
+                "role": .string("user"),
+                "_compressed_summary": .bool(true),
+                "content": .string("[CONTEXT COMPACTION 12:04 — 48% of window used]")
+            ]),
+            // The flag may survive only inside the record's metadata object
+            // after a persistence round-trip.
+            .object([
+                "id": .number(82),
+                "role": .string("user"),
+                "metadata": .object(["_compressed_summary": .bool(true)]),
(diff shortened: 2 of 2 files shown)


kaishi00:
Now let me analyze the new code carefully for correctness issues.

Let me focus on the key logic in `visibleContentRemovingCompaction` and `hasCompactionSummaryPrefix`.

Key observations:

1. **`hasCompactionSummaryPrefix` bounds the head to 32 chars** — but a `[CONTEXT SUMMARY]:` is 18 chars, `[CONTEXT COMPACTION` is 19 chars. 32 is fine. However, `[CONTEXT SUMMARY]` (without colon) is 17 chars and wouldn't match. The PR says it matches legacy `[CONTEXT SUMMARY]:`. OK.

2. **`hasCompactionSummaryPrefix(trimmed)` in the merged branch** — it checks the *whole* `trimmed` prefix. But note the trimming is bounded at 32 chars via `.prefix(32)`. Fine.

3. **Bug in merged branch wrapper-stripping**: It checks `hasPrefix(priorContextWrapperHeader)` on `trimmed` and then searches for `wrapperRange` in `visible` — that's fine since `visible` still contains it.

Wait, there's a subtle logic issue. When a record is flagged with `_compressed_summary == true` AND contains a delimiter with genuine content... the merged branch (delimiter check) is checked FIRST. So a flagged merged record keeps genuine content — that matches test `testMergedRecordKeepsGenuineContentEvenWhenFlagged`. Good, that's intentional.

Let me look at potential issues:

**Issue A: `hasCompactionSummaryPrefix` head length vs `[CONTEXT SUMMARY]:` with leading whitespace.** The `trimmingCharacters` trims leading whitespace then takes prefix(32). Good.

**Issue B: Potential false-positive / memory concern in delimiter search.** `content.range(of: compactionSummaryDelimiter)` scans the entire content. For a legitimate large user message that contains the delimiter string anywhere, it would truncate. But that's the intended behavior — a merged record. However, the bounded-head optimization only applies to prefix detection, not the delimiter search which does a full scan of potentially multi-MB content. That's acceptable (range(of:) is a search, not a copy).

**Issue C: The real concern — a genuine user message containing the delimiter text mid-message gets truncated.** This is arguably intended behavior per the tests. But consider: a user genuinely writes/sends a message containing `[END OF PRIOR CONTEXT — COMPACTION SUMMARY BELOW]` as part of their actual text, followed by more real content. This code would keep only the text before the delimiter and silently drop everything after. This is a potential false-positive but it's a deliberate tradeoff acknowledged in the design. Worth a minor note.

**Issue D: `range(of:)` default options.** `.range(of:)` default does a literal/bytewise search matching case and locale. The delimiter is a constant so no case sensitivity issue. Fine.

**Issue E (more significant): The wrapper-strip check inconsistency.** Look:
```swift
if visible.trimmingCharacters(in: .whitespacesAndNewlines)
    .hasPrefix(priorContextWrapperHeader),
    let wrapperRange = visible.range(of: priorContextWrapperHeader) {
    visible.removeSubrange(wrapperRange)
}
```
It checks `hasPrefix` on the trimmed string but then does `visible.range(of:)` on the untrimmed string and removes only the header text but NOT the leading whitespace (e.g., `\n\n` before the header). So after removing just the header, `visible` may still start with `"\n\n"` — but then it's trimmed at the return. Fine, `trimmed` is returned. But what if the wrapper appears and after stripping, the genuine content still has the wrapper's trailing content? Not an issue.

Wait, but there's a real bug: it removes the wrapper header but leaves the content after it. But the header check happens only if the *trimmed* visible starts with header. But `visible.range(of: priorContextWrapperHeader)` could match a *second* occurrence inside the genuine content (e.g., mid-message quoted copy) even when the leading position doesn't match... no, because `if` requires BOTH the hasPrefix AND the range(of:) to succeed. If hasPrefix is true (leading), range(of:) finds *the first* occurrence which is the leading one. Actually `range(of:)` returns the FIRST occurrence. Since hasPrefix is satisfied, the first occurrence is at the start. So it removes the correct leading header. Good.

But consider: what if the genuine content has the wrapper quoted mid-message AND the leading content is empty so hasPrefix is false? Then it won't strip. Good — that's the "quoted wrapper preserved" test.

Let me reconsider a real edge though. In the merged branch, after splitting at the delimiter, `visible` = prefix. The `range(of:)` finds first occurrence of the header. If a user's genuine content legitimately begins with something then... fine.

**Issue F — Potential correctness bug: `hasPrefix` check uses the fully-trimmed string but the removal targets the first occurrence in untrimmed `visible`.** Suppose visible = `"\n[PRIOR CONTEXT...]\nreal message"`. trimmed starts with header → hasPrefix true. range(of:) finds header at index 1. Removes it → visible = `"\n\nreal message"`. Returns trimmed = "real message". Correct.

Now, one more notable concern:

**Issue G — Double-compaction prefix: `hasCompactionSummaryPrefix(trimmed)`.** After the double-compaction, retained prefix is `[CONTEXT COMPACTION 12:04]...`. `hasCompactionSummaryPrefix` checks `.prefix(32).lowercased().hasPrefix("[context compaction")`. `[CONTEXT COMPACTION 12:04]` → `[context compaction 12:04]` prefix matches. Good. Test `testMergedRecordWhosePrefixIsItselfASummaryIsOmitted` passes.

But what about a merged record whose genuine prefix starts with `[CONTEXT SUMMARY]:`? That would be treated as a summary and dropped. That's a false-positive edge but matches design (a genuine user message can't start with that programmatic marker meaningfully).

Let me now consider the most important real issue: **the delimiter search is unbounded and case-sensitive.** The bounded-head optimization was carefully added for the prefix detector but NOT for the delimiter detector. However the delimiter `.range(of:)` doesn't copy, so no perf issue. Fine.

Let me think about whether there's an actual bug around tool-results bypass. The `guard` at line 1447 uses `visibleContentRemovingCompaction` which returns `content` for tool results. Good.

Now consider: **reasoning/thinking roles**. `isToolResult` is false for reasoning roles, so reasoning content that begins with `[CONTEXT COMPACTION` would be dropped. But reasoning content is ephemeral, unlikely. Minor.

Let me consider **the `check` on merged records flagged AND where the genuine prefix is genuinely empty but there's no summary after delimiter.** Handled.

Let me think about whether there's an actual bug worth flagging as critical/major.

Actually, here's a real concern: **`hasCompactionSummaryPrefix` is only checked on the *retained prefix* in the merged branch, and on full content in the non-merged branch. But in the non-merged branch for a *flagged* summary, the header match isn't even consulted — flagged short-circuits via `||`.** That's fine, `return nil`.

Let me reconsider the `[CONTEXT COMPACTION` detection: it's `head.hasPrefix("[context compaction")` — 32 chars head. `[CONTEXT COMPACTION` is 19 chars including `[` and space. In the strict case, "compaction" is 10 letters → `[CONTEXT COMPACTION` = 1+8+1+10=20 chars, within 32. Good.

Now the potential major issue: **case where genuine merged content ALSO contains the delimiter later but the record also has an earlier genuine prompt after more delimiter occurrences** — only the first delimiter is split. Fine.

Let me consider **iOS 15 `.prefix(32)` / String '\.prefix'** — works anywhere. Fine.

Let me evaluate the delimiter's use of the em-dash "—" count in head. Not relevant to delimiter search.

Now, is there a genuine bug? Let me re-examine the **wrapper-strip when the genuine content is only the header + nothing, i.e., record like wrapper header then immediately delimiter (no genuine text).** visible = "[PRIOR CONTEXT...]" or just the header, delimiter found, visible = header text, trim → header, hasPrefix true → strip → empty → trimmed empty → returns nil. Good.

Let me look more carefully at a subtle issue: **In the merged branch, `hasCompactionSummaryPrefix(trimmed)` is applied to the trimmed prefix. But the prefix detector only checks a 32-char head and requires exact `[context summary]:` with colon.** If the retained older summary uses the *legacy* `[CONTEXT SUMMARY]` WITHOUT colon (17 chars, matches head but the hasPrefix requires `"[context summary]:"` with colon)... `"[context summary]" < 32`, then `hasPrefix("[context summary]:")` — the head is `[context summary]` (no colon) so it doesn't match! So a legacy summary without colon retained as the prefix of a merged/double-compaction record would NOT be detected as a summary and would be shown as a user bubble. 

But that's an edge case; the current format uses the colon. And the standalone (non-merged) detection for legacy is `[context summary]:` too, so a legacy-no-colon summary would slip through in both paths. Minor inconsistency but this matches the PR's stated scope.

I think the findings are mostly minor. Let me also verify no retain cycles / concurrency issues — this is pure static func normalization, no closures retaining self, no concurrency. Good.

Let me finalize.

One more: **`content.range(of:)` on a String — fine.** 

Let me also double-check: **an assistant reasoning-bubble concern** — not relevant.

Let me reconsider severity. The main substantive thing worth a "major" or "minor" note:

1. **Unbounded full-content delimiter scan** — actually that's O(n) single pass, fine. But the bounded-32 optimization is inconsistently applied: the comment claims "so a summary-sized payload is never copied" — that's only about the prefix. The delimiter search iterates the whole thing. Not a real problem.

2. **The delimiter false-positive on legitimate user content (Issue C/D)** — a genuine user message containing the exact delimiter string followed by more text would be silently truncated. Low probability but silent data loss. This is a deliberate tradeoff. I'll note as minor.

3. **`[CONTEXT SUMMARY]` without colon legacy edge** — minor.

Let me also check: is there a concern that premature `continue` (return nil) for a flagged summary skips the `callsById` bookkeeping? Yes — if an assistant row is flagged as a summary, any tool calls on it are dropped entirely. But summaries wouldn't carry meaningful tool calls. Fine.

Actually, one more real consideration: **order of the guard vs. tool result extraction.** The guard is placed at line 1447, BEFORE the tool-result block at line 1472. But for tool results it returns content unchanged at line 1628. However — the guard `continue`s for a conversational (non-tool) role that's a summary. If such a record ALSO had tool calls that targeted earlier calls, skipping it is fine.

Wait, actually there's a subtle thing: **a role that's a summary but contains a `_compressed_summary` flag AND has tool output expected to update a prior ToolActivity card.** Rare, ignore.

I think overall this is a well-tested, focused change. I'll give APPROVE with a couple of minor notes.

Let me verify the file compiles logically — `range(of:)` returns `Range<String.Index>?`. Yes.

Let me write the review.

Let me double check `hasCompactionSummaryPrefix` head length relative to legacy `[CONTEXT SUMMARY]:`. Count: `[`(1)`CONTEXT`(7)=8, ` `(1)=9, `SUMMARY`(7)=16, `:`(1)=17, `]`? No — the string is `[CONTEXT SUMMARY]:` = `[CONTEXT SUMMARY]` (17) + `:`(1) = 18. Within 32. Good.

Also for merged wear, `trimmed` prefix check on a retained genuine content that begins with `[CONTEXT SUMMARY]:` — a genuine user message literally starting with that string. False positive, but that's the point of these markers.

I'm confident. Writing the review.## Summary
This PR adds a normalization-time filter in `MessageNormalizer.visibleContentRemovingCompaction` to strip Hermes context-compaction summary artifacts from resumed transcripts before they reach `ChatMessage`/the renderer, so multi-MB summaries no longer render as chat bubbles. It is a pure static-function change with no view/concurrency impact, well covered by 14 new tests.

## Findings

### [severity: minor] Delimiter truncation is a silent-loss false positive for genuine user content
**File:** Conduit/Services/HermesClient.swift:1630
**Issue:** `content.range(of: compactionSummaryDelimiter)` splits/truncates ANY conversational record that contains the exact delimiter string anywhere in its body, keeping only the text before it. A genuine user message that incidentally contains `[END OF PRIOR CONTEXT — COMPACTION SUMMARY BELOW]` followed by real content would be silently truncated with no visible signal. The bounded-32-char optimization (`hasCompactionSummaryPrefix`) was carefully applied to the prefix detector precisely to avoid this class of false positive, but the delimiter detector does an unbounded full-body match with no anchoring or bounded-head guard.
**Suggestion:** Consider bounding/anchoring the delimiter match (e.g., only trust it near a line-start, or only when followed by a summary prefix), or explicitly document this as an accepted false-positive tradeoff. At minimum it is asymmetric with the care taken in the prefix detector.

### [severity: minor] Legacy `[CONTEXT SUMMARY]` without colon slips through
**File:** Conduit/Services/HermesClient.swift:1672
**Issue:** `hasCompactionSummaryPrefix` matches only `"[context summary]:"` (with colon). A retained prefix (in the merged/double-compaction branch) that is an older-legacy summary written as `[CONTEXT SUMMARY] ...` without the colon will not match `head.hasPrefix("[context summary]:")`, so the record is kept and rendered as a user bubble instead of dropped.
**Suggestion:** If legacy variants without the colon exist, consider matching `[context summary]` with an optional trailing `:]`; otherwise this is consistent with the declared scope but worth a note.

### [severity: minor] Reasonable but behavior-changing: flagged merged records keep genuine content via ordering
**File:** Conduit/Services/HermesClient.swift:1630-1652
**Issue:** The delimiter branch is evaluated before the `_compressed_summary` flag check, so a `_compressed_summary = true` record containing a delimiter keeps its genuine prefix. This is intentional and tested (`testMergedRecordKeepsGenuineContentEvenWhenFlagged`), but it means the flag does not authoritatively drop a row once a delimiter is present — a slightly surprising precedence worth keeping in mind if the flag is ever expected to win.
**Suggestion:** No change required; optionally add a comment noting the intentional precedence.

## Verdict
APPROVE

The change is well-scoped, safe (pure static normalization, no closures/retain-cycles/concurrency concerns), and thoroughly tested for its edge cases. The issues above are minor, deliberate-tradeoff observations rather than blockers.

<a href="https://opencode.ai/s/u3e8umhi">New%20session%20-%202026-08-19T21%3A22%3A04.175Z</a>
[opencode session](https://opencode.ai/s/u3e8umhi)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32303407686)

kaishi00:
Thanks — triaging the three minor notes (all deliberate tradeoffs, leaving the diff as-is):

1. **Delimiter false-positive on genuine content** — intentional per the upstream contract: Hermes marks merged records by exactly this delimiter, and the handling is specified as containment-based ("if a record contains that delimiter"). A conversational message embedding the exact 49-char programmatic marker (em-dash included) mid-text is vanishingly unlikely from real users, while the shipped behavior this fixes (multi-MB summary bubbles stalling rendering on every compacted session) hits every reopened compacted session. Side note: the bounded head on the prefix detector is about avoiding a full-string *copy*, not a stricter matching bar — the delimiter search is a single `range(of:)` pass with no copy.

2. **Colon-less legacy `[CONTEXT SUMMARY]`** — the legacy format is defined as `[CONTEXT SUMMARY]:` *with* the colon; there's no known Hermes build emitting the colon-less form. Matching the documented formats exactly is deliberate (see `testOrdinaryCompactionDiscussionIsNotHidden`); loosening to an optional colon starts eating messages that merely reference the marker.

3. **Flag-vs-delimiter precedence** — agreed, intentional: the flag marks the row as *carrying* a summary, not as lacking a genuine prompt. Pinned by `testMergedRecordKeepsGenuineContentEvenWhenFlagged`, which is where a semantics change would surface if the flag ever needs to win.

kaishi00:
I now have full context. My analysis confirms:

1. **No retain cycles / concurrency issues** — this is pure static pure-function logic, no closures capturing self, no `@MainActor` concerns.

2. **The `continue` at line 1452** correctly skips a conversational summary record before it can participate in `callsById` bookkeeping. Since summaries aren't tool results (tool results bypass via `isToolResult` return at top), and this skip happens before any assistant tool-call registration, no orphaned call entries. However — worth noting: if a flagged *assistant* summary row had carried tool calls, those calls would be dropped entirely. But summaries don't carry meaningful tool calls; acceptable.

3. **Consistency of accessor usage** — `boolValue`/`objectValue` match existing patterns.

I've identified the key tradeoffs. Let me write the final review based on my full analysis of the code.

## Summary
This PR adds a normalization-time filter in `MessageNormalizer` that removes Hermes context-compaction summary artifacts (standalone, flagged, and merged-onto-prompt records) from resumed transcripts before they become `ChatMessage`s, preventing multi-MB summary text from stalling the Markdown/UI renderer. The change is a pure static-function addition to the normalization boundary with 14 new tests.

## Findings

### [severity: minor] Delimiter detection silently truncates genuine content matching the marker mid-body
**File:** Conduit/Services/HermesClient.swift:1620
**Issue:** `content.range(of: compactionSummaryDelimiter)` does an unbounded, unanchored search over the entire body. Any conversational record that contains the exact 49-char delimiter string anywhere — including a genuine user message that embeds it as a quoted/echoed string and continues with real content — is split, keeping only the text before the delimiter. Note the asymmetry: `hasCompactionSummaryPrefix` was deliberately bounded to a 32-char head and start-anchored to avoid false positives, but the delimiter path applies a similarly broad heuristic with no such guard. The quoted-wrapper case is protected (`testQuotedWrapperInsideGenuineContentIsPreserved`) but the delimiter-echo case is not.
**Suggestion:** This is a stated, deliberate tradeoff (matches the upstream contract where Hermes marks merged records by this exact delimiter). Acceptable as designed, but consider anchoring the delimiter match to a line boundary (e.g., requiring a preceding newline) to narrow the false-positive surface, since the delimiter is a programmatic constant that "would legitimately be printed" per the PR's own tool-bypass rationale — the same argument could apply to a conversational echo.

### [severity: minor] A summary row carrying tool calls would have its tool-call bookkeeping silently dropped
**File:** Conduit/Services/HermesClient.swift:1451
**Issue:** The `continue` at line 1452 bypasses the entire remainder of the loop for a flagged/recognized conversational summary — including the assistant tool-call extraction at line 1470. If an assistant-role, `_compressed_summary`-flagged row ever carried tool calls (calls whose results arrive in later rows), those call cards would never be created, and any later `tool` result row targeting them would find no `callsById` match and fall back to the name-based match, potentially attaching output to an unrelated earlier call.
**Suggestion:** Low risk in practice (compaction summaries are not expected to carry tool invocations), and consistent with the "summary is not a real message" intent. No change strictly required; a one-line comment documenting that a flagged row intentionally drops any tool-call cargo would help future maintainers.

### [severity: minor] Legacy `[CONTEXT SUMMARY]` without the trailing colon is not recognized
**File:** Conduit/Services/HermesClient.swift:1672
**Issue:** `hasCompactionSummaryPrefix` only matches `"[context summary]:"` (with colon). The merged/double-compaction branch re-classifies the retained prefix through this same helper, so a legacy prefix written `[CONTEXT SUMMARY] ...` without the colon (17 chars, well within the 32-char head but not matching the colon-suffixed predicate) would be kept as a visible bubble instead of dropped. The colon-less form is likewise missed in the standalone case.
**Suggestion:** The PR explicitly scopes this out ("Dash/case-tolerant delimiter matching deliberately declined") and the legacy format is defined as colon-terminated. Consistent with the stated contract — no change needed. Noting only for completeness.

## Verdict
APPROVE

The change is well-scoped, correct in its core logic, entirely side-effect-free (static pure function, no closures/retain cycles/concurrency exposure), and thoroughly tested across the stated transcript shapes. All findings are deliberate-tradeoff observations with negligible real-world impact, none blocking.

I have completed the code review. All findings are accurate based on direct reading of the diff and surrounding context, and the accessors and existing patterns I referenced were verified against the codebase.

<a href="https://opencode.ai/s/6IRD6RjU">New%20session%20-%202026-08-19T21%3A57%3A46.568Z</a>
[opencode session](https://opencode.ai/s/6IRD6RjU)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32306412068)

kaishi00:
Round-2 triage — all three noted as minor/no-change-required by the review itself; leaving the diff as-is:

1. **Unanchored delimiter match** — same tradeoff as round 1 (see [previous reply](https://github.com/kaishi00/hermes-conduit/pull/86#issuecomment-5348237932)): containment-based splitting is what the upstream Hermes contract specifies, and this follow-up commit is exactly what narrows the exposure — standalone rows now drop at the bounded prefix check before any delimiter scan, so the unanchored search only runs on records that aren't classifiable from their head.

2. **Tool-call cargo on a flagged summary row** — agreed low risk: compaction summaries are model-context bookkeeping and don't carry tool invocations; a flagged row being dropped wholesale (calls included) is the "not a real message" intent. Documenting-level only; happy to fold a note in if a future change ever makes summary rows carry calls.

3. **Colon-less legacy form** — same as round 1: the legacy format is defined as `[CONTEXT SUMMARY]:` with the colon; matching documented formats exactly is deliberate.

PR #87, opened 2026-08-19
Coalesce live reasoning publication to defuse scene-update watchdog

kaishi00:
## Problem

TestFlight build 135 hit reproducible **0x8BADF00D scene-update watchdog SIGKILLs** (iOS 27 beta) when the user expanded the live "Thinking" card while reasoning was still streaming.

## Root cause

`.reasoningDelta` mutated the published `messages` array on **every raw WebSocket frame**. Assistant streaming already coalesces publication (`scheduleStreamingPublish`, 33 ms) precisely to avoid main-actor saturation — reasoning bypassed that machinery entirely. With the card expanded (kept mounted across updates since PR #33's stable row identity), every delta drove the full loop: `SelectableTextView` attributed-text restyle → `UITextView` intrinsic-size invalidation → `sizeThatFits` measurement → `LazyVStack`/`ScrollView` relayout → next delta. Main thread pinned in AttributeGraph/ScrollView layout until the 10-second scene-update allowance expired.

## Fix (AppState.swift only)

Mirror the assistant-streaming discipline with one clearly-owned reasoning state machine:

- **`reasoningBuffer`** — authoritative merged trace; every raw delta merges immediately via the unchanged `mergedReasoning(existing:incoming:)` (handles incremental deltas *and* cumulative snapshots).
- **First delta mounts the card immediately** so the thinking box appears promptly; subsequent content republishes at a **50 ms cadence (~20 fps)** — reasoning card updates mutate the transcript, which is heavier than the `streamingText` projection, so the cadence is relaxed vs streaming's 33 ms.
- **`flushReasoningPublish()`** at every semantic boundary: `messageComplete` (at `scheduleStreamingCompletion` entry), `toolStart`, `toolComplete`, `messageError`, `messageInterrupted`, `messageStart`, and `finalizeStreamingCompletion` (covers deltas racing the completion drain window). No buffered text is lost at any boundary.
- **`resetReasoningStream()`** at every transcript replacement: `openSession`, resume apply, `clearActiveSessionIfNeeded`, new-conversation create, plus `prepareChatResumeForConnection` and `disconnect` (found in local review — two `messages = []` sites the original audit missed).
- **Stale publishes are inert by construction**: the publish task captures the card id at schedule time and validates it against the live card before mutating, so nothing can write into a finalized or replaced transcript.
- **Sidebar drawer parity**: opening cancels the pending publish; closing force-flushes — same suppression discipline as streaming text.
- `activeReasoningMessageId` stays stable within a segment (ThinkingCard expansion state preserved); a segment after a tool boundary still mounts a fresh card, matching pre-fix behavior.

## Tests (`AppStateReasoningStreamTests`, 12 cases)

- 75 rapid deltas → exactly **1** transcript publication (was 75), exact final content after flush
- cumulative snapshots (`abc`→`abcdefgh`) and incremental deltas — no duplication
- stable card id across coalesced publications
- `messageComplete` / `toolStart` / `messageError` / `messageInterrupted` flush pending reasoning synchronously, keep ordering, and emit **zero** post-boundary publications
- reasoning delta racing the completion drain mounts a fresh card after the assistant message (pre-fix behavior, nothing lost)
- completion-carried trace does not duplicate a streamed card
- session switch discards pending reasoning; a forced flush cannot leak it into the new transcript
- sidebar suppression + close-flush

Verified RED→GREEN on the coalescing test; full local pass on AppStateChatResume / BufferedEventDeduplication / SessionYolo / ResponseHaptic / SessionPresentationCache / TurnState suites.

## Deliberately out of scope

- `SelectableTextView` rewrite (per-frame attributed-string rebuild remains, but now runs at 20 fps max instead of per-frame)
- `ChatView`'s dual observation of `appState.messages` + `chatTranscriptRevision` (secondary amplifier, separate PR)
- ThinkingCard redesign

## Summary by CodeRabbit

* **Improvements**
  * Live reasoning updates now appear more smoothly and consistently during responses.
  * Reasoning content is flushed promptly before responses complete or transition to tools, errors, interruptions, and other conversation boundaries.
  * Opening the sidebar immediately preserves the latest reasoning content while avoiding delayed or duplicate updates.

* **Bug Fixes**
  * Prevented duplicate reasoning content after tool transitions or completed responses.
  * Pending reasoning is cleared when switching conversations, preventing content from appearing in the wrong session.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -1553,7 +1553,7 @@ final class AppState: ObservableObject {
         activeSessionTitle = "New conversation"
         messages = []
         clearStreamingText()
-        resetReasoningStream()
+        resetReasoningTurn()
         activeSessionTitlesByProfile = [:]
         pinnedSessionIDsByProfile = [:]
         pinnedSessionIDs = []
@@ -1992,7 +1992,7 @@ final class AppState: ObservableObject {
         messages = []
         setActiveSessionState(id: nil, title: "New conversation")
         clearStreamingText()
-        resetReasoningStream()
+        resetReasoningTurn()
         turnState = .idle
         defaults.removeObject(forKey: activeSessionTitlesByProfileKey)
         defaults.removeObject(forKey: pinnedSessionIDsByProfileKey)
@@ -2848,7 +2848,7 @@ final class AppState: ObservableObject {
             noteChatViewportTranscriptReplacement()
             clearStreamingText()
             activeAssistantMessageId = nil
-            resetReasoningStream()
+            resetReasoningTurn()
             turnState = .idle
             errorMessage = nil
             // A fresh conversation has no per-session override and no
@@ -3045,7 +3045,7 @@ final class AppState: ObservableObject {
             streamingText = recoveredText
         }
         activeAssistantMessageId = nil
-        resetReasoningStream()
+        resetReasoningTurn()
         applyRuntime(
             result.snapshot,
             for: result.sessionId
@@ -4285,7 +4285,7 @@ final class AppState: ObservableObject {
         messages = []
         clearStreamingText()
         activeAssistantMessageId = nil
-        resetReasoningStream()
+        resetReasoningTurn()
         turnState = .idle
         finishChatViewportTransition(generation: transitionGeneration)
     }
@@ -4363,7 +4363,7 @@ final class AppState: ObservableObject {
         messages = []
         clearStreamingText()
         activeAssistantMessageId = nil
-        resetReasoningStream()
+        resetReasoningTurn()
         updateActiveSessionTitle(for: sessionId)
         let reconciled = await reconcile(
             sessionId: sessionId,
@@ -7238,7 +7238,7 @@ final class AppState: ObservableObject {
             // A new turn ends any live reasoning card; flush first so the
             // previous segment keeps its exact buffered text.
             flushReasoningPublish()
-            resetReasoningStream()
+            resetReasoningTurn()
             setRunning(true)
             notifyVoiceAssistant(.started(sessionID: streamSessionId))
 
@@ -7266,7 +7266,7 @@ final class AppState: ObservableObject {
 
         case .messageError(_, let message):
             flushReasoningPublish()
-            resetReasoningStream()
+            resetReasoningTurn()
             clearStreamingText()
             errorMessage = message
             setRunning(false)
@@ -7274,7 +7274,7 @@ final class AppState: ObservableObject {
 
         case .messageInterrupted:
             flushReasoningPublish()
-            resetReasoningStream()
+            resetReasoningTurn()
             clearStreamingText()
             setRunning(false)
             notifyVoiceAssistant(.interrupted(sessionID: streamSessionId))
@@ -7309,8 +7309,10 @@ final class AppState: ObservableObject {
             if name.lowercased() == "clarify" { break }
             // The tool card must land after a complete reasoning card; flush
             // the coalesced buffer before the boundary reorders the transcript.
+            // A tool ends the reasoning SEGMENT only — the turn flag survives
+            // so completion-carried reasoning cannot duplicate this segment.
             flushReasoningPublish()
-            resetReasoningStream()
+            resetReasoningSegment()
             flushStreamingPartial()
             messages.append(ChatMessage(
                 id: "tool-start-\(Date().timeIntervalSince1970)",
@@ -7323,7 +7325,7 @@ final class AppState: ObservableObject {
         case .toolComplete(_, let name, let output):
             if name.lowercased() == "clarify" { break }
             flushReasoningPublish()
-            resetReasoningStream()
+            resetReasoningSegment()
             // Update the matching running tool card in place instead of
             // appending a duplicate. This keeps input + output together in
             // one chronological entry, matching how the HTTP API returns
@@ -7544,17 +7546,26 @@ final class AppState: ObservableObject {
         publishReasoningBuffer(liveCardID: activeReasoningMessageId)
     }
 
-    /// Restore the ENTIRE per-turn reasoning state machine to its initial
-    /// condition. The transcript is being replaced (session open/resume/new
-    /// conversation/disconnect), so the old card must not receive further
-    /// updates — including from an in-flight publish — and the next turn's
-    /// completion-carried reasoning must not look already-streamed.
-    private func resetReasoningStream() {
+    /// End the active reasoning SEGMENT without publishing. Used at tool
+    /// boundaries: the tool card ends the current thinking card, but the
+    /// assistant TURN continues — reasoning may resume in a fresh segment, and
+    /// reasoning already streamed this turn still counts at completion.
+    private func resetReasoningSegment() {
         reasoningPublishTask?.cancel()
         reasoningPublishTask = nil
         hasScheduledReasoningPublish = false
         reasoningBuffer = ""
         activeReasoningMessageId = nil
+    }
+
+    /// Restore the ENTIRE per-turn reasoning state machine to its initial
+    /// condition. Used when the turn or transcript itself is being replaced
+    /// (message boundaries, completion, disconnect, session switch): the old
+    /// card must not receive further updates — including from an in-flight
+    /// publish — and the next turn's completion-carried reasoning must not
+    /// look already-streamed.
+    private func resetReasoningTurn() {
+        resetReasoningSegment()
         receivedReasoningForCurrentTurn = false
     }
 
@@ -8376,7 +8387,7 @@ final class AppState: ObservableObject {
 
         clearStreamingText()
         activeAssistantMessageId = nil
-        resetReasoningStream()
+        resetReasoningTurn()
         setRunning(false)
         // Cancel any pending coalesced flush and write immediately — the
         // turn is complete so all messages are in their final state.
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -465,13 +465,19 @@ final class AppState: ObservableObject {
                 streamingPublishTask?.cancel()
                 streamingPublishTask = nil
                 hasScheduledStreamingPublish = false
-            } else if !streamingBuffer.isEmpty {
-                lastStreamingPublishBurst = max(
-                    streamingBuffer.count - streamingText.count,
-                    0
-                )
-                lastStreamingPublishDate = Date()
-                streamingText = streamingBuffer
+                reasoningPublishTask?.cancel()
+                reasoningPublishTask = nil
+                hasScheduledReasoningPublish = false
+            } else {
+                if !streamingBuffer.isEmpty {
+                    lastStreamingPublishBurst = max(
+                        streamingBuffer.count - streamingText.count,
+                        0
+                    )
+                    lastStreamingPublishDate = Date()
+                    streamingText = streamingBuffer
+                }
+                flushReasoningPublish()
             }
         }
     }
@@ -585,6 +591,13 @@ final class AppState: ObservableObject {
     private var activeAssistantMessageId: String?
     private var activeReasoningMessageId: String?
     private var receivedReasoningForCurrentTurn = false
+    /// Authoritative merged reasoning for the live thinking card. Raw deltas
+    /// merge into this buffer immediately; the published transcript is only
+    /// republished at a coalesced cadence so an expanded ThinkingCard cannot
+    /// monopolize main-actor layout work during a live reasoning stream.
+    private var reasoningBuffer = ""
+    private var reasoningPublishTask: Task<Void, Never>?
+    private var hasScheduledReasoningPublish = false
     private var streamingBuffer = ""
     /// Gateway deltas can arrive much faster than SwiftUI can lay out a chat
     /// transcript. Keep the authoritative buffer intact, but publish at a
@@ -1540,6 +1553,7 @@ final class AppState: ObservableObject {
         activeSessionTitle = "New conversation"
         messages = []
         clearStreamingText()
+        resetReasoningStream()
         activeSessionTitlesByProfile = [:]
         pinnedSessionIDsByProfile = [:]
         pinnedSessionIDs = []
@@ -1978,6 +1992,7 @@ final class AppState: ObservableObject {
         messages = []
         setActiveSessionState(id: nil, title: "New conversation")
         clearStreamingText()
+        resetReasoningStream()
         turnState = .idle
         defaults.removeObject(forKey: activeSessionTitlesByProfileKey)
         defaults.removeObject(forKey: pinnedSessionIDsByProfileKey)
@@ -2833,7 +2848,7 @@ final class AppState: ObservableObject {
             noteChatViewportTranscriptReplacement()
             clearStreamingText()
             activeAssistantMessageId = nil
-            activeReasoningMessageId = nil
+            resetReasoningStream()
             receivedReasoningForCurrentTurn = false
             turnState = .idle
             errorMessage = nil
@@ -3031,7 +3046,7 @@ final class AppState: ObservableObject {
             streamingText = recoveredText
         }
         activeAssistantMessageId = nil
-        activeReasoningMessageId = nil
+        resetReasoningStream()
         receivedReasoningForCurrentTurn = false
         applyRuntime(
             result.snapshot,
@@ -4272,7 +4287,7 @@ final class AppState: ObservableObject {
         messages = []
         clearStreamingText()
         activeAssistantMessageId = nil
-        activeReasoningMessageId = nil
+        resetReasoningStream()
         receivedReasoningForCurrentTurn = false
         turnState = .idle
         finishChatViewportTransition(generation: transitionGeneration)
@@ -4351,7 +4366,7 @@ final class AppState: ObservableObject {
         messages = []
         clearStreamingText()
         activeAssistantMessageId = nil
-        activeReasoningMessageId = nil
+        resetReasoningStream()
         updateActiveSessionTitle(for: sessionId)
         let reconciled = await reconcile(
             sessionId: sessionId,
@@ -7223,7 +7238,10 @@ final class AppState: ObservableObject {
         switch event {
         case .messageStart:
             finalizePendingStreamingCompletion()
-            activeReasoningMessageId = nil
+            // A new turn ends any live reasoning card; flush first so the
+            // previous segment keeps its exact buffered text.
+            flushReasoningPublish()
+            resetReasoningStream()
             receivedReasoningForCurrentTurn = false
             setRunning(true)
             notifyVoiceAssistant(.started(sessionID: streamSessionId))
@@ -7251,15 +7269,17 @@ final class AppState: ObservableObject {
             notifyVoiceAssistant(.completed(sessionID: streamSessionId, content: content))
 
         case .messageError(_, let message):
+            flushReasoningPublish()
+            resetReasoningStream()
             clearStreamingText()
-            activeReasoningMessageId = nil
             errorMessage = message
             setRunning(false)
             notifyVoiceAssistant(.failed(sessionID: streamSessionId, message: message))
 
         case .messageInterrupted:
+            flushReasoningPublish()
+            resetReasoningStream()
             clearStreamingText()
-            activeReasoningMessageId = nil
             setRunning(false)
             notifyVoiceAssistant(.interrupted(sessionID: streamSessionId))
 
@@ -7291,7 +7311,10 @@ final class AppState: ObservableObject {
 
         case .toolStart(_, let name, let input):
             if name.lowercased() == "clarify" { break }
-            activeReasoningMessageId = nil
+            // The tool card must land after a complete reasoning card; flush
+            // the coalesced buffer before the boundary reorders the transcript.
+            flushReasoningPublish()
+            resetReasoningStream()
             flushStreamingPartial()
             messages.append(ChatMessage(
                 id: "tool-start-\(Date().timeIntervalSince1970)",
@@ -7303,7 +7326,8 @@ final class AppState: ObservableObject {
 
         case .toolComplete(_, let name, let output):
             if name.lowercased() == "clarify" { break }
-            activeReasoningMessageId = nil
+            flushReasoningPublish()
+            resetReasoningStream()
             // Update the matching running tool card in place instead of
             // appending a duplicate. This keeps input + output together in
             // one chronological entry, matching how the HTTP API returns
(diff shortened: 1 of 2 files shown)


kaishi00:
I have enough to write the review.

## Summary
This PR coalesces live reasoning-delta publication in `AppState` so an expanded ThinkingCard no longer triggers per-frame transcript re-layout during reasoning streams (defusing the iOS 27 `0x8BADF00D` scene-update watchdog). It adds a `reasoningBuffer` state machine mirroring the existing assistant-streaming cadence (50ms/~20fps), flushes at every semantic boundary, resets at every transcript replacement, and ships 12 test cases.

## Findings

### [minor] Unstructured `Task` relies on inherited `@MainActor` isolation
**File:** Conduit/Services/AppState.swift:7511
**Issue:** `scheduleReasoningPublish` creates `Task { [weak self] in ... }` without an explicit `@MainActor`, and the post-`Task.sleep` body mutates `self.hasScheduledReasoningPublish`, `self.reasoningPublishTask`, and calls `publishReasoningBuffer`. It currently compiles because the unstructured task inherits MainActor from the enclosing `@MainActor` method, but this is implicit and fragile. Elsewhere in the same file (line 8309) the streaming-completion task uses the explicit `Task { @MainActor [weak self] in ... }` form.
**Suggestion:** Write `Task { @MainActor [weak self] in ... }` for consistency with `streamingCompletionTask` and to keep the isolation explicit under Swift 6 strict concurrency (avoids a silent data race if the enclosing method ever loses `@MainActor`).

### [minor] First reasoning delta mounts the card even while the sidebar is open
**File:** Conduit/Services/AppState.swift:7491-7502
**Issue:** `appendReasoning` mutates `messages` (appends the card) unconditionally on the first delta of a segment, bypassing the `showSidebar` suppression that both `scheduleReasoningPublish` and `scheduleStreamingPublish` honor. The stated discipline in the `showSidebar.didSet` is that live buffers are authoritative and republished only on drawer close; the first-delta mount breaks that symmetry (though the on-close `flushReasoningPublish` recovers the content). Impact is a single mutation, so low risk.
**Suggestion:** When `showSidebar` is true on the first delta, buffer without mounting and let the close-path `flushReasoningPublish` create the card, keeping suppression uniform.

### [nit] Time-based card id can collide across segment resets
**File:** Conduit/Services/AppState.swift:7492
**Issue:** `id = "reasoning-\(Date().timeIntervalSince1970)"` could collide if a fresh card is mounted twice within the same fractional second (e.g., a reasoning segment immediately after a `toolStart`/`toolComplete` flush+reset). `publishReasoningBuffer` matches on this id, so a collision would mutate the wrong card. This is a pre-existing id scheme, but the new per-tool-boundary reset makes same-second remounts more likely.
**Suggestion:** Use a monotonic counter or UUID for the reasoning card id (make the change part of this fix since it now touches these code paths).

## Verdict
APPROVE

The core mechanism is sound: `reasoningBuffer` is MainActor-confined (AppState is `@MainActor`), the publish task captures the card id and validates it before mutating so stale publishes are inert, cancellation of in-flight publishes is handled correctly on both the cancelled-during-sleep and cancelled-after-sleep paths, all five `messages = []` replacement sites are paired with `resetReasoningStream()`, and the flush/reset ordering at every semantic boundary is correct. `finalizeReasoning` cannot regress the streamed buffer because it only runs when `receivedReasoningForCurrentTurn == false` (buffer already empty). The findings above are minor robustness/consistency suggestions, not blockers.

<a href="https://opencode.ai/s/7edOmuDj">New%20session%20-%202026-08-19T21%3A26%3A22.180Z</a>
[opencode session](https://opencode.ai/s/7edOmuDj)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32303769371)

kaishi00:
Taken, @coderabbitai — good catch that every test either stayed inside the synchronous burst or force-flushed via the sidebar toggle, so a regression that scheduled the 50 ms publish but never landed it (frozen expanded card mid-stream) would have slipped through. Added `testScheduledReasoningPublishFiresWithoutForcedFlush`: feeds two chunks, waits out one cadence interval with generous CI slack, and asserts the coalesced content arrives without any forced flush or boundary event. 13/13 suite passes locally; CI re-running.

kaishi00:
OpenCode triage:

**[minor] Explicit `@MainActor` on the publish task — taken.** Applied in 1050408's follow-up commit; zero behavioral change, matches `streamingCompletionTask`'s form, and keeps the isolation explicit under Swift 6 strict concurrency. 13/13 suite passes locally.

**[minor] First delta mounts card while sidebar open — declined.** The immediate mount is an explicit requirement of this fix: the first reasoning delta must make the Thinking card appear promptly, and the boundary/flush contract (including `finalizeReasoning`'s completion-only-trace path) relies on `appendReasoning` mounting synchronously. It is exactly one publication per reasoning segment — not a cadence vector — and the drawer-close flush already recovers full content. The suppression discipline targets the repeated 50 ms republishing, which is honored.

**[nit] Timestamp card-id collision — declined, out of scope.** Pre-existing id scheme (microsecond-precision `TimeInterval`); a collision needs two segment mounts within the same `Date()` tick across a tool boundary, which is not a realistic sequence, and re-keying the id scheme would ripple into persistence/presentation-cache keying for no user-visible gain. Happy to revisit as a standalone hardening change if you disagree.

kaishi00:
I now have enough to write the review. Let me analyze the key concerns raised and check one more subtle issue: the `appendReasoning` mounts the card when `showSidebar` is true (first delta during sidebar open). Also, the `messageStart` calling `flushReasoningPublish()` then `resetReasoningStream()` — but note that `messageStart` also calls `finalizePendingStreamingCompletion()` which itself calls `resetReasoningStream()`. Let me check for a potentially subtle issue in `finalizeReasoning`'s first branch.

Let me look at the double mount concern: in `finalizeReasoning`, when `activeReasoningMessageId` is non-nil and card exists, it updates content but doesn't go through `appendReasoning`/buffer. This is fine since it's the final trace.

Now let me verify one potential bug: in `finalizeReasoning`, the first branch has redundant logic (`if text.hasPrefix(...) { content = text } else if content != text { content = text }`) — both branches do the same thing. This is a pre-existing redundancy though.

Let me now examine the most important potential real issue: **`testReasoningDeltaDuringCompletionDrainMountsFreshCardAfterAssistant`** ordering and the interaction between `flushReasoningPublish()` in `finalizeStreamingCompletion` and the fresh mount.

Let me think about the ordering logic. When the last reasoning delta arrives during drain: `appendReasoning` runs, `activeReasoningMessageId` is nil (reset at completion), so it mounts a fresh card. Then `finalizePendingStreamingCompletion → finalizeStreamingCompletion` runs `flushReasoningPublish()` which publishes the fresh card (correct content), then appends the assistant message, then `resetReasoningStream()`. This produces [reasoning, assistant, reasoning] as the test verifies. That's correct.

Now let me focus on the real deliverables and concerns.

**Concern 1: The `showSidebar`-open first-delta mount.** When `showSidebar` is true and a new reasoning segment starts while the drawer is open, `appendReasoning` mounts the card unconditionally (bypassing the `showSidebar` guard in `scheduleReasoningPublish`). This is a deliberate design decision per the author. The known-impact is a single mutation. This matches the assistant streaming behavior? Actually no — assistant streaming's `scheduleStreamingPublish` checks `!showSidebar`, but `appendReasoning`'s first-mount bypasses the guard. However the author explicitly declined fixing it. This is a genuine asymmetry in the new code worth noting, though low-risk.

**Concern 2: potential redundant condition in `finalizeReasoning`** — pre-existing.

Let me now verify one more thing: potential data race or the `Task { @MainActor }` — the author already applied that fix in commit 570bc5d. Good.

Let me now consider whether there's an actual defect. One thing to check: `appendReasoning` merges into `reasoningBuffer` and mounts with `content: reasoningBuffer`. But there's an asymmetry: `scheduleReasoningPublish` is only called when `activeReasoningMessageId != nil`. What if the first delta mounts the card, but then the *second* delta arrives while `showSidebar` is true? Then `scheduleReasoningPublish` returns immediately (guard `!showSidebar`), merges into buffer. When sidebar closes, `flushReasoningPublish()` publishes. Correct.

Let me check one subtle thing about `testScheduledReasoningPublishFiresWithoutForcedFlush` measurement — but I don't need to run tests.

Let me also double-check: is there a scenario where `reasoningPublishTask` is reassigned (a second task scheduled) while the first one's `guard hasScheduledReasoningPublish` correctly prevents it. Yes — `hasScheduledReasoningPublish` guards re-scheduling. Good.

Now let me consider the concurrency of `Task.sleep` cancellation: In `scheduleReasoningPublish`, the `do/catch` catches cancellation from `Task.sleep` and returns. But `flushReasoningPublish`/`resetReasoningStream` call `reasoningPublishTask?.cancel()` — after the task completes and sets `reasoningPublishTask = nil`, a subsequent `cancel()` on the stale reference is harmless. Good.

Potential subtle issue: In the publish task, after `Task.sleep` succeeds but before the guard, if `resetReasoningStream` set `hasScheduledReasoningPublish = false` and `reasoningPublishTask = nil` and the task runs the body, it re-sets `hasScheduledReasoningPublish = false` (no-op) and `reasoningPublishTask = nil`, then calls `publishReasoningBuffer(liveCardID: cardID)`. If a new card was mounted with a different id, `cardID != activeReasoningMessageId` → inert. If a *new* reasoning segment already scheduled a new publish with a new card id and the old task runs late... but `hasScheduledReasoningPublish` would be guarding. Actually there's a subtle interleaving: old task is cancelled but `Task.sleep` already returned (not cancelled), then new segment's `scheduleReasoningPublish` sets `hasScheduledReasoningPublish=true` and `reasoningPublishTask = newTask`. Then old task body runs: sets `hasScheduledReasoningPublish = false` (wrongly clearing the new flag) and `reasoningPublishTask = nil` (clearing the new task reference), then `publishReasoningBuffer(liveCardID: oldCardID)` — inert if ids differ, but it wrongly cleared the new segment's scheduling state, so the new segment's coalesced content may never auto-publish (only on next delta or flush). This is a genuine but narrow race. However, it requires the old task's `Task.sleep` to have returned (not cancelled) at the exact same time a reset + new segment happens. Since `resetReasoningStream` cancels the task, and `Task.sleep` throws cancellation immediately when cancelled before/during sleep, the body's `guard !Task.isCancelled` also guards. The window where `Task.sleep` completed but the body hasn't run yet while a reset occurs is tiny but real. Still, the `guard !Task.isCancelled` catches the common case. This is a very minor theoretical race — not worth elevating given MainActor confinement makes both run on the same actor (they can't truly interleave mid-instruction; the reset runs to completion, then the body runs). Since both are MainActor, they're serialized. The body would run after reset completed, so `hasScheduledReasoningPublish` from a *later* scheduleReasoningPublish call... but scheduleReasoningPublish is also MainActor. Order of execution on the main actor: if resetReasoningStream runs before old-task-body, and old-task-body was already past sleep and queued... MainActor serializes. The body sets hasScheduledReasoningPublish=false only if it runs. Since everything is MainActor and the flag is set true by schedule then false by the body, there is proper serialization and no torn state — the only risk is logical: old body clearing the flag that a newer schedule set. But both are main-actor serialized, so if old body runs after newer schedule, old body clears it → then new task is also scheduled and will fire. Actually new task would still fire its publish at its own sleep. And hasScheduledReasoningPublish being cleared is fine because the new task already scheduled. So harmless on main actor. Good — no real issue.

I'm confident. Let me write the review.

Key findings:
1. (minor) First-delta card mount bypasses sidebar suppression — asymmetry (author declined).
2. (nit) Pre-existing redundancy in `finalizeReasoning` first branch.
3. (nit) Timestamp id collision (author declined, out of scope).

Also worth noting the positive: the `@MainActor` fix was applied; tests are strong. But there's a gap worth noting: tests that check coalescing count rely on synchronous bursts; the `testScheduledReasoningPublishFiresWithoutForcedFlush` was added.

Let me also double-check whether `flushReasoningPublish()` being called in `showSidebar.didSet` else-branch could double-publish (it's called, and then publishes if buffer differs from content). That's fine.

One more: `messageStart` calls `flushReasoningPublish()` then `resetReasoningStream()`. But `finalizePendingStreamingCompletion()` (called at line 7240, before flush) itself calls `finalizeStreamingCompletion` → `flushReasoningPublish()` + `resetReasoningStream()`. So by the time line 7243's `flushReasoningPublish()` runs, the reasoning state is already reset (buffer empty, id nil). So line 7243-7244 are effectively no-ops in the common case where a completion was pending. That's just redundant but harmless. Actually wait — `finalizePendingStreamingCompletion` resets the stream only if there's a pending completion. If no pending completion (normal new turn after prior messageComplete already drained), then the reasoning card from the prior turn was already flushed at completion. Either way line 7243/7244 is mostly defensive. Fine.

Now, is there a case where reasoningBuffer could be flushed *twice* causing content issues? `publishReasoningBuffer` updates `messages[index].content` to `reasoningBuffer` but doesn't clear `reasoningBuffer`. So subsequent flushes republish the same buffer (idempotent due to `!=` guard). Fine.

I'm confident in the review. Let me write it.

## Summary
This PR coalesces live reasoning-delta publication in `AppState` so an expanded ThinkingCard no longer triggers per-frame transcript re-layout during reasoning streams, defusing the iOS 27 `0x8BADF00D` scene-update watchdog. It introduces a `reasoningBuffer` state machine (immediate first-delta mount + 50ms/20fps coalesced republish), flushes and resets reasoning at every semantic/transcript/session boundary, and adds 13 test cases.

## Findings

### [minor] First-delta card mount bypasses sidebar suppression
**File:** Conduit/Services/AppState.swift:7491-7502
**Issue:** `appendReasoning` mounts the reasoning card unconditionally (`messages.append`) on the first delta of a segment, while `scheduleReasoningPublish` (and its streaming twin) gates on `showSidebar`. So a new reasoning segment starting while the sidebar drawer is open still mutates the transcript, breaking the stated "suppressed while open, flush on close" symmetry. The close-path `flushReasoningPublish` manages to recover the content, so impact is a single mutation, but it's an inconsistency in the new code.
**Suggestion:** When `showSidebar` is true on the first delta, buffer without mounting and let the drawer-close `flushReasoningPublish()` create the card, keeping suppression uniform.

### [nit] Redundant branch in `finalizeReasoning`
**File:** Conduit/Services/AppState.swift:7578-7582
**Issue:** The two branches do the same thing: `if text.hasPrefix(...) { content = text } else if content != text { content = text }` both set `content = text`. The conditional structure is confusing and suggests an intended difference (e.g., only replace when the trace is a strict superset) that isn't implemented.
**Suggestion:** Collapse to a single `messages[index].content = text` (or implement the intended superset check) and add a brief comment.

### [nit] Time-based card id can collide across segment resets
**File:** Conduit/Services/AppState.swift:7492
**Issue:** `id = "reasoning-\(Date().timeIntervalSince1970)"` can collide if two reasoning segments are mounted within the same fractional `Date()` tick (e.g., a fresh segment immediately after a same-second `toolStart`/`toolComplete` flush+reset). Because `publishReasoningBuffer` matches on this id, a collision would mutate the wrong card. Pre-existing scheme, but the new per-tool-boundary reset increases same-second remount likelihood.
**Suggestion:** Use a monotonic counter or UUID for card ids.

## Verdict
APPROVE

The mechanism is sound and well-tested: `reasoningBuffer` is MainActor-confined, stale publish tasks are inert via the captured card-id check, cancellation is handled on both pre- and post-sleep paths, every `messages = []` replacement site pairs with `resetReasoningStream()`, and flush/reset ordering at each boundary is correct (`finalizeReasoning` cannot regress a streamed card because it only runs when the buffer is already empty). The `@MainActor` isolation on the publish task is explicit. All findings above are minor robustness/consistency nits, not blockers; two were already consciously declined by the author for good reasons.

<a href="https://opencode.ai/s/dEQqKrdi">New%20session%20-%202026-08-19T21%3A44%3A15.107Z</a>
[opencode session](https://opencode.ai/s/dEQqKrdi)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32305225215)

kaishi00:
Deployment target is iOS 17, so `Task.sleep(for:)` is fine. The codebase already uses it elsewhere.

Let me verify the `messageStart` flow concern about double flush is truly a no-op in normal cases, and check whether the `finalizePendingStreamingCompletion` inside `reasoningDelta` could cause an issue where the reasoning buffer state is reset mid-stream and content lost.

Actually, let me focus on a specific edge case in the `reasoningDelta` case combined with `scheduleStreamingCompletion`. When a `messageComplete` arrives while reasoning is still streaming, `scheduleStreamingCompletion` calls `flushReasoningPublish()` (line 8269) and sets a pending completion (does NOT reset the reasoning stream). Then if another `reasoningDelta` arrives during the drain window, the `reasoningDelta` case calls `finalizePendingStreamingCompletion()` which drains the completion and calls `resetReasoningStream()`, then `appendReasoning` mounts a fresh card B. Test `testReasoningDeltaDuringCompletionDrainMountsFreshCardAfterAssistant` covers this. Fine.

Now, there's a subtle loss scenario worth examining: In `messageComplete` → `scheduleStreamingCompletion` → flushReasoningPublish. But note that `scheduleStreamingCompletion` does NOT reset the reasoning stream, so `activeReasoningMessageId` and `reasoningBuffer` stay. Then in the `reasoningDelta` case during drain, `finalizePendingStreamingCompletion` (which resets via finalizeStreamingCompletion) runs first, THEN appendReasoning. So the drained (previous) reasoning is flushed and reset, then the new delta re-mounts. Good, matches design.

But consider: `messageComplete` followed immediately by `messageStart` (no more deltas). messageStart → `finalizePendingStreamingCompletion` → finalizeStreamingCompletion → flush (no-op) + reset. Then back in messageStart at line 7240 `flushReasoningPublish()` (no-op) + `resetReasoningStream()` (no-op). The reasoning card was already flushed at scheduleStreamingCompletion and preserved (card stays in messages). Good.

I've verified the core logic is sound. Let me now write the review.

Let me also double check one real concern CodeRabbit raised that was NOT fully addressed: the `mergedReasoning` incremental/cumulative ambiguity. Since this is pre-existing and not changed, but the PR now makes it apply to a persistent buffer, the risk of wrong merge on an incremental delta that starts with the buffer remains. The test only covers the clean cases ("abc"→"abcdef"→"abcdefgh" and incremental non-overlapping). The ambiguous case (incremental chunk that happens to start with current buffer, e.g. "a" then "ab") is not covered and would merge to "ab" instead of "aab". This is a legitimate gap, though pre-existing behavior. Worth flagging as minor.

Let me write my review now.

## Summary
This PR coalesces live reasoning-delta publication in `AppState` so an expanded ThinkingCard no longer drives per-frame transcript relayout during reasoning streams (defusing the iOS 27 `0x8BADF00D` scene-update watchdog). It introduces a MainActor-confined `reasoningBuffer` state machine: first delta mounts the card immediately, subsequent deltas republish at a 50ms/~20fps cadence, flush at every semantic boundary, and reset at every transcript/session replacement, plus 13 test cases.

## Findings

### [minor] `appendReasoning` first-delta mount bypasses sidebar suppression
**File:** Conduit/Services/AppState.swift:7487-7498
**Issue:** `appendReasoning` mounts the reasoning card unconditionally on the first delta of a segment even when `showSidebar` is true, while `scheduleReasoningPublish` (line 7503) gates on `!showSidebar` — an asymmetry with the streaming twin and with the `showSidebar.didSet` discipline that live buffers are authoritative and republished only on drawer close. Impact: a reasoning segment starting while the drawer is open still mutates the published transcript once (the close-flush recovers the content, so no loss).
**Suggestion:** When `showSidebar` is true on the first delta of a segment, buffer without mounting and let the drawer-close `flushReasoningPublish()` create the card, keeping suppression uniform.

### [minor] Uncovered ambiguous incremental-merge case in `mergedReasoning`
**File:** Conduit/Services/AppState.swift:7561-7567
**Issue:** The prefix heuristic can't distinguish cumulative snapshots from incremental deltas. If an incremental delta chunk happens to start with the current buffer (e.g. buffer `"a"` then incremental `"ab"`), `incoming.hasPrefix(existing)` matches and replaces to `"ab"`, losing content that should be `"aab"`. This logic is pre-existing and unchanged, but the PR now routes all reasoning through the persistent `reasoningBuffer`, widening its exposure. The new tests only cover clean non-overlapping cases.
**Suggestion:** Carry the event mode/offset from `StreamEventParser` into the merge, and add a regression case for the ambiguous incremental-chunk-overlapping-buffer scenario.

### [nit] Redundant branch in `finalizeReasoning`
**File:** Conduit/Services/AppState.swift:7577-7581
**Issue:** The two branches both assign `messages[index].content = text` (`if hasPrefix` → replace; `else if !=` → replace), so the first condition is dead and the intent (e.g., only replace when a superset) is unclear.
**Suggestion:** Collapse to a single `messages[index].content = text` or implement the intended superset check.

### [nit] Time-based card id can collide after tool-boundary resets
**File:** Conduit/Services/AppState.swift:7488
**Issue:** `"reasoning-\(Date().timeIntervalSince1970)"` can collide if two reasoning segments mount within the same `Date()` tick (a fresh segment immediately after the new per-tool-boundary flush+reset). `publishReasoningBuffer` matches on this id, so a collision would mutate the wrong card. Pre-existing scheme, but the new per-boundary resets increase same-second remount likelihood.
**Suggestion:** Use a monotonic counter or UUID for reasoning card ids.

## Verdict
APPROVE

The mechanism is sound and well-tested. `reasoningBuffer` and all reasoning state are MainActor-confined; the publish task captures the card id and validates it against the live card before mutating, so stale publishes are structurally inert (the `guard !Task.isCancelled, let self` + MainActor serialization makes the reset/new-segment interleaving correct — the body between the guard and final line has no suspension points, so no torn state). Cancellation is handled on both pre- and post-`Task.sleep` paths, every `messages = []` replacement site is paired with `resetReasoningStream()`, and the final commit correctly resets `receivedReasoningForCurrentTurn` in `resetReasoningStream()` (addressing the completion-trace-duplication concern). `finalizeReasoning` cannot regress a streamed card because `finalizeStreamingCompletion` only calls it when `receivedReasoningForCurrentTurn == false`. All findings are minor robustness/consistency items, not blockers.

<a href="https://opencode.ai/s/rV2bW0JA">New%20session%20-%202026-08-19T22%3A19%3A45.697Z</a>
[opencode session](https://opencode.ai/s/rV2bW0JA)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32308210110)

kaishi00:
Second-round triage (turn-flag fix + main merge):

**[minor] Ambiguous incremental-merge case in `mergedReasoning` — declined, explicitly out of scope.** Correct analysis, but this PR's mandate is to preserve the pre-existing merge heuristic unchanged; `StreamEventParser` erases the upstream cumulative-vs-incremental distinction, so carrying mode/offset into the merge is a protocol-normalization change that deserves its own PR (and a gateway-side look at what the protocol can actually express). Adding a regression test for the ambiguous case now would just pin behavior we intend to replace.

**[nit] Redundant branch in `finalizeReasoning` — declined, cleanup only.** Both branches assign `content = text`, so the condition is dead but behaviorally identical and harmless; collapsing it is cosmetic churn outside this PR's correctness scope.

**[minor] first-delta mount while sidebar open / [nit] timestamp card id — declined** in the previous round (see triage above): the immediate mount is an explicit requirement of this fix and a single publication per segment, not a cadence vector; the id scheme is pre-existing with microsecond-precision collision odds, and re-keying ripples into persistence.

All three checks green on 82a7026 (full suite 11m1s, CodeRabbit, review). PR is ready for human review.

kaishi00:
`handleStreamEvent` is called inside `Task { @MainActor }`, so all reasoning state mutations happen on the MainActor. Combined with the explicit `@MainActor` on the publish task and `Task.sleep` cancellation, there is no data race. The concurrency is sound.

I have completed a thorough independent verification of the code. My findings align with the well-handled prior review rounds, and the core logic is correct. Let me write the final review emphasizing the one genuinely novel additional observation (the drain-window second-card + odd ordering tradeoff) and confirm the strong points.

## Summary
This PR coalesces live reasoning-delta publication in `AppState` so an expanded ThinkingCard no longer drives per-frame transcript relayout during reasoning streams, defusing the iOS 27 `0x8BADF00D` scene-update watchdog. It adds a MainActor-confined `reasoningBuffer` state machine: the first delta of a segment mounts the card immediately, subsequent deltas republish at a 50 ms/~20 fps cadence (mirroring the existing assistant-streaming discipline), with synchronous flushes at every semantic boundary and resets at every transcript/session replacement, plus 13 test cases.

## Findings

### [minor] Drain-window reasoning delta still drops coalescing and mounts a stray card after the answer
**File:** Conduit/Services/AppState.swift:7252-7255, 8350-8390
**Issue:** When a `reasoningDelta` arrives during the completion drain window, `finalizePendingStreamingCompletion()` → `finalizeStreamingCompletion()` runs `resetReasoningTurn()` (clearing buffer + card id), then the `reasoningDelta` case re-appends, so the buffered delta mounts a *fresh* card *after* the assistant message, yielding `[reasoning, assistant, reasoning]`. This targets exactly the heavy-layout path the PR exists to eliminate and produces a card that renders below the final answer. The test `testReasoningDeltaDuringCompletionDrainMountsFreshCardAfterAssistant` pins this as "pre-fix behavior," but the pre-fix behavior was precisely the per-delta overload this PR fixes — preserving it here contradicts the fix's intent.
**Suggestion:** Consider routing the drained delta into the existing (pre-completion) card A instead of resetting on this path — merge `late thought` into card A's content before the assistant message is appended — so the coalescing guarantee holds even on the completion race and avoids the below-answer card. If preserving exact pre-fix ordering is a hard requirement, at least document the tradeoff in the PR.

### [minor] First-delta card mount bypasses sidebar suppression
**File:** Conduit/Services/AppState.swift:7489-7501
**Issue:** `appendReasoning` mounts the reasoning card unconditionally on the first delta of a segment (`messages.append`), while `scheduleReasoningPublish` gates on `!showSidebar`. So a segment starting while the sidebar drawer is open still mutates the transcript — an asymmetry the author consciously declined. It is a single mutation per segment (not a cadence vector), and the close-path flush recovers content, so risk is low.
**Suggestion:** When `showSidebar` is true on the first delta, buffer without mounting and let the drawer-close `flushReasoningPublish()` create the card, keeping suppression uniform (flagged so it stays a conscious decision).

### [nit] Ambiguous incremental-merge case in `mergedReasoning` is now the live path
**File:** Conduit/Services/AppState.swift:7572-7578
**Issue:** The prefix heuristic cannot distinguish cumulative snapshots from incremental deltas. If an incremental chunk begins with the current buffer (e.g. buffer `"a"` then incremental `"ab"`), `incoming.hasPrefix(existing)` matches and replaces to `"ab"`, losing content that should be `"aab"`. Pre-existing logic, but the PR now routes all reasoning through a persistent `reasoningBuffer`, widening exposure. Correctly declined as out of scope, but worth a follow-up with `StreamEventParser` mode preservation.

### [nit] Time-based card id can collide after tool-boundary resets
**File:** Conduit/Services/AppState.swift:7490
**Issue:** `"reasoning-\(Date().timeIntervalSince1970)"` can collide if two segments mount within the same `Date()` tick across a tool-boundary reset; `publishReasoningBuffer` matches on this id, so a collision would mutate the wrong card. Pre-existing scheme, low probability. Declined as out of scope.

## Verdict
APPROVE

I independently verified the state machine and found it sound. `handleStreamEvent` always runs on `Task { @MainActor }` (AppState.swift:2028), the publish task is explicitly `@MainActor` with `Task.sleep` cancellation handled on both pre- and post-sleep paths, and every one of the six `messages = []` sites is paired with `resetReasoningTurn()`. Stale publishes are structurally inert via the captured-card-id check, `flushReasoningPublish`/`resetReasoningSegment*` at each boundary preserve ordering (verified the tool-boundary segment-vs-turn distinction keeps `receivedReasoningForCurrentTurn` intact so completion traces don't duplicate), and `finalizeReasoning` cannot regress a streamed card because it only runs when `receivedReasoningForCurrentTurn == false`. Prior review findings (explicit `@MainActor`, `resetReasoningStream` owning the turn flag, scheduled-publish and session-switch async test coverage) are all addressed. The remaining items are minor/cosmetic and mostly consciously declined; none block merge.

Verify the drain-window finding above against the test before deciding — it's a legitimate functional tradeoff, not a defect in the current implementation's correctness.

<a href="https://opencode.ai/s/l1bgEmk6">New%20session%20-%202026-08-19T22%3A52%3A36.855Z</a>
[opencode session](https://opencode.ai/s/l1bgEmk6)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32310714766)

kaishi00:
Third-round triage (segment/turn split, head 4c1235b):

**[minor] Drain-window delta mounting a fresh card after the assistant — declined, documented tradeoff.** This conflates the ordering behavior with the publication storm the PR eliminates. Pre-fix, a delta racing the drain also finalized the pending completion and mounted a fresh card after the assistant message — that ordering is what the test pins as preserved, not the per-frame overload (a single late delta is one publication, not a cadence vector). Routing the drained delta back into card A would require re-ordering finalized transcript content and new state-machine surgery in the completion path for a degenerate gateway race that loses no text and creates no duplicate. The tradeoff is documented in the test name and PR description; happy to revisit if a real gateway exhibits post-completion reasoning regularly.

**Sidebar first-delta mount, mergedReasoning ambiguity, timestamp id — unchanged declines** (see prior triage rounds): spec-mandated immediate mount / explicitly out of scope per the task brief / pre-existing scheme with effectively zero collision odds.

Segment-vs-turn split verified by three local reviewers + this review: the flag now has exactly one producer (.reasoningDelta) and one turn-scoped clearer (resetReasoningTurn), and survives tool boundaries. 16-test suite, full CI, all checks green on 4c1235b.

PR #88, opened 2026-08-20
perf(chat): bound rendering work for very large messages

kaishi00:
## Origin

A legitimate but very large user or assistant message (a huge pasted prompt, a huge streamed response, a huge response loaded from history) could freeze the UI and risk an iOS watchdog kill: the Markdown pipeline performed whole-document parsing, attributed-string construction, and UIKit layout synchronously — and while streaming, repeated all of it at 30 fps.

Two adjacent problems are already solved elsewhere and deliberately untouched here: compaction summaries (#86) and reasoning-publication coalescing (#87).

## Measured bottlenecks first (iPhone 17 simulator, best-of-N, ms @1MB)

| Stage | plain | manyMarkdown | hugeCode | mixed | single giant paragraph |
|---|---|---|---|---|---|
| `MarkdownParser.parseDocument` | 144 | 192 | 15 | 158 | 82 |
| `MarkdownSelectionFormatter` | 204 | 692 | — | — | 157 |
| UITextView whole-doc layout | 476 | 958 | — | — | **2,422** |
| `SyntaxHighlighter` | — | — | **1,516** | — | — |

- Dominant: whole-document TextKit layout of the single selectable string; giant-code highlighting; per-fragment Foundation parses at ~10K blocks.
- Streaming multiplied every stage by 30 fps → **0.8–3.3 s per frame** at 1 MB; 250 KB already paid 220–800 ms/frame.
- Measured **non**-problems (so left alone): render-cache key construction with the embedded source costs 0.1 ms at 1 MB; fade copies ≤0.7 ms.

## Architecture

The path fork is centralized in `MarkdownText.body` — no length checks sprinkled through `ChatView`. Below threshold, the ordinary fast cached path runs **verbatim** (the previous body moved to `normalBody`). At/above it, `LargeMarkdownDocumentView`:

```
collapsed:  ≤16 KB preview rendered synchronously through the ordinary
            fast path + "This message is very large" banner
expanded:   one off-main structural parse + chunk plan, then progressive
            rendering of ~8 KB chunks (memoized attributed strings, one
            SelectableTextView per flow chunk); rich blocks reuse the
            ordinary block renderer with their original selection ids
```

Chunking (not a collapsed preview alone, not a second Markdown engine) because it reuses the existing parser/formatter/block views and bounds *every* stage — expand never re-enters a whole-document synchronous path.

## Policy (all thresholds utf8.count, measurement-justified)

- **100 KB** document threshold: ~50 KB passes cost 45–160 ms (today's tolerable high end); 250 KB enters 220 ms+/stage freeze territory.
- **16 KB** preview (byte-precise, grapheme-safe cut): ≤ ~35 ms synchronous first paint — session load and history scroll never parse a 1 MB source (asserted by parse-count tests).
- **8 KB** flow chunks: ~8 ms parse+format+layout each; a single oversized block (a 1 MB paragraph measured 2.4 s of layout) is split at word/item boundaries.
- **32 KB** code threshold, **800-line** slices: highlighting measured 67 ms @50 KB and 1.5 s @1 MB, so large code renders as slices with highlighting computed off-main and swapped in; Copy always uses the complete source.
- **480-chunk mounted cap**: documents made of pathological tiny blocks (a mini-table after every sentence → thousands of chunks) get a "Continue reading" control instead of unbounded mounting — plain-VStack `onAppear` is insertion-gated, so uncapped window growth cascades.

## Streaming above the threshold

O(delta) per tick instead of O(document): an append-only buffer, an incremental fence-safe boundary scanner (never splits fences/math/directives; resumable across arbitrary delta shapes), a stable prefix promoted into immutable chunks rendered once through the cached path, and only an ~8 KB tail re-parsing per frame. The character-fade stays on the tail; sub-threshold streaming code is unchanged.

## Cache

Unchanged. Key-build cost measured negligible, and large documents never enter `MarkdownRenderCache`, so no giant entries or eviction pressure are created. The hash-key redesign suggested in the task brief is declined with that evidence.

## What remains on the MainActor

Per-chunk attributed construction and TextKit layout (bounded ~8 KB each) and SwiftUI diffing. Off-main: the one-time expand parse (cancellation-safe via `.task(id:)` + source-identity guard), large-code line splitting, and syntax highlighting.

## Tests

25 new deterministic tests in `MarkdownLargeDocumentTests.swift` — work-count/bound assertions (a DEBUG `MarkdownParser.parseSourceSizes` counter proves the whole source parses exactly once on expand and never during large streaming), planner content-preservation round-trips, fence-safety, promotion invariants, segment-plan slicing parity, Dynamic Type memoization, staleness on source swap, CJK byte-budget precision. No wall-clock thresholds.

Full suite: **848 tests, 0 failures** (`ComposerPasteTextViewTests` skipped locally — the known flaky simulator XPC stall; CI covers it).

## Local review (2 rounds, 6 blockers found → fixed → regression-tested)

1. Stale `String.Index` reused across `largeAccumulated` reassignment → re-derived from the tracked UTF-8 offset.
2. Tautological stale-highlight guard in code slices → snapshot comparison.
3. + 4. Two chunk-window overflows (Continue button stale capture; unclamped initial batch when a large doc yields <12 chunks) → clamped via `nextWindowCount` / live-state reads.
5. Character-based preview window tripling the byte budget for CJK → byte-precise grapheme-safe cut.
6. Scanner left an unclosed-construct state for streams ending without a trailing newline → `finish()`.

## Deliberate tradeoffs

- Settled large messages open collapsed (including right after a stream settles); the full content is one tap away and Copy/branch/read-aloud/persistence always operate on the complete `ChatMessage.content`.
- Reference-style links render literally *while streaming* (definitions elsewhere in the stream aren't visible to the tail); they resolve message-wide the moment the message settles.
- Cross-chunk selection rides the shared coordinator (same architecture as cross-block selection today); selection *across* chunk boundaries is coordinator-driven rather than seamless. Normal-message selection is untouched.
- Streaming identity uses `utf8.count + hashValue` (per-process stable; collision risk astronomically small).

## Summary by CodeRabbit

* **New Features**
  * Added efficient rendering for large Markdown documents with bounded previews, progressive expansion, and incremental loading.
  * Added bounded handling for oversized tables, code, callouts, columns, math, and Mermaid content.
  * Improved streaming of long Markdown responses with safe chunk boundaries and reduced-motion support.
  * Preserved rich content, selection, references, copying, list numbering, and Dynamic Type support.

* **Bug Fixes**
  * Reduced memory use and rendering work while preventing content overflow.
  * Improved handling of long content, line endings, graphemes, cancellation, and source updates.

### DIFF
--- a/Conduit/Views/Components/MarkdownLargeDocument.swift
+++ b/Conduit/Views/Components/MarkdownLargeDocument.swift
@@ -258,40 +258,60 @@ struct MarkdownStableBoundaryScanner {
     /// Trailing partial line awaiting its newline.
     private var pendingLine = ""
     private var fenceMarker: String?
-    private var inMath = false
+    /// The close marker the open math region expects (`$$` or `\]`), so a
+    /// block opened with one cannot close on the other.
+    private var mathClose: String?
     private var inDirective = false
 
     /// True when the scanner sits inside a fenced/math/directive region at
     /// the end of everything consumed — callers avoid promoting a prefix
     /// that ends mid-construct.
-    var isInOpenConstruct: Bool { fenceMarker != nil || inMath || inDirective }
+    var isInOpenConstruct: Bool { fenceMarker != nil || mathClose != nil || inDirective }
 
     /// Feeds an append-only delta and returns any newly discovered safe
-    /// boundary offsets (absolute, UTF-8).
+    /// boundary offsets (absolute, UTF-8). The delta's complete lines are
+    /// processed in place — only the trailing partial is buffered — so cost
+    /// is O(delta) with no repeated prefix shifts even for one-shot bulk
+    /// feeds (threshold crossing, branch-swap reseed).
     mutating func append(_ delta: String) -> [Int] {
         var newBoundaries: [Int] = []
-        pendingLine += delta
+        var remainder = delta[...]
 
-        while let newline = pendingLine.firstIndex(of: "\n") {
-            let line = String(pendingLine[..<newline])
-            let lineBytes = line.utf8.count + 1 // including the newline
-
-            updateState(for: line.trimmingCharacters(in: .whitespaces))
-
-            // A blank line outside every construct ends a block: everything
-            // after this line is a fresh block, so the split point moves to
-            // the next line's start. Consecutive blank lines keep moving it.
-            if line.trimmingCharacters(in: .whitespaces).isEmpty,
-               fenceMarker == nil, !inMath, !inDirective {
-                let boundary = consumedOffset + lineBytes
-                lastSafeBoundary = boundary
-                newBoundaries.append(boundary)
+        // A buffered partial continues with the delta's first line.
+        if !pendingLine.isEmpty {
+            if let newline = remainder.firstIndex(of: "\n") {
+                pendingLine += remainder[..<newline]
+                consumeCompleteLine(pendingLine, boundaries: &newBoundaries)
+                pendingLine = ""
+                remainder = remainder[remainder.index(after: newline)...]
+            } else {
+                pendingLine += remainder
+                return newBoundaries
             }
-
-            consumedOffset += lineBytes
-            pendingLine.removeSubrange(..<pendingLine.index(after: newline))
         }
+
+        while let newline = remainder.firstIndex(of: "\n") {
+            consumeCompleteLine(String(remainder[..<newline]), boundaries: &newBoundaries)
+            remainder = remainder[remainder.index(after: newline)...]
+        }
+        pendingLine += remainder
         return newBoundaries
+    }
+
+    private mutating func consumeCompleteLine(_ line: String, boundaries: inout [Int]) {
+        let lineBytes = line.utf8.count + 1 // including the newline
+        updateState(for: line.trimmingCharacters(in: .whitespaces))
+
+        // A blank line outside every construct ends a block: everything
+        // after this line is a fresh block, so the split point moves to the
+        // next line's start. Consecutive blank lines keep moving it.
+        if line.trimmingCharacters(in: .whitespaces).isEmpty,
+           fenceMarker == nil, mathClose == nil, !inDirective {
+            let boundary = consumedOffset + lineBytes
+            lastSafeBoundary = boundary
+            boundaries.append(boundary)
+        }
+        consumedOffset += lineBytes
     }
 
     /// Processes the trailing partial line as if the stream had ended with
@@ -305,7 +325,7 @@ struct MarkdownStableBoundaryScanner {
         pendingLine = ""
         updateState(for: line.trimmingCharacters(in: .whitespaces))
         if line.trimmingCharacters(in: .whitespaces).isEmpty,
-           fenceMarker == nil, !inMath, !inDirective {
+           fenceMarker == nil, mathClose == nil, !inDirective {
             // Match append()'s arithmetic: the boundary sits after the
             // (virtual) newline of the blank line.
             lastSafeBoundary = consumedOffset + line.utf8.count
@@ -320,8 +340,8 @@ struct MarkdownStableBoundaryScanner {
             }
             return
         }
-        if inMath {
-            if trimmedLine == "$$" || trimmedLine == "\\]" { inMath = false }
+        if let close = mathClose {
+            if trimmedLine == close { mathClose = nil }
             return
         }
         if inDirective {
@@ -330,7 +350,8 @@ struct MarkdownStableBoundaryScanner {
         }
         if trimmedLine.hasPrefix("```") { fenceMarker = "```" }
         else if trimmedLine.hasPrefix("~~~") { fenceMarker = "~~~" }
-        else if trimmedLine == "$$" || trimmedLine == "\\[" { inMath = true }
+        else if trimmedLine == "$$" { mathClose = "$$" }
+        else if trimmedLine == "\\[" { mathClose = "\\]" }
         else if trimmedLine.hasPrefix(":::") {
             let name = trimmedLine.dropFirst(3).trimmingCharacters(in: .whitespaces).lowercased()
             if ["note", "info", "tip", "hint", "warning", "caution", "danger", "error", "important", "columns"].contains(name) {
--- a/Conduit/Views/Components/MarkdownLargeDocument.swift
+++ b/Conduit/Views/Components/MarkdownLargeDocument.swift
@@ -279,33 +279,59 @@ struct MarkdownStableBoundaryScanner {
 
         // A buffered partial continues with the delta's first line.
         if !pendingLine.isEmpty {
-            if let newline = remainder.firstIndex(of: "\n") {
-                pendingLine += remainder[..<newline]
-                consumeCompleteLine(pendingLine, boundaries: &newBoundaries)
+            if let separator = Self.firstLineSeparator(in: remainder) {
+                pendingLine += remainder[..<separator.index]
+                consumeCompleteLine(pendingLine, separatorBytes: separator.byteLength, boundaries: &newBoundaries)
                 pendingLine = ""
-                remainder = remainder[remainder.index(after: newline)...]
+                remainder = remainder[remainder.index(after: separator.index)...]
             } else {
                 pendingLine += remainder
                 return newBoundaries
             }
         }
 
-        while let newline = remainder.firstIndex(of: "\n") {
-            consumeCompleteLine(String(remainder[..<newline]), boundaries: &newBoundaries)
-            remainder = remainder[remainder.index(after: newline)...]
+        while let separator = Self.firstLineSeparator(in: remainder) {
+            consumeCompleteLine(
+                String(remainder[..<separator.index]),
+                separatorBytes: separator.byteLength,
+                boundaries: &newBoundaries
+            )
+            remainder = remainder[remainder.index(after: separator.index)...]
         }
         pendingLine += remainder
         return newBoundaries
     }
 
-    private mutating func consumeCompleteLine(_ line: String, boundaries: inout [Int]) {
-        let lineBytes = line.utf8.count + 1 // including the newline
-        updateState(for: line.trimmingCharacters(in: .whitespaces))
+    /// Swift treats "\r\n" as a single grapheme, so a plain
+    /// `firstIndex(of: "\n")` finds nothing in CRLF text and the whole
+    /// document reads as one line. A line separator here is the "\n"
+    /// grapheme (1 UTF-8 byte) or the "\r\n" grapheme (2 bytes); a lone
+    /// "\r" is not a separator, matching `parseDocument`'s CRLF-only
+    /// normalization.
+    private static func firstLineSeparator(in text: Substring) -> (index: Substring.Index, byteLength: Int)? {
+        var index = text.startIndex
+        while index < text.endIndex {
+            let character = text[index]
+            if character == "\n" { return (index, 1) }
+            if character == "\r\n" { return (index, 2) }
+            index = text.index(after: index)
+        }
+        return nil
+    }
+
+    private mutating func consumeCompleteLine(_ line: String, separatorBytes: Int, boundaries: inout [Int]) {
+        let lineBytes = line.utf8.count + separatorBytes // including the separator
+        // CRLF input (parseDocument normalizes it, so upstream can deliver
+        // it): the \r rides the line content, and .whitespaces trimming
+        // does not remove it — strip it for state/blank checks while the
+        // byte accounting above keeps counting it.
+        let normalized = line.hasSuffix("\r") ? String(line.dropLast()) : line
+        updateState(for: normalized.trimmingCharacters(in: .whitespaces))
 
         // A blank line outside every construct ends a block: everything
         // after this line is a fresh block, so the split point moves to the
         // next line's start. Consecutive blank lines keep moving it.
-        if line.trimmingCharacters(in: .whitespaces).isEmpty,
+        if normalized.trimmingCharacters(in: .whitespaces).isEmpty,
            fenceMarker == nil, mathClose == nil, !inDirective {
             let boundary = consumedOffset + lineBytes
             lastSafeBoundary = boundary
@@ -323,8 +349,9 @@ struct MarkdownStableBoundaryScanner {
         guard !pendingLine.isEmpty else { return }
         let line = pendingLine
         pendingLine = ""
-        updateState(for: line.trimmingCharacters(in: .whitespaces))
-        if line.trimmingCharacters(in: .whitespaces).isEmpty,
+        let normalized = line.hasSuffix("\r") ? String(line.dropLast()) : line
+        updateState(for: normalized.trimmingCharacters(in: .whitespaces))
+        if normalized.trimmingCharacters(in: .whitespaces).isEmpty,
            fenceMarker == nil, mathClose == nil, !inDirective {
             // Match append()'s arithmetic: the boundary sits after the
             // (virtual) newline of the blank line.
--- a/Conduit/Views/Components/MarkdownLargeDocument.swift
+++ b/Conduit/Views/Components/MarkdownLargeDocument.swift
@@ -43,6 +43,31 @@ enum MarkdownLargeDocumentPolicy {
     /// Lines per rendered slice of a large code block; bounds each slice's
     /// TextKit layout height (~800 lines ≈ 13K pt) and highlighting pass.
     static let codeSliceLineCount = 800
+    // --- Hardening additions (adversarial shapes) ---
+    /// Byte ceilings for code presentation: a block with a few very long
+    /// lines (minified JS, base64 blobs) must bound work by bytes as well
+    /// as by line count.
+    static let codePreviewLineCount = 200
+    static let codePreviewBytes = 16_000
+    static let codeSliceBytes = 64_000
+    /// Hard maximum for one promoted streaming stable chunk. Kept far below
+    /// the document threshold so a stable chunk can never route back into
+    /// the large-document renderer, and so the MarkdownText render of a
+    /// chunk is bounded regardless of how far ahead safe boundaries sit.
+    static let maxStreamChunkBytes = 2 * chunkTargetBytes
+    /// Tables at/above this estimated size use the paged presentation
+    /// (bounded column measurement on a sampled prefix + row batches).
+    static let largeTableBytes = 32_000
+    /// Textual rich blocks (callouts, columns) above this size reproject
+    /// their bodies into bounded inner text pieces.
+    static let largeTextBlockBytes = chunkTargetBytes
+    /// Math and Mermaid sources above this size drop the render action (the
+    /// renderers are not chunkable); the source stays previewable/copyable.
+    static let mathGuardBytes = 100_000
+    static let mermaidGuardBytes = 100_000
+    /// Byte budget for the per-chunk subset of message-wide reference
+    /// definitions appended at parse time (see MarkdownReferenceResolver).
+    static let referenceSubsetBudgetBytes = 8_000
 
     static func isLargeDocument(_ source: String) -> Bool {
         source.utf8.count >= documentThresholdBytes
@@ -68,6 +93,132 @@ enum MarkdownLargeDocumentPolicy {
 enum MarkdownLargeChunk: Equatable {
     case flow(blocks: [MarkdownBlock])
     case block(MarkdownBlock, originalIndex: Int)
+}
+
+/// Pure slicer for large code blocks: contiguous source slices bounded by
+/// BOTH a line ceiling and a UTF-8 byte ceiling. Concatenating the slices
+/// reproduces the original source exactly (giant lines are cut at
+/// whole-character boundaries with nothing inserted), which keeps Copy and
+/// round-trip tests trivially sound.
+enum MarkdownCodeSlicer {
+    static func slice(_ source: String, maxLines: Int, maxBytes: Int) -> [String] {
+        guard !source.isEmpty else { return [] }
+        var slices: [String] = []
+        var sliceStart = source.startIndex
+        var lineStart = source.startIndex
+        var lines = 0
+        var bytes = 0
+        var index = source.startIndex
+
+        func flush(through end: String.Index) {
+            guard sliceStart < end else { return }
+            slices.append(String(source[sliceStart..<end]))
+            sliceStart = end
+            lines = 0
+            bytes = 0
+        }
+
+        func emitGiantLinePieces(_ range: Range<String.Index>) {
+            var pieceStart = range.lowerBound
+            var pieceBytes = 0
+            var walk = range.lowerBound
+            while walk < range.upperBound {
+                pieceBytes += String(source[walk]).utf8.count
+                if pieceBytes >= maxBytes {
+                    let cut = source.index(after: walk)
+                    slices.append(String(source[pieceStart..<cut]))
+                    pieceStart = cut
+                    pieceBytes = 0
+                }
+                walk = source.index(after: walk)
+            }
+            if pieceStart < range.upperBound {
+                slices.append(String(source[pieceStart..<range.upperBound]))
+            }
+        }
+
+        while index < source.endIndex {
+            if source[index] == "\n" {
+                let lineEnd = source.index(after: index) // includes the newline
+                let lineBytes = source[lineStart..<lineEnd].utf8.count
+                if lineBytes > maxBytes {
+                    // A single line larger than the whole slice budget: emit
+                    // it as its own grapheme-aligned pieces.
+                    flush(through: lineStart)
+                    emitGiantLinePieces(lineStart..<lineEnd)
+                    sliceStart = lineEnd
+                    lines = 0
+                    bytes = 0
+                } else {
+                    if lines + 1 > maxLines || bytes + lineBytes > maxBytes {
+                        flush(through: lineStart)
+                    }
+                    lines += 1
+                    bytes += lineBytes
+                }
+                lineStart = lineEnd
+                index = lineEnd
+                continue
+            }
+            index = source.index(after: index)
+        }
+
+        if lineStart < source.endIndex {
+            let lineBytes = source[lineStart...].utf8.count
+            if lineBytes > maxBytes {
+                flush(through: lineStart)
+                emitGiantLinePieces(lineStart..<source.endIndex)
+                return slices
+            }
+            if lines + 1 > maxLines || bytes + lineBytes > maxBytes {
+                flush(through: lineStart)
+            }
+        }
+        flush(through: source.endIndex)
+        return slices
+    }
+}
+
+/// Incremental inline-marker balance for flow-text splitting. A cut is
+/// "safe" when no inline construct visibly spans it: even backtick count
+/// (inline code), zero open bracket depth (links/images/labels), and even
+/// asterisk/underscore/tilde parity (emphasis/strikethrough). Escaped
+/// characters (\*) do not count. This is deliberately a heuristic — it can
+/// only choose between cut points, never alter text — and a pathological
+/// unbalanceable window falls back to the plain target cut (rendered inline
+/// markers may then appear literal in that one piece).
+struct MarkdownInlineBalance {
+    private var backticks = 0
+    private var bracketDepth = 0
+    private var asterisks = 0
+    private var underscores = 0
+    private var tildes = 0
+    private var escaped = false
+
+    mutating func consume(_ character: Character) {
+        if escaped {
+            escaped = false
+            return
+        }
+        switch character {
+        case "\\": escaped = true
+        case "`": backticks += 1
+        case "[": bracketDepth += 1
+        case "]": bracketDepth = max(0, bracketDepth - 1)
+        case "*": asterisks += 1
+        case "_": underscores += 1
+        case "~": tildes += 1
+        default: break
+        }
+    }
+
+    var isBalanced: Bool {
+        bracketDepth == 0
+            && backticks % 2 == 0
+            && asterisks % 2 == 0
+            && underscores % 2 == 0
+            && tildes % 2 == 0
+    }
 }
 
 /// Groups a parsed document's blocks into bounded chunks. Pure and cheap:
@@ -115,14 +266,20 @@ enum MarkdownLargeChunkPlanner {
     /// quote lines, or word boundaries inside a paragraph/heading) so every
     /// piece stays within the chunk target. Heading text is small by nature
     /// but shares the paragraph fallback for completeness.
+    ///
+    /// Ordered lists: only the FIRST chunk of a split list keeps the real
+    /// `.orderedList` rendering; continuation groups render as paragraphs
+    /// with the ordinal baked into the text ("7. …") so numbering never
+    /// restarts at 1. The baked-number styling differs from the styled list
+    /// marker — a pathological-only formatting tradeoff.
     private static func splitOversized(block: MarkdownBlock) -> [MarkdownLargeChunk] {
         switch block {
         case .unorderedList(let items):
-            return groupItems(items) { .unorderedList($0) }
+            return splitListItems(items, ordered: false)
         case .orderedList(let items):
-            return groupItems(items) { .orderedList($0) }
+            return splitListItems(items, ordered: true)
         case .quote(let lines):
-            return groupItems(lines) { .quote($0) }
+            return splitQuoteLines(lines)
         case .heading(let level, let text):
             return splitText(text) { .heading(level: level, text: $0) }
         case .paragraph(let text):
@@ -132,32 +289,111 @@ enum MarkdownLargeChunkPlanner {
         }
     }
 
-    private static func groupItems<T>(_ items: [T], make: ([T]) -> MarkdownBlock) -> [MarkdownLargeChunk] {
-        func itemBytes(_ item: T) -> Int {
-            if let string = item as? String { return string.utf8.count }
-            if let line = item as? MarkdownQuoteLine { return line.text.utf8.count }
-            return 0
-        }
+    private static func splitListItems(_ items: [String], ordered: Bool) -> [MarkdownLargeChunk] {
+        let target = MarkdownLargeDocumentPolicy.chunkTargetBytes
         var chunks: [MarkdownLargeChunk] = []
-        var group: [T] = []
+        var group: [String] = []
+        var groupStartIndex = 0
         var groupBytes = 0
-        for item in items {
-            let bytes = itemBytes(item) + 32 // marker/separator allowance
-            if groupBytes + bytes > MarkdownLargeDocumentPolicy.chunkTargetBytes, !group.isEmpty {
-                chunks.append(.flow(blocks: [make(group)]))
-                group = []
-                groupBytes = 0
+        var emittedFirstGroup = false
+
+        func flush() {
+            guard !group.isEmpty else { return }
+            if ordered && emittedFirstGroup {
+                // Continuation of an ordered list: bake the real ordinals.
+                let paragraphs = group.enumerated().map { offset, item in
+                    MarkdownBlock.paragraph("\(groupStartIndex + offset + 1). \(taskStripped(item))")
+                }
+                chunks.append(.flow(blocks: paragraphs))
+            } else {
+                chunks.append(.flow(blocks: [ordered ? .orderedList(group) : .unorderedList(group)]))
+                emittedFirstGroup = true
+            }
+            group = []
+            groupBytes = 0
+        }
+
+        for (index, item) in items.enumerated() {
+            let bytes = item.utf8.count + 32 // marker/separator allowance
+            if bytes > target {
+                // One pathologically huge item: its first piece keeps the
+                // list semantics (or the baked ordinal); continuations are
+                // plain paragraphs.
+                flush()
+                groupStartIndex = index + 1
+                let task = MarkdownParser.taskItem(item)
+                let baseText = task?.text ?? item
+                let prefix = ordered ? "\(index + 1). " : ""
+                let taskMarker = task.map { $0.complete ? "[x] " : "[ ] " } ?? ""
+                let marked = taskMarker + baseText
+                // Uniform paragraph presentation for oversized items and
+                // their continuations keeps the pieces visually coherent
+                // (the ordinal or task marker is baked into the text).
+                chunks.append(contentsOf: splitText(marked) { piece in
+                    .paragraph(prefix + piece)
+                })
+                continue
+            }
+            if groupBytes + bytes > target, !group.isEmpty {
+                flush()
+                groupStartIndex = index
             }
             group.append(item)
             groupBytes += bytes
         }
-        if !group.isEmpty { chunks.append(.flow(blocks: [make(group)])) }
+        flush()
+        return chunks
+    }
+
+    private static func taskStripped(_ item: String) -> String {
+        MarkdownParser.taskItem(item)?.text ?? item
+    }
+
+    private static func splitQuoteLines(_ lines: [MarkdownQuoteLine]) -> [MarkdownLargeChunk] {
+        let target = MarkdownLargeDocumentPolicy.chunkTargetBytes
+        var chunks: [MarkdownLargeChunk] = []
+        var group: [MarkdownQuoteLine] = []
+        var groupBytes = 0
+
+        func flush() {
+            guard !group.isEmpty else { return }
+            chunks.append(.flow(blocks: [.quote(group)]))
+            group = []
+            groupBytes = 0
+        }
+
+        for line in lines {
+            let bytes = line.text.utf8.count + 8
+            if bytes > target {
+                // A pathologically huge quote line splits into several
+                // same-depth quote lines — quote chrome and depth semantics
+                // are preserved for every piece.
+                flush()
+                chunks.append(contentsOf: splitText(line.text) { piece in
+                    .quote([MarkdownQuoteLine(depth: line.depth, text: piece)])
+                })
+                continue
+            }
+            if groupBytes + bytes > target, !group.isEmpty {
+                flush()
+            }
+            group.append(line)
+            groupBytes += bytes
+        }
+        flush()
         return chunks
     }
 
     /// Word-boundary text split into ~chunkTargetBytes pieces. Each piece
     /// renders as its own paragraph; soft wrapping makes consecutive pieces
     /// read as one flowing body with a little extra paragraph spacing.
+    ///
+    /// Cut selection prefers the latest whitespace boundary whose prefix
+    /// has balanced inline markers (code spans, links, emphasis), so a
+    /// `**bold**`/`` `code` ``/`[link](…)` construct is not visibly broken
+    /// across pieces. A window with no balanced candidate falls back to the
+    /// plain whitespace cut — pathological inline soup may then show literal
+    /// markers at one seam, never altered text.
     static func splitText(_ text: String, make: (String) -> MarkdownBlock) -> [MarkdownLargeChunk] {
         guard text.utf8.count > MarkdownLargeDocumentPolicy.chunkTargetBytes else {
             return [.flow(blocks: [make(text)])]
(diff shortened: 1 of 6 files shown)


kaishi00:
The tests rely on the global counter. This confirms the data race is real (shared mutable static accessed across MainActor and detached-task threads).

Let me verify one thing about the tests that reset the counter — a concurrent test run could clear the array mid-assertion. But this is minor test-only.

I have enough. Let me write the review.

## Summary
This PR adds a bounded-rendering path for very large chat messages: a collapsed ≤16 KB preview with expand, chunked progressive rendering with per-chunk memoization, off-main parsing/highlighting, and an O(delta) append-only streaming mode above a 100 KB threshold. The design is well-measured and reuses existing parser/renderer components.

## Findings

### [severity: major] Unbounded tail re-render reintroduces the freeze it aims to prevent
**File:** Conduit/Views/Components/StreamingText.swift:218 (and 202-240, 309-345)
**Issue:** `largeTailSource` returns `String(largeAccumulated[revealedEnd...])` — everything from the reveal cursor to the end of the document — and this is fed to `MarkdownText(isStreaming: true)` every frame, whose fast path is cached by source and therefore *re-parses and re-lays-out the whole tail string on cache miss each tick*. Reveal advances at most 18 chars/tick (~540 chars/sec, `revealBatchSize`), but incoming content can arrive far faster (a big pasted prompt or a fast stream that crosses the threshold after `enterLargeMode`). As soon as the accumulated text outpaces the reveal cursor, the un-revealed remainder (which can be the bulk of a 1 MB message) becomes the tail string and is re-parsed at 30 fps — exactly the whole-document-per-frame freeze this PR claims to bound. The "only ~8 KB tail re-parsing" invariant only holds if reveal keeps pace with arrival, which is not enforced.
**Suggestion:** Bound `largeTailSource` to the revealed-but-unpromoted window regardless of how far `revealedEnd` trails, or treat backlogged un-revealed content as more stable chunks rather than the live tail, so the string handed to the re-rendering `MarkdownText` is clamped to ~`tailWindowBytes` independent of reveal lag.

### [severity: minor] DEBUG data race on `parseSourceSizes`
**File:** Conduit/Views/Components/MarkdownText.swift:1835
**Issue:** `nonisolated(unsafe) static var parseSourceSizes` is mutated in `parseDocument`, which now runs on **both** the MainActor (small path, previews) and inside `Task.detached` in `LargeMarkdownPreparedDocument.prepare` (MarkdownLargeDocumentView.swift:285). Concurrent appends from multiple threads are a data race. `nonisolated(unsafe)` silences the compiler but not the runtime hazard, and the tests read/clear this global (MarkdownLargeDocumentTests.swift:464, 608, 702), so a parallel or background parse can corrupt or empty it mid-assertion.
**Suggestion:** Make the counter `nonisolated(unsafe)` plus an `os_unfair_lock`/atomic, or restrict the expanded off-main parse from recording (e.g., a main-actor-only flag), or guard the array with synchronization. Keep it DEBUG/test-scoped.

### [severity: minor] Flow-chunk memo cache ignores foreground/accent/reference inputs
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:436-460
**Issue:** `LargeFlowChunkBox` caches on Dynamic Type category only; it does not key on `foregroundStyle`, `usesAccentSurface`, or `references`. If the view instance is reused (same `sourceIdentity-index` id) while those parameters change — e.g., a theme change that switches a bubble onto an accent surface — the box returns stale-colored attributed text until the font category changes.
**Suggestion:** Include `foregroundStyle`, `usesAccentSurface`, and `references` in the cache key, or invalidate the box when those change (`onChange` clearing `cachedText = nil`).

### [severity: minor] Character-fade is effectively invisible in large streaming mode
**File:** Conduit/Views/Components/StreamingText.swift:218
**Issue:** The reveal cursor (`revealedEnd`) advances past newly revealed characters, so the fade opacities (`characterOpacities`) are applied to `largeTailSource`, which begins *at* `revealedEnd` and therefore excludes the most recently revealed characters — they render unfaded until promoted into stable chunks (rendered with `isStreaming: false`, no fade). The PR states "the character-fade stays on the tail," but the fade window and the tail string are disjoint.
**Suggestion:** Make the fade target the recently-revealed slice (e.g., `largeAccumulated[max(promotedEnd, previousRevealedEnd)...revealedEnd]`) rather than the unresolved tail.

### [severity: nit] Detached prepare task is not actually cancelled on teardown
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:285
**Issue:** `Task.detached(priority:)` does not inherit cancellation from the `.task(id:)` scope that spawns it, so abandoning/expanding a new source while an old 1 MB prepare is in flight lets the background parse run to completion (wasted work). The `.value`/source-identity guard discards the result, so there's no correctness bug — just avoidable work.
**Suggestion:** Create the task with `Task(priority:)` inside the `.task` scope so it inherits cancellation, or check `Task.isCancelled` inside the loop of `prepare`.

## Verdict
REQUEST_CHANGES

The architecture is thoughtful and well-tested, but finding #1 undermines the PR's central invariant (bounded per-frame work) under exactly the fast-arrival scenario the feature targets; it should be addressed before merge. The remaining items are minor/nit.

<a href="https://opencode.ai/s/lEQklggI">New%20session%20-%202026-08-20T01%3A12%3A01.625Z</a>
[opencode session](https://opencode.ai/s/lEQklggI)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32320077480)

kaishi00:
Triage of the 11 actionable comments — all real findings except where noted; fixes landed in 860eb2f with a regression test per item.

**Critical — `largeTailSource` returned the unrevealed remainder** ✅ taken. Confirmed: the slice ran from the reveal cursor to the end, so unrevealed text rendered immediately while the revealed-unpromoted window was skipped — the character-paced reveal was effectively inverted in large mode. The tail is now the `[promoted, revealed)` window, extracted into a pure helper (`StreamingText.tailSource`) with a dedicated test (`testTailWindowExcludesUnrevealedText`).

**Critical — failed re-derive retains a stale `String.Index`** ✅ taken. The alignment invariant makes the failure unreachable in practice, but unreachable ≠ total: the fallback path now calls `revealAllLargeImmediately()` rather than keeping an index into a discarded string.

**Major — hard-cut boundary can land mid-grapheme and stall promotion** ✅ taken. Real for CJK/emoji streams with no blank lines (the failure path returned without advancing, so promotion stalled forever). Offsets now step back to the previous whole-character boundary (`alignedIndex`), covered by `testPromotionProgressesForUnbrokenCJKText` and `testAlignedIndexStepsBackToGraphemeBoundaryForCJK`.

**Major — scanner quadratic prefix-shift + whole-document reseed on main** ✅ taken. `append()` now processes complete lines in place and buffers only the trailing partial, so one-shot bulk feeds (threshold crossing, branch-swap reseed) are O(n) instead of O(n²); the rescan itself is bounded by the reseed semantics (append-only verification requires a full pass) and now linear.

**Major — `parseSourceSizes` cross-thread mutation** ✅ taken. DEBUG-only but a genuine data race (MainActor rendering vs. detached preparation). Now lock-guarded.

**Major — line-count scans per body evaluation** ✅ taken. Stats computed once off-main; the Copy state toggle no longer re-walks a megabyte block.

**Minor — math close markers not paired to opener** ✅ taken, tracked like `fenceMarker` (`testScannerMathCloseMarkersArePaired`).

**Minor — force unwraps in tests** ✅ taken (`XCTUnwrap`).

**Minor — tautological assertion** ✅ taken; now asserts the precondition (fewer chunks than one batch) that exercises the clamp, using the `initialChunkBatch` constant.

**Minor — `finish()` test didn't exercise the blank-partial branch** ✅ taken; input now ends with a real partial blank line and the boundary is asserted to advance past the pre-`finish` value.

**Nit — SwiftLint `optional_data_string_conversion` (tests)** — the repo's CI doesn't gate on SwiftLint and the flagged conversion is inside a test; left as-is.

kaishi00:
The unit test `testFlowChunkBoxMemoizesAndInvalidatesOnDynamicTypeChange` tests the box at the `LargeFlowChunkBox` level with explicit `contentCategory`. But the SwiftUI integration is the problem: `LargeFlowChunkView` doesn't wire the environment's Dynamic Type into `contentCategory`, so the box's category invalidation never fires in the actual view. The unit test passes but the real view is broken — exactly the integration gap CodeRabbit flagged and the author didn't verify.

Let me confirm the Dynamic Type is indeed the concern — let me check whether `MarkdownBlockView` (the ordinary path, reused for rich blocks) responds to Dynamic Type to see the expected behavior, and whether the flow chunk's `SelectableTextView` re-lays out with the new preferred font via `traitCollectionDidChange`. The baked attributed font is the issue regardless.

I have enough to write the review. Let me quickly verify the `isStreaming` forwarding nit — not addressed (line 42-47 drops isStreaming), but it's a nit with no current caller affected.

Also let me confirm the `groupItems` runtime-cast nit (still present, line 136-140) and the `itemBytes` returning 0 fallback — real but nit, future-proofing.

Let me also double check the detached prepare task cancellation nit — `Task.detached` doesn't inherit cancellation (line 285), and `guard !Task.isCancelled` only checks after awaiting. Still present.

Let me verify the whole test count/state and check one more potential concern: the `alignedIndex` in `promoteLargeChunks` / `tailSource` is O(offset) with the `while offset >= 0` stepping, but for aligned offsets it's O(1). For hard-cut misaligned it walks back at most ~3 bytes. Fine.

Let me finalize. The most significant remaining issue is the Dynamic Type integration gap. The major issues from the original review are fixed. Let me write the review.

Let me confirm the Dynamic Type concern is genuinely valid for the flow chunk by checking how `SelectableTextView` handles the attributed string vs `preferredFont`. The baked attributed string uses fonts from `MarkdownSelectionFormatter.attributedText`. If that formatter uses `.preferredFont(forTextStyle:)`, then the attributed string has preferred font baked at build time. When Dynamic Type changes, the attributed string's fonts don't update, and the memo cache returns it unchanged. Even SelectableTextView's `font` param is just the fallback; the attributed run styles dominate. So the flow chunk text truly doesn't scale. Confirmed as a genuine a11y bug.

Let me write the review now.

## Summary
This PR adds a bounded-rendering path for very large chat messages: a collapsed ≤16 KB preview, off-main chunk planning, progressive mounting of ~8 KB memoized flow chunks, sliced/async-highlighted large code blocks, and an O(delta) append-only streaming mode above a 100 KB threshold. The review-fix commit 860eb2f resolved all six major/critical findings (tail window, stale String.Index, CJK hard-cut alignment, scanner quadratic cost, parseSourceSizes data race, per-body line scans).

## Findings

### [severity: major] Flow-chunk memo cache never invalidates on Dynamic Type change in the actual view
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:466-501 (and LargeFlowChunkBox 436-460)
**Issue:** `LargeFlowChunkBox.attributedText` only rebuilds when `contentCategory` changes, but `LargeFlowChunkView.body` does not read `@Environment(\.dynamicTypeSize)` nor observe `UIContentSizeCategory.didChangeNotification`, and there is no `onChange` clearing `cachedText`. So when the user changes the accessibility text size, SwiftUI never re-evaluates `body`, `attributedText` is called with the same (default-argument) category, and the box returns the stale string whose fonts were baked at build time. The attributed string's run-level `preferredFont(forTextStyle:)` styles do not rescale, so large flow chunks stay at their old size — a real Dynamic Type/accessibility regression. The unit test `testFlowChunkBoxMemoizesAndInvalidatesOnDynamicTypeChange` only exercises the box directly with an explicit `contentCategory`, so it passes while the view integration is broken.
**Suggestion:** Read `@Environment(\.dynamicTypeSize)` (or `sizeCategory`) in `LargeFlowChunkView.body` and pass it to `box.attributedText(…, contentCategory:)`, so the environment change re-evaluates body and invalidates the cache. Note that reading a Dynamic Type environment value is what triggers SwiftUI to re-render on size change.

### [severity: minor] Memo box ignores foreground/accent/reference inputs (stale color on theme change)
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:436-460
**Issue:** `LargeFlowChunkBox` caches solely on `contentCategory`. The chunk's `.id` is `sourceIdentity-index`, which is stable across a theme change, so when `foregroundStyle`/`usesAccentSurface`/`references` change on the same view the box returns the old-colored attributed text. Distinct from the Dynamic Type case above, a theme change to an accent surface would leave the flow chunk in stale colors indefinitely.
**Suggestion:** Include `foregroundStyle`, `usesAccentSurface`, and `references` in the cache key, or invalidate the box via `onChange` of those parameters.

### [severity: minor] `isStreaming` is dropped before the large-document presentation
**File:** Conduit/Views/Components/MarkdownText.swift:41-47
**Issue:** When a large source routes to `LargeMarkdownDocumentView`, `isStreaming` is not forwarded, so the collapsed preview built inside `collapsedView → MarkdownText(...)` renders with `isStreaming: false` and populates `MarkdownRenderCache`. Current production callers route streaming through `StreamingText.largeBody` (never `MarkdownText` itself), so there is no live bug today — this is a latent footgun for a future changing preview.
**Suggestion:** Forward `isStreaming` into `LargeMarkdownDocumentView` and preserve it on the preview's `MarkdownText`.

### [severity: nit] Detached prepare/highlight tasks don't inherit cancellation
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:285, 597, 716
**Issue:** `Task.detached(priority:)` doesn't inherit cancellation from the `.task(id:)` scope that spawns it; the `guard !Task.isCancelled` runs only after the off-main work completes. Abandoning/expanding a new source while an old 1 MB prepare or highlight is in flight wastes the work to completion. No correctness bug (results are discarded), just avoidable CPU.
**Suggestion:** Use `Task(priority:)` inside the `.task` scope to inherit cancellation, or check `Task.isCancelled` inside the work loop.

### [severity: nit] `groupItems` uses runtime casts with a silent zero-byte fallback
**File:** Conduit/Views/Components/MarkdownLargeDocument.swift:135-156
**Issue:** `itemBytes` falls back to `0` for any payload unlike `String`/`MarkdownQuoteLine`; a future flow-block payload would measure 0 bytes and group every item into one chunk — silently recreating the oversized-chunk the planner exists to prevent.
**Suggestion:** Replace the casts with a small `chunkSourceBytes` protocol adopted by `String` and `MarkdownQuoteLine` (compile-time enforcement), or type-specific overloads.

## Verdict
COMMENT

The six major/critical review findings are fixed and regression-tested in 860eb2f. The one genuinely impactful remaining issue is the Dynamic Type integration gap (finding #1): the memoized flow-chunk box won't rescale under accessibility text-size changes because the view never observes the Dynamic Type environment — the unit test masks it by bypassing the view. That is an accessibility/correctness concern worth fixing before merge. The rest are minor/nit.

<a href="https://opencode.ai/s/nhwJp5wh">New%20session%20-%202026-08-20T01%3A30%3A28.256Z</a>
[opencode session](https://opencode.ai/s/nhwJp5wh)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32321182862)

kaishi00:
Follow-up on the CRLF finding — taken, and it ran deeper than the marker-comparison framing: Swift treats `"\r\n"` as a **single grapheme cluster**, so `firstIndex(of: "\n")` never matched in CRLF text at all — the scanner read an entire CRLF document as one line (no boundaries, no construct tracking, only hard-cut promotion). Line splitting now recognizes both the `"\n"` and `"\r\n"` graphemes with byte-accurate separator accounting, and the state/blank checks strip the trailing `\\r` that `.whitespaces` trimming leaves behind. Covered by `testScannerHandlesCRLFInput` (blank boundaries + `23693`/`:::` closes on CRLF, directive/math paired markers). Landed in the follow-up commit.

kaishi00:
Confirmed: the test exercises `LargeFlowChunkBox` directly with explicit `contentCategory` parameters. The view integration (`LargeFlowChunkView.body`) still does NOT wire the Dynamic Type environment — the default-arg approach only works if body re-evaluates on a size change, and nothing in the expanded view subtree reads a Dynamic Type-dependent environment/property, so body does not re-evaluate. This is the genuine accessibility gap that remains unfixed.

Let me also verify the `isStreaming` forwarding observation — the review said `LargeMarkdownDocumentView` doesn't take `isStreaming`. Confirmed (MarkdownText.swift:41-47 does not pass isStreaming). It's a latent nit.

Also, let me double-check the memo cache color staleness concern (minor 2 from review round 1). `LargeFlowChunkBox.attributedText` caches based only on `contentCategory`, not `foregroundStyle`/`usesAccentSurface`/`references`. So a theme change (dark/light or accent) could return stale colors. However — is `LargeFlowChunkView` in an identity-preserving ForEach with `.id("\(sourceIdentity)-\(index)")`? Yes (line 325). If `usesAccentSurface`/`foregroundStyle`/`references` change while the id stays the same and category doesn't change, the box returns the stale colored string. But does body re-evaluate on those changes? `usesAccentSurface` and `foregroundStyle` are `let` params passed down; SwiftUI re-renders when the parent passes new values if the property changes trigger it. The box cache would then return stale colored text since only category is in the key. This was flagged in round 1 as minor and remains.

Actually the code comment at line 238 says foregroundStyle is only ever .primary/.white tied to usesAccentSurface — and MarkdownRenderCache keys only on usesAccentSurface too. But here the box doesn't include usesAccentSurface in the cache key at all. So if a bubble's palette changes (e.g., wallpaper/theme switch changing accent), the memoized string keeps old colors. This is a real but minor concern.

Now let me also verify the round-2 "clear old prepared state" finding and the "persist aligned promotion offset" finding (both outside-diff). 

For "persist aligned promotion offset": In `promoteLargeChunks` (line 366-387), `largePromotedBytes = next` (line 385) — it stores the **unaligned** `next`, not the aligned `end`. Wait but the reviewer's round-2 finding says storing the raw next duplicates content when `alignedIndex` moved back. Let me think.

`alignedIndex` steps `end` back to a character boundary if `next` lands mid-grapheme. Then the chunk `[start..<end]` is appended, and `largePromotedBytes = next` (raw). Next iteration, `tailSource`/next promotion re-derives `alignedIndex(largePromotedBytes)` which steps back to the same `end` — so the tail `[end, revealedEnd)` would include characters that are already in the promoted chunk (up to `next` was included in the chunk since chunk went to `end` > boundary... wait).

Hmm. Actually `next >= end` (alignedIndex steps back, so `end <= next`). The chunk appended is `[start, end)`. If `largePromotedBytes = next` (raw), then on the next call, the tail starts at `alignedIndex(next)` = `end` again, but the chunk already covered up to `end`, so... Actually the chunk covered `[start, end)`. Setting promotedBytes = next > end means the next tail/growth window starts from `end` (alignedIndex(next)=end). So no duplication — the chunk covered to `end`, and the next slice starts at `end`. Setting promotedBytes=next just means it "claims" bytes up to `next` as promoted even though the chunk only contains to `end`. This could create a small gap between what's promoted-accounted vs. actually-rendered, but since end<next are all inside one grapheme, the gap is sub-character and the tail would start at `end` which is exactly where the chunk ended. So no duplication, no gap in rendered content. Actually this is fine — content is contiguous. The reviewer's concern about duplication doesn't materialize because alignedIndex is deterministic and idempotent. So this finding seems fine/as-addressed.

Wait, but could `largePromotedBytes = next` combined with the `promoteLargeChunks` inner `while` cause issues? Let me not over-analyze; the content is contiguous because both the chunk end and the next tail start derive from `end` = alignedIndex(next).

For "clear old prepared state": In `.task(id: sourceIdentity)` (lines 280-296), when sourceIdentity changes, `prepared` retains the old document until the detached prepare completes. During that interval the view shows the old `prepared` (old message content) — a stale-state window on source swap. The reviewer flagged this. Let me check: is it still present? Yes — lines 280-296 do not clear `prepared`/`renderedChunkCount`/selection before awaiting. However, `.task(id:)` cancels the previous task when id changes, and the guard checks `plan.sourceIdentity == sourceIdentity` — but during the run it shows old `prepared`. This is a real staleness concern but minor; it only matters on source swap of the same expanded view instance, which is unusual (expanded view keyed by source identity?). Let me check if LargeMarkdownExpandedView is identity-stable across source changes. It's shown from `LargeMarkdownDocumentView.body` keyed... there's no `.id`. The `.task(id: sourceIdentity)` re-runs on source change within the same view instance. So old content flashes until new parse completes. Minor.

Now let me also double check the `isStreaming` forwarding nit and the `groupItems` casts still present.

`groupItems` (line 135-156) still uses `as? String` / `as? MarkdownQuoteLine` with `return 0` fallback. That's the CodeRabbit nit that was explicitly left as-is.

Now let me verify the round-1 `isStreaming` dropped finding - yes still present (MarkdownText lines 41-47 no isStreaming). 

Let me also confirm the `LargeStreamPromotion` hard-cut / `promoteLargeChunks` unaligned `largePromotedBytes=next` reconsidered with the CRLF change. Also I want to confirm whether the loop in `promoteLargeChunks` being a `while` can infinite-loop. `next` returned is always > promotedBytes (since stableTarget - promotedBytes > chunkTargetBytes, minimumNext = promotedBytes + chunkTargetBytes). After appending chunk and `largePromotedBytes = next`, the loop recomputes. Since next >= promotedBytes + chunkTarget, each iteration strictly advances. But there's a subtle issue: `largePromotedBytes = next` where `next` could be > what's actually revealed? No, next <= stableTarget = revealedBytes - tail. Fine.

Let me now check one more important concurrency thing: `LargeMarkdownPreparedDocument.prepare` runs in `Task.detached`, which calls `MarkdownParser.parseDocument` which appends to `parseSizes` (now lock-guarded — good). And references Foundation `AttributedString(markdown:)` off-main — thread-safe per Foundation. OK.

Let me also double-check `promoteLargeChunks` at line 378-381: the `guard ... else { return }` — if conversion fails, returns silently without advancing, could this stall? `alignedIndex` steps back and returns nil only for offsets past end. Since next <= revealedBytes and revealedEnd is at a real character boundary, alignedIndex(next) should succeed (stepping back to valid char). So fine.

Now let me verify the deployment target for NSLock.withLock — they used manual lock()/defer unlock(), so no iOS16 requirement issue. Good.

Let me finalize. Let me quickly confirm there's no `.environment` dynamicTypeSize anywhere — already confirmed grep for dynamicTypeSize returned nothing across whole repo. This confirms the Dynamic Type gap is real relative to how the app handles it elsewhere (only via UIKit preferredFont / preferredContentSizeCategory, which get triggered for MarkdownText via the cache-key read in body; but for the flow chunk box the `preferredContentSizeCategory` default-arg only runs when body already re-evaluates).

Actually, wait. Let me reconsider whether `LargeFlowChunkView.body` DOES get re-evaluated on Dynamic Type change. The default argument `UIApplication.shared.preferredContentSizeCategory` is read as part of evaluating `body`'s statement. For SwiftUI to re-evaluate `body` on Dynamic Type change, the body must have a dependency on Dynamic Type. Reading `UIApplication.shared.preferredContentSizeCategory` directly does NOT establish a SwiftUI dependency (it's not an @Environment read). The official way to invalidate is reading `@Environment(\.dynamicTypeSize)` or `@Environment(\.sizeCategory)`. Without that, SwiftUI won't re-render on text-size change. So the finding stands and is real.

Now I have enough to write the review. The most significant remaining issue is the Dynamic Type integration gap (major, accessibility). Remaining minors/nits: memo box ignores color/reference inputs, isStreaming not forwarded, detached tasks don't inherit cancellation, stale prepared state on source swap, groupItems runtime casts.

Let me write the review.

## Summary
This PR adds a bounded-rendering path for very large chat messages: a collapsed ≤16 KB preview, off-main chunk planning, progressive ~8 KB memoized flow chunks, sliced/async-highlighted oversized code, and an O(delta) append-only streaming mode above the 100 KB threshold. Review-round fixes (tail window, CJK promotion alignment, scanner cost, parseSourceSizes locking, CRLF handling) are all in place.

## Findings

### [severity: major] Flow-chunk attributed strings never rebuild under Dynamic Type, breaking accessibility text scaling
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:479-501 (and LargeFlowChunkBox 436-460)
**Issue:** `LargeFlowChunkBox.attributedText` memoizes and only invalidates when `contentCategory` changes, passed as `UIApplication.shared.preferredContentSizeCategory`. But `LargeFlowChunkView.body` does not read `@Environment(\.dynamicTypeSize)` or `@Environment(\.sizeCategory)` (grep confirms no view in the app reads a Dynamic Type environment), and nothing observes `UIContentSizeCategory.didChangeNotification`. The default-argument read of `preferredContentSizeCategory` only runs when SwiftUI already decides to re-evaluate `body`; reading it inside body does not establish a SwiftUI dependency. So on an accessibility text-size change, SwiftUI never re-evals this subtree and the box returns the stale attributed string with fonts baked at the old size. This differs from the ordinary path, where `MarkdownRenderCache.rendering` is keyed on the size category and invoked from `MarkdownText.body`, so it rebuilds. The unit test `testFlowChunkBoxMemoizesAndInvalidatesOnDynamicTypeChange` only calls the box directly with explicit `contentCategory`, so it passes while the real view is broken — an integration gap.
**Suggestion:** Read `@Environment(\.dynamicTypeSize)` (or `sizeCategory`) in `LargeFlowChunkView.body` and pass the derived `UIContentSizeCategory` into `box.attributedText(..., contentCategory:)`. That environment read is what makes SwiftUI re-evaluate and drive the cache invalidation.

### [severity: minor] Memo box ignores foreground/accent/reference inputs, risking stale colors on theme/identity changes
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:436-460
**Issue:** `LargeFlowChunkBox` caches only on `contentCategory`; the chunk identity `.id("\(sourceIdentity)-\(index)")` is stable across a palette/theme change. If `foregroundStyle`/`usesAccentSurface`/`references` change while the category is unchanged, the box returns the old-colored attributed text (dark/light or accent flip, or a new reference definition) indefinitely.
**Suggestion:** Include `usesAccentSurface` (and, where meaningful, `references`) in the cache key, or invalidate via `onChange` of those parameters when they can change on a retained instance.

### [severity: minor] `isStreaming` is dropped before the large-document presentation
**File:** Conduit/Views/Components/MarkdownText.swift:41-47
**Issue:** `isStreaming` is not forwarded into `LargeMarkdownDocumentView`, so the collapsed preview's `MarkdownText` renders `isStreaming: false` and populates `MarkdownRenderCache`. Today, streaming crosses large mode through `StreamingText.largeBody` and never through `MarkdownText`, so there is no live bug — a latent footgun.
**Suggestion:** Forward `isStreaming` into `LargeMarkdownDocumentView` and preserve it on the bounded preview.

### [severity: minor] Detached prepare/highlight tasks don't inherit cancellation; stale state shown during source swap
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:280-296, 573-604, 705-726
**Issue:** `Task.detached(priority:)` doesn't inherit cancellation from the `.task(id:)` scope; the `guard !Task.isCancelled` checks run only after the off-main work returns, so abandoning a source wastes an in-flight 1 MB parse/highlight. Additionally, when `sourceIdentity` changes, `prepared`/`renderedChunkCount`/selection are not reset before awaiting, so the old message can be shown during the new 

<a href="https://opencode.ai/s/PkaHTSaH">New%20session%20-%202026-08-20T01%3A46%3A30.046Z</a>
[opencode session](https://opencode.ai/s/PkaHTSaH)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32322160243)

kaishi00:
Confirmed: `isStreaming` is indeed not forwarded to `LargeMarkdownDocumentView` (lines 41-47). This is a latent nit as described since streaming is routed through `StreamingText.largeBody`. This remains unaddressed but is harmless for current callers.

Now let me verify the memo-cache stale color concern in `LargeFlowChunkBox`. Currently it caches on `contentCategory` AND `usesAccentSurface` (lines 516-517, 527-529). So accent surface is now included. But `foregroundStyle` and `references` are not in the key. The `foregroundStyle` — let me check whether it can change for a surviving box. The chunk `.id("\(sourceIdentity)-\(index)")` is identity-stable. `foregroundStyle` is passed down from `LargeMarkdownExpandedView`'s `foregroundStyle`. If the theme changes and foregroundStyle changes while `usesAccentSurface` doesn't, the box returns stale-colored text. 

Earlier review noted the code comment at line 238 said foregroundStyle is only ever .primary/.white tied to usesAccentSurface. Let me check that claim - is that true? If foregroundStyle is fully determined by usesAccentSurface, then caching on usesAccentSurface alone suffices and there's no stale issue for foregroundStyle. But `references` could change on a surviving box — however references are tied to `prepared.referencesByChunk`, and a changed prepared plan means new sourceIdentity → new id → new box. So references can't change on a surviving box. That comment holds.

The Dynamic Type issue is now resolved (sizeCategory environment read at line 560). Let me confirm there's no other genuinely impactful issue. Let me check the `groupItems`/runtime casts — those are gone (splitText uses chunkTargetBytes, no `as?`). The `isStreaming` forwarding nit and detached-task-cancellation nit remain. Let me verify the `Task.detached` cancellation claim by checking prepare and how it's called.

The `LargeMarkdownPreparedDocu

<a href="https://opencode.ai/s/GMas4Q16">New%20session%20-%202026-08-20T02%3A40%3A52.178Z</a>
[opencode session](https://opencode.ai/s/GMas4Q16)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32325421599)

kaishi00:
The prepare/lineStats/sliceAsync all run as structured tasks under `.task(id:)` with cancellation guards. The detached-task concern is resolved.

I've completed a thorough review. The PR has been through multiple review rounds and the fixes are comprehensive. Let me write the final review.

## Summary
This PR adds a bounded-rendering path for very large chat messages: a collapsed ≤16 KB preview, off-main chunk planning, progressive ~8 KB memoized flow chunks, sliced/async-highlighted oversized code blocks, and an O(delta) append-only streaming mode above the 100 KB threshold. It reuses the existing parser/formatter/block renderer and preserves selection, references, Copy, and Dynamic Type support.

## Findings
The original review findings (tail window inversion, stale String.Index, CJK hard-cut alignment, scanner quadratic cost, parseSourceSizes data race, CRLF handling, table index-out-of-range crash, Dynamic Type integration, and the code/callout/column per-body rescans) are all resolved in the follow-up commits. What remains are latent/minor items with no reachable production bug.

### [severity: minor] `isStreaming` still not forwarded for the large-document preview
**File:** Conduit/Views/Components/MarkdownText.swift:41-47
**Issue:** `LargeMarkdownDocumentView` is constructed without `isStreaming`, so the collapse preview's `MarkdownText` renders `isStreaming: false` and would populate `MarkdownRenderCache`. Today no caller routes a live stream through `MarkdownText` for a large doc (streaming crosses via `StreamingText.largeBody`), so there is no current bug — but a future change to the preview would silently blow the cache with giant entries.
**Suggestion:** Forward `isStreaming` into `LargeMarkdownDocumentView` and preserve it on the bounded preview.

### [severity: nit] `largeReveal` loop lacks an explicit endIndex guard
**File:** Conduit/Views/Components/StreamingText.swift:363-368
**Issue:** The `for offset in 0..<revealCount` loop calls `largeAccumulated.index(after: revealedEnd)` without an `endIndex` bound. I traced the grapheme-merge-across-append scenario that could overshoot and concluded it is not reachable (a merged grapheme makes `newText.hasPrefix(largeAccumulated)` false, which reseeds and resets the character/byte counts to consistent values). It is defensive only, but a cheap safeguard against future changes to the delta derivation.
**Suggestion:** Add `guard revealedEnd < largeAccumulated.endIndex else { break }` at the top of the loop.

### [severity: nit] `parseSourceSizes` grows unbounded in DEBUG
**File:** Conduit/Views/Components/MarkdownText.swift:2040-2060
**Issue:** The array is now correctly lock-protected, but it will grow without limit during a long streaming session (roughly one entry per parse). Test-scoped instrumentation only, so no runtime impact outside DEBUG.
**Suggestion:** Optionally bound retention (e.g., keep the last N entries or halve when over a cap).

### [severity: nit] Force unwrap on `columnPieceRanges` in columns view
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:818-820
**Issue:** `content.columnPieceRanges![column]` with the pre-guard `content.columnPieceRanges?.indices.contains(column) == true` is safe by construction (the unwrap only happens after confirming non-nil), so no crash. Stylistic only.
**Suggestion:** No change required, or bind the range once to avoid the `!`.

## Verdict
COMMENT

The architecture is sound and every previously-flagged critical/major issue (including the tail-window inversion that undermined the core bounded-work invariant, the CJK/grapheme promotion stalling, and the table index-out-of-range crash) is genuinely fixed and regression-tested. Remaining items are latent or defensive with no reachable production defect. The `isStreaming` forwarding is the only one worth a low-priority follow-up; the rest can be left as-is.

<a href="https://opencode.ai/s/SxTcrndU">New%20session%20-%202026-08-20T08%3A41%3A14.636Z</a>
[opencode session](https://opencode.ai/s/SxTcrndU)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32349947450)

kaishi00:
Triage of the 17 comments on 6810426 — head is now `0b4ae67`.

**Stale re-anchors (fixed in 6810426 itself, verified present in the head at the given sites):** `emittedFirstGroup` in the oversized-item branch, the hard-cut candidate interval check (`isInsideConstruct` tests both the stable target and the candidate), callout/column split-once-during-preparation, the preview whole-block walk (`firstSlice`), the `renderedRowCount` init clamp, and per-piece descriptor registration. CodeRabbit reviewed the diff that *introduced* these fixes.

**New findings from this round — all real, all fixed in 0b4ae67 with regressions:**

- **`firstSlice` returned a non-prefix when a giant line followed normal lines** (it dropped the preceding lines), and could exceed the byte budget (preceding lines + a full 16 KB giant piece). Fixed: the giant piece now uses the *remaining* budget (`maxBytes − bytes`), and the result is always a source prefix — pinned by a new normal-lines-then-giant-line case in `testCodePreviewIsBoundedPrefixWork`.
- **`firstSlice` walked the entire newline-free blob** before classifying it as giant. It now stops as soon as the current line itself exceeds the ceiling — O(ceiling), never O(source).
- **`columnWidths` ran once per mounted row** (up to 26× per body evaluation, each rebuilding bounded-cell strings for the sampled measurement). Hoisted to a single `let widths` per body evaluation shared by the header and every row.
- **`splitTextCallCount` was lock-guarded on write only.** Now a lock-guarded computed property end to end (mirrors `parseSourceSizes`).

**Test hardening per review:** continuation ordinals now assert strict increase and no duplicates; the pre-fence scanner test's stable target moved past `promoted + minChunk` so the window logic actually runs (previously the assertion was vacuous); the stale-content test asserts the precondition that the old source rendered before swapping; force unwraps replaced with `XCTUnwrap`; the non-optional nil-coalescing on `text.prefix` cleaned up.

Suite: 55/55 in the large-document class, 878 tests 0 failures across the full unit suite.

kaishi00:
The `columnPieceRanges![column]` force unwrap at line 819 is guarded by the pre-check at line 818 (`indices.contains(column) == true`), so it's safe by construction — only a stylistic nit.

I've now completed a thorough review of the head commit. Let me compile my findings. The evidence:

1. **Major (real, unaddressed, untested):** Streaming grapheme-merge content loss in `updateLargeTarget` via `dropFirst(largeAccumulatedChars)`.
2. **Minor (latent, acknowledged):** `isStreaming` not forwarded to `LargeMarkdownDocumentView` → preview `MarkdownText` renders `isStreaming: false` and populates `MarkdownRenderCache`.
3. **Nit:** `alignedIndex` lacks the clamp (defensive only).
4. **Nit:** `columnPieceRanges![column]` force unwrap (safe by precondition).

Everything else from the many review rounds is genuinely fixed and I verified: tail window inversion, stale-prepared-state clearing, structured-task cancellation, parseSourceSizes + splitTextCallCount locking, table row clamp, Dynamic Type env integration, callout/columns per-piece descriptors/references, columnWidths once-per-body, code preview firstSlice prefix+bounded.

Let me write the review.

## Summary
This PR adds a bounded-rendering path for very large chat messages: a collapsed ≤16 KB preview with expand, off-main chunk planning, progressive ~8 KB memoized flow chunks, sliced/async-highlighted oversized code, and an O(delta) append-only streaming mode above the 100 KB threshold. It reuses the existing parser/formatter/block renderer and preserves selection, references, Copy, list numbering, and Dynamic Type.

## Findings

### [severity: major] Streaming can silently drop a grapheme cluster split across deltas
**File:** Conduit/Views/Components/StreamingText.swift:299
**Issue:** `updateLargeTarget` derives the new delta with `String(newText.dropFirst(largeAccumulatedChars))`. `dropFirst` drops whole `Character` graphemes, so when a single grapheme cluster is split across two WebSocket deltas (a combining mark, a ZWJ emoji sequence, or a regional-indicator flag), the first delta increments `largeAccumulatedChars` by the base character, and the delta that *completes* the cluster — if it contains only that completion — yields an empty `delta` and hits `guard !delta.isEmpty else { return }` before `largeAccumulated = newText` runs. The completed grapheme is then permanently absent from `largeAccumulated` and never renders. I reproduced this: streaming `"👩"` then `"👩\u{200D}"` then `"👩\u{200D}👧"` leaves `largeAccumulated == "👩"`; the family emoji is dropped. The same applies to `base + U+0300` and CP/CP/regional flag pairs when the completing fragment arrives alone. This is distinct from the small-stream path (which uses `Array(newText)` and is unaffected), it diverges from the O(delta) append invariant, and `testStreamingReconstructionExactForCJKAndEmoji` only exercises the static promotion arithmetic on a pre-baked corpus — it never drives `updateLargeTarget` with split graphemes, so the bug is uncovered.
**Suggestion:** Derive the delta and the accumulated bookkeeping from the tracked UTF-8 byte offset instead of `largeAccumulatedChars`, and reseed `largeAccumulated`/counters when the boundary falls inside a grapheme, or store the full `newText` and recompute the delta by a byte-accurate `hasPrefix` walk. Add a regression that appends a grapheme across multiple `updateLargeTarget` calls and asserts the reconstructed text matches the target.

### [severity: minor] `isStreaming` dropped before the large-document presentation
**File:** Conduit/Views/Components/MarkdownText.swift:41, Conduit/Views/Components/MarkdownLargeDocumentView.swift:117
**Issue:** `MarkdownText.body` routes large sources to `LargeMarkdownDocumentView` without forwarding `isStreaming`, and the collapsed preview builds its inner `MarkdownText` with `isStreaming` defaulting to `false`. For a streaming large message this would populate `MarkdownRenderCache` with a large entry (defeating the design intent of not caching live streams). Currently no production caller routes a live stream through `MarkdownText` for a large document (streaming crosses via `StreamingText.largeBody`), so this is latent — a future preview change would silently blow the cache.
**Suggestion:** Forward `isStreaming` into `LargeMarkdownDocumentView` and preserve it on the bounded preview's `MarkdownText`.

### [severity: nit] `alignedIndex` step-back loop isn't clamped to `utf8.count`
**File:** Conduit/Views/Components/StreamingText.swift:429
**Issue:** `var offset = utf8Offset` starts the decrementing loop unclamped; if a future caller passes an offset well past `string.utf8.count`, the loop performs many failing `utf8.index`/`String.Index` walks before returning nil. Current callers pass offsets ≤ `utf8.count`, so the loop runs at most a few times — defensive only.
**Suggestion:** `var offset = min(utf8Offset, utf8.count)` before the loop, preserving the existing backward-alignment and nil behavior.

### [severity: nit] Force unwrap on `columnPieceRanges`
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:819
**Issue:** `content.columnPieceRanges![column]` is guarded by `content.columnPieceRanges?.indices.contains(column) == true` on the preceding line, so the unwrap never fires under current construction. Stylistic only.
**Suggestion:** Bind the range once (e.g., `if let ranges = content.columnPieceRanges, ranges.indices.contains(column)`) to avoid the `!`.

## Verdict
COMMENT

The architecture is sound and nearly every issue raised across the six review rounds is genuinely resolved and regression-tested: the tail-window inversion, stale prepared-state clearing, structured-task cancellation, both DEBUG-counter data races (lock-guarded), the table `renderedRowCount` clamp, the Dynamic Type memo integration (`@Environment(\.sizeCategory)` now drives `attributedText`), per-piece callout/column descriptors and reference subsets, once-per-body `columnWidths`, and the bounded code `firstSlice`.

One finding merits attention before merge: the streaming `dropFirst(largeAccumulatedChars)` delta derivation can silently drop a grapheme cluster that is split across deltas (combining marks, ZWJ emoji, regional-indicator flags), which is a correctness regression in the very streaming feature this PR is meant to harden — and it is currently untested. The `isStreaming` forwarding is a worthwhile low-priority follow-up; the two nits are optional.

The finding at line 299 is the only substantive one; everything else verified clean.

<a href="https://opencode.ai/s/RfCuoODU">New%20session%20-%202026-08-20T08%3A56%3A31.257Z</a>
[opencode session](https://opencode.ai/s/RfCuoODU)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32351225961)

kaishi00:
The prepare/highlight tasks now run as structured tasks under `.task(id:)` with cancellation guards (per the author's later review notes), so that nit is resolved. The remaining findings are:

- `columnPieceRanges![column]` force unwrap — guarded by `indices.contains(column) == true` precondition, safe-by-construction nit.
- `alignedIndex` lacking the clamp — defensive-only nit.
- `isStreaming` not forwarded to `LargeMarkdownDocumentView` — latent minor.

All major/critical findings from the six review rounds have been addressed. Let me write the final review.

## Summary
This PR adds a bounded-rendering path for very large chat messages: a collapsed ≤16 KB preview with expand, off-main chunk planning, progressive ~8 KB memoized flow chunks, sliced/async-highlighted oversized code blocks, and an O(delta) append-only streaming mode above the 100 KB threshold. It reuses the existing parser/formatter/block renderer and preserves selection, references, Copy, list numbering, and Dynamic Type.

## Findings
Every critical/major finding raised across the six prior review rounds has been resolved and regression-tested in the head commit (`c067b35`): the streaming tail-window inversion, the stale `String.Index` re-derive and grapheme-merge content loss in `updateLargeTarget` (now via byte-boundary `largeAppendDelta` with full rescale), CJK/hard-cut promotion alignment, the scanner's quadratic cost, CRLF normalization, the table `renderedRowCount` index-out-of-range clamp, both DEBUG-counter data races (`parseSourceSizes`, `splitTextCallCount`), the Dynamic Type integration gap (`@Environment(\.sizeCategory)` now drives `LargeFlowChunkBox`), per-body `columnWidths`/`firstSlice` rescans, and structured-task cancellation. What remains are latent or cosmetic items with no reachable production defect.

### [severity: minor] `isStreaming` still not forwarded to the large-document preview
**File:** Conduit/Views/Components/MarkdownText.swift:41
**Issue:** `LargeMarkdownDocumentView` is constructed without `isStreaming`, so the collapsed preview's inner `MarkdownText` renders `isStreaming: false` and would populate `MarkdownRenderCache` with a large entry. Streaming currently crosses large mode via `StreamingText.largeBody` and never through `MarkdownText`, so there is no live bug — but a future preview change would silently defeat the no-giant-cache-entry design intent.
**Suggestion:** Forward `isStreaming` into `LargeMarkdownDocumentView` and preserve it on the bounded preview.

### [severity: nit] `alignedIndex` step-back loop is not clamped to `utf8.count`
**File:** Conduit/Views/Components/StreamingText.swift:464
**Issue:** `var offset = utf8Offset` enters the decrementing loop unclamped; a future caller passing an offset far past `utf8.count` would perform many failing UTF-8 walks before returning nil. All current callers pass offsets ≤ the end, so the loop runs at most a few times — defensive only.
**Suggestion:** `var offset = min(utf8Offset, utf8.count)` before the loop.

### [severity: nit] Force unwrap on `columnPieceRanges`
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:819
**Issue:** `content.columnPieceRanges![column]` is guarded by `content.columnPieceRanges?.indices.contains(column) == true` on the preceding ternary condition, so the unwrap is safe by construction. Stylistic only.
**Suggestion:** No change required, or bind the range once to avoid the `!`.

## Verdict
COMMENT

The head commit resolves the last substantive issue (the streaming grapheme-merge content loss) with byte-based delta derivation and a full rescale path plus regression tests. The remaining items are latent (`isStreaming` forwarding, worth a low-priority follow-up) or safe-by-construction nits. The PR is in good shape for merge.

<a href="https://opencode.ai/s/D788I5eB">New%20session%20-%202026-08-20T09%3A22%3A58.959Z</a>
[opencode session](https://opencode.ai/s/D788I5eB)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32353492500)

PR #89, opened 2026-08-20
perf(chat): bound transcript rendering work during streaming

kaishi00:
## Why

TestFlight `.ips` reports show `0x8BADF00D` scene-update watchdog terminations with the main thread inside SwiftUI (`ForEachState.estimatedCount` → `LazyStack.sizeThatFits`) — **with PR #88 already present**. #88 bounds pathological *individual* messages; the remaining failure mode is **cumulative transcript-level work**: with hundreds of settled Markdown messages and one active response, every streaming publish (~30 Hz) re-evaluated every mounted settled row through Markdown → SelectableTextView → TextKit, and several per-tick paths scaled with total transcript length or bytes.

This is a focused stabilization pass — no chat UX, Markdown architecture, viewport-controller, or protocol changes.

## The invariant

> Work caused by one streaming/layout tick must be proportional to the live/visible change — not to total transcript length or total transcript bytes.

## Changes (one commit per fix, reviewable in order)

| Commit | Fix |
|---|---|
| `c6d801b` | **Instrumentation** — DEBUG-only deterministic `TranscriptPerf` counters (settled Markdown/bubble body evals, STV updates/rebuilds, TextKit measurements, layout/transcript-change calls, stable-top scan size, fingerprinted messages/bytes). Zero-cost in Release. |
| `5c87460` | **Fix 1 — duplicate transcript reconciliation.** `ChatView` observed both `messages` and `chatTranscriptRevision`; `messages.didSet` publishes both, so every mutation ran `viewport.transcriptChanged` (and its fingerprint work) twice. The `messages` observer is now the single trigger; the revision stays in `AppState` for the settle handshake and rendered-scope versioning. |
| `6b400ee` | **Fix 2 — isolate settled presentation (highest value).** Settled content of every row type now lives behind `.equatable()` gates over immutable inputs (`message` + presentation values), so a streaming publish re-creating an identical row skips its body entirely. Read-aloud observation moved from every `AssistantBubble` to a 34pt `ReadAloudButton`; busy-state controls into `AssistantMessageActions`. Gateway `MEDIA:` resolution moved from a per-body-eval closure capturing `AppState` to a stable per-profile `GatewayMediaDataURLResolver` compared by identity. |
| `a79f009` | **Fix 3 — SelectableTextView memoization.** A presentation token (attributed content, font, color, spacing, line limit, wrapping, alignment, link color, self-sizing range) skips `configure()` entirely when unchanged — no styled copy, attribute enumeration, container mutation, or intrinsic invalidation. A `(presentation generation, width regime) → CGSize` cache on the coordinator skips TextKit `sizeThatFits` at unchanged width. Selection registration/unregistration and the coordinated-text-view swap stay on the fast path. |
| `5de803a` | **Fix 4 — stable-top O(rendered rows).** Rendered frames now carry their transcript order; stable-top picks the min-order frame intersecting the viewport instead of walking the full target list per geometry tick. Semantic ordering preserved exactly (min order, not `minY`). |
| `f19a623` | **Fix 5 — incremental fingerprints.** `ChatMessageScrollTargetCache.update` finds the longest common prefix of equal messages (value compares only), hashes only the changed suffix, and takes the incremental path only when duplicate-count semantics are provably local (no fingerprint crosses the prefix/suffix boundary in either direction; suffix internally duplicate-free). Duplicate-crossing appends, reorders, and full replacements fall back to the full rebuild. |
| `5a62cc2` | **Fixture** — 250 Markdown-heavy messages (headings, lists, quotes, links, moderate code; each well under the 100 KB threshold) + plain-text counterpart, hosting the real `ChatView`, driven by real `streamingText` publishes. Counters only, no wall-clock. |
| `605207f` | **Hardening pass** (review follow-ups, each verified before applying): Release-build compile fix (`Storage` type out of `#if DEBUG`; explicit Release build verified); text-view swap now fully configures the fresh `UITextView` (font/colors/line limits/wrapping/container/link attributes — previously skipped when the presentation gate held) with bidirectional swap tests; fixture `UIWindow` retained for the full measurement (was discarded — detached hierarchy) with explicit teardown; Dynamic Type is an explicit Equatable input fed by environment-reading shells (no unread-`@Environment` reliance) with a hosted regression test; the gateway resolver is `AppState`-owned and available from the first body pass, eliminating the nil → resolver double render; harness determinism (per-test counter resets, detach-then-flush teardown, index clamp, `XCTUnwrap`). |

## Measured result (deterministic counters, real ChatView hosted in a retained live window)

250 settled Markdown-heavy messages + one active streaming response, 10 streaming publishes:

| Path | Before (structural) | After (measured, live window) |
|---|---|---|
| Settled Markdown body evals | every mounted settled row × 10 (~60–80) | **0** |
| SelectableTextView updates | every mounted settled row per tick | **3** total (live row only) |
| Attributed-text rebuilds | every mounted settled row per tick | **1** total (live row only) |
| TextKit measurements | every revisited row | **1** total (measurement cache) |
| Stable-top scan per geometry tick | O(total messages) | **O(rendered frames)** (proven: 3 vs 500) |
| Fingerprints for append to 200-message transcript | 200 | **1** (tail edit: 1; unchanged: 0) |
| `transcriptChanged` per messages mutation | 2 | **1** |
| Settled renders on first appearance | 2 passes (nil → resolver sweep) | **1** |

Plain-text counterpart: same assertions green.

## Tests

- **34 new deterministic tests**: duplicate-trigger regression (2), settled isolation incl. gate-open cases + Dynamic Type regression (6), STV presentation/measurement cache incl. bidirectional swap configuration (7), stable-top bound (1), incremental-vs-full-rebuild reference incl. randomized 60-step seeded sweep (12), 250-message fixtures incl. first-appearance single-render (4) — plus `MarkdownSelectionCoordinator.isSegmentRegistered` as a behavior-level test accessor.
- **Full unit suite: 934 passed, 0 failed, 0 skipped** (single complete run on the final head).
- **Release configuration build verified green** (the instrumentation file could not compile in Release before `605207f`; CI Debug runs did not catch it).
- Focused suites on the final head: markdown rendering/selection/coordinator/table/reference/large-document/fallback 175/175; viewport/scroll/title/resume/incremental/reasoning 142/142; STV cache + settled isolation + fixture 16/16.

## Remaining transcript-size-dependent paths (all cheap value work, no parse/layout/measure)

- `Array(targets.enumerated())` allocation per `ChatView` body eval — O(count) struct copies.
- Common-prefix equality scan per mutation — O(prefix) compares, no hashing.
- `.equatable()` full-message compare per mounted row per tick — CoW-shared, bounded by the mounted subset.

Transcript windowing remains the documented next safety net if field reports persist on multi-thousand-message transcripts; the corrected live-window evidence says it is not needed for the reported failure mode.

## Manual validation

The automatable behaviors above are covered by hosted tests; a human pass on simulator/device before the next TestFlight is recommended (cold open of a long Markdown-heavy conversation, streaming while scrolling, cross-block selection, read aloud, Dynamic Type change with a settled transcript visible, profile switching, gateway `MEDIA:` images, row expand/collapse, background/foreground).

## Summary by CodeRabbit

  * Improved chat transcript updates, scrolling, and layout handling for large conversations.
  * Reduced unnecessary re-rendering during streaming, text updates, and app-state changes.
  * Improved selectable-text measurement and presentation through smarter caching.
  * Optimized incremental updates so new or changed messages require less processing.
  * Added performance tracking for transcript rendering and layout activity.

* **Reliability**
  * Improved media preview handling and gateway-hosted images.
  * Preserved stable scroll positioning during transcript changes.
  * Added safeguards for message edits, reordering, and duplicate content.
  * Ensured settled message content updates correctly when Dynamic Type settings change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ Package.resolved
 *.p8
 *.p12
 *.mobileprovision
+TestResults*.xcresult
+ci-logs/
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -333,6 +333,25 @@ final class AppState: ObservableObject {
     @Published var isConnecting = false
     @Published var profiles: [String] = []
     @Published private(set) var sessionFilterOrder: [SessionSource] = [.chat, .discord, .telegram, .api, .webhook, .other]
+    /// Stable per-profile gateway-media resolver for settled row content.
+    /// Created lazily on first read and reused while the active profile is
+    /// unchanged, so ChatView's first body pass already has a resolver
+    /// identity (no nil → resolver invalidation sweep over the settled
+    /// transcript) and rows' Equatable gates only open on genuine profile
+    /// changes. The resolver holds this AppState weakly, so caching it here
+    /// creates no retain cycle.
+    var gatewayMediaResolver: GatewayMediaDataURLResolver {
+        if let cached = cachedGatewayMediaResolver,
+           cached.profile == activeProfile {
+            return cached.resolver
+        }
+        let resolver = GatewayMediaDataURLResolver(appState: self, profile: activeProfile)
+        cachedGatewayMediaResolver = (activeProfile, resolver)
+        return resolver
+    }
+
+    private var cachedGatewayMediaResolver: (profile: String, resolver: GatewayMediaDataURLResolver)?
+
     @Published private(set) var activeProfile: String = "default" {
         didSet { refreshActiveChatScrollSessionIdentity() }
     }
--- a/Conduit/Services/ChatScrollState.swift
+++ b/Conduit/Services/ChatScrollState.swift
@@ -101,12 +101,23 @@ struct ChatRenderedScrollContent: Equatable {
 /// Global-space frame of one rendered stable message row, scoped to the
 /// rendered scroll scope that produced it. Only rows SwiftUI actually laid
 /// out report frames; this is how the viewport controller learns which
-/// stable row intersects the viewport without .scrollPosition.
+/// stable row intersects the viewport without .scrollPosition. `order` is
+/// the row's position in the transcript target list, so consumers can pick
+/// the semantic first visible row from the rendered subset alone — no scan
+/// of the full transcript.
 struct ChatRenderedRowFrame: Equatable {
     let id: String        // ChatMessageScrollTarget.id == message.id
     let minY: CGFloat
     let maxY: CGFloat
+    let order: Int
     let scope: ChatRenderedScrollScope
+}
+
+/// Frame + transcript order carried per rendered row inside the
+/// preference payload dictionaries.
+struct ChatRenderedRowGeometry: Equatable {
+    let frame: CGRect
+    let order: Int
 }
 
 /// A preference payload emitted only by targets SwiftUI has instantiated.
@@ -115,16 +126,17 @@ struct ChatRenderedRowFrame: Equatable {
 struct ChatRenderedScrollTargets: Equatable {
     private(set) var rowsByScope: [ChatRenderedScrollScope: Set<String>] = [:]
     private(set) var bottomsByScope: [ChatRenderedScrollScope: Set<String>] = [:]
-    private(set) var framesByScope: [ChatRenderedScrollScope: [String: CGRect]] = [:]
+    private(set) var framesByScope: [ChatRenderedScrollScope: [String: ChatRenderedRowGeometry]] = [:]
 
     static func row(
         semanticID: String,
         scope: ChatRenderedScrollScope,
-        frame: CGRect? = nil
+        frame: CGRect? = nil,
+        order: Int = 0
     ) -> ChatRenderedScrollTargets {
         var targets = ChatRenderedScrollTargets(rowsByScope: [scope: [semanticID]])
         if let frame {
-            targets.framesByScope = [scope: [semanticID: frame]]
+            targets.framesByScope = [scope: [semanticID: ChatRenderedRowGeometry(frame: frame, order: order)]]
         }
         return targets
     }
@@ -165,9 +177,9 @@ struct ChatRenderedScrollTargets: Equatable {
         bottomsByScope[scope]?.contains(anchorID) == true
     }
 
-    /// Global frames of rendered stable rows for a scope (rows that reported
-    /// geometry this pass; offscreen lazy rows are absent).
-    func rowFrames(in scope: ChatRenderedScrollScope) -> [String: CGRect] {
+    /// Global frames + transcript order of rendered stable rows for a scope
+    /// (rows that reported geometry this pass; offscreen lazy rows are absent).
+    func rowFrames(in scope: ChatRenderedScrollScope) -> [String: ChatRenderedRowGeometry] {
         framesByScope[scope] ?? [:]
     }
 
--- a/Conduit/Services/ChatScrollState.swift
+++ b/Conduit/Services/ChatScrollState.swift
@@ -61,6 +61,8 @@ struct ChatMessageScrollTargetCache: Equatable {
     @discardableResult
     mutating func update(for messages: [ChatMessage]) -> ChatMessageScrollTargetCacheUpdate {
         let updatedFingerprints = ChatMessageScrollTargets.fingerprints(for: messages)
+        TranscriptPerf.lastFingerprintedMessageCount = messages.count
+        TranscriptPerf.lastFingerprintedByteCount = messages.reduce(0) { $0 + $1.content.utf8.count + ($1.code?.utf8.count ?? 0) }
         if updatedFingerprints == fingerprints, targets.count == messages.count {
             guard targets.map(\.message) != messages else { return .unchanged }
             targets = zip(messages, targets).map { message, target in
--- a/Conduit/Services/ChatScrollState.swift
+++ b/Conduit/Services/ChatScrollState.swift
@@ -60,22 +60,78 @@ struct ChatMessageScrollTargetCache: Equatable {
 
     @discardableResult
     mutating func update(for messages: [ChatMessage]) -> ChatMessageScrollTargetCacheUpdate {
-        let updatedFingerprints = ChatMessageScrollTargets.fingerprints(for: messages)
-        TranscriptPerf.lastFingerprintedMessageCount = messages.count
-        TranscriptPerf.lastFingerprintedByteCount = messages.reduce(0) { $0 + $1.content.utf8.count + ($1.code?.utf8.count ?? 0) }
-        if updatedFingerprints == fingerprints, targets.count == messages.count {
-            guard targets.map(\.message) != messages else { return .unchanged }
-            targets = zip(messages, targets).map { message, target in
+        // Longest common prefix of equal messages: plain value compares,
+        // no hashing, no intermediate allocations.
+        var commonPrefix = 0
+        while commonPrefix < targets.count,
+              commonPrefix < messages.count,
+              targets[commonPrefix].message == messages[commonPrefix] {
+            commonPrefix += 1
+        }
+
+        // Identical transcripts: no work at all.
+        if commonPrefix == targets.count, commonPrefix == messages.count {
+            TranscriptPerf.lastFingerprintedMessageCount = 0
+            TranscriptPerf.lastFingerprintedByteCount = 0
+            return .unchanged
+        }
+
+        // Hash only the changed suffix.
+        let suffixStart = commonPrefix
+        let suffixMessages = Array(messages[suffixStart...])
+        let suffixFingerprints = ChatMessageScrollTargets.fingerprints(for: suffixMessages)
+        TranscriptPerf.lastFingerprintedMessageCount = suffixMessages.count
+        TranscriptPerf.lastFingerprintedByteCount = Self.fingerprintedBytes(of: suffixMessages)
+
+        // Same length and identical suffix fingerprints: a rendering-only
+        // replacement (equal semantics, different message objects). Swap the
+        // message values in place; semantic IDs and restoration metadata are
+        // untouched, so duplicate semantics cannot shift.
+        if targets.count == messages.count,
+           suffixFingerprints.elementsEqual(fingerprints[suffixStart...]) {
+            let replacement = zip(suffixMessages, targets[suffixStart...]).map { message, target in
                 ChatMessageScrollTarget(
                     message: message,
                     semanticID: target.semanticID,
                     restorationMetadata: target.restorationMetadata
                 )
             }
+            targets.replaceSubrange(suffixStart..., with: replacement)
             renderingRevision &+= 1
             return .renderingChanged
         }
 
+        // Incremental semantic rebuild of the suffix is safe only when
+        // duplicate-count semantics are provably local to the suffix: no
+        // fingerprint may cross the prefix/suffix boundary in either
+        // direction (old or new), and the suffix itself must be
+        // duplicate-free. Otherwise fall back to a full rebuild — correctness
+        // over exotic incremental cases.
+        let prefixFingerprints = Set(fingerprints[..<suffixStart])
+        let oldSuffixFingerprints = fingerprints[suffixStart...]
+        let canRebuildSuffixIncrementally =
+            suffixFingerprints.allSatisfy { !prefixFingerprints.contains($0) }
+            && oldSuffixFingerprints.allSatisfy { !prefixFingerprints.contains($0) }
+            && Set(suffixFingerprints).count == suffixFingerprints.count
+
+        if canRebuildSuffixIncrementally {
+            let suffixTargets = ChatMessageScrollTargets.make(
+                for: suffixMessages,
+                fingerprints: suffixFingerprints
+            )
+            fingerprints.replaceSubrange(suffixStart..., with: suffixFingerprints)
+            targets.replaceSubrange(suffixStart..., with: suffixTargets)
+            renderingRevision &+= 1
+            return .semanticsChanged
+        }
+
+        // Full rebuild fallback: mutation could affect duplicate-count
+        // semantics anywhere in the transcript.
+        let updatedFingerprints = commonPrefix == 0
+            ? suffixFingerprints
+            : ChatMessageScrollTargets.fingerprints(for: messages)
+        TranscriptPerf.lastFingerprintedMessageCount = messages.count
+        TranscriptPerf.lastFingerprintedByteCount = Self.fingerprintedBytes(of: messages)
         fingerprints = updatedFingerprints
         targets = ChatMessageScrollTargets.make(
             for: messages,
@@ -83,6 +139,10 @@ struct ChatMessageScrollTargetCache: Equatable {
         )
         renderingRevision &+= 1
         return .semanticsChanged
+    }
+
+    private static func fingerprintedBytes(of messages: [ChatMessage]) -> Int {
+        messages.reduce(0) { $0 + $1.content.utf8.count + ($1.code?.utf8.count ?? 0) }
     }
 }
 
--- a/Conduit/Services/ChatViewportController.swift
+++ b/Conduit/Services/ChatViewportController.swift
@@ -712,21 +712,21 @@ struct ChatViewportController: Equatable {
         )
     }
 
+    /// Stable-top detection operates purely over the rendered frames the
+    /// layout pass supplied: filter to rows intersecting the viewport, then
+    /// take the smallest transcript order. Work is O(rendered rows) — never
+    /// a scan of the full transcript target list, which in a deep
+    /// conversation is hundreds of entries per geometry tick.
     private mutating func updateStableTopMessage(rowFrames: [ChatRenderedRowFrame]) {
         guard let viewportMinY, let viewportMaxY else {
             stableTopMessageID = nil
             return
         }
-        let framesByID = Dictionary(rowFrames.map { ($0.id, $0) }) { first, _ in first }
-        TranscriptPerf.stableTopScanTargetCount = targetCache.targets.count
-        for target in targetCache.targets {
-            guard let frame = framesByID[target.id] else { continue }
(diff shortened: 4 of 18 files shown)


kaishi00:
I have enough context. Let me now also verify the Dynamic Type claim and how fonts flow, and confirm the concern about whether `.equatable()` blocks the environment-driven re-eval definitively by considering the existing pattern. Actually, I have enough for the review.

Let me reconsider the two main findings and their confidence levels:

1. **STV swap skips configure (major)** — High confidence. The code path is clear: swap bumps generation but doesn't clear `appliedPresentation`; unchanged presentation short-circuits `configure` on a freshly-created unconfigured text view.

2. **Dynamic Type regression via equatable gate (major)** — Medium-high confidence. The `.equatable()` gate's `==` ignores `sizeCategory`, so sizeCategory changes won't re-evaluate the settled content. The PR explicitly claims it's handled, but the code contradicts it.

Let me reconsider #2 once more. There's genuine debate about SwiftUI EquatableView + environment. Some sources indicate that `.equatable()` compares including environment-dependent values that the body actually reads... No. The standard behavior: EquatableView's diff uses the view's `Equatable` conformance. The custom `==` here ignores sizeCategory. Environment changes are tracked by the graph but the gate intercepts the update.

However, I should hedge appropriately: I'll present it as a real risk that contradicts the PR's claim, recommending a dynamic-type test. Since there's no dynamic-type test in the suite and the existing `UserMessageContent`/`SettledAssistantMessageContent` `==` ignores it, this is a valid finding.

Actually — wait. Let me reconsider one thing about EquatableView and environment more carefully, because if I'm wrong I want to be calibrated. 

When you use `.equatable()` on a view, SwiftUI wraps it in `EquatableView`. `EquatableView` stores the previous *value* (the wrapped view's value). On each update of the graph node, it computes the new value and calls `==`. If equal, it does NOT call `body`. 

The environment is a dependency. When `sizeCategory` changes, the graph marks the node dirty. On the next frame, EquatableView's Update compares old vs new value of the wrapped view type (`SettledAssistantMessageContent`). Its `==` compares message/displayName/avatarURL/resolver — all equal → skip body. So body is not re-evaluated, and the `@Environment(\.sizeCategory)` captured in the stored/previous body is stale. Dynamic Type won't re-size.

Yes — this is the well-known gotcha. Confirmed enough to report as major.

Actually, I realize I should double check: does the previous body's subviews (MarkdownText etc.) have their own environment dependency on sizeCategory that bypasses the gate? No — if parent body isn't re-evaluated, the subview hierarchy isn't re-projected, so no.

OK. I'm confident on both. Let me also scan for any force-unwrap issues in the new code. The new code in ChatView / SelectableTextView / resolvers — I didn't see new `!`. The `.min(by:)?.id` uses optional chaining. Good.

Let me also verify one concurrency concern: `GatewayMediaDataURLResolver` is `@MainActor`. `gatewayMediaDataURL` (the resolver method) is async. Rows call it in `.task`. Fine.

Let me also double-check the `CopyResponse` `Task` in `AssistantMessageActions` - `Task { try? await Task.sleep... }` captures `self` (the view struct) strongly in a Task that isn't cancelled... Actually SwiftUI Views are value types; adding to a Task captures a copy. The `Task` isn't tied to view lifetime (not `.task` on a view), so if the view disappears the task still runs and sets `copied = false` on a stale copy - harmless. Not a leak of anything significant. Skip.

I think I have my findings. Let me write the review. Let me also verify the `SettledAssistantMessageContent` reads sizeCategory but the equality ignores it — confirmed at lines 1091-1103.

Let me finalize. Severity: I'll make the STV swap configure bug **major**, the Dynamic Type **major**, and add a minor on the equatable gate not comparing sizeCategory being the cause.

Let me quickly double check whether the STV swap bug is actually reachable in practice — when would a block transition plain↔coordinated with unchanged presentation? The selection coordinator is shared globally (MarkdownSelectionChromeRoot). A block becomes coordinated when a selection involves it. This swap with unchanged attributedText is the normal case for settled blocks. So yes, reachable and visible.

I'm confident. Writing the review.

## Summary
This PR bounds transcript rendering work during streaming: isolates settled rows behind equatable gates, memoizes SelectableTextView presentation/sizing, bounds stable-top scanning to rendered frames, incrementalizes fingerprints, and removes duplicate transcript reconciliation.

## Findings

### [severity: major] Fresh text view swap skips `configure`, leaving the mounted view unstyled
**File:** Conduit/Views/Components/SelectableTextView.swift:169-198
**Issue:** When a block transitions between the plain and coordinated text view (`uiView.isUsingCoordinatedTextView != needsCoordinatedTextView`), the code swaps in a brand-new `UITextView`/`MarkdownSelectionTextView` copying only `attributedText` and `selectedRange`. It bumps `presentationGeneration` but does **not** clear `coordinator.appliedPresentation`. If the presentation inputs are unchanged (the normal case when attaching selection to an already-settled block), the new presentation gates to "unchanged" and `configure(textView)` is skipped — so the fresh view is never given its `font`, `textColor`, `textContainer` sizing/limits, `linkTextAttributes`, or paragraph style. It renders with `configureBaseTextView` defaults. Before this PR, `configure` was called unconditionally after the swap, so this is a regression introduced by the memoization (Fix 3).
**Suggestion:** On swap, force `configure` for the fresh view — e.g. set `coordinator.appliedPresentation = nil` (and re-bump generation) in the swap branch so the gate re-opens. Add a test that swaps plain↔coordinated with unchanged presentation and asserts the mounted `font`/paragraph styling.

### [severity: major] Dynamic Type no longer re-sizes settled rows (equatable gate ignores `sizeCategory`)
**File:** Conduit/Views/ChatView.swift:1091-1103 (and UserMessageContent 848-856)
**Issue:** `SettledAssistantMessageContent`/`UserMessageContent` read `@Environment(\.sizeCategory)` but their custom `==` compares only `message`, `displayName`, `avatarURL`, and `gatewayResolver` (identity) — not `sizeCategory`. With `.equatable()`, when Dynamic Type changes, the inputs compare equal, the gate short-circuits, and the body (which resolves `UIFont.preferredFont(...)` inside MarkdownText) is never re-evaluated. Result: settled transcript rows keep the old font size after a Dynamic Type change, contradicting the PR's claim that "Dynamic Type still invalidates via `\.sizeCategory` reads inside the gated views." The counter-only test suite cannot detect this — there is no Dynamic Type test.
**Suggestion:** Include `sizeCategory` in the `==` so the gate re-opens on Dynamic Type change, or test/rely on a font-read that bypasses the gate. Add a test that re-hosts a gate with a different `sizeCategory` and asserts re-render.

### [severity: minor] `.equatable()` gate means a settled row cannot respond to non-message presentation changes beyond the three listed inputs
**File:** Conduit/Views/ChatView.swift:1100-1103
**Issue:** The gate only re-opens for message/displayName/avatarURL/resolver identity changes. Any other environment-driven presentation change (e.g. contrast, other environment traits read by MarkdownText) will likewise be suppressed for settled rows. This is a refinement of the Dynamic Type finding: the gate is a hard stop on all non-listed dependencies.
**Suggestion:** Broaden the `==` to cover the environment traits the gated subtree actually consumes, or explicitly enumerate/document which environment changes must invalidate and include them.

## Verdict
REQUEST_CHANGES

(Findings 1 and 2 are verifiable regressions; no force-unwrap, retain-cycle, or concurrency issues found in the new code — the fingerprint incremental logic, stable-top min-by-order, and resolver identity design are correct.)

<a href="https://opencode.ai/s/KxdRRDK4">New%20session%20-%202026-08-20T23%3A10%3A44.066Z</a>
[opencode session](https://opencode.ai/s/KxdRRDK4)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32427450270)

kaishi00:
The index clamp (`max(0, ...)`) is fixed. All the coderabbit critical/major findings and the original review's two majors are addressed in `605207f`.

Remaining genuinely-open items are the three low-severity perf/maintainability nits on `ChatScrollState.swift`:
- Full-rebuild path re-hashes the suffix (lines 130-133) — but only on rare fallback cases (duplicate-crossing/reorder/full replace).
- Prefix `Set` build O(prefix) per mutation (lines 110) — for tail appends (the common streaming case), `commonPrefix > 0`, so the decision path reaches line 110 and builds `Set(fingerprints[..<suffixStart])` over all settled messages each tick. This is exactly the O(total transcript) work the PR is meant to eliminate.
- `order: Int = 0` default (line 195) — but sole production call site always supplies `order`.

Let me double-check the common streaming case: a streaming response means the last message is being edited, so `messages` changes at the tail. `commonPrefix` will equal `targets.count`/close to it (all settled messages equal, last changed). The rendering-only check at line 90 requires `targets.count == messages.count`. During streaming the last message content changes but count stays same — so actually for a pure content edit of the last message: suffixFingerprints would be `[fingerprint(last)]`, `fingerprints[suffixStart...]` = `[fingerprint(old last)]`. These disagree → falls through to line 104. Then builds `Set(fingerprints[..<suffixStart])` = all 250 settled fingerprints → builds a Set of N strings each streaming tick. That IS the O(N) per-tick work coderabbit flagged at line 104-115.

Wait — but is this path actually hit, or does the `.equatable()` isolation mean messages object doesn't change? During streaming, `appState.messages` last element content changes each tick → new message value → `.onChange(of: appState.messages)` fires → `viewport.transcriptChanged` → `targetCache.update(for: messages)` → reaches the incremental decision with `commonPrefix` large. So yes, the `Set(fingerprints[..<suffixStart])` of ~N elements is built per streaming tick. That's the coderabbit "poor tradeoff" note that was NOT marked addressed.

However — building a Set of N strings is O(N) hashing without the expensive fingerprinting (no CryptKit/string hashing of content, just a Set build of already-computed strings). It's cheap memcpy-ish value work on N strings, not the expensive per-message fingerprint/parse/measure. The PR's "remaining transcript-size-dependent paths" list explicitly acknowledges "Common-prefix equality scan per mutation — O(prefix) compares" but doesn't mention this Set build. Still, it's O(total transcript) per tick, which is what the invariant says should be avoided — though it's cheap value work. This is worth a minor finding.

The full-rebuild double-hash (lines 130-133) is also a genuine minor inefficiency but only in rare fallback cases.

These are minor, not blockers. The core acceptance criterion (no settled Markdown parse/measure/layout) is met. Let me write the review.

Overall verdict: APPROVE with minor notes. All critical/major issues from prior reviews are fixed. Remaining are minor perf/robustness nits.

## Summary
This PR bounds transcript rendering work during streaming by isolating settled rows behind equatable gates (with explicit Dynamic Type inputs), memoizing SelectableTextView presentation/measurement, bounding stable-top scanning to rendered frames, incrementalizing fingerprints, removing duplicate transcript reconciliation, and adding deterministic instrumentation plus fixture tests.

## Findings

### [severity: minor] Per-tick `Set(fingerprints[..<suffixStart])` rebuild is O(total transcript)
**File:** Conduit/Services/ChatScrollState.swift:110-114
**Issue:** During streaming the tail message changes each tick, so `commonPrefix` is large and control reaches the incremental-safety branch, which rebuilds `Set(fingerprints[..<suffixStart])` — inserting one entry per settled message on every mutation (~N string hashes/tick for the 250-message fixture). This is the O(total-transcript) per-tick class of work the PR otherwise removes; the PR's "remaining transcript-size-dependent paths" list omits this. It's cheap value work (no content hashing/parse/measure), so it doesn't violate the acceptance criterion, but it scales with transcript length.
**Suggestion:** Maintain an incremental fingerprint-count map (`[String: Int]`) next to `fingerprints`, updated wherever the array is mutated, and use it for prefix membership checks so validation is O(suffix).

### [severity: minor] Full-rebuild fallback hashes the suffix twice and under-reports
**File:** Conduit/Services/ChatScrollState.swift:130-133
**Issue:** When `commonPrefix > 0`, the fallback calls `ChatMessageScrollTargets.fingerprints(for: messages)` over the whole array even though `suffixFingerprints` was already computed at line 82, then reports `lastFingerprintedMessageCount = messages.count` while the real work is `messages.count + suffixMessages.count`. Only affects the rare duplicate-crossing/reorder/replacement fallback.
**Suggestion:** Reuse the unchanged prefix fingerprints with the already-computed suffix, and report the true amount of hashing work.

### [severity: nit] `order` defaults to 0 in `ChatRenderedScrollTargets.row`
**File:** Conduit/Services/ChatScrollState.swift:195
**Issue:** `order: Int = 0` means a call site that omits order silently claims transcript position 0 in stable-top selection. The sole production call site (ChatView.swift:157) always passes `index` with a frame, so not reachable today.
**Suggestion:** Pair `frame` and `order` into a single optional so the two cannot diverge.

No retain cycles, force unwraps, or concurrency violations found in the new code. The resolver uses a weak `appState` to avoid a cycle; all closures capture value-typed views or weak coordinator references; instrumentation is `@MainActor` and DEBUG-gated with the `Storage` type hoisted out of `#if DEBUG` (Release-safe).

## Verdict
APPROVE

The two previously-reported major issues (STV swap skipping `configure` on a fresh text view, and Dynamic Type not re-opening the equatable settled gate) are both correctly addressed in `605207f`: the swap branch now clears `appliedPresentation`/`cachedMeasurement` so the fresh view is configured, and `sizeCategory` is read in the `AssistantBubble` shell and passed as an explicit Equatable input included in `==`, with a hosted regression test. The Release-build `Storage` fix, resolver first-pass availability, and fixture harness determinism (window retention, setUp/tearDown counter reset, `XCTUnwrap`, index clamp) all verify. Remaining findings are minor perf/robustness nits on the fallback paths, not blockers.

<a href="https://opencode.ai/s/T2Fn1mez">New%20session%20-%202026-08-20T23%3A51%3A12.967Z</a>
[opencode session](https://opencode.ai/s/T2Fn1mez)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32430267584)

kaishi00:
The sole production call site (ChatView.swift:153-158) always passes both `frame` and `order` together. The `order: Int = 0` default is not reachable in production, and `order`/`frame` are always paired at the call site.

I have completed my review. All the major and critical findings from prior reviews are verifiably addressed in the final head:

1. **STV swap skipping `configure`** — Fixed at SelectableTextView.swift:185-187 (clears `appliedPresentation` and `cachedMeasurement` on swap), so a fresh text view is fully configured.
2. **Dynamic Type equatable gate** — Fixed: `sizeCategory` is read via `@Environment(\.sizeCategory)` in shell views (ChatView.swift:814/1241/1486/1678) and passed as an explicit `ContentSizeCategory` Equatable input included in `==` (lines 842-848, 1095-1103, etc.).
3. **Resolver nil on first body pass** — Fixed: `AppState.gatewayMediaResolver` (AppState.swift:343-351) is AppState-owned, lazily created per-profile, identity-stable, holds `AppState` weakly (no cycle).
4. **Release build compile** — Fixed: `Storage` type hoisted outside `#if DEBUG` in TranscriptPerformanceInstrumentation.swift:121-137.
5. **All test harness items** — Fixed: window retention, `setUp`/`tearDown` counter resets, index clamp, `XCTUnwrap`.

## Summary
This PR bounds transcript rendering work during streaming by isolating settled rows behind equatable gates (with explicit Dynamic Type + resolver inputs), memoizing SelectableTextView presentation/sizing, bounding stable-top scans to rendered frames, incrementalizing fingerprints, and removing duplicate transcript reconciliation — all verified against the final head.

## Findings
All critical/major issues from prior reviews are resolved at the final head. Two acknowledged low-severity perf/robustness nits remain, both non-blocking.

### [severity: minor] Per-tick `Set(fingerprints[..<suffixStart])` rebuild is O(total transcript)
**File:** Conduit/Services/ChatScrollState.swift:110-115
**Issue:** During streaming the tail message changes each tick while `messages.count` is stable, so `commonPrefix` is large and control reaches the incremental-safety branch, which rebuilds `Set(fingerprints[..<suffixStart])` — inserting one entry per settled message on every mutation (~N string hashes/tick for the 250-message fixture). This is O(total-transcript) per-tick work, the same class the PR removes elsewhere (and omitted from the PR's "remaining transcript-size-dependent paths" list). It's cheap value work (no content hashing/parse/measure), so the acceptance criterion (no settled parse/measure/layout) still holds.
**Suggestion:** Maintain an incremental `[String: Int]` fingerprint-count map beside `fingerprints`, updated wherever the array is mutated, and use it for prefix membership so validation is O(suffix).

### [severity: minor] Full-rebuild fallback re-hashes the suffix
**File:** Conduit/Services/ChatScrollState.swift:130-134
**Issue:** When `commonPrefix > 0`, the fallback calls `ChatMessageScrollTargets.fingerprints(for: messages)` over the whole array even though `suffixFingerprints` was already computed at line 82, and reports `messages.count` while the true work is larger. Only affects the rare duplicate-crossing/reorder/replacement fallback.
**Suggestion:** Reuse the unchanged prefix fingerprints with the already-computed suffix, and report the actual hashing work. `ChatRenderedScrollTargets.row` (`order: Int = 0`) defaults safely: the sole production call site (ChatView.swift:153-158) always supplies both `frame` and `order`.

No retain cycles, force unwraps, or concurrency violations found in the new code: the resolver holds `AppState` weakly and is `@MainActor`; the fingerprint incremental and stable-top min-by-order logic are correct; instrumentation is DEBUG-gated with zero Release cost and `Storage` hoisted for Release compilation.

## Verdict
APPROVE

The previously-reported major issues (STV swap skipping `configure`, and Dynamic Type no longer re-opening the equatable settled gate) are both correctly fixed in `605207f`: the swap branch clears `appliedPresentation`/`cachedMeasurement`, and `sizeCategory` is read in the shell and passed as an explicit Equatable input included in `==`, with a hosted regression test. The resolver-first-pass, Release-build `Storage`, and fixture-harness determinism fixes all verify. The remaining items are minor perf/robustness nits on fallback paths, not blockers.

<a href="https://opencode.ai/s/rT9vaHOC">New%20session%20-%202026-08-21T02%3A57%3A27.809Z</a>
[opencode session](https://opencode.ai/s/rT9vaHOC)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32441579398)

PR #92, opened 2026-08-22
feat: add native Kanban support

kaishi00:
## Summary

Native Kanban for Conduit: lane-chip + vertical-card-list board, task lifecycle (create/edit/move/delete/comments/reassign/reclaim), board picker with client-local selection, and Settings-hosted Capabilities. HTTP stays on the existing DashboardTicketBridge; HermesClient and all session/chat transport are untouched.

## Correctness safeguards

- **Concrete board identity pinned before fetch**: after GET /boards (and invalid-selection validation), one resolved slug drives both GET /board?board=<slug> and the recorded snapshot identity, so "server current" can never drift between the two requests.
- **Non-interactive stale snapshot during navigation**: while selected != loaded, old cards are read-only cached content (store rejects mutations with an explicit error; view disables cards and New Task) with a loading indicator.
- **Stale-server mutation isolation**: mutations capture an immutable context plus a generation-owned UI token; completions that lose ownership after a dashboard switch never refresh, flip flags, or surface errors - from either the store or view handlers.
- **Capabilities request/profile ownership**: loads return a request-scoped CapabilityLoadOutcome; monotonic generations reject A->B->A commits, a snapshot-profile gate masks foreign rows, and a view token keeps older completions from clearing newer loading state.
- **Locked lanes**: review/running/scheduled match upstream LOCKED_COLUMNS - visible, backend-controlled, never offered as create/move destinations.
- **Dispatcher nudges**: 400 ms debounced, fire-and-forget after successful writes; failures are non-events.
- **Detail polling preserves drafts**: comments/runs/metadata update live; unsaved title/body/status edits win, with a remote-change notice.
- **Fail-closed transport**: dot-segment ids and unencodable board/id values reject before transport; board scope is always an explicit query parameter.
- **Tolerant decoding**: exact integers only (2.5 -> nil); id-less task rows dropped so duplicate SwiftUI identities cannot occur.

## Shared cross-profile Kanban

Hermes Kanban is intentionally shared across profiles: kanban_db anchors boards at the shared Hermes root (profile homes resolve back via get_default_hermes_root()), making it the cross-profile coordination primitive. Conduit therefore sends no profile parameter, does not reload on UI profile switches, and scopes persisted board selection by server/root identity only.

## Validation

Full ios_test run at this head: **1010 test cases passed, 0 failed** across 58 discovered suites / 8 serialized batches (0 hangs, stalls, or uncovered suites). Includes focused regressions for every safeguard above.

## Deferred

- Live /events WebSocket support (polling only in this PR).

### DIFF
--- a/Conduit/Models/KanbanModels.swift
+++ b/Conduit/Models/KanbanModels.swift
@@ -12,7 +12,12 @@ private extension KeyedDecodingContainer {
 
     func decodeLossyInt(forKey key: Key) -> Int? {
         if let value = try? decode(Int.self, forKey: key) { return value }
-        if let value = try? decode(Double.self, forKey: key) { return Int(value) }
+        // Doubles must be finite and exactly representable; a fractional or
+        // huge value decodes as nil instead of silently truncating (or
+        // crashing on architectures where the cast traps).
+        if let value = try? decode(Double.self, forKey: key), value.isFinite, let exact = Int(exactly: value.rounded(.towardZero)) {
+            return exact
+        }
         if let value = try? decode(String.self, forKey: key) { return Int(value) }
         return nil
     }
--- a/Conduit/Models/KanbanModels.swift
+++ b/Conduit/Models/KanbanModels.swift
@@ -15,7 +15,10 @@ private extension KeyedDecodingContainer {
         // Doubles must be finite and exactly representable; a fractional or
         // huge value decodes as nil instead of silently truncating (or
         // crashing on architectures where the cast traps).
-        if let value = try? decode(Double.self, forKey: key), value.isFinite, let exact = Int(exactly: value.rounded(.towardZero)) {
+        if let value = try? decode(Double.self, forKey: key) {
+            // Exact integers only: finite AND integral. Fractional values
+            // (2.5), NaN, and overflow decode as nil instead of truncating.
+            guard value.isFinite, let exact = Int(exactly: value) else { return nil }
             return exact
         }
         if let value = try? decode(String.self, forKey: key) { return Int(value) }
@@ -233,7 +236,11 @@ struct KanbanColumn: Codable, Equatable, Identifiable {
     init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         name = (try? container.decode(String.self, forKey: .name)) ?? "unknown"
-        tasks = (try? container.decode([KanbanTask].self, forKey: .tasks)) ?? []
+        // Lossily decode every row, but DROP rows without an identity: two
+        // id-less tasks would collide on "" inside SwiftUI ForEach and the
+        // service rejects empty IDs anyway, so they can never be actionable.
+        let rawTasks = (try? container.decodeIfPresent([KanbanTask].self, forKey: .tasks)) ?? []
+        tasks = rawTasks.filter { !$0.id.isEmpty }
     }
 }
 
--- a/Conduit/Models/KanbanModels.swift
+++ b/Conduit/Models/KanbanModels.swift
@@ -1,0 +1,765 @@
+import Foundation
+
+// MARK: - Defensive JSON helpers
+
+private extension KeyedDecodingContainer {
+    func decodeLossyString(forKey key: Key) -> String? {
+        if let value = try? decode(String.self, forKey: key) { return value }
+        if let value = try? decode(Int.self, forKey: key) { return String(value) }
+        if let value = try? decode(Double.self, forKey: key) { return String(value) }
+        return nil
+    }
+
+    func decodeLossyInt(forKey key: Key) -> Int? {
+        if let value = try? decode(Int.self, forKey: key) { return value }
+        if let value = try? decode(Double.self, forKey: key) { return Int(value) }
+        if let value = try? decode(String.self, forKey: key) { return Int(value) }
+        return nil
+    }
+}
+
+// MARK: - Board and task models
+
+struct KanbanTaskWarnings: Codable, Equatable {
+    var count: Int = 0
+    var highestSeverity: String?
+
+    enum CodingKeys: String, CodingKey {
+        case count
+        case highestSeverity = "highest_severity"
+    }
+}
+
+struct KanbanLinkCounts: Codable, Equatable {
+    var parents: Int = 0
+    var children: Int = 0
+}
+
+struct KanbanProgress: Codable, Equatable {
+    var done: Int = 0
+    var total: Int = 0
+}
+
+struct KanbanDiagnosticAction: Codable, Equatable {
+    var kind: String
+    var label: String
+    var payload: [String: AnyCodable]?
+    var suggested: Bool?
+}
+
+struct KanbanDiagnostic: Codable, Equatable {
+    var kind: String
+    var severity: String
+    var title: String
+    var detail: String
+    var actions: [KanbanDiagnosticAction]
+    var count: Int
+    var lastSeenAt: Int?
+    var data: [String: AnyCodable]?
+
+    enum CodingKeys: String, CodingKey {
+        case kind, severity, title, detail, actions, count, data
+        case lastSeenAt = "last_seen_at"
+    }
+}
+
+/// The small, UI-facing slice of a Kanban task returned by the backend.
+/// Unknown backend fields are intentionally ignored so schema additions remain safe.
+struct KanbanTask: Codable, Identifiable, Equatable {
+    let id: String
+    var title: String
+    var body: String?
+    var status: String
+    var assignee: String?
+    var priority: Int?
+    var tenant: String?
+    var createdAt: Int?
+    var latestSummary: String?
+    var commentCount: Int?
+    var linkCounts: KanbanLinkCounts?
+    var progress: KanbanProgress?
+    var warnings: KanbanTaskWarnings?
+    var startedAt: Int?
+    var workerPid: Int?
+    var lastHeartbeatAt: Int?
+
+    // Detail-only fields. Keeping them on the same value type lets the board
+    // and drawer share one decoder while preserving the backend's flat shape.
+    var result: String?
+    var createdBy: String?
+    var modelOverride: String?
+    var providerOverride: String?
+    var reasoningEffort: String?
+    var completedAt: Int?
+    var lastFailureError: String?
+    var workspaceKind: String?
+    var workspacePath: String?
+    var branchName: String?
+    var consecutiveFailures: Int?
+    var diagnostics: [KanbanDiagnostic]?
+
+    enum CodingKeys: String, CodingKey {
+        case id, title, body, status, assignee, priority, tenant
+        case createdAt = "created_at"
+        case latestSummary = "latest_summary"
+        case commentCount = "comment_count"
+        case linkCounts = "link_counts"
+        case progress, warnings
+        case startedAt = "started_at"
+        case workerPid = "worker_pid"
+        case lastHeartbeatAt = "last_heartbeat_at"
+        case result
+        case createdBy = "created_by"
+        case modelOverride = "model_override"
+        case providerOverride = "provider_override"
+        case reasoningEffort = "reasoning_effort"
+        case completedAt = "completed_at"
+        case lastFailureError = "last_failure_error"
+        case workspaceKind = "workspace_kind"
+        case workspacePath = "workspace_path"
+        case branchName = "branch_name"
+        case consecutiveFailures = "consecutive_failures"
+        case diagnostics
+    }
+
+    init(
+        id: String,
+        title: String,
+        body: String? = nil,
+        status: String,
+        assignee: String? = nil,
+        priority: Int? = nil,
+        tenant: String? = nil,
+        createdAt: Int? = nil,
+        latestSummary: String? = nil,
+        commentCount: Int? = nil,
+        linkCounts: KanbanLinkCounts? = nil,
+        progress: KanbanProgress? = nil,
+        warnings: KanbanTaskWarnings? = nil,
+        startedAt: Int? = nil,
+        workerPid: Int? = nil,
+        lastHeartbeatAt: Int? = nil,
+        result: String? = nil,
+        createdBy: String? = nil,
+        modelOverride: String? = nil,
+        providerOverride: String? = nil,
+        reasoningEffort: String? = nil,
+        completedAt: Int? = nil,
+        lastFailureError: String? = nil,
+        workspaceKind: String? = nil,
+        workspacePath: String? = nil,
+        branchName: String? = nil,
+        consecutiveFailures: Int? = nil,
+        diagnostics: [KanbanDiagnostic]? = nil
+    ) {
+        self.id = id
+        self.title = title
+        self.body = body
+        self.status = status
+        self.assignee = assignee
+        self.priority = priority
+        self.tenant = tenant
+        self.createdAt = createdAt
+        self.latestSummary = latestSummary
+        self.commentCount = commentCount
+        self.linkCounts = linkCounts
+        self.progress = progress
+        self.warnings = warnings
+        self.startedAt = startedAt
+        self.workerPid = workerPid
+        self.lastHeartbeatAt = lastHeartbeatAt
+        self.result = result
+        self.createdBy = createdBy
+        self.modelOverride = modelOverride
+        self.providerOverride = providerOverride
+        self.reasoningEffort = reasoningEffort
+        self.completedAt = completedAt
+        self.lastFailureError = lastFailureError
+        self.workspaceKind = workspaceKind
+        self.workspacePath = workspacePath
+        self.branchName = branchName
+        self.consecutiveFailures = consecutiveFailures
+        self.diagnostics = diagnostics
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = container.decodeLossyString(forKey: .id) ?? ""
+        title = (try? container.decode(String.self, forKey: .title)) ?? "Untitled task"
+        body = try? container.decodeIfPresent(String.self, forKey: .body)
+        status = (try? container.decode(String.self, forKey: .status)) ?? "todo"
+        assignee = try? container.decodeIfPresent(String.self, forKey: .assignee)
+        priority = container.decodeLossyInt(forKey: .priority)
+        tenant = try? container.decodeIfPresent(String.self, forKey: .tenant)
+        createdAt = container.decodeLossyInt(forKey: .createdAt)
+        latestSummary = try? container.decodeIfPresent(String.self, forKey: .latestSummary)
+        commentCount = container.decodeLossyInt(forKey: .commentCount)
+        linkCounts = try? container.decodeIfPresent(KanbanLinkCounts.self, forKey: .linkCounts)
+        progress = try? container.decodeIfPresent(KanbanProgress.self, forKey: .progress)
+        warnings = try? container.decodeIfPresent(KanbanTaskWarnings.self, forKey: .warnings)
+        startedAt = container.decodeLossyInt(forKey: .startedAt)
+        workerPid = container.decodeLossyInt(forKey: .workerPid)
+        lastHeartbeatAt = container.decodeLossyInt(forKey: .lastHeartbeatAt)
+        result = try? container.decodeIfPresent(String.self, forKey: .result)
+        createdBy = try? container.decodeIfPresent(String.self, forKey: .createdBy)
+        modelOverride = try? container.decodeIfPresent(String.self, forKey: .modelOverride)
+        providerOverride = try? container.decodeIfPresent(String.self, forKey: .providerOverride)
+        reasoningEffort = try? container.decodeIfPresent(String.self, forKey: .reasoningEffort)
+        completedAt = container.decodeLossyInt(forKey: .completedAt)
+        lastFailureError = try? container.decodeIfPresent(String.self, forKey: .lastFailureError)
+        workspaceKind = try? container.decodeIfPresent(String.self, forKey: .workspaceKind)
+        workspacePath = try? container.decodeIfPresent(String.self, forKey: .workspacePath)
+        branchName = try? container.decodeIfPresent(String.self, forKey: .branchName)
+        consecutiveFailures = container.decodeLossyInt(forKey: .consecutiveFailures)
+        diagnostics = try? container.decodeIfPresent([KanbanDiagnostic].self, forKey: .diagnostics)
+    }
+}
+
+struct KanbanColumn: Codable, Equatable, Identifiable {
+    var name: String
+    var tasks: [KanbanTask]
+    var id: String { name }
+
+    init(name: String, tasks: [KanbanTask] = []) {
+        self.name = name
+        self.tasks = tasks
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        name = (try? container.decode(String.self, forKey: .name)) ?? "unknown"
+        tasks = (try? container.decode([KanbanTask].self, forKey: .tasks)) ?? []
+    }
+}
+
+struct KanbanBoard: Codable, Equatable {
+    var columns: [KanbanColumn]
+    var tenants: [String]
+    var assignees: [String]
+    var latestEventID: Int?
+    var now: Int?
+
+    enum CodingKeys: String, CodingKey {
+        case columns, tenants, assignees
+        case latestEventID = "latest_event_id"
+        case now
+    }
+
+    init(columns: [KanbanColumn], tenants: [String] = [], assignees: [String] = [], latestEventID: Int? = nil, now: Int? = nil) {
+        self.columns = columns
+        self.tenants = tenants
+        self.assignees = assignees
+        self.latestEventID = latestEventID
+        self.now = now
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        columns = (try? container.decode([KanbanColumn].self, forKey: .columns)) ?? []
+        tenants = (try? container.decode([String].self, forKey: .tenants)) ?? []
+        assignees = (try? container.decode([String].self, forKey: .assignees)) ?? []
+        latestEventID = container.decodeLossyInt(forKey: .latestEventID)
+        now = container.decodeLossyInt(forKey: .now)
+    }
+}
+
+struct KanbanBoardMetadata: Codable, Identifiable, Equatable {
+    let slug: String
+    var name: String?
+    var description: String?
+    var isCurrent: Bool?
+    var total: Int?
+    var defaultWorkdir: String?
+    var defaultWorkspaceKind: String?
+    var projectID: String?
+    var projectName: String?
+
+    var id: String { slug }
+
+    enum CodingKeys: String, CodingKey {
+        case slug, name, description
+        case isCurrent = "is_current"
+        case total
+        case defaultWorkdir = "default_workdir"
+        case defaultWorkspaceKind = "default_workspace_kind"
+        case projectID = "project_id"
+        case projectName = "project_name"
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        slug = container.decodeLossyString(forKey: .slug) ?? "default"
+        name = try? container.decodeIfPresent(String.self, forKey: .name)
+        description = try? container.decodeIfPresent(String.self, forKey: .description)
+        isCurrent = try? container.decodeIfPresent(Bool.self, forKey: .isCurrent)
+        total = container.decodeLossyInt(forKey: .total)
+        defaultWorkdir = try? container.decodeIfPresent(String.self, forKey: .defaultWorkdir)
+        defaultWorkspaceKind = try? container.decodeIfPresent(String.self, forKey: .defaultWorkspaceKind)
+        projectID = try? container.decodeIfPresent(String.self, forKey: .projectID)
+        projectName = try? container.decodeIfPresent(String.self, forKey: .projectName)
+    }
+
+    init(slug: String, name: String? = nil, description: String? = nil, isCurrent: Bool? = nil, total: Int? = nil, defaultWorkdir: String? = nil, defaultWorkspaceKind: String? = nil, projectID: String? = nil, projectName: String? = nil) {
+        self.slug = slug
+        self.name = name
+        self.description = description
+        self.isCurrent = isCurrent
+        self.total = total
+        self.defaultWorkdir = defaultWorkdir
+        self.defaultWorkspaceKind = defaultWorkspaceKind
+        self.projectID = projectID
+        self.projectName = projectName
+    }
+}
+
+struct KanbanBoardsResponse: Codable, Equatable {
+    var boards: [KanbanBoardMetadata]
+    var current: String
+
+    init(boards: [KanbanBoardMetadata] = [], current: String = "default") {
+        self.boards = boards
+        self.current = current
+    }
+}
+
+// MARK: - Detail collections
+
+struct KanbanComment: Codable, Identifiable, Equatable {
+    let id: String
+    var taskID: String?
+    var author: String
+    var body: String
+    var createdAt: Int?
+
+    enum CodingKeys: String, CodingKey {
+        case id
+        case taskID = "task_id"
+        case author, body
+        case createdAt = "created_at"
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = container.decodeLossyString(forKey: .id) ?? UUID().uuidString
+        taskID = try? container.decodeIfPresent(String.self, forKey: .taskID)
+        author = (try? container.decode(String.self, forKey: .author)) ?? "Hermes"
+        body = (try? container.decode(String.self, forKey: .body)) ?? ""
+        createdAt = container.decodeLossyInt(forKey: .createdAt)
+    }
+}
+
+struct KanbanEvent: Codable, Identifiable, Equatable {
+    let id: String
+    var taskID: String?
+    var kind: String
+    var payload: AnyCodable?
+    var createdAt: Int?
+    var runID: String?
+
+    enum CodingKeys: String, CodingKey {
+        case id
+        case taskID = "task_id"
+        case kind, payload
+        case createdAt = "created_at"
+        case runID = "run_id"
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = container.decodeLossyString(forKey: .id) ?? UUID().uuidString
+        taskID = try? container.decodeIfPresent(String.self, forKey: .taskID)
+        kind = (try? container.decode(String.self, forKey: .kind)) ?? "event"
+        payload = try? container.decodeIfPresent(AnyCodable.self, forKey: .payload)
+        createdAt = container.decodeLossyInt(forKey: .createdAt)
+        runID = container.decodeLossyString(forKey: .runID)
+    }
+}
+
+struct KanbanAttachment: Codable, Identifiable, Equatable {
+    let id: String
+    var taskID: String?
+    var filename: String
+    var contentType: String?
+    var size: Int?
+    var uploadedBy: String?
+    var createdAt: Int?
+
+    enum CodingKeys: String, CodingKey {
+        case id
+        case taskID = "task_id"
+        case filename
+        case contentType = "content_type"
+        case size
+        case uploadedBy = "uploaded_by"
+        case createdAt = "created_at"
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = container.decodeLossyString(forKey: .id) ?? UUID().uuidString
+        taskID = try? container.decodeIfPresent(String.self, forKey: .taskID)
+        filename = (try? container.decode(String.self, forKey: .filename)) ?? "Attachment"
+        contentType = try? container.decodeIfPresent(String.self, forKey: .contentType)
+        size = container.decodeLossyInt(forKey: .size)
+        uploadedBy = try? container.decodeIfPresent(String.self, forKey: .uploadedBy)
+        createdAt = container.decodeLossyInt(forKey: .createdAt)
+    }
+}
+
+struct KanbanRun: Codable, Identifiable, Equatable {
+    let id: String
+    var taskID: String?
+    var profile: String?
+    var stepKey: String?
+    var status: String
+    var outcome: String?
+    var summary: String?
+    var error: String?
+    var metadata: AnyCodable?
+    var workerPID: Int?
+    var startedAt: Int?
+    var endedAt: Int?
+
+    enum CodingKeys: String, CodingKey {
+        case id
+        case taskID = "task_id"
+        case profile
+        case stepKey = "step_key"
+        case status, outcome, summary, error, metadata
+        case workerPID = "worker_pid"
+        case startedAt = "started_at"
+        case endedAt = "ended_at"
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = container.decodeLossyString(forKey: .id) ?? UUID().uuidString
+        taskID = container.decodeLossyString(forKey: .taskID)
+        profile = try? container.decodeIfPresent(String.self, forKey: .profile)
+        stepKey = try? container.decodeIfPresent(String.self, forKey: .stepKey)
+        status = (try? container.decode(String.self, forKey: .status)) ?? "unknown"
+        outcome = try? container.decodeIfPresent(String.self, forKey: .outcome)
+        summary = try? container.decodeIfPresent(String.self, forKey: .summary)
+        error = try? container.decodeIfPresent(String.self, forKey: .error)
+        metadata = try? container.decodeIfPresent(AnyCodable.self, forKey: .metadata)
+        workerPID = container.decodeLossyInt(forKey: .workerPID)
+        startedAt = container.decodeLossyInt(forKey: .startedAt)
+        endedAt = container.decodeLossyInt(forKey: .endedAt)
+    }
+}
+
+struct KanbanTaskDetail: Codable, Equatable {
+    var task: KanbanTask
+    var comments: [KanbanComment]
+    var events: [KanbanEvent]
+    var attachments: [KanbanAttachment]
+    var links: KanbanTaskLinks
+    var childResults: [KanbanChildResult]
+    var runs: [KanbanRun]
+
+    init(task: KanbanTask, comments: [KanbanComment] = [], events: [KanbanEvent] = [], attachments: [KanbanAttachment] = [], links: KanbanTaskLinks = KanbanTaskLinks(), childResults: [KanbanChildResult] = [], runs: [KanbanRun] = []) {
+        self.task = task
+        self.comments = comments
+        self.events = events
+        self.attachments = attachments
+        self.links = links
+        self.childResults = childResults
+        self.runs = runs
+    }
+
+    enum CodingKeys: String, CodingKey {
+        case task, comments, events, attachments, links
+        case childResults = "child_results"
+        case runs
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        task = (try? container.decode(KanbanTask.self, forKey: .task)) ?? KanbanTask(id: "", title: "Untitled task", status: "todo")
+        comments = (try? container.decode([KanbanComment].self, forKey: .comments)) ?? []
+        events = (try? container.decode([KanbanEvent].self, forKey: .events)) ?? []
+        attachments = (try? container.decode([KanbanAttachment].self, forKey: .attachments)) ?? []
+        links = (try? container.decode(KanbanTaskLinks.self, forKey: .links)) ?? KanbanTaskLinks()
+        childResults = (try? container.decode([KanbanChildResult].self, forKey: .childResults)) ?? []
+        runs = (try? container.decode([KanbanRun].self, forKey: .runs)) ?? []
+    }
+}
+
+struct KanbanTaskLinks: Codable, Equatable {
+    var parents: [String]
+    var children: [String]
+    init(parents: [String] = [], children: [String] = []) {
+        self.parents = parents
+        self.children = children
+    }
+}
+
+struct KanbanChildResult: Codable, Equatable, Identifiable {
+    let id: String
+    var title: String?
+    var status: String?
+    var latestSummary: String?
+    var result: String?
+
+    enum CodingKeys: String, CodingKey {
+        case id, title, status
+        case latestSummary = "latest_summary"
+        case result
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        id = container.decodeLossyString(forKey: .id) ?? UUID().uuidString
+        title = try? container.decodeIfPresent(String.self, forKey: .title)
+        status = try? container.decodeIfPresent(String.self, forKey: .status)
+        latestSummary = try? container.decodeIfPresent(String.self, forKey: .latestSummary)
+        result = try? container.decodeIfPresent(String.self, forKey: .result)
+    }
+}
+
+// MARK: - Auxiliary board data
+
+struct KanbanProfile: Codable, Identifiable, Equatable {
+    var name: String
+    var isDefault: Bool
+    var description: String
+    var descriptionAuto: Bool
+    var model: String?
+    var provider: String?
+    var skillCount: Int?
+    var id: String { name }
+
+    enum CodingKeys: String, CodingKey {
+        case name
+        case isDefault = "is_default"
+        case description
+        case descriptionAuto = "description_auto"
+        case model, provider
+        case skillCount = "skill_count"
+    }
+}
+
+struct KanbanProject: Codable, Identifiable, Equatable {
+    var id: String
+    var slug: String
+    var name: String
+    var primaryPath: String?
+    var icon: String?
+    var color: String?
+
+    enum CodingKeys: String, CodingKey {
+        case id, slug, name, icon, color
+        case primaryPath = "primary_path"
+    }
+}
+
+struct KanbanOrchestrationSettings: Codable, Equatable {
+    var orchestratorProfile: String
+    var defaultAssignee: String
+    var autoDecompose: Bool
+    var autoPromoteChildren: Bool?
+    var resolvedOrchestratorProfile: String
+    var resolvedDefaultAssignee: String
+
+    enum CodingKeys: String, CodingKey {
+        case orchestratorProfile = "orchestrator_profile"
+        case defaultAssignee = "default_assignee"
+        case autoDecompose = "auto_decompose"
+        case autoPromoteChildren = "auto_promote_children"
+        case resolvedOrchestratorProfile = "resolved_orchestrator_profile"
+        case resolvedDefaultAssignee = "resolved_default_assignee"
+    }
+}
+
+struct KanbanTaskEstimate: Codable, Equatable {
+    var ok: Bool
+    var reason: String?
+    var estimatedTokens: Int?
+    var complexity: String?
+    var rationale: String?
+    var model: String?
+
+    enum CodingKeys: String, CodingKey {
+        case ok, reason
+        case estimatedTokens = "est_tokens"
+        case complexity, rationale, model
+    }
+}
+
+struct KanbanWorkerLog: Codable, Equatable {
+    var exists: Bool
+    var sizeBytes: Int
+    var content: String
+    var truncated: Bool
+
+    enum CodingKeys: String, CodingKey {
+        case exists
+        case sizeBytes = "size_bytes"
+        case content, truncated
+    }
+}
+
+// MARK: - Request bodies
+
+struct KanbanCreateTaskRequest: Encodable, Equatable {
+    var title: String
+    var body: String?
+    var assignee: String?
+    var tenant: String?
+    var priority: Int
+    var workspaceKind: String
+    var workspacePath: String?
+    var parents: [String]
+    var triage: Bool
+    var idempotencyKey: String?
+    var maxRuntimeSeconds: Int?
+    var skills: [String]?
+    var goalMode: Bool
+    var goalMaxTurns: Int?
+    var modelOverride: String?
+    var providerOverride: String?
+    var reasoningEffort: String?
+    var projectID: String?
+
+    enum CodingKeys: String, CodingKey {
+        case title, body, assignee, tenant, priority
+        case workspaceKind = "workspace_kind"
+        case workspacePath = "workspace_path"
+        case parents, triage
+        case idempotencyKey = "idempotency_key"
+        case maxRuntimeSeconds = "max_runtime_seconds"
+        case skills
+        case goalMode = "goal_mode"
+        case goalMaxTurns = "goal_max_turns"
+        case modelOverride = "model_override"
+        case providerOverride = "provider_override"
+        case reasoningEffort = "reasoning_effort"
+        case projectID = "project_id"
+    }
+
+    init(title: String, body: String? = nil, assignee: String? = nil, tenant: String? = nil, priority: Int = 0, workspaceKind: String = "scratch", workspacePath: String? = nil, parents: [String] = [], triage: Bool = false, idempotencyKey: String? = nil, maxRuntimeSeconds: Int? = nil, skills: [String]? = nil, goalMode: Bool = false, goalMaxTurns: Int? = nil, modelOverride: String? = nil, providerOverride: String? = nil, reasoningEffort: String? = nil, projectID: String? = nil) {
+        self.title = title
+        self.body = body
+        self.assignee = assignee
+        self.tenant = tenant
+        self.priority = priority
+        self.workspaceKind = workspaceKind
+        self.workspacePath = workspacePath
+        self.parents = parents
+        self.triage = triage
+        self.idempotencyKey = idempotencyKey
+        self.maxRuntimeSeconds = maxRuntimeSeconds
+        self.skills = skills
+        self.goalMode = goalMode
+        self.goalMaxTurns = goalMaxTurns
+        self.modelOverride = modelOverride
+        self.providerOverride = providerOverride
+        self.reasoningEffort = reasoningEffort
+        self.projectID = projectID
+    }
+}
+
+struct KanbanTaskPatch: Encodable, Equatable {
+    var status: String?
+    var assignee: String?
+    var priority: Int?
+    var title: String?
+    var body: String?
+    var result: String?
+    var blockReason: String?
+    var summary: String?
+    var metadata: [String: AnyCodable]?
+    var modelOverride: String?
+    var providerOverride: String?
+    var clearModelOverride: Bool
+    var reasoningEffort: String?
+    var clearReasoningEffort: Bool
+
+    enum CodingKeys: String, CodingKey {
+        case status, assignee, priority, title, body, result
+        case blockReason = "block_reason"
+        case summary, metadata
+        case modelOverride = "model_override"
+        case providerOverride = "provider_override"
+        case clearModelOverride = "clear_model_override"
+        case reasoningEffort = "reasoning_effort"
+        case clearReasoningEffort = "clear_reasoning_effort"
+    }
+
+    init(status: String? = nil, assignee: String? = nil, priority: Int? = nil, title: String? = nil, body: String? = nil, result: String? = nil, blockReason: String? = nil, summary: String? = nil, metadata: [String: AnyCodable]? = nil, modelOverride: String? = nil, providerOverride: String? = nil, clearModelOverride: Bool = false, reasoningEffort: String? = nil, clearReasoningEffort: Bool = false) {
+        self.status = status
+        self.assignee = assignee
+        self.priority = priority
+        self.title = title
+        self.body = body
+        self.result = result
+        self.blockReason = blockReason
+        self.summary = summary
+        self.metadata = metadata
+        self.modelOverride = modelOverride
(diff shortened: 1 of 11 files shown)


kaishi00:
## Summary
This PR replaces the sidebar's Capabilities destination with a native Kanban board (typed models, KanbanStore/KanbanService backed by the authenticated `DashboardTicketBridge`), migrates the persisted sidebar tab value, and relocates the Capabilities browser into Settings while preserving AppState-driven loading and mutations.

## Findings

### [major] `reload` can leave `isLoading` permanently `true`, wedging the Kanban tab in an infinite spinner
**File:** Conduit/Services/KanbanStore.swift:49-62
**Issue:** `reload` increments `loadGeneration`, then bails early if `service == nil` — after a *previous* in-flight reload may have already set `isLoading = true`. The early-return branch never resets `isLoading`, and the superseded load's `defer { if generation == loadGeneration { isLoading = false } }` is skipped because the generation no longer matches. Scenario: a board load is in flight when the dashboard bridge goes nil (`.task(id: bridgeIdentity)` fires, `configure(nil)` runs, new reload hits the `guard let service` return). The tab then shows `store.isLoading && store.board == nil` → "Loading board…" forever (KanbanViews.swift:37), with no error surfaced.
**Suggestion:** In the `guard let service` else-branch, set `isLoading = false` (and clear `loadGeneration`/generation bookkeeping) before returning, or set `isLoading = true` *after* the guard so the early return never leaves it true from a prior generation.

### [minor] Mutation failures are swallowed silently
**File:** Conduit/Views/KanbanViews.swift:201-216
**Issue:** `move(_:to:)` and `delete(_:)` catch `error` and do nothing, and the Kafka: `KanbanStore.updateTask`/`deleteTask` throw the service error *before* calling `reload`, so `store.errorMessage` is never populated. The comment "the store surfaces the mutation error" is therefore incorrect — a failed move or delete gives the user zero feedback.
**Suggestion:** Set `errorMessage` on mutation failure (e.g., in the catch of `move`/`delete`, or have the store assign `errorMessage` when the mutation throws), so failures are visible in the overlay used elsewhere in the view.

### [minor] Card and ellipsis controls are not accessibility-accessible
**File:** Conduit/Views/KanbanViews.swift:356, 298-318, 523-525
**Issue:** The board card is opened via `.onTapGesture` but exposes no button trait/action to VoiceOver (the inner labels are separate elements, so the tap-to-open affordance is invisible to assistive tech). The card's `ellipsis` menu and the detail sheet's toolbar `Menu` have image-only labels with no `.accessibilityLabel`. The board picker/refresh/new-task buttons and the Archived toggle are correctly labelled, which makes this inconsistency notable.
**Suggestion:** Add `.accessibilityElement(children: .combine)` + `.accessibilityAddTraits(.isButton)` (and a `.accessibilityHint`) to the card, and `accessibilityLabel`s to the two ellipsis menus.

### [minor] Move-to menu omits custom/unknown statuses, making some tasks immovable from the card
**File:** Conduit/Views/KanbanViews.swift:281-285
**Issue:** `statusOptions` is built from `knownStatuses` (minus `running`) only. A task whose status is a server-defined custom value (all unknown statuses fall back through `forStatus` to `sortOrder: 100`) cannot be moved back to its own column via the card menu. The detail picker works because it appends the current status, so the two UIs disagree about what is movable.
**Suggestion:** Include unknown statuses present on the board (union of column names / task statuses) in the card's move menu, or persist "other" status options alongside `knownStatuses`.

### [minor] `KanbanStore` strongly retains the bridge (and its WKWebView) after AppState releases it
**File:** Conduit/Services/KanbanStore.swift:37-47
**Issue:** `configure` stores `service` which strongly holds the `any DashboardJSONRequester`. When AppState disconnects (`dashboardTicketBridge = nil`), the store keeps the invalidated bridge, its WKWebView, and script-message handlers alive until `KanbanView` is torn down (tab switch / sheet dismissal). Not a cycle, but a heavy transient hold.
**Suggestion:** In the `guard let requester else` branch the service is already nilled; also consider clearing the existing service when `appState.dashboardTicketBridge` goes nil via the next `.task(id:)` cycle (it does), so this is mainly a note to keep tab teardown prompt or drop the webview on invalidation.

### [minor] Task creation results in two full board reloads (≈10 HTTP calls)
**File:** Conduit/Views/KanbanViews.swift:442-453 and Conduit/Services/KanbanStore.swift:124-137
**Issue:** For non-"todo" target columns, `save()` calls `store.createTask(...)` (which reloads board + profiles + projects + orchestration) and then a second `store.updateTask(...)` which triggers a second full reload — 8-10 requests for a single create. Additionally, if the create response omits `task` (optional), the second status patch is skipped and the task stays "todo" despite being created in another column.
**Suggestion:** Batch mutations and do a single `reload` after the final mutation; and/or include the target status in the create request rather than a follow-up PATCH.

### [minor] Archived toggle is not persisted
**File:** Conduit/Views/KanbanViews.swift:9
**Issue:** `includeArchived` is a plain `@State` defaulting to false, so the preference resets every time the sidebar/tab is reopened — inconsistent with board selection which is persisted in `UserDefaults`.
**Suggestion:** Back `includeArchived` with `@AppStorage` (e.g., `conduit.kanbanIncludeArchived`).

### [nit] `createBoard`/`updateBoard` are unreachable dead code
**File:** Conduit/Services/KanbanService.swift:146-177
**Issue:** Both methods are public on the service but no caller exists in `KanbanStore` or any view, so the service's board-creation surface is untested/unexercised.
**Suggestion:** Either wire them into the board-picker menu (create/edit board) or remove them for now.

### [nit] `RelativeDateTimeFormatter` allocated per comment row
**File:** Conduit/Views/KanbanViews.swift:713-718
**Issue:** `relativeDate` constructs a formatter on every call inside `ForEach(comments)`. Cheap, but trivially hoistable.
**Suggestion:** Use a lazily-created static/shared formatter.

## Verdict
REQUEST_CHANGES

Strong, defensive PR overall — `@MainActor` isolation is consistent, no force unwraps in new code, board selection correctly stays client-local, and the sidebar/`SidebarTab` raw-value migration is explicit and well-tested. The two blocking items are the `KanbanStore.reload` early-return leaving `isLoading` stuck in a race (major), and the silent swallowing of move/delete mutation errors together with the accessibility gaps on the primary card interaction; the rest are follow-ups.

<a href="https://opencode.ai/s/NlYh6wj6">New%20session%20-%202026-08-22T00%3A45%3A38.797Z</a>
[opencode session](https://opencode.ai/s/NlYh6wj6)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32541303776)

kaishi00:
I have reviewed the current HEAD of the PR (`c80927a` + hardening commits `92473b9`, `b4e8cc6`), verifying each prior review comment against the actual code. Findings below reflect the state in HEAD, not earlier commits (several earlier findings — mutation error surfacing via `store.showMutationError`, the `includeArchived` observer placement — are already fixed).

## Summary
This PR replaces the sidebar's Capabilities tab with a native Kanban workspace (new typed models, `KanbanStatus` presentation, `KanbanStore`, and `KanbanService` over the existing `DashboardTicketBridge`), migrates the persisted `SidebarTab` raw value, and moves the Capabilities browser into Settings. Strong overall shape: consistent `@MainActor` isolation, explicit sidebar raw-value migration, client-local board selection, and all 58 tests passing.

## Findings

### [critical] `decodeLossyInt` traps on out-of-range/non-finite `Double` values — crashes the app
**File:** Conduit/Models/KanbanModels.swift:15
**Issue:** `Int(value)` is the non-failable `Int(_: Double)` initializer. It traps at runtime when the double is NaN, infinite, or outside `Int` range. A backend field such as `created_at: 1e20` or `priority: 9e18` decodes as `Double` (Int decode fails first), then `Int(value)` traps and hard-crashes the process. This defeats the file's entire tolerant-decoding purpose. The helper feeds every numeric field (`createdAt`, `priority`, `workerPid`, `lastHeartbeatAt`, `completedAt`, `commentCount`, `board total`, `size`, …), so one odd server value crashes the app.
**Suggestion:** Use a guarded conversion, e.g. `if let value = try? decode(Double.self, forKey: key), value.isFinite, value >= Double(Int.min), value <= Double(Int.max) { return Int(value) }`, or `Int(exactly:)`.

### [major] Task-detail polling overwrites unsaved edits in the editor
**File:** Conduit/Views/KanbanViews.swift:581-590 and 721-739
**Issue:** `.task(id: initialTask.id)` polls `loadDetail()` every 4 seconds, and `loadDetail` unconditionally assigns `title = loaded.task.title`, `taskBody = loaded.task.body ?? ""`, and `status = loaded.task.status` back into the `@State` bindings. The guard only checks `isSaving`/`isAddingComment`, not whether the user is currently editing or has changed the Status picker. So while a user types in the Title/body fields or selects a new status without saving, the next poll (≤4s) silently wipes their draft and resets the picker — unsaved data loss.
**Suggestion:** Gate polling writes on field focus / dirty state (e.g. `@FocusState`, or track a "dirty" flag reset on save) and skip `loadDetail` while any editor field is being edited.

### [minor] `reload`'s early returns can silently drop required reloads (latent loading wedge)
**File:** Conduit/Services/KanbanStore.swift:98-106
**Issue:** `guard !isLoading else { return }` swallows any reload request while a load is in flight, and the `guard let service else` branch returns without touching `isLoading`. The specific "infinite spinner" scenario from the earlier review is largely mitigated because every `configure` path that bumps `loadGeneration` also resets `isLoading = false` — but a `.task(id: pollingKey)` restart triggered while the cancelled predecessor's `isLoading` is still `true` (cancellation not yet processed) makes the new `reload` bail at line 98, so an archived-toggle or board-selection change does nothing until the next 8s poll, and the stale-board window is visible. The `guard let service` else-branch also sets `errorMessage` without a generation check, so a superseded load can still stamp the error text.
**Suggestion:** Set `isLoading` after the service guard (or reset it in the else branch), and consider making reload coalesce via the generation number instead of bailing at the `!isLoading` gate so a pending request is never dropped.

### [minor] Double network round-trip and duplicate-create risk in task creation
**File:** Conduit/Views/KanbanViews.swift:479-503 and Conduit/Services/KanbanStore.swift:205-233
**Issue:** For non-todo/triage columns, creating a task issues `POST /tasks` followed by a `PATCH` status update (two HTTP round-trips) plus one full reload per mutation — roughly twice the requests a single create needs. If the create response omits `task`, the guard throws `taskCreatedButMoveFailed` and the sheet stays open; the user can tap "Create" again, which creates a duplicate since `idempotencyKey` is never populated.
**Suggestion:** Set a stable `idempotencyKey` for each save attempt and/or dismiss the sheet after a successful create, reporting the status-move failure separately instead of as a full failure.

### [minor] Kanban cards expose no accessibility button affordance; ellipsis menus are unlabeled
**File:** Conduit/Views/KanbanViews.swift:350-356, 394-395, 572-579
**Issue:** The card's tap-to-open is `.onTapGesture` with no `.accessibilityAddTraits(.isButton)` or hint, so VoiceOver users cannot discover the open action. The card's `ellipsis` `Menu` (with Move/Delete) and the detail sheet's `ellipsis.circle` toolbar `Menu` (Delete) have image-only labels with no `.accessibilityLabel`. This is inconsistent with the board picker/refresh/new-task buttons, which are all labelled.
**Suggestion:** Add `.accessibilityElement(children: .combine)` + `.isButton` trait + hint to the card, and `accessibilityLabel`s to both ellipsis menus.

### [minor] Card move menu omits custom/unknown statuses, so some tasks can't be moved back to their column
**File:** Conduit/Views/KanbanViews.swift:322-324
**Issue:** `statusOptions` is built only from `manuallySelectableStatuses` (the nine known statuses). A task in a server-defined custom column (unknown status → `forStatus` fallback, `isManuallySelectable == false`) cannot be returned to its own column from the card; only the detail picker (which inserts the current status) can. The two UIs disagree about what is movable.
**Suggestion:** Include the current column/task status (and other statuses present on the board) in the card's move menu, mirroring the detail picker.

### [minor] Empty-`id`/random-`UUID` fallbacks break `Identifiable` identity across reloads
**File:** Conduit/Models/KanbanModels.swift:187, 343, 369, 399, 436, 512
**Issue:** A task missing `id` decodes to `""`, colliding in `ForEach(column.tasks)`; `KanbanComment`/`KanbanEvent`/`KanbanAttachment`/`KanbanRun`/`KanbanChildResult` all fall back to a fresh `UUID().uuidString` per decode. Since the detail view re-decodes after every save/comment/poll, every row gets a new identity and SwiftUI rebuilds all rows each reload.
**Suggestion:** Drop id-less tasks at board decode time (as `KanbanService` already rejects them), and derive deterministic fallback identifiers from stable fields (e.g. `"\(taskID ?? "")-\(createdAt ?? 0)-\(author)"`).

### [minor] `KanbanStore` strongly retains the bridge (and its WKWebView) after disconnect
**File:** Conduit/Services/KanbanStore.swift:45-58, 72
**Issue:** `service` strongly holds the `any DashboardJSONRequester`. When AppState releases `dashboardTicketBridge`, the store keeps the invalidated bridge, its WKWebView, and script-message handlers alive until `KanbanView` is torn down. Not a cycle, but a heavy transient hold.
**Suggestion:** Clear the existing service in the `guard let requester else` branch (it already nils it), and/or drop references promptly when the identity changes.

### [minor] `withBoard` silently drops the board query parameter on encoding failure
**File:** Conduit/Services/KanbanService.swift:264-278
**Issue:** `compactMap` removes any key whose `encodedQueryComponent` returns nil, and `pathComponent` falls back to the raw value. If `board` (or a task/board slug) encoding ever fails, the request is sent unscoped — a mutation would target the server's current board, breaking the "board always sent locally" contract. Practically unreachable with the current character set, but the silent failure-mode violates the file's own stated contract.
**Suggestion:** Throw instead of dropping the key when encoding fails, and remove the raw-value fallback.

### [nit] `addComment` silently succeeds for a whitespace-only body
**File:** Conduit/Services/KanbanService.swift:136-137
**Issue:** The `guard !trimmed.isEmpty else { return }` returns success without sending anything or erroring. The caller then clears the text field and reloads, so the user's "comment" disappears with no explanation.
**Suggestion:** Throw a validation error for whitespace-only bodies (the caller already pre-validates, so this only hides the earlier defect).

### [nit] Auxiliary load failures are silently swallowed
**File:** Conduit/Services/KanbanStore.swift:126-129
**Issue:** `try? service.fetchProfiles/Projects/Orchestration` leaves those states empty or stale on failure, and `KanbanTaskEditorView`'s Assignment picker shows no explanation.
**Suggestion:** Capture and route each failure through the store's error-reporting path instead of `try?`.

### [nit] `includeArchived` preference is not persisted
**File:** Conduit/Views/KanbanViews.swift:14
**Issue:** `includeArchived` is a plain `@State`, so the toggle resets every time the sidebar tab is reopened — inconsistent with board selection, which is persisted in `UserDefaults`.
**Suggestion:** Back it with `@AppStorage` (e.g. `conduit.kanbanIncludeArchived`).

### [nit] `createBoard`/`updateBoard` are unreachable dead code
**File:** Conduit/Services/KanbanService.swift:191-222
**Issue:** No caller exists in `KanbanStore` or any view, so the board-creation/edit surface is unexercised. Note `updateBoard(slug:)`'s `pathComponent` also allows URL dot-segments (slug `".."` normalizes to the parent path), worth fixing if ever wired up.
**Suggestion:** Wire them into the board-picker menu or delete them.

### [nit] `RelativeDateTimeFormatter` allocated per comment row
**File:** Conduit/Views/KanbanViews.swift:790-795
**Issue:** `relativeDate` constructs a formatter on every call inside `ForEach(comments)`.
**Suggestion:** Hoist a private static/shared `RelativeDateTimeFormatter` with `.abbreviated` units.

## Verdict
REQUEST_CHANGES

The two blockers are the `decodeLossyInt` `Int(Double)` trap (a server-data crash in the file whose whole purpose is tolerance) and the task-detail polling that clobbers unsaved edits. Everything else is minor-to-nit hardening. Prior review items around mutation-error surfacing and the archived-toggle observer are genuinely fixed in HEAD, and the `@MainActor` story, sidebar migration, and board-scoping design are solid.

<a href="https://opencode.ai/s/Taot6QpF">New%20session%20-%202026-08-22T01%3A25%3A53.731Z</a>
[opencode session](https://opencode.ai/s/Taot6QpF)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32543334713)

kaishi00:
Adversarial audit vs NousResearch/hermes-agent@b6bcb3e complete; repairs pushed in 8709668. Key correction: the earlier profile= query was fabricated semantics - the plugin API has no profile parameter, and upstream Desktop reaches other profiles only via per-profile backend processes (each profile is its own HERMES_HOME). Conduit now documents this constraint and scopes board persistence by server identity instead. Locked lanes (review/running/scheduled), fire-and-forget dispatcher nudges, draft-preserving detail polls, fail-closed query building, and hostile-number decoding are now covered by tests. Full local suite: 58/58.

kaishi00:
## Summary
This PR converts the sidebar's Capabilities tab into a native iPhone Kanban workspace (typed models, `KanbanStatusPresentation`, `KanbanStore`, `KanbanService` over the existing `DashboardTicketBridge`), migrates the persisted `SidebarTab` raw value, and relocates Capabilities into Settings. Prior hardening commits have already fixed the CRT/`Int` trap, draft-clobbering detail polls, mutation-error surfacing, nudge blocking, fail-closed query building, and the archived-toggle observer.

I verified the current HEAD (hardened state) by reading each changed file; the findings below reflect what is still present.

## Findings

### [major] Board card's primary interaction is invisible to VoiceOver, and several icon-only controls are unlabeled
**File:** Conduit/Views/KanbanViews.swift:439, 394-400, 657-659, 107-110
**Issue:** The card opens via `.onTapGesture(perform: onOpen)` with no `.accessibilityAddTraits(.isButton)`, no combined element, and no hint — VoiceOver treats the title/summary as separate static elements and offers no "open" action (only the implicit context-menu fallback). The card `ellipsis` menu (Move to / Delete), the detail toolbar's `ellipsis.circle` menu (Delete), and the mutation-error dismiss `xmark.circle.fill` button all have image-only labels with no `.accessibilityLabel`, so VoiceOver announces raw symbol names. This is inconsistent with the correctly labelled board-picker/refresh/new-task chips and lane chips.
**Suggestion:** Add `.accessibilityElement(children: .combine)` + `.accessibilityAddTraits(.isButton)` + a hint ("Opens task") to `KanbanCardView`, and `.accessibilityLabel`s to both ellipsis menus and the dismiss button.

### [minor] A requested reload is silently dropped while another load is in flight
**File:** Conduit/Services/KanbanStore.swift:115, 172-180
**Issue:** `reload` bails on `guard !isLoading else { return }` with no re-queue. `selectBoard(_:includeArchived:)` sets and persists the new slug then calls `reload`; if a poll reload is mid-flight the call returns, the header/subtitle updates to the new board but the columns stay on the previous board until the next 8 s poll tick. The permanent-spinner race is gone (every `configure` path resets `isLoading`), but a momentary stale-board window remains for board switches and archived toggles made during a load.
**Suggestion:** Capture `let wasLoading = isLoading` in `selectBoard` and reload again after the in-flight load completes, or coalesce reloads via `loadGeneration` instead of bailing at the flag.

### [minor] Auxiliary-load failures are swallowed silently
**File:** Conduit/Services/KanbanStore.swift:143-146
**Issue:** `async let loadedProfiles = try? service.fetchProfiles()` (and projects/orchestration) turns any failure into `nil`, leaving the editor's Profile picker stale or empty with no explanation while the store reports a successful reload. A transient failure can hide a missing profile/project list in `KanbanTaskEditorView`.
**Suggestion:** Route each failure through `errorMessage`/`mutationErrorMessage` (or a dedicated partial-error state) instead of `try?`, or at least leave a visible banner.

### [minor] Card move menu omits unknown/custom statuses, disagreeing with the detail picker
**File:** Conduit/Views/KanbanViews.swift:366-368
**Issue:** `statusOptions` is built only from `manuallySelectableStatuses` (the nine known statuses). A task in a server-defined custom column (unknown status, `isManuallySelectable == false`) cannot be moved back to its own column from the card menu, while the detail picker inserts the current status into its options. The two UIs disagree about what is movable.
**Suggestion:** Union in the task's current status (and other statuses present on the board) into the card's move menu, mirroring `KanbanTaskDetailView.statusOptions`.

### [minor] Empty-id tasks and random-UUID fallbacks break `Identifiable` identity across reloads
**File:** Conduit/Models/KanbanModels.swift:192, 348, 374, 404, 441, 517
**Issue:** A task missing `id` decodes to `""`, colliding in `ForEach(column.tasks)`. `KanbanComment`/`KanbanEvent`/`KanbanAttachment`/`KanbanRun`/`KanbanChildResult` fall back to a fresh `UUID()` per decode, so the detail view (re-decoded after every save/comment/poll) builds new identities for every row each reload.
**Suggestion:** Drop id-less tasks at board decode (the service already rejects them at mutation time) and derive deterministic fallbacks from stable fields (e.g. `\(taskID ?? "")-\(createdAt ?? 0)-\(author)`) for the child collections.

### [minor] `includeArchived` preference is not persisted
**File:** Conduit/Views/KanbanViews.swift:44
**Issue:** It is a plain `@State` defaulting to false, so the toggle resets each time the sidebar tab is reopened — inconsistent with board selection, which is persisted in `UserDefaults`.
**Suggestion:** Back it with `@AppStorage("conduit.kanbanIncludeArchived")` and fold it into the persistence (it already feeds `pollingKey`, so the load/refresh path keeps working).

### [nit] Force unwrap in the migration path
**File:** Conduit/Services/KanbanStore.swift:95
**Issue:** `selectedBoardSlug = defaults.string(forKey: persistenceKey!) ?? ""` force-unwraps `persistenceKey`. It is assigned two lines earlier and is non-nil by construction, so it cannot trap today, but it is an unnecessary `!` in a codebase that otherwise avoids them.
**Suggestion:** `guard let key = persistenceKey else { return }` or use `persistenceKey ?? Self.selectedBoardKey`.

### [nit] `addComment` returns success for a whitespace-only body
**File:** Conduit/Services/KanbanService.swift:153-154
**Issue:** The `guard !trimmed.isEmpty else { return }` returns without sending or erroring; the caller then clears the field and forces a reload, so a whitespace comment silently vanishes.
**Suggestion:** Throw a validation error (callers already pre-validate, so this is purely defensive/explicit).

### [nit] `createBoard`/`updateBoard` are unreachable dead code
**File:** Conduit/Services/KanbanService.swift:214-245
**Issue:** No caller exists in `KanbanStore` or any view, so the board-management surface is unexercised and untested. Note `updateBoard` routes the slug through `pathComponent`, which can normalize a `..` slug to a parent path — a latent dot-segment issue if ever wired up.
**Suggestion:** Either wire them into the board-picker menu or delete them (and the related `pathComponent` scrutiny) until needed.

### [nit] `RelativeDateTimeFormatter` allocated per comment row
**File:** Conduit/Views/KanbanViews.swift:901-906
**Issue:** `relativeDate` constructs a formatter on every call inside `ForEach(comments)` on each poll.
**Suggestion:** Hoist a shared `static let` formatter configured with `.abbreviated` units.

## Verdict
REQUEST_CHANGES

The architecture is solid — consistent `@MainActor` isolation, no force unwraps in new code beyond the safe `persistenceKey!`, frozen mutation contexts that snapshot board/requester identity, and the previously flagged crash/draft/data-integrity issues are genuinely fixed in HEAD. The remaining blocker is the accessibility gap on the Kanban card (the core interaction of this feature), with the reload-drop, silent auxiliary failures, status-option disagreement, and identity-churn items as follow-ups.

<a href="https://opencode.ai/s/47m0xA94">New%20session%20-%202026-08-22T02%3A35%3A17.027Z</a>
[opencode session](https://opencode.ai/s/47m0xA94)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32546604044)

kaishi00:
Final correctness pass pushed as 79d9546. Highlights: explicit board selection now supersedes an in-flight poll (generation-forced reload) and mutations bind to the LOADED board snapshot identity, so an A card can never PATCH board B while B is pending; stale mutation completions after a server reconfigure are fully inert (no UI error, no flag flip, no refresh) via a generation-owned mutation token; Capabilities loading now returns a request-scoped CapabilityLoadOutcome instead of inferring from global state; lossy integer decoding is exact (2.5 -> nil, no truncation); id-less task rows are dropped during column decoding so duplicate SwiftUI identities are impossible. Documentation corrected: Hermes Kanban is a shared cross-profile board anchored at the shared Hermes root (kanban_db.py resolves profiles back through get_default_hermes_root()) - hence no profile query and no profile-scoped persistence, by design. Full local suite: 58/58 passed, 8/8 batches.

kaishi00:
## Summary
Adds a native iPhone Kanban workspace (typed models, `KanbanStatusPresentation`, `KanbanStore`, `KanbanService` on the existing `DashboardTicketBridge`) replacing the sidebar Capabilities tab, with explicit migration of the persisted `SidebarTab` raw value, plus a Capabilities browser relocated inside Settings. Prior hardening in `8709668`/`79d9546` correctly fixed the `Int(Double)` trap, draft-clobbering detail polls, fail-closed query building, locked-lane creation, stale-mutation isolation, and id-less task identity.

## Findings

### [minor] Board card's primary "open" action is invisible to VoiceOver
**File:** Conduit/Views/KanbanViews.swift:441
**Issue:** The card opens the detail sheet via `.onTapGesture(perform: onOpen)` with no `.accessibilityElement(children: .combine)` and no `.accessibilityAddTraits(.isButton)`. VoiceOver exposes the title/summary as static text with no action; the only discovery path is the context-menu actions rotor. The ellipsis menus (line 402) and dismiss button (line 111) are now correctly labelled, making this the remaining gap on the feature's core interaction.
**Suggestion:** Add `.accessibilityElement(children: .combine)` + `.accessibilityAddTraits(.isButton)` + an "Opens task" hint to the card.

### [minor] Failed first-load error is never cleared after the detail view recovers
**File:** Conduit/Views/KanbanViews.swift:805-843
**Issue:** `loadDetail`'s success path sets `refreshErrorMessage = nil` but never clears `errorMessage`. If the very first detail fetch fails (`detail == nil`, so `errorMessage` is set at line 838), the 4 s poll loop keeps retrying and on recovery populates `detail` but the stale error text stays pinned at the bottom of the sheet forever.
**Suggestion:** Clear `errorMessage` in the success branch (alongside `refreshErrorMessage`) once `detail` is populated.

### [minor] Post-mutation reload can be dropped by an in-flight poll, showing stale columns up to one poll interval
**File:** Conduit/Services/KanbanStore.swift:134, 399
**Issue:** `reload` returns when `isLoading` is true unless `superseding` is passed. `selectBoard` uses `superseding: true`, but `performMutation` calls plain `reload(...)`. If the 8 s poll holds the load slot when a mutation completes, the refresh bails and the board keeps pre-mutation data until the next poll tick (up to ~8 s of staleness).
**Suggestion:** Pass `superseding: true` from `performMutation`'s refresh, or coalesce a pending reload flag that re-runs once the active load completes.

### [minor] `loadedBoardSlug` can mismatch the actually-loaded board after a stale selection is discarded
**File:** Conduit/Services/KanbanStore.swift:146, 160-179
**Issue:** `requestedSlug` is frozen at line 146 before `fetchBoards`. If the new boards list no longer contains `selectedBoardSlug`, it is reset to `""` (line 161) and the board is fetched as server-current, but `loadedBoardSlug = requestedSlug ?? currentServerBoardSlug` still returns the frozen stale slug (line 179) — mutations would then target `board=<stale>` against content actually loaded from server-current.
**Suggestion:** After the reset, recompute the binding from the effective selection (e.g., `let appliedSlug = selectedBoardSlug.isEmpty ? nil : selectedBoardSlug`) and use that for both the fetch and `loadedBoardSlug`.

### [minor] Card move menu excludes the task's own custom/unknown status
**File:** Conduit/Views/KanbanViews.swift:367-369
**Issue:** `statusOptions` is built only from `manuallySelectableStatuses`. A task in a server-defined custom column (unknown status, fallback `sortOrder: 100`) cannot be returned to its own column from the card menu; the detail picker inserts but also disables the unknown option, so the limitation holds there too. Both UIs silently forbid an otherwise-valid move.
**Suggestion:** Union the task's current status (and other statuses present on the board) into the card menu when selectable, mirroring the detail picker's insertion.

### [minor] `configure(nil)` leaves connection-scoped auxiliary state stale
**File:** Conduit/Services/KanbanStore.swift:80-91
**Issue:** The nil-requester branch clears `board`, `boards`, `loadedBoardSlug`, but not `profiles`, `projects`, `orchestration`, `currentServerBoardSlug`, `errorMessage`, or `mutationErrorMessage`. A store that stays alive across a disconnect can retain the previous dashboard's profiles/orchestration and error text.
**Suggestion:** Mirror the full reset used in the non-nil reconfiguration branch.

### [nit] Force unwraps of `persistenceKey`
**File:** Conduit/Services/KanbanStore.swift:111, 116
**Issue:** `defaults.string(forKey: persistenceKey!)` and `defaults.set(legacy, forKey: persistenceKey!)` — safe today (assigned two lines above) but a needless `!` in a codebase that otherwise avoids them; a future reorder would crash.
**Suggestion:** Bind the key to a local constant (as CodeRabbit's `scopedKey` suggestion) and drop the unwrap.

### [nit] `pathComponent` still permits `.` and `..` URL dot-segments
**File:** Conduit/Services/KanbanService.swift:307-312
**Issue:** `DashboardPath.encodedQueryComponent` preserves dots, so an id/slug of `..` normalizes to the parent collection route on request. Reachability is limited (`updateBoard(slug:)` is dead code; task ids are server-issued), but the fail-closed contract stated at lines 290-292 is not fully met.
**Suggestion:** Reject the literal `"."`/`".."` values before encoding, or percent-encode dots.

### [nit] `addComment` silently succeeds for whitespace-only bodies
**File:** Conduit/Services/KanbanService.swift:155
**Issue:** The guard returns success without sending anything; a caller that missed the pre-validation would see its comment vanish with no error.
**Suggestion:** Throw a validation error for a whitespace-only trimmed body.

### [nit] Test harness timing and fixtures
**File:** ConduitTests/KanbanTests.swift:290, 474-479
**Issue:** `flushNudge` uses a 250 ms wall-clock sleep instead of the existing `awaitPendingDispatcherNudgeForTesting()` seam (timing-dependent under load), and `testLossyIntDecodingHandlesExtremeAndMalformedValues` uses `1e999`, which modern Foundation may reject at parse time before `decodeLossyInt` runs; `1e300` exercises the intended finite-but-unconvertible path deterministically.
**Suggestion:** Await the pending-nudge seam via the store, and switch to `1e300`.

## Verdict
APPROVE

The critical (crash on hostile numerics) and major (draft clobbering, silent mutation errors, locked-lane semantics, stale-server mutation isolation, identity collisions) items from earlier reviews are genuinely fixed in HEAD, and the defensive work is notable: frozen mutation contexts bound to the loaded-board snapshot, generation-owned mutation tokens, request-scoped capability outcomes, and forced-supersede navigation. The remaining items are minor-to-nit hardening — most notably the VoiceOver gap on the card tap and the never-cleared detail-load error — none of which block this feature's merge.

<a href="https://opencode.ai/s/SQ5kPg0Q">New%20session%20-%202026-08-22T03%3A32%3A35.083Z</a>
[opencode session](https://opencode.ai/s/SQ5kPg0Q)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32549202769)

kaishi00:
## Summary
A well-hardened PR that adds a native iPhone Kanban workspace (typed models, `KanbanStore`/`KanbanService` over the existing `DashboardTicketBridge`), migrates the persisted `SidebarTab` raw value, and relocates the Capabilities browser into Settings. The critical/major issues flagged in earlier review rounds (trapping `Int(Double)`, draft-clobbering detail polls, silent mutation errors, un-scoped board writes, dot-segment path traversal) are genuinely fixed in HEAD; I verified each against the current code. Remaining findings are minor-to-nits.

## Findings

### [minor] Post-mutation reload can be dropped by an in-flight poll, leaving stale columns up to one poll interval
**File:** Conduit/Services/KanbanStore.swift:149, 426-428, 438
**Issue:** `reload` bails at `if !superseding && isLoading { return }`. `performMutation` calls plain `reload(...)` (not `superseding: true`) on success and error. If the 8 s background poll holds the load slot when a move/delete/create completes, the refresh is silently dropped and the board keeps pre-mutation data until the next poll tick (~8 s stale window). The board selection path (`selectBoard`, line 226) correctly uses `superseding: true`; mutations do not.
**Suggestion:** Pass `superseding: true` from `performMutation`, or coalesce a pending-reload flag so a dropped request re-runs once the active load completes.

### [minor] The board card's primary "open" action is invisible to VoiceOver
**File:** Conduit/Views/KanbanViews.swift:450-451
**Issue:** The card opens the detail sheet via `.onTapGesture(perform: onOpen)` with no `.accessibilityElement(children: .combine)`, no `.accessibilityAddTraits(.isButton)`, and no hint. VoiceOver exposes title/summary as static text with no button trait or action beyond the context-menu rotor. This is the core interaction of the feature and inconsistent with the properly labelled chips, ellipsis menus, and toolbar buttons.
**Suggestion:** Add `.accessibilityElement(children: .combine)` + `.accessibilityAddTraits(.isButton)` + `.accessibilityAction(named: "Open task")` on the card.

### [minor] Tasks in server-defined custom statuses can never return to their own column
**File:** Conduit/Views/KanbanViews.swift:377-379 (card menu), 614-620 (detail picker); Conduit/Models/KanbanStatus.swift:82-85
**Issue:** The card move menu offers only `manuallySelectableStatuses` (known statuses). A task in an unknown/custom column (fallback `isManuallySelectable == false`) is neither offered in the card menu nor selectable in the detail picker (the inserted "current" option is `.disabled`). Both UIs therefore forbid a valid move back to the task's own column.
**Suggestion:** Union statuses actually present on the board (column names / task statuses) into the card menu when the task is in one, mirroring the detail picker — and if the backend accepts a same-column transition, keep it selectable.

### [minor] `configure(nil)` leaves connection-scoped auxiliary state from the previous dashboard
**File:** Conduit/Services/KanbanStore.swift:94-104
**Issue:** The nil-requester branch clears `board`, `boards`, `loadedBoardSlug`, and the service, but not `profiles`, `projects`, `orchestration`, `currentServerBoardSlug`, `errorMessage`, or `mutationErrorMessage`. A store alive across a disconnect can show the prior dashboard's editor profile list and stale server-board slug until reconfigure.
**Suggestion:** Mirror the full reset used in the non-nil reconfigure branch (also reset the auxiliary arrays and error text).

### [minor] Detail-collection identity falls back to a fresh `UUID()` per decode, churning SwiftUI rows on every poll
**File:** Conduit/Models/KanbanModels.swift:355, 381, 411, 448, 524
**Issue:** `KanbanComment`/`KanbanEvent`/`KanbanAttachment`/`KanbanRun`/`KanbanChildResult` fall back to `UUID().uuidString` when `id` is missing. The detail view re-decodes after every save/comment/poll, so id-less rows get a new identity each reload and `ForEach` rebuilds them (and duplicate random ids could collide). The task/column path was fixed by dropping id-less rows; the collections were not.
**Suggestion:** Derive deterministic fallback identities from stable fields (e.g. `"\(taskID ?? "")-\(createdAt ?? 0)-\(author)"`), matching the board-level approach.

### [nit] Archived-toggle preference is not persisted and can lag until the next poll
**File:** Conduit/Views/KanbanViews.swift:44, 58, 148-164
**Issue:** `includeArchived` is a plain `@State`, so the preference resets each time the sidebar tab reopens — inconsistent with board selection, which is persisted. Also, toggling while a load is in flight triggers a non-superseding reload that is dropped, so the archived state only applies on the next 8 s tick.
**Suggestion:** Back it with `@AppStorage("conduit.kanbanIncludeArchived")` and make the toggle's refresh supersede an in-flight load (or route it through `selectBoard`-style `superseding: true`).

### [nit] `relativeDate` allocates a `RelativeDateTimeFormatter` on every call
**File:** Conduit/Views/KanbanViews.swift:914-919
**Issue:** Runs once per comment row per 4 s render pass; formatter construction is not free.
**Suggestion:** Hoist a `static let` formatter with `.abbreviated` units.

### [nit] Status picker options use `.disabled` instead of `.selectionDisabled`
**File:** Conduit/Views/KanbanViews.swift:707
**Issue:** Per-option `.disabled` also greys the entire picker row set in some iOS versions and is the wrong API on this deployment target.
**Suggestion:** Use `.selectionDisabled(!value.isManuallySelectable)`.

### [nit] `addComment` silently succeeds for a whitespace-only body
**File:** Conduit/Services/KanbanService.swift:155-158
**Issue:** The trimmed-empty guard returns success without sending anything; a caller that missed pre-validation would see the comment vanish with no error.
**Suggestion:** Throw a validation error on whitespace-only bodies.

### [nit] `createBoard`/`updateBoard` are unreachable dead code
**File:** Conduit/Services/KanbanService.swift:218-249
**Issue:** No caller exists in the store or any view, so the board-management surface is untested/unexercised.
**Suggestion:** Wire them into the board-picker menu or remove them until needed.

### [minor] Bridges held past disconnect while Kanban tab is covered; plus `KanbanStore` retains the service
**File:** Conduit/Services/KanbanStore.swift:36, 93-104
**Issue:** `service` strongly holds the requester (and its WKWebView). When the sidebar is on a different tab at disconnect, `.task(id:)` is cancelled and `configure(nil)` does not run, so the invalidated bridge + webview stay alive (via the still-`@StateObject`-owned store) until the tab reappears or the store deallocs. Not a cycle, but a heavy transient hold.
**Suggestion:** Consider publishing/dispatch to clear the bridge when it goes nil even while the tab is hidden, or just note the retention window.

### [nit] Test timing/flags nondeterminism
**File:** ConduitTests/KanbanTests.swift:290, 615-620
**Issue:** `flushNudge` still uses a 250 ms wall-clock sleep instead of the `awaitPendingDispatcherNudgeForTesting()` seam, and the hostile-number test uses `1e999`, which some Foundation builds reject at JSON parse time (making the test throw regardless of `decodeLossyInt`). `1e300` exercises the intended finite-but-unconvertible path deterministically.
**Suggestion:** Await the pending-nudge seam (expose via `KanbanStore`) and switch the fixture value to `1e300`.

## Verdict
APPROVE

No critical or major issues remain in HEAD. The previously blocking correctness/data-integrity items (numeric crash on hostile JSON, draft-clobbering detail polls, silent mutation-error swallowing, stale-server mutation isolation, board-identity drift, dot-segment path traversal, id-less task identity collisions) are fixed and, where applicable, covered by tests. The remaining items are minor UX/correctness polish — most notably the post-mutation refresh drop and the VoiceOver gap on the card tap — that do not block this feature's merge.

<a href="https://opencode.ai/s/52i87ASH">New%20session%20-%202026-08-22T12%3A26%3A21.440Z</a>
[opencode session](https://opencode.ai/s/52i87ASH)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32572937703)

kaishi00:
## Summary
This PR adds a native iPhone Kanban workspace (typed models, `KanbanStore`/`KanbanService` over the existing `DashboardTicketBridge`), migrates the persisted `SidebarTab` raw value (Capabilities→Kanban), relocates the Capabilities browser into Settings, and hardens loads/mutations with generations, frozen board contexts, and request-scoped outcomes. The overall design (MainActor isolation, board-scoping, fail-closed query building, stale-generation isolation) is strong — but HEAD does not compile.

## Findings

### [critical] Undefined `CapabilityLoadPolicy.resolvePresentation` — HEAD does not build
**File:** Conduit/Views/CapabilitiesView.swift:19 (also ConduitTests/KanbanTests.swift:665-696)
**Issue:** Commit `c40bdb0` rewrote the Capabilities rendering boundary to `switch CapabilityLoadPolicy.resolvePresentation(...)` with cases `.loading/.failure/.emptySuccess/.list(...)`, and the tests assert on the same function. But the `CapabilityLoadPolicy` enum in Conduit/Services/AppState.swift:8646 only declares `canCommit` and `shouldPresentRows` — `resolvePresentation` and its return type are defined nowhere in the repo (a repo-wide search finds only references). This is a compile error in the app and test targets, so the "1008 test cases passed" claim cannot be true for this head.
**Suggestion:** Add `static func resolvePresentation(snapshotProfile:activeProfile:isLoading:loadError:hasRows:) -> CapabilityPresentation` (and the enum with `.loading`, `.failure(String?)`, `.emptySuccess`, `.list(banner: String?)`, plus the row/foreign-snapshot gating behavior the tests encode) to `CapabilityLoadPolicy`, or revert the `c40bdb0` render-boundary rewrite.

### [minor] Detail-collection identity falls back to a fresh `UUID()` per decode
**File:** Conduit/Models/KanbanModels.swift:355, 381, 411, 448, 524
**Issue:** `KanbanComment`/`KanbanEvent`/`KanbanAttachment`/`KanbanRun`/`KanbanChildResult` default `id` to `UUID().uuidString` when the field is missing. The detail sheet re-decodes every 4 s poll, giving every id-less row a new identity so `ForEach` rebuilds all rows each poll; duplicate random ids could collide. The board-level path was fixed by dropping id-less tasks; the collections were not.
**Suggestion:** Derive deterministic fallbacks from stable fields (e.g. `"\(taskID ?? "")-\(createdAt ?? 0)-\(author)"`), matching the column decode approach.

### [minor] Board card's primary "open" action is invisible to VoiceOver
**File:** Conduit/Views/KanbanViews.swift:450-451
**Issue:** The card opens the detail sheet via `.onTapGesture(perform: onOpen)` with no `.accessibilityElement(children: .combine)`, no `.accessibilityAddTraits(.isButton)`, and no hint. VoiceOver exposes title/summary as static text with no action. This is the feature's core interaction and is inconsistent with the correctly labelled chips, ellipsis menus (`Task actions`), and toolbar buttons.
**Suggestion:** Add `.accessibilityElement(children: .combine)` + `.accessibilityAddTraits(.isButton)` + `.accessibilityAction(named: "Open task", onOpen)` to the card.

### [minor] Tasks in server-defined custom statuses can't be moved back to their own column
**File:** Conduit/Views/KanbanViews.swift:377-379 (card), 614-620/707 (detail picker); Conduit/Models/KanbanStatus.swift:82-85
**Issue:** The card menu offers only `manuallySelectableStatuses` (known statuses); the detail picker inserts the current unknown status but marks it `.disabled(!value.isManuallySelectable)`. A task in an unknown/custom lane is therefore not offered (card) and not selectable (detail) — the two UIs agree only on forbidding a valid same-column move.
**Suggestion:** Union statuses present on the board (and the task's own status) into the card menu, and use `.selectionDisabled` (+ same-column allowance) rather than `.disabled` on picker rows.

### [minor] `configure(nil)` leaves connection-scoped auxiliary state stale
**File:** Conduit/Services/KanbanStore.swift:94-104
**Issue:** The nil-requester branch clears `board`, `boards`, `loadedBoardSlug`, and the service, but not `profiles`, `projects`, `orchestration`, `currentServerBoardSlug`, `errorMessage`, or `mutationErrorMessage`. A store alive across a disconnect can retain the previous dashboard's editor profile list and stale error text until the next non-nil reconfigure.
**Suggestion:** Mirror the full reset in the non-nil reconfiguration branch (also reset the auxiliary arrays and error text).

### [minor/nit] Post-mutation reconciliation duplicated for create-in-other-lane; idempotency key unused
**File:** Conduit/Services/KanbanService.swift:126-133 / 364-371; Conduit/Views/KanbanViews.swift:542-558
**Issue:** Creating into a non-todo/triage lane issues POST + PATCH (two round trips) plus one full superseding reload; `KanbanCreateTaskRequest.idempotencyKey` is never set. The `didCreate` flag mitigates duplicate re-POSTs on the partial-success path, so this is no longer a correctness hole — just inefficiency and an unused safety field.
**Suggestion:** Include the target status in the create body when the API supports it, and populate `idempotencyKey` per save attempt so a retried save is deduplicated server-side.

### [nit] Whitespace-only comment silently succeeds
**File:** Conduit/Services/KanbanService.swift:155-158
**Issue:** `guard !trimmed.isEmpty else { return }` returns success without sending anything; a caller that missed pre-validation would see its comment vanish with no error.
**Suggestion:** Throw a validation error for a trimmed-empty body.

### [nit] `includeArchived` preference is not persisted
**File:** Conduit/Views/KanbanViews.swift:44
**Issue:** It is a plain `@State` defaulting to false, so the toggle resets whenever the sidebar tab reopens — inconsistent with board selection, which is persisted in `UserDefaults`.
**Suggestion:** Back it with `@AppStorage("conduit.kanbanIncludeArchived")` (it already feeds `pollingKey`, so refresh behavior is preserved).

### [nit] `relativeDate` allocates a `RelativeDateTimeFormatter` per comment row per 4 s render
**File:** Conduit/Views/KanbanViews.swift:914-919
**Issue:** Formatter construction runs once per comment row on every poll pass.
**Suggestion:** Hoist a `static let` formatter configured with `.abbreviated` units.

### [nit] Test timing/fixture nondeterminism
**File:** ConduitTests/KanbanTests.swift:290, 717-722
**Issue:** `flushNudge` still sleeps 250 ms of wall-clock instead of using the `awaitPendingDispatcherNudgeForTesting()` seam the service already exposes, and the hostile-number test uses `1e999`, which newer Foundation JSON parsers reject at parse time (making the assertion moot) — `1e300` deterministically exercises the intended finite-but-unconvertible path.
**Suggestion:** Expose the nudge-await seam via `KanbanStore` and assert on it; change the fixture to `1e300`.

### [nit] `.disabled` on status Picker rows instead of `.selectionDisabled`
**File:** Conduit/Views/KanbanViews.swift:707
**Issue:** Per-option `.disabled` greys the whole picker set on some iOS versions and is the wrong API here.
**Suggestion:** Use `.selectionDisabled(!value.isManuallySelectable)`.

## Verdict
REQUEST_CHANGES

The architecture and most prior hardening is genuinely in place at HEAD (frozen mutation contexts, generation-owned stale-state isolation, fail-closed query building, draft-preserving detail polls, id-less task dropping — all verified against current code). The blocker is strictly the undefined `CapabilityLoadPolicy.resolvePresentation` introduced by the final commit `c40bdb0`, which breaks the build; everything else is minor-to-nit polish that can follow.

<a href="https://opencode.ai/s/Ldqt0QRw">New%20session%20-%202026-08-22T14%3A14%3A26.977Z</a>
[opencode session](https://opencode.ai/s/Ldqt0QRw)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32578058313)

kaishi00:
## Summary
This PR replaces the sidebar's Capabilities tab with a native iPhone Kanban workspace (typed models, `KanbanStatusPresentation`, `KanbanStore`/`KanbanService` over the existing `DashboardTicketBridge`), migrates the persisted `SidebarTab` raw value, and relocates the Capabilities browser into Settings. The design is highly defensive — MainActor isolation is consistent, mutation-aware generations/frozen board contexts are sound, and the previously flagged `Int(Double)` trap, draft-clobbering polls, and fail-closed query building are genuinely fixed in HEAD.

## Findings

### [major] Capabilities Settings page renders an unkillable spinner on any failed first load
**File:** Conduit/Views/CapabilitiesView.swift:19-25 (with Conduit/Services/AppState.swift:6526-6533)
**Issue:** `CapabilityLoadPolicy.resolvePresentation` returns `.loading` whenever `shouldPresentRows(snapshotProfile:, activeProfile:)` is false, i.e. whenever `appState.capabilitiesProfile != activeProfile`. That value is nil until a load *succeeds* (it is only assigned on `.success`, AppState.swift:6531). So when the first load for a profile fails — which is the common "no dashboard connected" case — `appState.loadCapabilities()` returns `.unavailable`, the view sets `loadError = "Connect to a Hermes dashboard to load capabilities."`, but `resolvePresentation` short-circuits on the nil snapshot and renders the `.loading` ProgressView. The `.failure` case is unreachable for any profile that never had a successful load: the error message never displays and the loading state has no `.refreshable`, so the user can only leave and re-enter (which fails identically). The tests only exercise foreign-`"A"`/`"B"` snapshots, not the nil-snapshot-with-loadError case.
**Suggestion:** `loadError` is always current-request-scoped (the view clears it at request start), so surface it before the snapshot gate — e.g. return `.failure(loadError)` when `loadError != nil` even without a matching snapshot, falling through to `.loading` only when `loadError == nil`. Add a regression test for `resolvePresentation(snapshotProfile: nil, activeProfile: "p", isLoading: false, loadError: "...", hasRows: false) == .failure(...)`.

### [minor] Board card's primary tap-to-open action is invisible to VoiceOver
**File:** Conduit/Views/KanbanViews.swift:450-451
**Issue:** The card opens the detail sheet via `.onTapGesture(perform: onOpen)` with no `.accessibilityElement(children: .combine)`, no `.accessibilityAddTraits(.isButton)`, and no action — VoiceOver exposes only static text (title/summary as separate elements) with no "open" affordance beyond the context-menu rotor. This is the core interaction of the feature, and inconsistent with the correctly labelled chips, ellipsis menus, and toolbar buttons elsewhere in the same file.
**Suggestion:** Add `.accessibilityElement(children: .combine)`, `.accessibilityAddTraits(.isButton)`, and `.accessibilityAction(named: "Open task", onOpen)` plus a hint to the card.

### [minor] Tasks in server-defined custom statuses can never return to their own column
**File:** Conduit/Views/KanbanViews.swift:377-379 (card menu) and 614-620, 707 (detail picker); Conduit/Models/KanbanStatus.swift:82-85
**Issue:** The card's move menu offers only `manuallySelectableStatuses` (the nine known statuses). A task in an unknown/custom lane (`isManuallySelectable == false` via the `default` fallback) appears in no menu option; the detail picker inserts the current status but marks it `.disabled`, so both UIs forbid a same-column move. The two surfaces agree only on blocking a valid operation.
**Suggestion:** Union the task's own status (and other statuses present on the board) into the card's move options when not in `manuallySelectableStatuses`, and use `.selectionDisabled` (rather than `.disabled`) on the picker rows.

### [minor] `refresh()` (pull-to-refresh) is silently dropped while a passive poll holds the load slot
**File:** Conduit/Services/KanbanStore.swift:229-232, 148-149
**Issue:** `refresh()` calls `reload` without `superseding: true`, and `reload` returns early when `!superseding && isLoading`. A pull-to-refresh initiated while the 8-second poll is mid-flight is discarded with no follow-up, so the board stays on pre-pull data until the next tick. Mutations and board selection already use `superseding: true`; the manual refresh path does not.
**Suggestion:** Coalesce a pending-reload flag in `reload` and re-run once the active load completes, or route `refresh()` through the superseding path.

### [minor] First detail-load failure message is never cleared after the poll recovers
**File:** Conduit/Views/KanbanViews.swift:846-852
**Issue:** `loadDetail`'s catch sets `errorMessage` when `detail == nil`. If the first fetch fails and a later poll succeeds, the success path clears `refreshErrorMessage` but never `errorMessage`, so the stale error text stays pinned at the bottom of the sheet for the sheet's lifetime.
**Suggestion:** Clear `errorMessage` in the success branch once `detail` is populated (alongside `refreshErrorMessage`).

### [minor] `configure(nil)` leaves connection-scoped auxiliary state from the previous dashboard
**File:** Conduit/Services/KanbanStore.swift:94-104
**Issue:** The nil-requester branch clears `board`, `boards`, `loadedBoardSlug`, and the service, but not `profiles`, `projects`, `orchestration`, `currentServerBoardSlug`, `errorMessage`, or `mutationErrorMessage`. A store alive across a disconnect can keep showing the prior dashboard's editor profile list and stale error text until reconfigured.
**Suggestion:** Mirror the full reset used in the non-nil reconfiguration branch.

### [minor] Detail-collection identity falls back to a fresh `UUID()` per decode, churning SwiftUI rows on each poll
**File:** Conduit/Models/KanbanModels.swift:355, 381, 411, 448, 524
**Issue:** `KanbanComment`/`KanbanEvent`/`KanbanAttachment`/`KanbanRun`/`KanbanChildResult` default `id` to `UUID().uuidString` when the field is missing. The detail sheet re-decodes every 4-second poll, so id-less rows get a new identity every pass and `ForEach` rebuilds them. The board-level path was fixed by dropping id-less tasks; the collections were not.
**Suggestion:** Derive deterministic fallback identities from stable fields (e.g. `"\(taskID ?? "")-\(createdAt ?? 0)-\(author)"`), matching the column-decode approach.

### [nit] `addComment` silently succeeds for a whitespace-only body
**File:** Conduit/Services/KanbanService.swift:157-158
**Issue:** `guard !trimmed.isEmpty else { return }` returns success without sending anything; a caller that misses pre-validation sees its comment vanish with no error.
**Suggestion:** Throw a validation error for a trimmed-empty body.

### [nit] `includeArchived` preference is not persisted
**File:** Conduit/Views/KanbanViews.swift:44
**Issue:** It is a plain `@State` defaulting to false, so the toggle resets whenever the sidebar tab reopens — inconsistent with board selection, which is persisted in `UserDefaults`.
**Suggestion:** Back it with `@AppStorage("conduit.kanbanIncludeArchived")` (it already feeds `pollingKey`, so refresh behavior is preserved).

### [nit] `relativeDate` allocates a `RelativeDateTimeFormatter` per comment row per 4-second poll
**File:** Conduit/Views/KanbanViews.swift:914-919
**Issue:** Formatter construction runs once per comment row on every render pass.
**Suggestion:** Hoist a `static let` formatter configured with `.abbreviated` units.

### [nit] `createBoard`/`updateBoard` remain unreachable dead code
**File:** Conduit/Services/KanbanService.swift:218-249
**Issue:** No caller exists in the store or any view, so the board-management surface is unexercised and untested.
**Suggestion:** Wire them into the board-picker menu or remove them until needed.

### [nit] Test timing/fixture nondeterminism
**File:** ConduitTests/KanbanTests.swift:717-722, 290
**Issue:** `flushNudge` still sleeps 250 ms of wall-clock instead of awaiting the `awaitPendingDispatcherNudgeForTesting()` seam the service exposes, and the hostile-number test uses `1e999`, which stricter Foundation JSON parsers may reject at parse time (rather than exercising `decodeLossyInt`'s finite-but-unconvertible path).
**Suggestion:** Await the pending-nudge seam via `KanbanStore` and use `1e300` in the fixture.

## Verdict
REQUEST_CHANGES

No critical or major data-integrity issues remain — the generation/frozen-context isolation, fail-closed transport, tolerant decoding, and draft-preserving polls are well-executed and tested. The blocker is the newly introduced capabilities render boundary: `.failure` is unreachable whenever the active profile has never had a successful load, so a failed or offline first load leaves the Settings Capabilities page on an eternal spinner with its error message suppressed. The remaining items are minor-to-nit hardening.

<a href="https://opencode.ai/s/egOJyAFl">New%20session%20-%202026-08-22T14%3A21%3A28.666Z</a>
[opencode session](https://opencode.ai/s/egOJyAFl)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32578403959)

kaishi00:
Two-race cleanup pushed (c40bdb0 + f6d4efb). 1) Post-mutation reconciliation now uses the superseding reload path, so a completed create/update/delete can never be dropped behind an in-flight passive poll, and the stale poll discards by generation when it lands - covered by testMutationReconciliationSupersedesInFlightPassivePoll (held-poll race, asserts both the fresh board fetch occurring and no overwrite). 2) Capabilities rendering now goes through CapabilityLoadPolicy.resolvePresentation as the single boundary: foreign snapshots can only ever resolve to loading (never rows/toggles), empty success shows an explicit No-capabilities state instead of a blank list, and empty-failure vs populated-refresh-banner are distinguished - policy covered by four deterministic tests; commit/ABA protections retained. CI green at f6d4efb (Build & Test pass, review pass, CodeRabbit pass); full local suite 1008 cases passed / 0 failed.

kaishi00:
## Summary
Replaces the sidebar Capabilities tab with a native Kanban workspace (typed models, `KanbanStore`/`KanbanService` over the existing `DashboardTicketBridge`, lane-chip + card-list UI), migrates the persisted `SidebarTab` raw value, and moves the Capabilities browser into Settings with a request-scoped loading/ownership boundary. Earlier review rounds' critical issues (trapping `Int(Double)`, draft-clobbering detail polls, silent mutation errors, stale-server mutation isolation, dot-segment traversal, compile break) are verified fixed in HEAD.

## Findings

### [minor] First detail-load failure message is never cleared after the poll recovers
**File:** Conduit/Views/KanbanViews.swift:851 (success path of `loadDetail`, 826-859)
**Issue:** On the first fetch failure `detail == nil` so `errorMessage` is set (line 854). Every later successful 4 s poll sets `refreshErrorMessage = nil` but never clears `errorMessage`, so the stale red text stays pinned at the bottom of the sheet for its entire lifetime.
**Suggestion:** In the success branch, once `detail` is populated, also set `errorMessage = nil`.

### [minor] Non-superseding reloads are silently dropped during an in-flight load
**File:** Conduit/Services/KanbanStore.swift:149
**Issue:** `if !superseding && isLoading { return }` bails with no re-queue. Board selection and post-mutation reconciliation use `superseding: true`, so those are safe, but pull-to-refresh (`refresh()`) and the Archived-toggle path (`.task(id: pollingKey)` restart → plain `reload`) are not. Toggling Archived or pulling to refresh while an 8 s poll holds the load slot leaves stale columns (wrong archived state) until the next tick.
**Suggestion:** Coalesce a `reloadRequestedWhileLoading` flag and re-run one reload when the active load completes, or route `refresh()`/the toggle through the superseding path.

### [minor] `configure(nil)` leaves connection-scoped auxiliary state stale
**File:** Conduit/Services/KanbanStore.swift:94-104
**Issue:** The nil-requester branch clears `board`, `boards`, `loadedBoardSlug`, and the service, but not `profiles`, `projects`, `orchestration`, `currentServerBoardSlug`, `errorMessage`, or `mutationErrorMessage`. A store surviving a disconnect keeps the previous dashboard's editor profile list, server-current slug, and error text (the non-nil branch at 135-142 resets all of these, so the two branches are asymmetric).
**Suggestion:** Mirror the full reset used in the non-nil reconfiguration branch.

### [minor] Tasks in server-defined custom statuses can never be moved back to their own column
**File:** Conduit/Views/KanbanViews.swift:377-379, 713-714; Conduit/Models/KanbanStatus.swift:84
**Issue:** The card move menu only offers `manuallySelectableStatuses`; the detail picker inserts the current unknown status but marks it `.disabled`. A task in an unknown/custom lane (fallback `isManuallySelectable == false`) is neither offered (card) nor selectable (detail) — both UIs forbid a move back to its own column.
**Suggestion:** Union the task's current status (and other statuses present on the board) into the card menu, and use `.selectionDisabled(!value.isManuallySelectable)` (iOS 17) rather than per-option `.disabled`.

### [minor] `createTask` never populates `idempotencyKey`, so a retry can duplicate
**File:** Conduit/Services/KanbanStore.swift:268; Conduit/Models/KanbanModels.swift:626
**Issue:** The create request is a POST with no idempotency key. In `KanbanTaskEditorView.save()` (KanbanViews.swift:541-565) only `.taskCreatedButMoveFailed` sets `didCreate`; a transport-level failure (e.g. timeout after the POST landed) leaves the sheet open and a retry issues a second POST.
**Suggestion:** Set a stable `idempotencyKey` per save attempt so a retried save is deduplicated server-side, and/or dismiss the sheet once creation succeeds.

### [minor] Id-less detail-collection identities fall back to a fresh `UUID()` per decode
**File:** Conduit/Models/KanbanModels.swift:355, 381, 411, 448, 524
**Issue:** `KanbanComment`/`KanbanEvent`/`KanbanAttachment`/`KanbanRun`/`KanbanChildResult` default missing `id` to `UUID().uuidString`. The detail view re-decodes every 4 s poll, so id-less rows get a new identity each pass and `ForEach` rebuilds them (duplicate random ids could also collide). The board-level path was fixed by dropping id-less tasks; the collections were not.
**Suggestion:** Derive deterministic fallbacks from stable fields (e.g. `"\(taskID ?? "")-\(createdAt ?? 0)-\(author)"`), matching the column-decode approach.

### [nit] Capabilities load failure writes the global `appState.errorMessage`
**File:** Conduit/Services/AppState.swift:6533
**Issue:** On failure `loadCapabilities` sets the shared `errorMessage` even though the view now consumes the returned `CapabilityLoadOutcome`. This can stamp a capabilities error into a global error surface owned by unrelated flows.
**Suggestion:** Drop the global write now that the request-scoped outcome carries the message.

### [nit] Archived preference is not persisted
**File:** Conduit/Views/KanbanViews.swift:44
**Issue:** `includeArchived` is a plain `@State`, so the toggle resets whenever the sidebar tab is reopened — inconsistent with board selection, which is persisted in `UserDefaults`.
**Suggestion:** Back it with `@AppStorage` (e.g. `conduit.kanbanIncludeArchived`).

### [nit] `RelativeDateTimeFormatter` allocated per comment row per render
**File:** Conduit/Views/KanbanViews.swift:920-925
**Issue:** `relativeDate` constructs a formatter on every call, once per comment row on each 4 s render pass.
**Suggestion:** Hoist a `static let` formatter configured with `.abbreviated` units.

### [nit] `addComment` silently succeeds for a whitespace-only body
**File:** Conduit/Services/KanbanService.swift:157-158
**Issue:** The trimmed-empty guard returns success without sending or erroring; a caller that misses pre-validation sees its comment vanish.
**Suggestion:** Throw a validation error instead of returning silently.

### [nit] Search-clear button has no accessibility label
**File:** Conduit/Views/KanbanViews.swift:336-341
**Issue:** The `xmark.circle.fill` clear button (plus the error-dismiss and ellipsis controls elsewhere in the file are now labelled) has no `.accessibilityLabel`, so VoiceOver announces the raw symbol name.
**Suggestion:** Add `.accessibilityLabel("Clear search")`.

### [nit] Unreachable `do`/`catch` around `try?` in the dispatcher nudge task
**File:** Conduit/Services/KanbanService.swift:196-209
**Issue:** Both throwing expressions inside the task are `try?`, so the `catch` block is unreachable (compiler warning) and the "acceleration hint" comment is detached from the actual failure-swallowing site.
**Suggestion:** Drop the `do`/`catch` and keep the comment next to the `request` call.

## Verdict
APPROVE

No critical or major issues remain in HEAD. The previously blocking concerns — `Int(Double)` trap, compile-break from the missing `resolvePresentation`, eternal-spinner on failed first capability load, draft-clobbering detail polls, stale-server mutation isolation, board-identity drift, and dot-segment traversal — are genuinely fixed and covered by deterministic tests. Remaining items are minor-to-nit polish (stale error text in the detail sheet, a non-superseding reload drop on pull-to-refresh/archive toggle, asymmetric `configure(nil)` reset, custom-status move gaps, unused `idempotencyKey`), none of which block merge.

<a href="https://opencode.ai/s/ajLiHCop">New%20session%20-%202026-08-22T16%3A30%3A44.091Z</a>
[opencode session](https://opencode.ai/s/ajLiHCop)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32584861404)

PR #97, opened 2026-08-24
feat(kanban): V3D — live /events invalidation stream

kaishi00:
## CI-hang correction (commit d8c806f)
- The always-failing-minter reconnect test spun a non-suspending loop on the MainActor and exceeded the CI budget; the backoff path now guarantees a Task.yield per cycle and the test bounds the minter (30 failures then parks on a cancellation-safe sleep). Full ios_test: 64/64 suites, 1238 test cases, 0 failed/skipped.
## Review-pass correction (commit 52eca6d)
- BLOCKER: run() installs withTaskCancellationHandler - cancelling the owning SwiftUI task retires the socket/loop/delays immediately (regression: owner.cancel() -> socket cancelled, zero reconnects/mints).
- BLOCKER: malformed /events array elements are consumed/discarded so the walk always advances (previously a bare scalar blocked every later valid event forever while the cursor advanced past them; regression frame [5,{task_id:A}],cursor12 -> A delivered).
- Show Archived is part of the live-stream identity: toggling it restarts the stream so event refreshes use the fresh filter.
- Deferred retry is EXACTLY ONE delayed re-attempt composed with a fresh revision and folded with newly-pending events; still-busy drops to polling (no indefinite reschedule).
- Heartbeat ping is timeout-bounded (10s); minor diagnostic cleanups. Full ios_test: 64/64 suites, 1238 test cases, 0 failed/skipped.

## Summary by CodeRabbit

* **New Features**
  * Added live Kanban updates through an authenticated WebSocket connection.
  * Board changes are grouped and synchronized with the latest server state.
  * Task details refresh automatically when the displayed task is affected.
  * Reconnection and heartbeat handling maintain updates during temporary connection issues.

* **Bug Fixes**
  * Added safeguards for malformed, outdated, or irrelevant events.
  * Polling continues when live updates are unavailable.
  * Updates defer safely during loading or editing.
  * Improved cancellation handling prevents stale updates and hanging connections.

## Final V3D correction (commit 780ef7d)
- All lifecycle/ping tests pass; full suite green at 64/64 suites and 1238 test cases.
- CI-hang fix: bounded minter test + backoff yield prevents MainActor spin.


kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream (`KanbanEventStreamCoordinator`) that coalesces frames into batches and drives authoritative REST refreshes plus a touched-task detail invalidation, with polling left as the fallback. Includes models, store boundary, view glue, and 805 lines of deterministic tests.

## Findings

### [severity: critical] Old live stream is never torn down on board switch / view dismiss — zombie sockets accumulate
**File:** Conduit/Services/KanbanEventStream.swift:208-221 (and Conduit/Views/KanbanViews.swift:207-260)
**Issue:** `start()` creates `loopTask = Task { [weak self] in await self?.runLoop(myRun) }` — an **unstructured** task, which does not inherit cancellation from the parent `.task(id:)` closure. Cleanup relies entirely on `run()`: `await task.value` then `stop()`. But `task.value` only returns when the loop exits, and the loop only exits when `isCurrent` turns false — which requires `runID` bump (`stop()`) or loop cancellation. Neither ever happens when SwiftUI cancels the `.task(id: liveEventsKey)` task on a board switch or view dismiss; the view (KanbanViews.swift) never calls `coordinator.stop()`. Result: every stream-key change leaks the old socket, heartbeat pinger, and the pending `await task.value`, and the baseURL/slug unchanged-but-new stream co-exists. Rapid alpha→beta→alpha switching produces duplicate live sockets for the same stamped context, causing redundant `.refreshed` reloads and duplicate `publishLiveInvalidation` publications (the `.stale` guard only protects dead contexts, not the resurrected same-context one). This contradicts the PR's stated "context change is immediate FINAL cancellation."
**Suggestion:** Make teardown independent of loop exit, e.g. wrap `await coordinator.run()` in `withTaskCancellationHandler` that calls `coordinator.stop()` on the view side, or have `run()` spawn the loop as a tracked child and arm a cancellation handler that cancels/`stop()`s `loopTask` when the awaiting task is cancelled. Ideally add a view-level test asserting socket teardown when the stream key changes.

### [severity: major] Live stream captures a stale `includeArchived` flag
**File:** Conduit/Views/KanbanViews.swift:207-214, 243-247
**Issue:** `liveEventsKey` is built from bridge identity, baseURL, configuration generation, and `loadedBoardSlug`, but does **not** include `includeArchived`. Toggling "Show Archived" changes `pollingKey` (line 200-202) and reloads the board, but leaves `liveEventsKey` unchanged, so `.task(id: liveEventsKey)` does not restart and the `onBatch` closure keeps the `includeArchived` value captured at coordinator creation. Every subsequent live event batch then calls `store.refreshFromEvent(includeArchived: <stale value>)`, overwriting `store.board` with data fetched under the wrong archived flag until the next 8s poll corrects it — visible task set flicker/loss after toggling the filter.
**Suggestion:** Include `includeArchived` in `liveEventsKey` (matching `pollingKey`), or read the current `includeArchived` at batch time from a stable source rather than capturing it in the closure.

### [severity: major] Documented "retry ONCE" for deferred batches is not enforced — unbounded retry loop
**File:** Conduit/Services/KanbanEventStream.swift:97-99, 422-431
**Issue:** `flushNow()` re-merges the batch and calls `scheduleFlushIfIdle()` on every `.retrySoon` outcome and schedules a fresh 300ms window each time, with no retry budget. If the store remains busy (`isMutating`/`isLoading`, e.g. a long-running bulk mutation), the same batch is retried every ~1.05s indefinitely, directly contradicting the documented policy ("retries ONCE … if still busy it is dropped"). The store-boundary tests exercise `.deferred` only via direct `refreshFromEvent` calls, never the coordinator's `.retrySoon` path, so the unbounded loop is untested.
**Suggestion:** Track a per-batch retry count (e.g. re-merge at most once, then drop), or make the re-merge accumulate only genuinely new events, and add a coordinator test where the store stays busy through two retries.

### [severity: minor] Heartbeat ping has no timeout — silently-dead peers may not surface via ping
**File:** Conduit/Services/KanbanEventStream.swift:52-61, 335-352
**Issue:** `ping()` uses `withCheckedThrowingContinuation` and only resumes when the URLSession pongReceiveHandler fires. For a half-open connection (peer died without FIN/RST), `sendPing` may not complete for minutes (OS TCP keepalive/host-reachability timeouts), so the "20s heartbeat surfaces silently-dead sockets into the backoff path" intent is not met — both `receive()` and `ping()` block until OS-level timeouts. Also, if the continuation's awaiter is cancelled before the pong handler fires, the continuation is only unwound because `stop()` cancels the socket (which fires pending handlers); a continuation ping in flight never self-resolves.
**Suggestion:** Arm a time-bounded race (e.g. `Task.sleep` + cancel the ping payload / use an explicit pong timeout) so the heartbeat fails within a bounded interval, and resume the continuation on task cancellation to avoid a leaked/never-resumed continuation.

### [severity: minor] Duplicate diagnostic-cap statement and dead `recordDiagnostic`
**File:** Conduit/Services/KanbanEventStream.swift:165-170, 354-358
**Issue:** `recordDiagnostic(_:)` is defined but never called (all arrays are capped inline), and in `backoff()` the `recordedBackoffs` cap statement appears twice (lines 357-358). Harmless but indicates copy/paste drift.
**Suggestion:** Delete `recordDiagnostic` and the duplicate cap line (or route the inline append+cap sites through the helper).

### [severity: nit] Single-use tickets retained in diagnostics
**File:** Conduit/Services/KanbanEventStream.swift:257-258, 279-280
**Issue:** `mintedTickets` and `issuedURLs` (which embed the single-use ticket in the query string) retain up to 50 mints/URLs in memory. The comment acknowledges URLs embed tickets. They are single-use and capped, and `private(set)`, so exposure is limited.
**Suggestion:** Acceptable as-is; if desired, drop ticket strings and redact the URL query before storing diagnostics.

## Verdict
REQUEST_CHANGES

Strengths worth noting: no force-unwraps or implicitly-unwrapped optionals in the new code; everything is MainActor-isolated (no data races); `weak self` used in the spawned tasks; per-element tolerant decoding is well done; the monotonic cursor and context-stamp ownership checks are sound, and the `.stale` zero-request path is a nice fail-closed design. The critical teardown gap, stale-`includeArchived` capture, and enforced-once retry should be fixed before merge.

<a href="https://opencode.ai/s/a32WKSEp">New%20session%20-%202026-08-24T01%3A15%3A36.628Z</a>
[opencode session](https://opencode.ai/s/a32WKSEp)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32679152672)

kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream (`KanbanEventStreamCoordinator`) that coalesces frames into batches, drives authoritative REST refreshes plus touched-task detail invalidation, keeps polling as fallback, and adds models, view glue, and ~800 lines of deterministic tests.

## Findings

### [severity: critical] Zombie stream — unstructured loop task is never torn down on board switch / view dismiss
**File:** Conduit/Services/KanbanEventStream.swift:208-221, Conduit/Views/KanbanViews.swift:532-534, 259
**Issue:** `start()` spawns `loopTask = Task { [weak self] in ... }` — an **unstructured** task that does not inherit cancellation from the `.task(id:)` closure SwiftUI manages. `run()` exits only via `await task.value` → `stop()`, but `task.value` returns only when the loop exits, and the loop exits only when `isCurrent` turns false via `runID` bump or the loop task's own cancellation. Neither happens when SwiftUI cancels the awaiting `.task(id: liveEventsKey)` task on a key change or view dismissal: the view never calls `coordinator.stop()`, so the old socket, heartbeat pinger, flush task, and the pending `await task.value` leak and keep running indefinitely. Reconnecting the same stamped context (e.g., alpha→beta→alpha, or dismiss→present) yields duplicate live sockets for the same context — the `.stale` guard only protects differing contexts, so these zombie streams still publish duplicate refreshes/invalidations. Contradicts the documented "cancellation is FINAL" contract.
**Suggestion:** Make teardown independent of loop exit. Arm a `withTaskCancellationHandler` on the view side that calls `coordinator.stop()` when `.task(id:)` is cancelled, or have `run()`/`stop()` propagate cancellation into `loopTask` (e.g. `run()` replaces `await task.value` with `while !Task.isCancelled { ... }` plus a cancellation handler that cancels `loopTask`). Add a view-level test asserting socket teardown when the stream key changes.

### [severity: major] Live stream captures stale `includeArchived`
**File:** Conduit/Views/KanbanViews.swift:207-214, 243-247
**Issue:** `liveEventsKey` omits `includeArchived` (unlike `pollingKey`), and `onBatch` captures it by value at coordinator creation. Toggling "Show Archived" restarts the poll task but leaves `liveEventsKey` unchanged, so the socket and `onBatch` keep the pre-toggle flag; every subsequent event batch calls `store.refreshFromEvent(includeArchived: <stale>)`, reloading the board under the wrong archived filter until the next 8s poll corrects it — visible task-set flicker.
**Suggestion:** Include `includeArchived` in the stream identity (extend `KanbanLiveUpdateSupport.streamKey` with it), so the toggle retires the socket (which also fixes the zombie issue) and re-binds `onBatch` with the current flag.

### [severity: major] "Retry ONCE" deferred batches actually retry unboundedly
**File:** Conduit/Services/KanbanEventStream.swift:97-99, 426-435
**Issue:** `flushNow()` on `.retrySoon` re-merges the same batch and calls `scheduleFlushIfIdle()`, which schedules a fresh 300ms window with no retry budget. A store that stays busy (long-running bulk mutation) makes the same batch retry every ~300ms forever, contradicting the documented "retries ONCE … if still busy it is dropped," and `deferredRetryNanoseconds` (750ms) is stored but never actually used. No test exercises the coordinator's `.retrySoon` path.
**Suggestion:** Track a per-batch retry count (at most one re-merge, then drop), and actually use `deferredRetryNanoseconds` for the retry delay. Add a coordinator test where the store remains busy through two retries.

### [severity: major] Tolerant frame decoder drops all later events when one element is a non-object
**File:** Conduit/Models/KanbanModels.swift:1328-1335
**Issue:** `nested.decode(KanbanLiveEvent.self)` under `try?` does not advance `currentIndex` on failure (a well-known `UnkeyedDecodingContainer` behavior). For `{"events":[5, {"task_id":"A"}, ...]}`, the first non-object element fails to decode, `currentIndex` stays at 0, and the fixed-count loop re-decodes the same element every pass — every subsequent valid event (including the frame's invalidation) is skipped. Because `ingest` advances the cursor before checking `events.isEmpty`, those events are permanently lost and the board is never invalidated, silently.
**Suggestion:** On decode failure, consume the element with a guaranteed-success dummy decode (e.g. `try? nested.decode(AnyDecodable.self)` as done in `KanbanDiagnostic.init(from:)`) before continuing; add a test with a non-object element followed by valid ones.

### [severity: minor] Heartbeat ping has no timeout and can leak its continuation
**File:** Conduit/Services/KanbanEventStream.swift:49-61, 335-352
**Issue:** `ping()` only resumes when the URLSession pong handler fires. For a half-open connection the ping can block for OS-level keepalive timeouts, so the "20s heartbeat surfaces silently-dead sockets" intent isn't met; also, if the pinger is cancelled between `sendPing` and the pong, the continuation only unwinds because `stop()` cancels the socket's pending handlers — an in-flight ping never self-resolves.
**Suggestion:** Race the ping against a bounded `Task.sleep`/explicit pong timeout and resume the continuation on cancellation.

### [severity: minor] Dead `recordDiagnostic` and duplicate cap line
**File:** Conduit/Services/KanbanEventStream.swift:165-170, 357-358
**Issue:** `recordDiagnostic(_:)` is never called (all three arrays are capped inline at 257-258, 279-280, 356-358), and the `recordedBackoffs` trim appears twice at 357-358.
**Suggestion:** Route the appends through the helper (or remove it) and delete the duplicate cap.

### [severity: minor] Tests force-unwrap in helpers and hang-prone spin-waits
**File:** ConduitTests/KanbanV3DTests.swift:33-34, 93, 126, 150; 618-622, 649-653, 696-701, 809-813
**Issue:** `frameJSON` uses `try!`/`!` and the `queryItems!` assertions trap the whole process instead of failing one test readably. The spin helpers (`waitUntilHeld`, `waitUntilAwaited`, `waitForSuspension`, `waitUntilURLCount`) loop with no iteration limit, so an unreachable condition hangs CI without naming the failing test; `waitUntilHeld(0)` at line 347 is also a no-op.
**Suggestion:** Use `XCTUnwrap`/`throws` for the unwraps, and add bounded loops with `XCTFail` on exhaustion.

### [severity: nit] Docs attribute types to the wrong file
**File:** docs/KANBAN_V3D_PLAN.md:8
**Issue:** `KanbanLiveUpdatePolicy`, `KanbanEventInvalidation`, `KanbanEventBatchOutcome` are declared in `Conduit/Services/KanbanEventStream.swift`, not `KanbanLiveUpdateSupport.swift` (which holds only `streamKey`, `shouldRefreshDetail`, `initialWatermark`).
**Suggestion:** Correct the file attribution.

### [severity: nit] Single-use tickets retained in diagnostics
**File:** Conduit/Services/KanbanEventStream.swift:257-258, 279-280
**Issue:** `mintedTickets` and `issuedURLs` (query strings embed the single-use ticket) retain up to 50 entries; capped and `private(set)`, so exposure is limited.
**Suggestion:** Acceptable as-is; consider redacting the ticket from stored URLs.

## Verdict
REQUEST_CHANGES

Positive: no force-unwraps in production code, all new types are `@MainActor`-isolated (no data races), `weak self` is used in the spawned tasks, monotonic cursor and context-stamp ownership checks are sound, and the `.stale` zero-request path is a good fail-closed design. The zombie-stream teardown leak (critical), the stale `includeArchived` capture, the unbounded retry, and the frame-decoder element-consumption bug should all be fixed before merge.

<a href="https://opencode.ai/s/SZdxgNmJ">New%20session%20-%202026-08-24T02%3A09%3A16.638Z</a>
[opencode session](https://opencode.ai/s/SZdxgNmJ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32682033818)

kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream (`KanbanEventStreamCoordinator`): tickets minted fresh per connect, coalesced batches driving authoritative REST refreshes plus touched-task detail invalidation, polling as fallback, with models, store/view glue, docs, and a ~800-line deterministic test suite. The correction commit (52eca6d) explicitly addresses the previous review-pass blockers, each with a regression test.

## Findings

### [severity: minor] Heartbeat "timeout-bound" is not actually enforced for a fully silent peer
**File:** Conduit/Services/KanbanEventStream.swift:362-384
**Issue:** `pingBounded` races `socket.ping()` against a 10s sleeper, but `withTaskGroup` waits for *all* children to complete before the body returns. On timeout the losing `socket.ping()` child is only cancelled, not forced to finish — it stays suspended in the `withCheckedThrowingContinuation` at line 52-61 until URLSession/OS-level timeouts fire (default request timeout ≈60s). So for a half-open peer the advertised 10s dead-peer bound isn't a hard bound. The in-flight ping continuation also has no cancellation-aware self-resolution; it only unwinds because `stop()` cancels the socket.
**Suggestion:** Make `ping` cancellation-aware (e.g. wrap in `withTaskCancellationHandler` and `resume(throwing: CancellationError())`), or lower the racing losing child into a truly cancellable path so the group exits immediately. Then the documented bound becomes real.

### [severity: minor] Test helpers still force-unwrap and spin without bounds (CI-hang risk)
**File:** ConduitTests/KanbanV3DTests.swift:33-34, 93, 126, 150; 740-744, 771-775, 818-823, 931-935
**Issue:** `frameJSON` uses `try!`/`!` and the `queryItems!` assertions (93/126/150) trap the whole process instead of failing a readable test. The spin-wait helpers (`waitUntilHeld`, `waitUntilAwaited`, `waitForSuspension`, `waitUntilURLCount`) loop on `Task.yield()` with no iteration cap — the exact class of MainActor starvation hang the PR body says was fixed in the minter test, but a regression making a condition unreachable would still hang CI without naming the failing test.
**Suggestion:** Use `XCTUnwrap`/`throws` for the unwraps; bound each wait loop (`for _ in 0..<N { if cond return; await Task.yield() }; XCTFail(...)`).

### [severity: nit] Duplicate diagnostic cap on `mintedTickets`
**File:** Conduit/Services/KanbanEventStream.swift:264-265
**Issue:** The `if mintedTickets.count > 50 { … removeFirst … }` trim is present twice; the second is a no-op (the `recordDiagnostic` helper was removed in the cleanup, but this duplicate survived).
**Suggestion:** Delete line 265.

### [severity: nit] No-op `waitUntilHeld(0)` asserts nothing
**File:** ConduitTests/KanbanV3DTests.swift:467
**Issue:** `waitUntilHeld(0)` returns immediately (`0 < 0` is false) and verifies no condition.
**Suggestion:** Replace with an explicit `XCTAssertEqual(gates.heldCount, 0)` or remove the call.

### [severity: nit] Docs drift after the correction pass
**File:** docs/KANBAN_V3D_PLAN.md:8, 10
**Issue:** Line 8 still attributes `KanbanLiveUpdatePolicy` / `KanbanEventInvalidation` / `KanbanEventBatchOutcome` to `KanbanLiveUpdateSupport.swift` — they're declared in `KanbanEventStream.swift`. Line 10's `liveEventsKey` description omits the now-included `includeArchived` component.
**Suggestion:** Fix the file attribution and note the archived flag in the stream identity.

### [severity: nit] Single-use tickets retained in diagnostics
**File:** Conduit/Services/KanbanEventStream.swift:263-265, 286-287
**Issue:** `mintedTickets` and `issuedURLs` (query strings embed the single-use ticket) retain up to 50 entries. Acknowledged in the source comment; capped and `private(set)`, exposure limited.
**Suggestion:** Acceptable as-is; optionally redact the ticket from stored URLs.

## Verdict
APPROVE

All previously blocking findings are fixed and covered by regression tests: owner-cancellation now tears the stream down (run() → `withTaskCancellationHandler` → stop() cancels the socket, which resolves the suspended receive loop — testOwnerTaskCancellationRetiresStreamFinally); `includeArchived` is part of the stream identity (testStreamKeyIncludesArchivedFlag); deferred retry is bounded to exactly one re-attempt with a fresh revision (testDeferredRetryBoundedToOneShotWithFreshRevision); malformed array elements are consumed via `AnyCodable` so later events are delivered (testMalformedLeadingElementConsumedAndLaterEventsDelivered). Production code has no force-unwraps, everything is `@MainActor`-isolated (no data races), `weak self` is used in all spawned tasks, and the `.stale` zero-request path is a sound fail-closed design. Remaining items are non-blocking minors/nits.

<a href="https://opencode.ai/s/60klNLDE">New%20session%20-%202026-08-24T03%3A31%3A49.052Z</a>
[opencode session](https://opencode.ai/s/60klNLDE)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32686695376)

kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream: a `@MainActor` coordinator mints fresh single-use tickets, coalesces frames into batches, drives authoritative REST refreshes plus touched-task detail invalidation, keeps polling as fallback, and lands ~1,400 lines of deterministic tests plus docs.

The previous critical/major blockers (zombie socket teardown, stale `includeArchived` capture, unbounded deferred retry, non-consuming tolerant decoder, unbounded ping) are all genuinely addressed in the current HEAD with regression tests. Remaining findings are cleanup-grade.

## Findings

### [severity: minor] Leftover debug test and `print` noise shipped in the suite
**File:** ConduitTests/KanbanV3DTests.swift:343-397, 322, 476, 912
**Issue:** `testZZDebugDeferredTrace` ("run-last", obviously a debugging scaffold) contains six `print("DEBUG-ZZ ...")` statements and ends in a weak `XCTAssertGreaterThanOrEqual(batches.count, 1)`. Additional stray `print("TRACE ...")` lines survive at 322 (`TRACE deferred: frame injected`), 476 (`TRACE in-flight released batch-A park; batches=...`), and 912 (`TRACE connect-silent start`). These survived the "review-pass" correction commit and pollute the deterministic test output.
**Suggestion:** Delete `testZZDebugDeferredTrace` and all `print("TRACE ...")`/`print("DEBUG-ZZ ...")` calls.

### [severity: minor] Docs drift not fully corrected
**File:** docs/KANBAN_V3D_PLAN.md:8, :10
**Issue:** Line 8 still attributes `KanbanLiveUpdatePolicy`, `KanbanEventInvalidation`, and `KanbanEventBatchOutcome` to `Conduit/Views/Kanban/KanbanLiveUpdateSupport.swift` — they are declared in `Conduit/Services/KanbanEventStream.swift` (lines 107-122), which now also hosts `KanbanLiveUpdatePolicy` (line 126). Line 10's `liveEventsKey` description (`bridge identity|baseURL|generation|loadedBoardSlug`) omits the now-included `includeArchived` component.
**Suggestion:** Fix the file attribution on line 8 and note the archived filter in the stream-identity description.

### [severity: minor] `testEmptyEventsFrame...` still uses the no-op `waitUntilHeld(0)` and never asserts the stream stayed idle
**File:** ConduitTests/KanbanV3DTests.swift:557-558
**Issue:** `await gates.waitUntilHeld(0)` evaluates `while heldCount < 0` and never loops, so it verifies nothing. The earlier review's suggested replacement — asserting `gates.heldCount == 0` to prove an empty frame never opens a coalescing window — was not applied.
**Suggestion:** Replace with `XCTAssertEqual(gates.heldCount, 0, "an empty frame never opens a coalescing window")` (asserting after `drain()`), or drop the call.

### [severity: minor] "Late receive completion" test is vacuous — parked continuations are wiped by `cancel()`
**File:** ConduitTests/KanbanV3DTests.swift:1004-1031, 1067-1071, 1121-1127
**Issue:** `testStopCancelsSocketAndSilencesLateCompletions` intends to prove a late post-retirement receive is ignored, but `MockSocket.cancel()` resumes and *removes* every parked continuation (lines 1123-1125). By the time `emitLate` (line 1026) runs, `parked` is empty, so nothing is delivered and the coordinator never observes a late completion — the assertions pass for the wrong reason.
**Suggestion:** Keep the parked receive continuations across `cancel()` so `emitLate` can actually deliver the late frame (per the earlier suggestion), or rename the test / assert only the socket-cancellation behavior it really exercises. Add a guard `XCTAssertFalse(parked.isEmpty, "emitLate requires a pending receive")`.

### [severity: minor] Deferred-retry test is missing the parked-mutation handshake (flaky-by-scheduling)
**File:** ConduitTests/KanbanV3DTests.swift:288-341
**Issue:** `testDeferredRetryBoundedToOneShotWithFreshRevision` creates the `mutation` Task and immediately injects the frame / asserts the deferred path with no `await requester.waitForSuspension()` (unlike the sibling busy test at 493-519 which uses the handshake). If the mutation hasn't reached its parked POST when attempt-1 calls `refreshFromEvent`, the disposition would be `.refreshed`, `waitUntilHeld(2)` would burn its 20k spin budget, and the test fails spuriously. It happens to pass under MainActor FIFO scheduling but is fragile.
**Suggestion:** Add `await requester.waitForSuspension()` right after creating the mutation Task, mirroring `testDeferredWhileMutatingThenConvergesWithOneRetry`.

### [severity: nit] Dead `recordBackoff` helper
**File:** Conduit/Services/KanbanEventStream.swift:366-369
**Issue:** `recordBackoff(_:)` is never called — `backoff(myRun:)` (492-495) inlines the append + cap instead. Dead code left over from the cleanup that removed `recordDiagnostic`.
**Suggestion:** Route `backoff`'s append through `recordBackoff` or delete the helper.

### [severity: nit] Implicitly unwrapped optional `pinger` in production code
**File:** Conduit/Services/KanbanEventStream.swift:299
**Issue:** `private var pinger: Pinger!` is the one IUO in the new production code. It is always assigned in `init`, so it is functionally safe, but it contradicts the "no force-unwraps / no IUOs" quality bar the rest of the PR meets.
**Suggestion:** Declare `private let pinger: Pinger` and initialize it in `init` (the default closure only captures `timeout`, which is available).

### [severity: nit] Weak/near-tautological assertions in a cancellation test
**File:** ConduitTests/KanbanV3DTests.swift:842, 841-843
**Issue:** `testOwnerCancelDuringInFlightBatchPreventsRetryAndPublication` asserts `XCTAssertTrue(batches.count == 1 || batches.count == 2, "handler state consistent")` — that passes in almost every case and the duplicate `XCTAssertNil(store.liveInvalidation)` is asserted twice.
**Suggestion:** Pin a concrete expected count (e.g. `XCTAssertEqual(batches.count, 1)`) and drop the duplicate assertion, or explain why the range is legitimate.

## Verdict
APPROVE

All previously blocking issues are fixed and covered by regression tests: owner-cancellation now tears the stream down via `withTaskCancellationHandler` in `run()` (with dedicated tests at KanbanV3DTests.swift:760-844, 969-1000); `includeArchived` is part of the stream identity (KanbanEventStream `streamKey` + testStreamKeyIncludesArchivedFlag); deferred retry is bounded to one re-attempt with a fresh revision; malformed array elements are consumed via `AnyCodable`; both connect-time and heartbeat pings are time-bounded. Production code is free of force-unwraps, fully `@MainActor`-isolated (no data races), uses `weak self` in all spawned tasks, and the `.stale` zero-request path is a sound fail-closed design. The remaining items are cleanup — notably the leftover `testZZDebugDeferredTrace`/`print` debug noise and the still-vacuous `emitLate` test — and should be cleaned before merge but do not block it.

<a href="https://opencode.ai/s/3M3DPaYU">New%20session%20-%202026-08-24T12%3A00%3A00.685Z</a>
[opencode session](https://opencode.ai/s/3M3DPaYU)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32724640092)

kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream (`KanbanEventStreamCoordinator`): cancellation-aware socket/loop/heartbeat/batch lifecycle, fresh single-use tickets per connect, coalesced batches driving authoritative REST refreshes plus a touched-task detail invalidation, polling kept as fallback, with ~1,400 lines of deterministic tests and docs. The review corrections committed at HEAD genuinely resolve the previously-blocking teardown, archived-filter, retry, and decoder issues.

## Findings

### [severity: minor] Stream retirement clears an existing mutation error banner
**File:** Conduit/Services/KanbanStore.swift:941-943
**Issue:** The `Task.isCancelled` path in `refreshFromEvent` restores only `errorMessage`; `mutationErrorMessage` is set to `nil` unconditionally. This refresh is passive (runs only while `!isMutating`), so a pre-existing mutation error can belong to an unrelated in-flight mutation. Stream retirement mid-refresh (e.g. toggling Show Archived, which now restarts `liveEventsKey`) silently wipes that banner, contradicting the adjacent comment's "restore the pre-cancellation error state" intent.
**Suggestion:** Capture `let previousMutationErrorMessage = mutationErrorMessage` before the await and restore it on the cancelled path, mirroring `errorMessage`.

### [severity: minor] Debugging test and print noise shipped in the suite
**File:** ConduitTests/KanbanV3DTests.swift:343-397 (also 322, 476, 906)
**Issue:** `testZZDebugDeferredTrace` is an obvious debug scaffold (ZZ prefix, six `print("DEBUG-ZZ …")` calls, and only a `XCTAssertGreaterThanOrEqual(batches.count, 1)` assertion). Stray `print("TRACE …")` lines remain at 322, 476, and 906. These survived the correction commit and pollute the deterministic suite output.
**Suggestion:** Delete `testZZDebugDeferredTrace` and all `TRACE`/`DEBUG-ZZ` prints.

### [severity: minor] "Late receive completion" test is vacuous
**File:** ConduitTests/KanbanV3DTests.swift:1004-1031, 1067-1071, 1121-1127
**Issue:** `testStopCancelsSocketAndSilencesLateCompletions` intends to prove a late post-retirement frame is ignored, but `MockSocket.cancel()` (1121-1127) resumes and *removes* every parked receive continuation. By the time `emitLate` (1026) runs, `parked` is empty, so nothing is ever delivered and the coordinator never observes a late completion — the assertions pass for the wrong reason.
**Suggestion:** Add `XCTAssertFalse(parked.isEmpty, "emitLate requires a pending receive")` to `emitLate`, or keep parked continuations across `cancel()` for this test so the late frame actually reaches the coordinator.

### [severity: minor] Owner-cancel test has an unasserted counter and near-tautological predicates
**File:** ConduitTests/KanbanV3DTests.swift:805, 823, 841-843
**Issue:** `publishedAfterRefresh` is incremented at 823 but never asserted (its sole purpose is to prove a retired generation performs no post-refresh publication). The duplicate `XCTAssertNil(store.liveInvalidation)` (841, 843) is redundant, and `XCTAssertTrue(batches.count == 1 || batches.count == 2, …)` passes in almost every case.
**Suggestion:** Assert `XCTAssertEqual(publishedAfterRefresh, 0, "a retired generation completes no refresh publication")`, keep one nil check, and pin a concrete batch count.

### [severity: minor] `waitUntilHeld(0)` verifies nothing in the empty-frame test
**File:** ConduitTests/KanbanV3DTests.swift:558
**Issue:** `waitUntilHeld(0)` evaluates `while heldCount < 0` and never loops, so it is a no-op. The empty-frame test never proves the intended property that an empty frame opens no coalescing window.
**Suggestion:** Assert `XCTAssertEqual(gates.heldCount, 0, "an empty frame never opens a coalescing window")` after `drain()`, per the earlier review suggestion.

### [severity: minor] `ContinuationGate.pendingReleases` is never consumed by `awaitRelease`
**File:** ConduitTests/KanbanV3DTests.swift:1206-1215, 1235-1241
**Issue:** `SleeperGate.sleeper` (1146-1149) consumes a pre-recorded release, but `ContinuationGate.awaitRelease` parks unconditionally. If `releaseAll()` runs before the first park, a later `releaseAll()` hits the `pendingReleases > 0` branch, decrements, and returns without resuming the parked continuation — a real release becomes a no-op and the handler stays parked until the 20k spin budget fails. Not triggered by current tests, but latent.
**Suggestion:** Consume one `pendingReleases` entry at the top of `awaitRelease` before parking, matching `SleeperGate`.

### [severity: minor] Deferred-retry test lacks the parked-mutation handshake
**File:** ConduitTests/KanbanV3DTests.swift:299-324
**Issue:** `testDeferredRetryBoundedToOneShotWithFreshRevision` creates the `mutation` Task then injects the frame without `await requester.waitForSuspension()` (the sibling test at 493-519 uses it). It relies on `waitUntilHeld` yields to let the mutation reach its parked POST under MainActor FIFO scheduling; if the mutation hasn't parked when attempt-1 calls `refreshFromEvent`, the disposition flips to `.refreshed` and the test fails spuriously.
**Suggestion:** Add `await requester.waitForSuspension()` immediately after creating the mutation Task.

### [severity: nit] Dead `recordBackoff` helper
**File:** Conduit/Services/KanbanEventStream.swift:366-369, 492-495
**Issue:** `recordBackoff(_:)` is never called — `backoff` still inlines the append + 50-cap at 494-495. Dead code from the cleanup that removed `recordDiagnostic`.
**Suggestion:** Route `backoff`'s append through `recordBackoff` or delete the helper.

### [severity: nit] Implicitly unwrapped `pinger` in production code
**File:** Conduit/Services/KanbanEventStream.swift:299
**Issue:** `private var pinger: Pinger!` is the only IUO in the new production code. Always assigned in `init`, so functionally safe, but it breaks the otherwise held no-IUO quality bar.
**Suggestion:** Declare `private let pinger: Pinger` and assign in `init`.

### [severity: nit] `cancelledSockets` double-counts when the heartbeat retires a connection
**File:** Conduit/Services/KanbanEventStream.swift:463-469, 484-488
**Issue:** When `heartbeatLoop` detects a dead peer it calls `retireConnection` (counts once); then the receive loop's failure path at 468 calls `retireConnection` again for the same socket, counting a second cancel. Diagnostic only — no functional impact.
**Suggestion:** Guard the second retirement with `if currentSocket?.socket === socket` (or make the counter idempotent per socket).

### [severity: nit] Docs drift after the identity change
**File:** docs/KANBAN_V3D_PLAN.md:8, :10; Conduit/Views/KanbanViews.swift:204-206
**Issue:** Line 8 still attributes `KanbanLiveUpdatePolicy`/`KanbanEventInvalidation`/`KanbanEventBatchOutcome` to `KanbanLiveUpdateSupport.swift` — they live in `KanbanEventStream.swift`. Line 10 and the `liveEventsKey` comment omit the now-included `includeArchived` component of the stream identity.
**Suggestion:** Fix the file attribution and note the archived filter in the identity description.

### [severity: nit] Single-use tickets retained in diagnostics
**File:** Conduit/Services/KanbanEventStream.swift:239-241, 263-265, 286-287
**Issue:** `mintedTickets` and `issuedURLs` (query strings embed the single-use ticket) retain up to 50 entries. Capped and `private(set)`, exposure is limited and the source comment acknowledges it.
**Suggestion:** Acceptable as-is; optionally redact the ticket query parameter before retention.

## Verdict
APPROVE

The prior blockers are genuinely resolved and regression-covered: owner-cancellation tears the generation down through the `withTaskCancellationHandler` in `run()` (KanbanEventStream.swift:320-334) with tests at 738-791, 969-1000; `includeArchived` is part of the stream identity (KanbanLiveUpdateSupport.swift:15-24); deferred retry is one re-attempt with a fresh revision; malformed elements are consumed via `AnyCodable` in `KanbanEventFrame`; both connect-time and heartbeat pings are time-bounded. Production code has no force-unwraps, is wholly `@MainActor`-isolated, uses `weak self` in every spawned task, and the `.stale` zero-request path is a sound fail-closed design. The remaining items are non-blocking cleanup — notably the leftover debug test/prints and the vacuous `emitLate` test.

<a href="https://opencode.ai/s/Oo4ptcPW">New%20session%20-%202026-08-24T12%3A07%3A35.325Z</a>
[opencode session](https://opencode.ai/s/Oo4ptcPW)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32725305799)

kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream (`KanbanEventStreamCoordinator`) with fresh single-use tickets per connect, coalesced invalidation batches driving authoritative store refreshes plus touched-task detail invalidation, polling kept as fallback, plus ~1,400 lines of deterministic tests and docs. Corrections for owner-cancellation teardown, `includeArchived` stream identity, bounded one-shot deferred retry, malformed-element consumption, and time-bounded ping are in place.

## Findings

### [severity: major] Final follow-up re-read/clear silently drops events that arrive during the last batch handler
**File:** Conduit/Services/KanbanEventStream.swift:589-594

**Issue:** After the inner `while` loop breaks, lines 590-593 copy and then *clear* `pendingTaskIDs`/`pendingBoardInvalidation` ("Re-check for arrivals that landed during the last handler call"), but the locals `ids`/`boardInvalidated` are never used — the outer loop at 547 immediately re-reads the now-cleared pending state, hits the guard at 552, and returns. The documented behavior ("outer loop picks up arrivals that merged into pending during a handler") is defeated: a frame that lands while `onBatch` is running is discarded from live invalidation and the board is not refreshed until the next frame or the 8 s poll. `ids`/`boardInvalidated` at 590-591 are dead stores; the `.removeAll()`/`= false` at 592-593 is the harmful half.
**Suggestion:** Delete lines 589-593 entirely. The outer loop's own re-read at 548-551 will then fold the fresh arrivals into a follow-up pass, matching the intent. Consider a test that injects a frame while a handler is parked and asserts the follow-up batch delivers it.

### [severity: minor] Cancelled refresh wipes an unrelated `mutationErrorMessage` banner
**File:** Conduit/Services/KanbanStore.swift:940

**Issue:** The `Task.isCancelled` path restores `previousErrorMessage` but sets `mutationErrorMessage = nil` unconditionally. This refresh runs only when `!isMutating`, so the mutation error shown can belong to a completed operation. When the stream task is retired mid-reload (e.g., toggling Show Archived), the user's mutation error banner silently disappears, contradicting the "restore the pre-cancellation error state" comment at 932-935.
**Suggestion:** Capture `previousMutationErrorMessage` before `refresh(...)` and restore it on the cancelled path, mirroring `errorMessage`.

### [severity: minor] Debug scaffold shipped in the test suite
**File:** ConduitTests/KanbanV3DTests.swift:322, 343-397, 901

**Issue:** `testZZDebugDeferredTrace` (ZZ prefix, six `print("DEBUG-ZZ ...")` calls, only a `XCTAssertGreaterThanOrEqual(batches.count, 1)` assertion) is a leftover debugging aid, and `print("TRACE ...")` lines remain at 322, 366-392, and 901. It duplicates `testDeferred...` coverage while adding noise and suite time.
**Suggestion:** Delete `testZZDebugDeferredTrace` and all TRACE/DEBUG prints.

### [severity: minor] `waitUntilHeld(0)` is a no-op; empty-frame test never proves no coalescing window
**File:** ConduitTests/KanbanV3DTests.swift:553

**Issue:** `while heldCount < 0` never executes, so the empty-frame test verifies nothing about the idle path, and the intended property (an empty frame opens no coalescing window) is unasserted.
**Suggestion:** Assert `XCTAssertEqual(gates.heldCount, 0, "an empty frame never opens a coalescing window")` after `drain()` and drop the `waitUntilHeld(0)` call.

### [severity: minor] `recordBackoff` helper is dead; `backoff` inlines the append+cap
**File:** Conduit/Services/KanbanEventStream.swift:366-369, 492-495

**Issue:** `recordBackoff(_:)` is never called — `backoff(myRun:)` re-implements the append and 50-entry cap inline at 494-495. Dead code left from the cleanup that removed `recordDiagnostic`.
**Suggestion:** Route `backoff`'s append through `recordBackoff` or delete the helper.

### [severity: minor] `cancelledSockets` double-counts heartbeat-initiated retirement
**File:** Conduit/Services/KanbanEventStream.swift:468, 484-487

**Issue:** When the heartbeat retires a dead socket, `retireConnection(socket)` counts one cancel and nulls `currentSocket`. The receive loop then observes the close and calls `retireConnection(socket)` again (line 468) — `socket.cancel()` is a no-op but `cancelledSockets` increments a second time. Diagnostic only.
**Suggestion:** Guard the second retirement with `if currentSocket?.socket === socket { cancelledSockets += 1 }` or track a per-socket cancelled flag.

### [severity: minor] `ContinuationGate.awaitRelease` never consumes a pre-recorded release
**File:** ConduitTests/KanbanV3DTests.swift:1203-1212

**Issue:** If `releaseAll()` runs before the first `awaitRelease()`, the pre-release is parked in `pendingReleases`; the later `awaitRelease()` still parks a continuation, and the next `releaseAll()` consumes the stale `pendingReleases` and returns without resuming it — a real release becomes a no-op and the handler stays parked until the 20k spin budget fails. Latent; not triggered by current tests (`SleeperGate` handles the same case correctly).
**Suggestion:** Consume one `pendingReleases` entry at the top of `awaitRelease` before parking, matching `SleeperGate`.

### [severity: nit] Implicitly unwrapped `pinger` in production code
**File:** Conduit/Services/KanbanEventStream.swift:299

**Issue:** `private var pinger: Pinger!` is the only IUO in the new production code. Always assigned in `init` (the default closure captures only local `timeout`), so it is functionally safe but breaks the otherwise held no-IUO bar.
**Suggestion:** Declare `private let pinger: Pinger` and assign in `init`.

### [severity: nit] Docs drift after the identity change
**File:** docs/KANBAN_V3D_PLAN.md:8, 10

**Issue:** Line 8 still attributes `KanbanLiveUpdatePolicy`, `KanbanEventInvalidation`, and `KanbanEventBatchOutcome` to `KanbanLiveUpdateSupport.swift` — they live in `KanbanEventStream.swift`, which hosts only `streamKey`/`shouldRefreshDetail`/`initialWatermark`. Line 10's `liveEventsKey` description omits the now-included `includeArchived` component.
**Suggestion:** Correct the file attribution (noting `KanbanLiveUpdateSupport` holds the view-layer helpers) and add the archived flag to the stream-identity description.

## Verdict
REQUEST_CHANGES

Production code contains no force-unwraps, is fully `@MainActor`-isolated (no data races), uses `weak self` in all spawned tasks, the corrupted-cursor/`stale` fail-closed paths are sound, and the previously blocking teardown/archived-identity/retry/decoder/ping bugs are genuinely fixed with regression coverage. The one substantive remaining issue is the event-drop at KanbanEventStream.swift:589-594 — a small, clearly-correct fix — plus a handful of minor cleanup items in the cancelled-refresh error restore and the test suite. The rest are nits.

<a href="https://opencode.ai/s/GF3yeDtR">New%20session%20-%202026-08-24T12%3A52%3A42.609Z</a>
[opencode session](https://opencode.ai/s/GF3yeDtR)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32729353436)

kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream (`KanbanEventStreamCoordinator`) that coalesces frames into batches, drives authoritative REST refreshes plus a touched-task detail invalidation, keeps polling as the fallback, and lands ~1,400 lines of deterministic tests and docs. The review-pass commit resolves the prior blockers (owner-cancellation teardown, `includeArchived` in stream identity, one-shot deferred retry, malformed-element consumption, bounded ping).

## Findings

### [severity: major] Follow-up flush clears arrivals that land during the last handler — events are silently dropped
**File:** Conduit/Services/KanbanEventStream.swift:589-594
**Issue:** After the inner `while` loop breaks, lines 590-593 copy pending state into `ids`/`boardInvalidated` and then clear it ("Re-check for arrivals that landed during the last handler call"). The copied locals are never read — the outer loop at 548-551 re-reads `pendingTaskIDs`/`pendingBoardInvalidation` from scratch — so the only live effect is the harmful `.removeAll()`/`= false`. A frame ingested while `onBatch` runs merges into pending and is then wiped here; the outer loop hits the empty guard at 552 and returns. That event never invalidates the board until the next frame or the 8 s poll, contradicting the documented contract in `runBatchLifecycle` (lines 544-546) and `docs/KANBAN_V3D_PLAN.md:7` ("one follow-up flush after an in-flight batch"). The sibling test `testEventDuringInFlightRefreshRunsExactlyOneFollowUp` (ConduitTests/KanbanV3DTests.swift:438-480) injects frame B only *after* `drain()` completes batch A, so `scheduleFlushIfIdle` starts a fresh batch task and the true in-flight-arrival path is never exercised.
**Suggestion:** Delete lines 589-593 entirely; the outer loop's own re-read will fold the fresh arrivals into a follow-up pass. Add a test that parks the handler and injects a frame mid-handler, asserting the follow-up batch delivers the touched ID.

### [severity: minor] Cancelled event refresh wipes an unrelated `mutationErrorMessage`
**File:** Conduit/Services/KanbanStore.swift:936-941
**Issue:** The `Task.isCancelled` path restores `previousErrorMessage` but sets `mutationErrorMessage = nil` unconditionally. `refreshFromEvent` only runs when `!isMutating`, so a visible mutation error can belong to a completed operation; retiring the stream mid-reload (e.g. toggling Show Archived restarts `liveEventsKey`) silently clears the user's banner, contrary to the "restore the pre-cancellation error state" comment.
**Suggestion:** Capture `let previousMutationErrorMessage = mutationErrorMessage` before `refresh(...)` and restore it on the cancelled path, mirroring `errorMessage`.

### [severity: minor] Debug scaffold shipped in the test suite
**File:** ConduitTests/KanbanV3DTests.swift:343-397 (also 322, 894)
**Issue:** `testZZDebugDeferredTrace` (ZZ prefix, six `print("DEBUG-ZZ …")`, only `XCTAssertGreaterThanOrEqual(batches.count, 1)`) is a leftover debugging aid duplicated from `testDeferredRetryBoundedToOneShotWithFreshRevision`; stray `print("TRACE …")` lines survive at 322 and 894. These pollute the deterministic suite output.
**Suggestion:** Delete the test and all `DEBUG-ZZ`/`TRACE` prints.

### [severity: minor] "Late receive completion" test is vacuous — parked receives are wiped by cancel()
**File:** ConduitTests/KanbanV3DTests.swift:993-1020, 1056-1060, 1110-1116
**Issue:** `MockSocket.cancel()` resumes and removes every parked receive continuation with `CancellationError` (lines 1112-1114). By the time `emitLate` runs at 1015, `parked` is empty, so no late frame is ever delivered and the assertions pass for the wrong reason.
**Suggestion:** Add `XCTAssertFalse(parked.isEmpty, "emitLate requires a pending receive")` inside `emitLate`, or keep parked continuations across `cancel()` for this test so the late frame actually reaches the coordinator.

### [severity: minor] `waitUntilHeld(0)` is a no-op; empty-frame test never proves no-Coalescing-window
**File:** ConduitTests/KanbanV3DTests.swift:549-554
**Issue:** `while heldCount < 0` never executes, so the empty-frame test verifies nothing about the idle path — the intended property (an empty frame opens no coalescing window) is unasserted.
**Suggestion:** Replace with `XCTAssertEqual(gates.heldCount, 0, "an empty frame never opens a coalescing window")` after `drain()`.

### [severity: minor] `ContinuationGate.awaitRelease` never consumes a pre-recorded release
**File:** ConduitTests/KanbanV3DTests.swift:1195-1210
**Issue:** If `releaseAll()` runs before the first `awaitRelease()`, the pre-release sits in `pendingReleases`; the later `awaitRelease()` still parks, and the next `releaseAll()` consumes the stale `pendingReleases` and returns without resuming the parked continuation — a real release becomes a no-op and the handler stays parked until the spin budget trips. `SleeperGate.sleeper` handles this correctly at 1135-1138. Latent, not triggered by current tests.
**Suggestion:** Consume one `pendingReleases` entry at the top of `awaitRelease` before parking, matching `SleeperGate`.

### [severity: nit] Implicitly unwrapped `pinger` in production code
**File:** Conduit/Services/KanbanEventStream.swift:299
**Issue:** `private var pinger: Pinger!` is the only IUO in the new production code. Always assigned in `init` (the default closure captures only the local `timeout`), so it is functionally safe but breaks the otherwise-held no-IUO bar.
**Suggestion:** Declare `private let pinger: Pinger` and assign it in `init`.

### [severity: nit] Dead `recordBackoff` helper
**File:** Conduit/Services/KanbanEventStream.swift:366-369, 492-495
**Issue:** `recordBackoff(_:)` is never called — `backoff(myRun:)` inlines the append + 50-cap at 494-495. Dead code surviving the `recordDiagnostic` cleanup.
**Suggestion:** Route `backoff`'s append through the helper or delete it.

### [severity: nit] `cancelledSockets` double-counts heartbeat-initiated retirement
**File:** Conduit/Services/KanbanEventStream.swift:463-468, 484-487
**Issue:** When the heartbeat retires a dead peer, `retireConnection(socket)` counts once and nulls `currentSocket`; the receive loop then unwinds on the cancelled receive, `isCurrent(myRun)` is still true, and line 468 retires the same socket a second time — `socket.cancel()` is a no-op but `cancelledSockets` increments again. Diagnostics only.
**Suggestion:** Guard the second retirement with `if currentSocket?.socket === socket` before incrementing, or track per-socket cancellation.

### [severity: nit] Docs drift after the identity change
**File:** docs/KANBAN_V3D_PLAN.md:8, 10
**Issue:** Line 8 still attributes `KanbanLiveUpdatePolicy`, `KanbanEventInvalidation`, and `KanbanEventBatchOutcome` to `KanbanLiveUpdateSupport.swift` (they live in `KanbanEventStream.swift`); line 10's `liveEventsKey` description omits the now-included `includeArchived` component.
**Suggestion:** Fix the file attribution and note the archived filter in the stream-identity description.

## Verdict
REQUEST_CHANGES

The prior blockers are genuinely resolved and regression-covered, and the production code is strong: fully `@MainActor`-isolated (no data races), `weak self` in every spawned task, cancellation-aware ping via exactly-once gates, no force-unwraps (one IUO nit), and the `.stale` zero-request fail-closed path is sound. However, the in-flight-arrival drop at KanbanEventStream.swift:589-594 is a real, untested correctness bug in the core coalescing loop, and it should be fixed (delete the re-read/clear block) with a genuine in-flight-handler test before merge. The remaining items are minor cleanup in the store boundary and test suite.

<a href="https://opencode.ai/s/GZGozUun">New%20session%20-%202026-08-24T13%3A15%3A49.212Z</a>
[opencode session](https://opencode.ai/s/GZGozUun)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32731533934)

kaishi00:
## Summary
Adds a board-scoped live `/events` WebSocket invalidation stream (`KanbanEventStreamCoordinator`) that mints fresh single-use tickets per connect, coalesces frames into batches, drives authoritative REST refreshes plus touched-task detail invalidation, keeps polling as fallback, and lands ~1,340 lines of deterministic tests plus docs. Prior blockers (owner-cancellation teardown, `includeArchived` stream identity, one-shot deferred retry, malformed-element consumption, bounded ping) are resolved in the current HEAD.

## Findings

### [severity: minor] Cancelled event refresh silently wipes an unrelated mutation-error banner
**File:** Conduit/Services/KanbanStore.swift:940
**Issue:** The `Task.isCancelled` path in `refreshFromEvent` restores `errorMessage` but sets `mutationErrorMessage = nil` unconditionally. This refresh runs only while `!isMutating`, so a visible mutation error can have been set by an earlier, already-completed mutation. If the owning stream task is retired mid-reload (e.g. toggling Show Archived restarts `liveEventsKey`), the user's mutation-error banner disappears without user action, contradicting the adjacent "restore the pre-cancellation error state" comment.
**Suggestion:** Capture `let previousMutationErrorMessage = mutationErrorMessage` before `refresh(...)` and restore it on the cancelled path, mirroring `errorMessage`.

### [severity: minor] "Late receive completion" test is vacuous — `cancel()` wipes the parked receives beforehand
**File:** ConduitTests/KanbanV3DTests.swift:928-965, 1050-1060
**Issue:** `testStopCancelsSocketAndSilencesLateCompletions` intends to prove a late post-retirement frame is ignored, but `MockSocket.cancel()` resumes and removes every parked receive continuation with `CancellationError`. By the time `emitLate` (line 950) runs, `parked` is empty, so no frame is ever delivered and the coordinator never observes a late completion — the assertions pass for the wrong reason.
**Suggestion:** Add `XCTAssertFalse(parked.isEmpty, "emitLate requires a pending receive")` inside `emitLate`, or keep the parked continuation across `cancel()` so the late frame actually reaches the coordinator (a separate `cancelResumingPendingReceive()` helper where other tests need it).

### [severity: minor] Deferred-retry test omits the parked-mutation handshake (relies on FIFO scheduling)
**File:** ConduitTests/KanbanV3DTests.swift:299-341
**Issue:** `testDeferredRetryBoundedToOneShotWithFreshRevision` creates the mutation Task then immediately injects the frame with no `await requester.waitForSuspension()`, unlike the sibling `testDeferredWhileMutatingThenConvergesWithOneRetry` (line 435). It currently passes only because `waitUntilHeld` yields let the mutation reach its parked POST under MainActor FIFO ordering; if scheduling changes and attempt-1's `refreshFromEvent` sees an idle store, the disposition flips to `.refreshed` and the test fails spuriously.
**Suggestion:** Add `await requester.waitForSuspension()` right after creating the mutation Task.

### [severity: nit] `waitUntilHeld(0)` is a no-op; empty-frame test never proves no coalescing window
**File:** ConduitTests/KanbanV3DTests.swift:485
**Issue:** `while heldCount < 0` never executes, so the empty-frame test verifies nothing about the idle path — the intended property (an empty frame opens no coalescing window) is unasserted.
**Suggestion:** After `drain()`, assert `XCTAssertEqual(gates.heldCount, 0, "an empty frame never opens a coalescing window")` and drop the `waitUntilHeld(0)` call.

### [severity: nit] `frameJSON` masks a UTF-8 failure as a valid empty frame
**File:** ConduitTests/KanbanV3DTests.swift:38
**Issue:** `String(data:encoding:) ?? "{}"` converts a serialization/encoding failure into a well-formed empty frame, so a broken test fixture fails later with an unrelated message instead of at the helper.
**Suggestion:** `return try XCTUnwrap(String(data: data, encoding: .utf8), "frame JSON is not valid UTF-8")`.

### [severity: nit] `ContinuationGate.awaitRelease` never consumes a pre-recorded release
**File:** ConduitTests/KanbanV3DTests.swift:1140-1155
**Issue:** `SleeperGate.sleeper` consumes a pending pre-release before parking; `ContinuationGate.awaitRelease` does not. If `releaseAll()` runs before the first `awaitRelease()`, the later `releaseAll()` consumes the stale `pendingReleases` and returns without resuming the parked continuation — a real release becomes a no-op and the handler stays parked until the spin budget trips. Latent; not triggered by current tests.
**Suggestion:** Consume one `pendingReleases` entry at the top of `awaitRelease` before parking, matching `SleeperGate`.

### [severity: nit] Dead `recordBackoff` helper
**File:** Conduit/Services/KanbanEventStream.swift:366-369, 492-495
**Issue:** `recordBackoff(_:)` is never called — `backoff(myRun:)` re-implements the append and 50-entry cap inline (lines 494-495). Dead code surviving the `recordDiagnostic` cleanup.
**Suggestion:** Route `backoff`'s append through `recordBackoff` or delete the helper.

### [severity: nit] `cancelledSockets` double-counts heartbeat-initiated retirement
**File:** Conduit/Services/KanbanEventStream.swift:371-379, 463-468, 484-487
**Issue:** When the heartbeat retires a dead peer, `retireConnection` counts once and nulls `currentSocket`; the receive loop then unwinds on the cancelled receive, `isCurrent(myRun)` is still true (runID unchanged), and line 468 retires the same socket again — `socket.cancel()` is a no-op but `cancelledSockets` increments a second time. Diagnostics only; no functional impact.
**Suggestion:** Guard the second retirement with `if currentSocket?.socket === socket` before incrementing, or track per-socket cancellation.

### [severity: nit] Docs drift after the identity change
**File:** docs/KANBAN_V3D_PLAN.md:8, 10
**Issue:** Line 8 still attributes `KanbanLiveUpdatePolicy`, `KanbanEventInvalidation`, and `KanbanEventBatchOutcome` to `Conduit/Views/Kanban/KanbanLiveUpdateSupport.swift` — they are declared in `Conduit/Services/KanbanEventStream.swift`. Line 10's `liveEventsKey` description (`bridge identity|baseURL|generation|loadedBoardSlug`) omits the now-included `includeArchived` component.
**Suggestion:** Fix the file attribution (noting `KanbanLiveUpdateSupport` holds only `streamKey`/`shouldRefreshDetail`/`initialWatermark`) and add the archived flag to the stream-identity description.

## Verdict
APPROVE

The previously blocking correctness issues are genuinely fixed and regression-covered: owner-cancellation retires the generation through `withTaskCancellationHandler` in `run()` (KanbanEventStream.swift:320-334, with cancellation tests); `includeArchived` is part of the stream identity (KanbanLiveUpdateSupport.swift:15-24); deferred retry is bounded to exactly one re-attempt per outer pass via `deferredRetryAvailable`; malformed array elements are consumed via `AnyCodable` (KanbanModels.swift:1336-1338); and both connect-time and heartbeat pings go through the time-bounded `pingBounded`. The stale in-flight-batch re-read/clear was removed (lines 589-592), so the outer loop now correctly folds handler-time arrivals. Production code has no force-unwraps, is fully `@MainActor`-isolated (no data races), uses `weak self` in every spawned task, and the `.stale`/fail-closed paths are sound. Remaining items are non-blocking cleanup — chiefly the vacuous late-completion test, the missing mutation handshake in one deferred-retry test, and the `mutationErrorMessage` wipe in the cancelled-refresh path — and should be tidied before merge but do not block it.

<a href="https://opencode.ai/s/W2QuVbvM">New%20session%20-%202026-08-24T15%3A16%3A34.274Z</a>
[opencode session](https://opencode.ai/s/W2QuVbvM)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32743736671)

kaishi00:
## Summary

Adds a board-scoped live `/events` WebSocket invalidation stream: a `@MainActor` `KanbanEventStreamCoordinator` mints fresh single-use tickets per connect, coalesces frames into batches, drives authoritative REST refreshes plus touched-task detail invalidation, keeps the 8s poll as fallback, and lands ~1,590 lines of deterministic tests plus docs. The current HEAD (a7a5448) resolves the previously-blocking teardown, archived-identity, retry-bounding, decoder-consumption, and ping-bounding issues, each with regression coverage.

## Findings

### [severity: minor] `frameJSON` masks a UTF-8 failure as a valid empty frame
**File:** ConduitTests/KanbanV3DTests.swift:60
**Issue:** `return String(data: data, encoding: .utf8) ?? "{}"` converts a serialization/encoding failure into a well-formed empty frame, so a broken fixture fails later with an unrelated message instead of at the helper. This was flagged in a prior review pass and is still present.
**Suggestion:** `return try XCTUnwrap(String(data: data, encoding: .utf8), "frame JSON is not valid UTF-8")`.

### [severity: minor] `refreshFromEvent` reports `.refreshed` even when the authoritative reload failed or bailed
**File:** Conduit/Services/KanbanStore.swift:922-958
**Issue:** `refresh` → `reload` swallows errors internally (sets `errorMessage`, keeps the last board) and also early-returns when `service == nil`. `refreshFromEvent` then still returns `.refreshed` because `Task.isCancelled` is false, so the view publishes `liveInvalidation` (and reports `.completed`) as if fresh REST state was landed. A socket-driven refresh that actually failed therefore wakes detail surfaces and publishes invalidations computed from stale board state.
**Suggestion:** Have the cancelled/failure path distinguish outcome — e.g. return `.refreshed` only on a confirmed reload, or surface a `.refreshed`-with-error state, so publication after a failed background reload is skipped.

### [severity: minor] `cancelledSockets` double-counts heartbeat-initiated retirement
**File:** Conduit/Services/KanbanEventStream.swift:389-397, 486-487, 502-505
**Issue:** When `heartbeatLoop` detects a silent peer it calls `retireConnection(socket)` (counts once, nulls `currentSocket`). The receive loop then unwinds on the cancelled receive with `isCurrent` still true and calls `retireConnection(socket)` again at line 486 — `socket.cancel()` is a no-op but `cancelledSockets += 1` increments a second time. Diagnostics only, no functional impact.
**Suggestion:** Guard with `if currentSocket?.socket === socket` before incrementing, or track per-socket cancellation.

### [severity: nit] Dead `recordBackoff` helper
**File:** Conduit/Services/KanbanEventStream.swift:384-387, 510-513
**Issue:** `recordBackoff(_:)` is never called; `backoff(myRun:)` still inlines the append + 50-entry cap. Dead code surviving the `recordDiagnostic` cleanup.
**Suggestion:** Route `backoff`'s append through `recordBackoff` or delete the helper.

### [severity: nit] Docs drift after the identity change
**File:** docs/KANBAN_V3D_PLAN.md:8, 10
**Issue:** Line 8 still attributes `KanbanLiveUpdatePolicy`, `KanbanEventInvalidation`, and `KanbanEventBatchOutcome` to `KanbanLiveUpdateSupport.swift` — they are declared in `KanbanEventStream.swift`; `KanbanLiveUpdateSupport` holds only `streamKey`/`shouldRefreshDetail`/`initialWatermark`. Line 10's `liveEventsKey` description omits the now-included `includeArchived` component (the `KanbanLiveUpdateSupport.swift` header comment is correct, but the plan doc is not).
**Suggestion:** Fix the file attribution and note the archived filter in the stream-identity description.

### [severity: nit] Test closure uses an unread parameter and an under-asserted counter
**File:** ConduitTests/KanbanV3DTests.swift:837-885
**Issue:** In `testOwnerCancelDuringInFlightBatchPreventsRetryAndPublication` the `onBatch` closure names `invalidation` but only uses captured `stampA`; the `pingCallCount` equality after `boundedYields(5_000)` is the only post-retirement signal and is weaker than the sibling tests' explicit counter assertions.
**Suggestion:** Use `{ _ in ... }`, and consider pinning the pre/post ping counts as `XCTAssertEqual` after a bounded drain (already implicitly done).

## Verdict

APPROVE

Production code is strong: fully `@MainActor`-isolated (no data races), `weak self` in every spawned task, no force-unwraps or IUOs, cancellation-aware ping via the exactly-once `PingContinuationGate` (including the pre-registration-result race, covered by `testPingGateConsumesResultThatRacesAheadOfRegister`), the malformed-element decoder now consumes via `AnyCodable`, deferred retry is bounded to one re-attempt, `includeArchived` is part of the stream identity, and owner-cancellation retirement is regression-tested. The remaining items are minor/nit cleanup; none block merge.

<a href="https://opencode.ai/s/hUaNp6kq">New%20session%20-%202026-08-24T17%3A47%3A53.583Z</a>
[opencode session](https://opencode.ai/s/hUaNp6kq)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32758574707)

kaishi00:
## Summary
Adds `KanbanEventStreamCoordinator` — a `@MainActor` WebSocket `/events` invalidation stream that mints fresh per-connect tickets, coalesces frames into batches, drives authoritative REST refreshes plus touched-task detail invalidation, keeps the 8s poll as fallback, and lands ~1,640 lines of deterministic tests plus contract docs. The prior critical/major blockers (owner-cancellation teardown, `includeArchived` stream identity, one-shot deferred retry, malformed-element consumption, bounded ping, batch-handle ownership) are resolved at HEAD and regression-tested.

## Findings

### [severity: minor] `refreshFromEvent` reports `.refreshed` even when the authoritative reload failed or bailed
**File:** Conduit/Services/KanbanStore.swift:938-946
**Issue:** The cancelled path is handled, but otherwise the function returns `.refreshed` unconditionally. `refresh(includeArchived:)` (269-272) delegates to `reload`, which swallows failures internally (`catch { … errorMessage = error.localizedDescription }` and early-returns when `service == nil`), keeping the last board. A socket-triggered reload that actually failed therefore reports `.refreshed`, and the view (`KanbanViews.swift:255-257`) publishes `liveInvalidation` and reports `.completed` from stale board state — waking detail surfaces and potentially surfacing a background refresh error as a user-facing banner.
**Suggestion:** Have `refresh`/`reload` report whether a reload landed (e.g. return a success flag or check `Task.isCancelled`/error state), and only return `.refreshed` on a confirmed reload; otherwise skip publication.

### [severity: minor] `cancelledSockets` double-counts when the heartbeat retires a connection
**File:** Conduit/Services/KanbanEventStream.swift:486-487 (and `retireConnection` 389-397)
**Issue:** When `heartbeatLoop` retires a dead peer it calls `retireConnection` (counts once, nulls `currentSocket`); the receive loop then unwinds on the cancelled receive with `isCurrent(myRun)` still true and retires the same socket again at line 486 — `socket.cancel()` is a no-op but `cancelledSockets += 1` increments a second time. Diagnostics only; reconnect still happens exactly once.
**Suggestion:** Guard the second retirement with `if currentSocket?.socket === socket` before incrementing, or track per-socket cancellation.

### [severity: nit] Dead `recordBackoff` helper
**File:** Conduit/Services/KanbanEventStream.swift:384-387, 510-513
**Issue:** `recordBackoff(_:)` is never called; `backoff(myRun:)` still inlines the append + 50-entry cap. Dead code surviving the `recordDiagnostic` cleanup.
**Suggestion:** Route `backoff`'s append through `recordBackoff` or delete the helper.

### [severity: nit] `frameJSON` masks a UTF-8 conversion failure as a valid empty frame
**File:** ConduitTests/KanbanV3DTests.swift:60
**Issue:** `String(data:encoding:) ?? "{}"` turns a serialization/encoding failure into a well-formed empty frame, so a broken fixture fails later with an unrelated message instead of at the helper.
**Suggestion:** `return try XCTUnwrap(String(data: data, encoding: .utf8), "frame JSON is not valid UTF-8")`.

### [severity: nit] Test force-unwrap in an assertion-adjacent read
**File:** ConduitTests/KanbanV3DTests.swift:604
**Issue:** `boardOnlyBatches.first!.taskIDs` force-unwraps; it is guarded by the preceding `XCTAssertEqual(boardOnlyBatches.count, 1)`, so it cannot trap, but it is inconsistent with the `touchedBatches.first?.` pattern below and with the `XCTUnwrap` style used elsewhere.
**Suggestion:** Use `try XCTUnwrap(boardOnlyBatches.first)` or optional chaining.

### [severity: nit] Docs drift after the identity change
**File:** docs/KANBAN_V3D_PLAN.md:10
**Issue:** The `liveEventsKey` description (`bridge identity|baseURL|generation|loadedBoardSlug`) omits the now-included `includeArchived` component of the stream identity (line 8's file attribution is correctly fixed).
**Suggestion:** Add `includeArchived` to the stream-identity description.

### [severity: nit] Latent Swift 6 Sendable concern in the production ping
**File:** Conduit/Services/KanbanEventStream.swift:60-68
**Issue:** `task.sendPing(pongReceiveHandler:)` takes a `@Sendable` closure, and the closure captures `gate`, a `@MainActor`-isolated non-`Sendable` class. Compiles fine under Swift 5.9 mode (`project.yml` SWIFT_VERSION 5.9) but is a strict-concurrency violation if the project moves to Swift 6.
**Suggestion:** Hoist the MainActor hop fully out of the `@Sendable` closure (e.g., capture only `ObjectIdentifier`/a `Sendable` token and look up the gate on the actor) or mark the gate for main-actor-safe transfer.

## Verdict
APPROVE

The previously blocking issues are genuinely fixed with regression coverage: owner-cancellation retires the generation via `withTaskCancellationHandler` in `run()` (KanbanEventStream.swift:338-352, with tests at KanbanV3DTests.swift:274-300, 837-930); `includeArchived` is part of the stream identity (KanbanLiveUpdateSupport.swift:15-24); the deferred retry is bounded to one re-attempt per outer pass; malformed array elements are consumed via `AnyCodable`; both connect-time and heartbeat pings go through the uniformly bounded `pingBounded` with a pre-registration-safe `PingContinuationGate`; and `batchTask` cleanup is ownership-scoped (line 560-564). Production code has no force-unwraps or IUOs and is fully `@MainActor`-isolated (`weak self` in every spawned task); the spin-waits are all bounded; the late-completion and empty-frame tests now assert real properties. The remaining items are minor/nit cleanup (the `.refreshed` misreport on failed reload being the most substantive) and do not block merge.

<a href="https://opencode.ai/s/sENcatjV">New%20session%20-%202026-08-24T20%3A21%3A06.842Z</a>
[opencode session](https://opencode.ai/s/sENcatjV)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/32773276135)

PR #100, opened 2026-08-27
fix(presentation): hard cross-profile boundary for deferred presentation-cache writes

kaishi00:
## Root cause

`AppState.schedulePresentationCacheFlush(for:)` coalesces streaming presentation-cache writes into a delayed `Task` that sleeps up to 2 s and then calls `cacheMessagePresentation(for: [sessionId])`. The task captured only the session ID while reading everything else at fire time from live mutable state — `messages`, reconciliation IDs, restoration guards, and crucially `activeProfile`, which selects the cache namespace (`SessionPresentationCache` keys are `"<profile>|<sessionID>"`). `switchProfile(to:)` wrote the outgoing transcript but never cancelled `presentationCacheFlushTask`, so a flush parked under profile A could wake after the switch completed and save A's session IDs — sometimes with A's stale transcript still in memory — through **profile B's cache namespace**, cross-contaminating presentation metadata between profiles.

## Fix

Three independent layers, so no single caller can regress the boundary:

1. **Self-defending deferred write** — every scheduled flush captures `(scheduledProfile, scheduledEpoch)` and drops itself after waking if either no longer matches. The epoch is bumped only by the sole mutator `setActiveProfile(_:)`, so an **A→B→A round trip** (which a string guard alone would pass) still fences the task.
2. **Fence-bearing identity mutator** — all three runtime `activeProfile` mutation sites funnel through `setActiveProfile(_:)`. Deliberately fence-only: on failed-switch rollbacks the still-active identity does *not* own the restored in-memory content, and a trace-instrumented regression test caught exactly that mis-attribution when a flush was temporarily centralized here.
3. **Explicit forward boundary flushes** — `switchProfile(to:reusing:)` *and* `connect(with:profile:)`'s profile-change branch call `flushPendingPresentationCache()` synchronously while the outgoing profile provably owns the in-memory transcript, cancelling the parked debounce and preserving its latest presentation metadata in its own namespace before any identity change. The failed-switch catch path restores prior state *before* flipping back, and writes nothing during rollback.

**Preserved behavior:** stream coalescing math (2 s debounce / 5 s ceiling), turn-completion flushes, ordinary session switching, pending clarification/approval restoration semantics, and background/foreground resume are untouched. `SessionPresentationCache` itself is not redesigned.

## Files changed

- `Conduit/Services/AppState.swift` — epoch capture + stale-execution fence + handle cleanup in `schedulePresentationCacheFlush`; `setActiveProfile(_:)` sole-mutator fencing (with wrap/redundancy documentation); forward boundary flushes at both switch entry points; `#if DEBUG` test accessor for the pending flush task.
- `ConduitTests/CrossProfilePresentationCacheTests.swift` — new deterministic regression suite (gate-controlled debounce parking, zero wall-clock sleeps).

## Regression coverage

| Test | Pins |
|---|---|
| `testStreamFlushAcrossProfileSwitchCannotContaminateTargetProfile` | reported bug: parked flush across A→B must not populate B's namespace; A's latest transcript survives exactly |
| `testDeferredFlushStillLandsWhenProfileUnchanged` | anti-overcorrection guard: unchanged-profile flushes still land |
| `testConnectProfileChangePreservesOutgoingAndFencesStaleFlush` | `connect(with:profile:)` upholds the same boundary as `switchProfile` |
| `testFailedSwitchRollbackLeavesOnlyRestoredProfileNamespace` | aborted switch leaves exactly the restored profile's namespace |
| `testStaleFlushIsFencedAcrossRoundTripBackToOriginProfile` | A→B→A round trip: only the epoch fence drops the parked write |

## Validation

- Focused `CrossProfilePresentationCacheTests`: **5/5 passed** on the final head.
- Full suite via `ios_test` on the final head: **65 units, 0 failed / 0 skipped / 0 hangs**, coverage reconciled, no simulator recovery.
- Full suite was additionally green twice on intermediate heads (same configuration).

## Review gate

- Initial three-model local review (`workflows/multi-model-review.js`; deepseek-v4-flash/opencode-go, step-3.7-flash/stepfun, mimo-v2.5/xiaomi-token-plan-sgp, consolidation on deepseek-v4-flash): **APPROVE 3/3, zero BLOCKERs** at `d31783e`, one majority WARNING (`connect(with:profile:)` flush asymmetry).
- Remediation introduced a mis-attribution on the rollback path that the new failed-switch test caught via stack-trace instrumentation; corrected design (fence-only setter, per-site forward flushes) verified by the same suite.
- Gate **re-run on the final design**: no BLOCKERs; majority APPROVE with a required comment fix plus a nil-out hardening and an A→B→A test request — all applied in `e5d0a17` (comment/doc/test-only in nature, so no third gate run; full suite re-verified above). Remaining non-blocking suggestions (DEBUG fenced-drop log line, schema-key extraction, connect-failure rollback pin) are tracked as follow-ups.

## Commit series

Nine substantive commits plus one paired temporary-instrumentation commit (`ea6df94` adds a debug tracer, removed again in `3f9a055`; net-zero effect). Squash-merge recommended for a clean history.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented delayed presentation-cache updates from being written to the wrong profile after switching profiles.
  - Ensured pending transcript data is saved before profile transitions.
  - Improved cache consistency when profile connections fail or users return to a previous profile.

- **Tests**
  - Added comprehensive coverage for profile switching, failed transitions, deferred updates, and transcript preservation.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -860,6 +860,7 @@ final class AppState: ObservableObject {
             // operation safe on its own so no caller can forget it.
             guard self.activeProfile == scheduledProfile,
                   self.presentationCacheProfileEpoch == scheduledEpoch else {
+                presentationCacheFlushTask = nil
                 return
             }
             self.cacheMessagePresentation(for: [sessionId])
@@ -888,8 +889,9 @@ final class AppState: ObservableObject {
     /// transition site (`switchProfile`, `connect(with:profile:)`) where
     /// that ownership is guaranteed. Both fence conditions in
     /// `schedulePresentationCacheFlush` are intentionally redundant — string
-    /// equality is the readable guard, the wrapping `&+= 1` (~2^63 switches,
-    /// unreachable) catches A→B→A round-trips — do not simplify either away.
+    /// equality is the readable guard, the wrapping `&+= 1` (overflows only
+    /// after ~9.2×10^18 switches — unreachable) catches A→B→A round-trips —
+    /// do not simplify either away.
     private func setActiveProfile(_ newValue: String) {
         guard newValue != activeProfile else { return }
         activeProfile = newValue
@@ -6291,12 +6293,14 @@ final class AppState: ObservableObject {
             projectsLoading = false
             restoreActiveSessionState(for: previousProfile)
             restorePinnedSessions(for: previousProfile)
-            // Restore the outgoing transcript state BEFORE flipping the
-            // identity back: setActiveProfile(_:)'s boundary flush persists
-            // whatever is in memory under the outgoing (previous) profile, so
-            // it must already describe that profile here — otherwise the
-            // partially-loaded target transcript would be written into the
-            // restored profile's namespace.
+            // restoreActiveSessionState/restorePinnedSessions are
+            // synchronous and cache-write-free; the pre-flight boundary
+            // flush (above) already persisted the outgoing transcript under
+            // the previous profile's namespace. setActiveProfile(_:)
+            // deliberately does NOT flush: flipping the identity back only
+            // re-arms the namespace and bumps the fence epoch, making any
+            // flush scheduled under the failed target during the aborted
+            // attempt stale.
             setActiveProfile(previousProfile)
             connection = savedConnection
             client?.disconnect()
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -877,23 +877,35 @@ final class AppState: ObservableObject {
         cacheMessagePresentation()
     }
 
-    /// Assigns the active profile and fences deferred presentation-cache
-    /// work: any coalesced flush scheduled while another profile was active
-    /// becomes stale the moment the identity changes. Call this instead of
-    /// assigning `activeProfile` directly so no mutation site can skip the
-    /// fence.
+    /// Assigns the active profile and enforces the hard presentation-cache
+    /// profile boundary: any coalesced flush scheduled while another profile
+    /// was active is flushed synchronously here — while the outgoing profile
+    /// is still active — and then fenced by the epoch bump, so no mutation
+    /// site can skip the boundary or lose the outgoing transcript. The
+    /// flush-before-bump ordering is what makes the epoch a complete fence;
+    /// `presentationCacheProfileEpoch` uses wrapping arithmetic because ~2^63
+    /// switches are unreachable, and the profile string is compared
+    /// independently in the task anyway, so a wrap can never resurrect a
+    /// stale flush. Both fence conditions in
+    /// `schedulePresentationCacheFlush` are deliberately redundant with this
+    /// single mutator (string equality is the readable guard; the epoch
+    /// catches A→B→A round-trips) — do not simplify either away.
     private func setActiveProfile(_ newValue: String) {
         guard newValue != activeProfile else { return }
+        flushPendingPresentationCache()
         activeProfile = newValue
         presentationCacheProfileEpoch &+= 1
     }
 
+#if DEBUG
     /// Deterministic-test access to the pending coalesced flush. Awaiting its
     /// value settles every deferred-write decision (cancelled, fenced, or
-    /// completed) without wall-clock timing.
+    /// completed) without wall-clock timing. Test-only plumbing; compiled out
+    /// of release builds.
     var presentationCacheFlushOperationForTesting: Task<Void, Never>? {
         presentationCacheFlushTask
     }
+#endif
 
     // MARK: - Persistence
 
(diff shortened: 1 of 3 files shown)


kaishi00:
## Summary
This PR prevents cross-profile contamination of the presentation cache by fencing coalesced deferred writes: it captures `(profile, epoch)` at schedule time, drops stale executions after wake, funnels all runtime `activeProfile` mutations through `setActiveProfile(_:)`, and adds synchronous forward-boundary flushes in both `switchProfile` and `connect(with:profile:)`, with five deterministic regression tests.

## Findings

### [minor] Stale-fence reads participate in the same unsynchronized off-main access the class already has
**File:** Conduit/Services/AppState.swift:848-862
**Issue:** `AppState` is a plain `final class: ObservableObject` (Conduit/Services/AppState.swift:326) with no `@MainActor` isolation and no `Sendable` annotations, and it is `@Published`-observed from SwiftUI. The flush `Task` is created in a synchronous (main-thread-convention), non-isolated context, so after `suspendForDebounce` returns it may resume on a non-main executor while all `activeProfile`/`presentationCacheProfileEpoch` writes happen on the main thread. The fence's guarantee — "reads `activeProfile`/epoch and compares — is therefore backed by unsynchronized (data-race) reads that Swift's memory model does not order. The wake path can observe a stale identity or stale epoch. In practice the pre-flight `flushPendingPresentationCache()` cancellation makes the dangerous interleaving hard to hit, and the read-race pattern (on `messages`, `reconciliation`, etc.) predates this change — but the *new* code makes the fix's core invariant depend on it.
**Suggestion:** Not blocking given the pre-existing pattern, but the fence would be airtight if the compare happens while isolated, e.g. hop to the main actor for the guard (`await` capture + compare) or have `presentationCacheProfileEpoch`/`activeProfile` reads funneled through a `@MainActor`-isolated accessor instead of being read straight off-main. Worth revisiting when the class is eventually made MainActor-isolated.

### [minor] The round-trip test does not exercise the epoch fence it claims to pin
**File:** ConduitTests/CrossProfilePresentationCacheTests.swift:455-513
**Issue:** The PR and the test comment state `testStaleFlushIsFencedAcrossRoundTripBackToOriginProfile` is *the* distinguishing case for the epoch guard ("only the epoch fence drops the parked write"). But both `switchProfile(to: "work")` and the return leg call `flushPendingPresentationCache()` (AppState.swift:6189), which cancels the parked task at the first switch. When `gate.release()` resumes the suspended continuation, the flush wakes into `guard !Task.isCancelled` (AppState.swift:857) and returns before ever reaching the epoch comparison — so the epoch guard never executes in this test. The epoch bump is, however, *not* redundant in production: a flush scheduled in the async window between `flushPendingPresentationCache()` and `setActiveProfile(target)` inside `switchProfile`'s `mintChatResumeTicket` gap survives cancellation and is dropped only by the epoch fence.
**Suggestion:** Add a variant that bumps the epoch without cancelling the parked task — e.g. perform profile mutation *without* any pre-flush (or count on the injection seam so no `cancel()` runs) and only then release the gate — asserting the parked write is dropped solely because `epoch` changed. Update the comment to say the epoch fence is "redundant defense" rather than "the distinguishing fence", since as-written the test pins the cancellation path.

### [nit] Guarded clear of `presentationCacheFlushTask` can still race a successor handle
**File:** Conduit/Services/AppState.swift:862
**Issue:** The fenced-drop branch assigns `presentationCacheFlushTask = nil` inside the off-main task. Because the wake may lag behind a subsequent `schedulePresentationCacheFlush`, a fence-dropped old task can clear the reference to a *newer* in-flight flush, delaying/cancelling-coalescing that newer write's visibility to `lastPresentationCacheFlushDate`. This is a pre-existing flavor of the handle-management race (the original code already had `presentationCacheFlushTask = nil` at execution end), so it's not a regression.
**Suggestion:** Move the nil-out to the main thread (e.g. capture a sequence number at schedule time and only clear the handle if it still refers to this task), or compare-and-clear against the task identity.

### [nit] `let scheduledProfile = activeProfile` unconditionally captured before sim-first delivery
**File:** Conduit/Services/AppState.swift:838
**Issue:** Harmless today because all bump sites run a synchronous `flushPendingPresentationCache()` (which cancels) first, leaving a genuine epoch-only fence window only inside `switchProfile`'s `mintChatResumeTicket` await. With `setActiveProfile` documented as "fence-only" and the redundant-guard doc comment, readers may over-estimate how often the epoch guard fires. No defect, just doc/reality drift.
**Suggestion:** Optionally note in the `setActiveProfile` doc that the fence is primarily defense-in-depth because every bump site cancels first.

### [nit] Helper pastes a production storage-key string constant
**File:** ConduitTests/CrossProfilePresentationCacheTests.swift:202
**Issue:** `persistedCache(from:)` hardcodes `"conduit.sessionPresentation.v1"`; if the storage key changes, this decode silently returns `[:]` and tests can start passing vacuously.
**Suggestion:** Acceptable given the comment pins the version; asserted at least once via `XCTAssertNotNil`/seed-then-read so an empty decode can't silently pass.

## Verdict
APPROVE

The fix is well-designed and layered: pre-flight flush + cancel covers the reported bug deterministically, the epoch fence adds real defense for the narrow non-cancelled window, and `setActiveProfile` succeeds in centralizing the three mutation sites (confirmed: the only other `activeProfile` writers are in `init`, which runs before any flush can be scheduled). All five tests are deterministic and timing-free, `[weak self]` + cancellation handling is correct, and no force unwraps, retain cycles, or UI/accessibility surface changes were introduced. The listed findings are non-blocking; the unsynchronized fence read and the round-trip test's overstated coverage are worth a follow-up, and the reported full-suite pass plus the 5/5 focused suite support merging.

<a href="https://opencode.ai/s/ef5In7yA">New%20session%20-%202026-08-27T01%3A50%3A26.825Z</a>
[opencode session](https://opencode.ai/s/ef5In7yA)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33031317426)

PR #101, opened 2026-08-27
fix(presentation): deduplicate logical cache snapshots across alias lookups

kaishi00:
## Duplicated-candidate root cause

`SessionPresentationCache.merge(...)` resolved every supplied session ID and flattened each hit into one candidate pool:

```swift
let cached = sessionIDs.compactMap { stored[key(profile:sessionID:)] }.flatMap { $0.messages }
```

`save(...sessionIDs:)` intentionally writes equivalent records under every alias, and reconciliation feeds `[resolvedId, requestedId]` back in (`AppState.applyChatResume`). When aliases pointed at the same logical snapshot, the pool held that snapshot twice - `A B C D A B C D`. Because `bestMatch` tracked candidates through a `Set` of indices and resolved scores with an order-arbitrary `max(by:)`, repeated messages / tool calls (equal names, empty gateway inputs -> uniform `+5` preview scores) could consume duplicate copies or stale generations as though they were distinct history: cached timestamps, tool input/output presentation, attachments crossed onto wrong gateway rows. Merge still iterated the authoritative gateway list, so no transcript rows were duplicated - the corruption was entirely in the candidate pool.

**Deterministic pre-fix failures captured on unfixed main** (the first commit in this branch is tests-only for exactly that purpose):

- dual-alias text rows enriched `[stale-1, stale-2, fresh-3, fresh-4, fresh-5]` - the first rows attached metadata from the older generation;
- repeated same-name tools returned `stale-input-1/2` where fresh generations existed.

## How logical records are deduplicated

`merge()` now resolves aliases **in supplied order** and collapses them by a stable logical identity before flattening:

1. each unique ID string resolves at most one record (`seenCacheKeys`);
2. identity = FNV fingerprint of the JSON-encoded row matrix `[role | id | content-signature | toolName]` per row - value-derived, never object identity, deliberately excluding volatile stamps (timestamps/previews) so a re-stamped rewrite of the same transcript is recognized as the *same* snapshot;
3. among copies of one identity the freshest write wins (`updatedAt`, then message count; strict monotonic clock in tests), so stamp-drift rewrites collapse to their freshest generation;
4. only surviving logical snapshots flatten once into the pool.

Adjacent fix in the same symptom class (surfaced by the new tests): `bestMatch` broke score ties via `Set<Int>` iteration order - order-arbitrary even after dedup. It now scans remaining indices ascending and keeps the first highest score (`strict >`), so equal-scoring repeats deterministically enrich first-cached-first, and the absolute-maximum exact-id match returns immediately.

## Why alias lookup still works

* `save` / `removePendingDecision` / `recordPendingDecision` keep writing every supplied alias - unchanged;
* lookup follows all supplied IDs; an absent alias simply contributes nothing;
* single-alias merges are behaviorally identical to before (identity map holds one entry) - covered by test;
* genuinely divergent snapshots have different identities, are NOT collapsed, and both stay available to the matcher - pinned by `testDifferentSnapshotsRemainAvailableToMatcher`.

Documented contract (review-gate W1 fallback): when two *divergent* snapshots tie at equal score, precedence follows argument order - earliest position wins. Production passes `[resolvedId, requestedId]`, so the live write leads.

## Which regression tests failed before the fix

Verified red against unmodified main (commit ordering in this branch preserves the evidence):

* `testDualAliasMergeYieldsFreshestSnapshotOncePerRow`
* `testRepeatedToolCallsKeepDistinctFreshMetadataAcrossAliases`

Also added and green before-and-after (compatibility guards): duplicate strings inside `sessionIDs`, single-alias behavior unchanged, distinct-snapshots-stay-available, argument-order tie contract pinning, plus an integration test driving the real `applyChatResume` path with `[resolvedId, requestedId]` and five repeated rows (`testReconciliationMergeDeduplicatesRequestedAndResolvedAliases`).

## Validation (final head, via Harness ios_test)

| Run | Result |
|---|---|
| Focused: SessionPresentationCacheTests + AppStateChatResumeTests | passed |
| Full Conduit suite | **64 units / 64 passed / 0 failed / 0 skipped / 0 hangs** |

(One earlier full run lost eight suites to a simulator host-bootstrap crash - infrastructure, no attributable cases. All eight passed identically re-run isolated, and a subsequent complete full-suite run was green end-to-end.)

## Scope notes

* No persistence migration; storage format untouched.
* PR #100 profile-switch behavior untouched: branch diff has zero deltas outside `SessionPresentationCache.swift` and two test files.
* Pending clarification/approval restoration flows run unchanged downstream of the deduplicated pool; supersede/expiry suites remain green.
* Follow-ups noted in review but intentionally out of scope here: freshness-weighted tie-breaking across divergent snapshots (currently argument-order per documented contract), DEBUG log line on fenced drops, monotonic per-write order counter replacing wall-clock `updatedAt` comparison.

Gate history: local three-model review via `workflows/multi-model-review.js` (deepseek-v4-flash/opencode-go, step-3.7-flash/stepfun, mimo-v2.5/xiaomi-token-plan-sgp; consolidation deepseek-v4-flash/opencode-go) - **APPROVE 3/3, zero blockers**, on this branch's final diff modulo its own prescribed remediations (JSON-encoded identity key = reviewer A's W2 fix, ordering-contract documentation = W1 fallback, exact-id early exit = S1).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session resume reconciliation when multiple session IDs refer to the same session.
  * Prevented duplicate messages and ensured the freshest cached presentation data is used.
  * Made matching results deterministic when candidates have equal scores.
  * Preserved distinct session snapshots while avoiding duplicate cache entries.

### DIFF
--- a/Conduit/Services/SessionPresentationCache.swift
+++ b/Conduit/Services/SessionPresentationCache.swift
@@ -169,6 +169,12 @@ final class SessionPresentationCache {
         // are resolved in supplied order; when several aliases hold the same
         // logical snapshot — including one rewritten later with only fresh
         // presentation stamps — the freshest write wins.
+        //
+        // Ordering contract (review-gate W1): DIVERGENT snapshots that
+        // outscore equally tie-break by earliest pool position, which is
+        // this call's argument order. The sole production caller passes
+        // [resolvedId, requestedId] so the live write leads; keep callers
+        // passing the most-current alias first.
         var resolutionOrder: [String] = []
         var snapshotByID: [String: CachedSession] = [:]
         var seenCacheKeys = Set<String>()
@@ -603,7 +609,8 @@ final class SessionPresentationCache {
             var score = Int.min
 
             if candidate.id == message.id {
-                score = 1_000
+                // Absolute maximum; no later candidate can outrank it.
+                return index
             } else if let tool = message.tool {
                 guard candidate.toolName == normalized(tool.name) else { continue }
                 score = 50
@@ -625,6 +632,8 @@ final class SessionPresentationCache {
                 score = 20 - distance
             }
 
+            // Strict > : the first highest-scoring candidate in ascending
+            // index order wins the tie, never arbitrary Set iteration order.
             if score > bestScore {
                 bestScore = score
                 bestIndex = index
@@ -704,12 +713,22 @@ final class SessionPresentationCache {
     private static func logicalSnapshotIdentity(
         for messages: [CachedMessage]
     ) -> String {
-        let rowIdentity = messages.map { message -> String in
-            String(describing: message.role)
-                + "|" + message.id
-                + "|" + message.signature
-                + "|" + (message.toolName ?? "")
+        // Review-gate W2: encode the row matrix as JSON before fingerprinting
+        // so no separator/field content can ever alias two structurally
+        // different sequences onto one identity.
+        let rowIdentity = messages.map { message -> [String] in
+            [
+                String(describing: message.role),
+                message.id,
+                message.signature,
+                message.toolName ?? ""
+            ]
         }
-        return fingerprint(rowIdentity.joined(separator: "\u{1F}"))
+        guard let data = try? JSONEncoder().encode(rowIdentity) else {
+            // Encoding cannot fail for [ [String] ]; failing open can only
+            // ever ADD a candidate, never silently drop a real one.
+            return UUID().uuidString
+        }
+        return fingerprint(String(decoding: data, as: UTF8.self))
     }
 }
--- a/Conduit/Services/SessionPresentationCache.swift
+++ b/Conduit/Services/SessionPresentationCache.swift
@@ -160,9 +160,40 @@ final class SessionPresentationCache {
         includePendingApprovals: Bool = false
     ) -> [ChatMessage] {
         let stored = load()
-        let cached = sessionIDs
-            .compactMap { stored[key(profile: profile, sessionID: $0)] }
-            .flatMap { session in
+        // Resolve every supplied alias, but let each LOGICAL cached snapshot
+        // enter the matching pool exactly once. save(...sessionIDs:) writes
+        // equivalent records under every alias, so a requested+resolved
+        // lookup used to flatten the same rows twice and the matcher could
+        // consume stale duplicates as though they were distinct historical
+        // messages (fingerprint and positional scoring especially). Aliases
+        // are resolved in supplied order; when several aliases hold the same
+        // logical snapshot — including one rewritten later with only fresh
+        // presentation stamps — the freshest write wins.
+        var resolutionOrder: [String] = []
+        var snapshotByID: [String: CachedSession] = [:]
+        var seenCacheKeys = Set<String>()
+        for rawSessionID in sessionIDs {
+            let trimmed = rawSessionID.trimmingCharacters(in: .whitespacesAndNewlines)
+            guard seenCacheKeys.insert(trimmed).inserted else { continue }
+            guard let session = stored[
+                key(profile: profile, sessionID: trimmed)
+            ] else { continue }
+            let identity = Self.logicalSnapshotIdentity(for: session.messages)
+            if let existing = snapshotByID[identity] {
+                let isFresher =
+                    session.updatedAt == existing.updatedAt
+                    ? session.messages.count > existing.messages.count
+                    : session.updatedAt > existing.updatedAt
+                if isFresher {
+                    snapshotByID[identity] = session
+                }
+            } else {
+                resolutionOrder.append(identity)
+                snapshotByID[identity] = session
+            }
+        }
+        let cached = resolutionOrder.compactMap { snapshotByID[$0] }
+            .flatMap { session -> [CachedMessage] in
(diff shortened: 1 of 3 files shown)


kaishi00:
I reviewed the full diff of `SessionPresentationCache.swift` and both test files, plus the sole production call site in `AppState.swift`.

## Summary
This PR deduplicates logically-identical cached presentation snapshots that `merge(...)` used to flatten once per alias (so `[resolvedId, requestedId]` lookups no longer double the candidate pool and the matcher can't consume stale duplicates), and makes `bestMatch` tie-breaking deterministic by scanning candidates in ascending pool order with strict `>` instead of order-arbitrary `Set` iteration via `max(by:)`.

## Findings

### [minor] "Freshest wins" relies on second-precision wall-clock `updatedAt`; the count tiebreak is dead code
**File:** Conduit/Services/SessionPresentationCache.swift:188-192 (persistence at 573-575)
**Issue:** `updatedAt` is a `Date` persisted through the default `JSONEncoder` strategy (`.deferredToDate`, seconds since reference date). Two saves in the same second encode to the *same* value, so `session.updatedAt > existing.updatedAt` cannot distinguish recency. When times compare equal the code falls back to `session.messages.count > existing.messages.count` — but two snapshots with the same logical identity have, by construction, the identical JSON row matrix, hence identical row counts, so that branch is *always false*. Under a same-second rewrite the "freshest" landing copy is actually decided by argument order, not recency — which can prefer a stale-listed alias. Low likelihood because the reported bug shape spans separations of at least a heartbeat, but the fallback logic does not do what it appears to (and the PR body itself lists a monotonic counter as a follow-up).
**Suggestion:** Store/compare a monotonic per-write counter, or encode `Date` with `millisecondsSince1970`, or at minimum tighten the fallback so it can never silently prefer an argument-order-older write.

### [minor] Identity omits tool I/O and decision payloads — divergence guarantee is overstated
**File:** Conduit/Services/SessionPresentationCache.swift:704-733
**Issue:** `logicalSnapshotIdentity` keys on `(role, id, content-signature, toolName)`. Two snapshots whose rows genuinely differ only in tool input/output signatures, tool previews, or clarify/approval payloads (same role, id, and empty-content fingerprint) get the *same* identity and are collapsed to the freshest write, silently discarding the other copy's fields. The doc comment claims "snapshots whose rows genuinely differ keep separate identities and all stay available" — over-broad. In practice `preservingPresentation` forwards `toolInputSignature`/`toolOutputSignature`/`preview`/decision cards on save, so the risk is an edge case (a re-stamp write that intentionally changes a tool's input/output for an unchanged id), but the comment over-promises.
**Suggestion:** Document the exclusion explicitly (opening a row's value under an unchanged id is collapsed), or fold `toolInputSignature`/`toolOutputSignature`/`decisionKey` into the identity if those must never be overwritten across alias writes.

### [minor] Fail-open identity generation silently disables the fix
**File:** Conduit/Services/SessionPresentationCache.swift:727-731
**Issue:** On `JSONEncoder` failure, `logicalSnapshotIdentity` returns a fresh `UUID().uuidString`, which defeats dedup entirely and reproduces the exact duplicated-candidate bug this PR fixes — with zero diagnostics. It cannot fail for `[[String]]`, so the guard is defensive, but failing open masks bugs rather than surfacing them.
**Suggestion:** Use `preconditionFailure`/`assertionFailure` in DEBUG (or document that this path is unreachable), instead of silently reverting to buggy behavior.

### [nit] Test name overclaims the tie contract it is supposed to pin
**File:** ConduitTests/SessionPresentationCacheTests.swift:369 (`testDivergentSnapshotTieResolvesByAliasArgumentOrder`)
**Issue:** The test name asserts argument-order tie precedence, but the assertions only check that both timestamps are present (`Set`) and differ — they never verify *which* row wins per gateway position. The inline comment is also internally contradictory (first says identities diverge only via stamps, then says the opposite). So the documented contract it claims to pin is not actually pinned.
**Suggestion:** Assert `merged[0].timestamp == "newer-ts"` given `[resolvedId, requestedId]` order, or rename the test and drop the misleading comment.

### [nit] Cross-file test helper coupling
**File:** ConduitTests/SessionPresentationCacheTests.swift:2020 (`DeterministicClock`), used in ConduitTests/AppStateChatResumeTests.swift:3425
**Issue:** The shared `DeterministicClock` is defined inside `SessionPresentationCacheTests.swift` and consumed by another test source file. It compiles only because both files share one test module; any future target split silently breaks the reference.
**Suggestion:** Move it to a shared test-support file (e.g. `TestSupport.swift`) explicitly included in the test target.

## Verdict
APPROVE

No critical or major issues. The fix is sound, well-scoped (zero deltas outside the cache service and two test files), and the deterministic tie-break plus dedup logic are correct. The findings above are robustness/robustness-of-claim nits that don't block merge; none involve retain cycles, force unwraps, or new concurrency hazards in production code (the only `@unchecked Sendable` is test-scoped and lock-guarded).

<a href="https://opencode.ai/s/X7009iUo">New%20session%20-%202026-08-27T03%3A32%3A09.502Z</a>
[opencode session](https://opencode.ai/s/X7009iUo)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33036598524)

PR #102, opened 2026-08-27
fix(kanban): responsive selection chrome on iPhone (#98)

### LINKED ISSUE #98: Kanban UI texts are broken
Mostly, take a look at the screenshot. The menu is completely unreadable. The bottom menu has two lines instead of a single line.

iPhone 17, iOS 26.6

Image

kaishi00:
# fix(kanban): responsive selection chrome on iPhone (#98)

Fixes #98 — "Kanban UI texts are broken" (reproduced on iPhone 17 / iOS 26.6).

## Root cause

Two independent horizontal-constraint problems:

1. **Selection-mode header.** The V3C selection controls ('Cancel' + 'N tasks selected') were rendered by an else branch *inside* the ordinary board-header HStack. With the board picker, overflow menu, refresh and add buttons still occupying layout, SwiftUI had almost no width left at phone sizes — producing Ca/nc/el and 0/tas/ks/se-/lec/ted.
2. **Bottom bulk-action bar.** A fixed labeled HStack (Move / Assign / More) wrapped every action label into stacked fragments once the count text + busy indicator claimed their space.

### 1. Selection mode REPLACES the header row

- While isSelectionActive the whole top row renders a dedicated compact KanbanSelectionHeaderBar: ONE full-width glass capsule (mirroring the legacy selection pill and the search/filter bar idiom) containing the Cancel control (exit stays idle-only — disabled while bulkBusy), a Spacer, ONE logical lineLimit(1) count label, and an optional ProgressView when busy.
- Normal mode renders standardBoardHeader: the previous interactive header verbatim (visually unchanged).
- The search/filters row hides with the board chrome during selection (it was already inert there).
- The select-tasks entry button now renders unconditionally in the standard header (previously mutually exclusive with the inline selection cluster).
- Accessibility preserved: 'Exit selection mode' cancel label, count label, identifiers on both chrome components.

### 2. Adaptive bottom bar (ViewThatFits)

- KanbanBulkActionsCluster composes Move/Assign/More through ViewThatFits(in: .horizontal): labeled icon+title candidate when roomy, icon-only candidate (36pt touch targets) when narrow. Fallback spacing equals primary spacing so the fallback candidate can never be the wider one under extreme width pressure.
- ONE shared control factory builds both candidates, so staging payloads (.move(""), .assign(nil), Priority/Archive/Delete overflow), .disabled gating (!canRunBulk || bulkBusy) and VoiceOver labels are identical by construction across variants.
- Count text gets lineLimit(1); bar geometry/material/safe-area insets untouched.

### 3. Pure policy extraction

enum KanbanSelectionChromePolicy owns count singular/plural (incl. '0 tasks selected'), the header-variant mapping, the idle-only cancel rule, and full/compact BulkActionDescriptor arrays plus their shared accessibility strings — making every responsive decision unit-testable without UI, following this file's existing policy-enum convention.

## Preserved V3C behavior (unchanged code paths)

selectionContext stamps & ownership (isSelectionOwned), frozen/staged selections, PendingBulkOperation/PendingBulkDelete flows, selection pruning + auto-exit, board/server-switch invalidation, confirmation semantics, bulkBusy gating everywhere it existed before, sheet presentation, safe-area handling of the bottom bar. No business logic moved; only view/policy extraction.

## Regression coverage (ConduitTests/KanbanSelectionLayoutTests.swift)

| Issue requirement | Covered by |
|---|---|
| Normal mode keeps board controls | header-variant mapping test |
| Selection mode uses compact replacement header | policy mapping + hosted-width tests |
| Compact actions preserve Move/Assign/More semantics + a11y labels | descriptor-equivalence tests (kinds, SF Symbols, per-count VoiceOver strings pinned variant-independent) |
| Count singular/plural incl. zero | policy label tests |
| Exit still possible while idle | idle-cancel gate test (+ visual-disabled parity in hosted busy test) |
| bulkBusy blocks inappropriate cancellation/actions | cancel-gate + gated controlsEnabled composition test |
| Real narrow-iPhone check | hosted-layout fitting-height proofs at 393pt for header AND composed bar, counts 0/1/3, Dynamic Type large/xLarge/xxLarge |

Note: runtime accessibility-tree introspection is deliberately not asserted in hosted tests — SwiftUI materializes the a11y tree lazily outside real assistive clients — so the VoiceOver strings are pinned at the policy layer instead.

## Validation

Branch tip `8c2001c` on top of current main (`19a3e60`, post #100/#101):

- ios_test on a clean Mac clone at the exact committed bytes: KanbanV2Tests + KanbanV3CTests + new suite green, repeatedly — including after the review-gate refinements landed.
- Full-suite run mid-development: 64/65 suites pass; sole failure is a pre-existing ConduitUITests-Runner bootstrap crash reproduced identically WITHOUT this diff (simulator environment, unrelated to the change).
- App-target build: zero compiler warnings from either changed file (remaining repo warnings pre-date this branch; e.g. the long-standing unused `board` binding in KanbanViews' loading else-if chain).

## Review gate disclosure

The house multi-model gate ran with pinned reviewers Step 3.7 Flash (stepfun) and MiMo V2.5 (xiaomi-token-plan-sgp): **APPROVE** and **COMMENT**, zero BLOCKERs. All four of their WARNING-class findings were adjudicated and folded into this branch (full-row header glass unification, equal fallback spacing, documented font-metric/UI-bridge couplings, added busy-state composed-bar coverage), plus the 36pt icon-target compromise from MiMo's suggestions.

DeepSeek V4 Flash (opencode-go) was UNAVAILABLE for this gate: its endpoint exceeded every execution ceiling across five attempts (~50 minutes). Both the third review and the consolidation had to be conducted by the harness instead of that pin; a follow-up full-trio re-run can be requested once the endpoint recovers if desired before merge.

## Out of scope (untouched)

Board redesign, selection semantics, bulk-operation behavior, backend/API behavior, ownership/context logic, snapshot test infrastructure.

## Summary by CodeRabbit

- **New Features**
  - Added responsive multi-select controls for Kanban boards.
  - Selection mode now shows a compact Cancel and selection-count header.
  - Bulk actions adapt between labeled and icon-only controls based on available space.
  - Search and filter controls are hidden during selection mode.
  - Preserved Move, Assign, Priority, Archive, and Delete workflows.
  - Improved accessibility labels and identifiers across selection controls.

- **Tests**
  - Added coverage for responsive layouts, Dynamic Type, accessibility, and busy states.

### DIFF
--- a/Conduit/Views/KanbanViews.swift
+++ b/Conduit/Views/KanbanViews.swift
@@ -165,15 +165,17 @@ enum KanbanSelectionChromePolicy {
     ]
 
     /// VoiceOver text — deliberately IDENTICAL for both variants: the labels,
-    /// not visual titles, carry the meaning when icons stand alone.
+    /// not visual titles, carry the meaning when icons stand alone. The noun
+    /// is singular for exactly one selected task, plural otherwise.
     static func bulkAccessibilityLabel(_ kind: BulkActionDescriptor.Kind, selectedCount: Int) -> String {
+        let noun = selectedCount == 1 ? "task" : "tasks"
         switch kind {
         case .move:
-            return "Move \(selectedCount) selected tasks"
+            return "Move \(selectedCount) selected \(noun)"
         case .assign:
-            return "Assign \(selectedCount) selected tasks"
+            return "Assign \(selectedCount) selected \(noun)"
         case .more:
-            return "More actions for \(selectedCount) selected tasks"
+            return "More actions for \(selectedCount) selected \(noun)"
         }
     }
 }
@@ -1532,13 +1534,14 @@ struct KanbanView: View {
         // picker / overflow / refresh / add controls alongside "Cancel" +
         // "N tasks selected", which on iPhone widths wrapped into unreadable
         // character fragments.
-        if isSelectionActive {
+        switch KanbanSelectionChromePolicy.headerVariant(isSelectionActive: isSelectionActive) {
+        case .compactSelection:
             KanbanSelectionHeaderBar(
                 selectedCount: selectedTaskIDs.count,
                 bulkBusy: bulkBusy,
                 onCancel: { exitSelectionMode() }
             )
-        } else {
+        case .standardBoardControls:
             standardBoardHeader
         }
 
--- a/Conduit/Views/KanbanViews.swift
+++ b/Conduit/Views/KanbanViews.swift
@@ -90,6 +90,254 @@ enum KanbanCardDeletePolicy {
     }
 }
 
+/// Issue #98 (Kanban selection chrome on narrow iPhones): pure presentation
+/// policy for the V3C multi-select header and the bottom bulk-action bar.
+/// Extracted so the responsive decisions — which chrome variant renders,
+/// what each bulk action shows, and how counts pluralize — stay
+/// deterministic and regression-testable without dragging the whole board
+/// screen into a test harness.
+enum KanbanSelectionChromePolicy {
+    /// Which top-row chrome is visible. Selection mode REPLACES the ordinary
+    /// interactive board controls instead of sharing their row: at iPhone
+    /// widths "Cancel" plus "N tasks selected" cannot coexist with the board
+    /// picker / overflow / refresh / add buttons without wrapping into
+    /// character fragments (issue #98).
+    enum HeaderVariant: Equatable {
+        case standardBoardControls
+        case compactSelection
+    }
+
+    static func headerVariant(isSelectionActive: Bool) -> HeaderVariant {
+        isSelectionActive ? .compactSelection : .standardBoardControls
+    }
+
+    /// Exiting selection mode is an idle-time action only: while a bulk
+    /// operation is in flight (`bulkBusy`) Cancel stays inert so an exit can
+    /// never race the in-flight mutation's frozen-context bookkeeping.
+    static func allowsCancelExit(bulkBusy: Bool) -> Bool {
+        !bulkBusy
+    }
+
+    /// ONE logical single-line label — never a vertical letter column.
+    static func selectedCountLabel(count: Int) -> String {
+        count == 1 ? "1 task selected" : "\(count) tasks selected"
+    }
+
+    /// One bulk-action control in the bottom bar. `title == nil` renders it
+    /// icon-only (the compact width variant).
+    struct BulkActionDescriptor: Equatable, Identifiable {
+        enum Kind: Equatable {
+            case move
+            case assign
+            case more
+        }
+
+        let kind: Kind
+        /// Visible text; nil means icon-only.
+        let title: String?
+        let systemImage: String
+
+        var id: Kind { kind }
+
+        /// Stable automation identifier shared by BOTH variants so tests (and
+        /// UI automation) see identical semantics for full and compact bars.
+        var accessibilityIdentifier: String {
+            switch kind {
+            case .move: return "kanban.bulk.move"
+            case .assign: return "kanban.bulk.assign"
+            case .more: return "kanban.bulk.more"
+            }
+        }
+    }
+
+    /// Roomy variant: icon + title per action.
+    static let fullBulkActions: [BulkActionDescriptor] = [
+        BulkActionDescriptor(kind: .move, title: "Move", systemImage: "arrow.left.arrow.right"),
+        BulkActionDescriptor(kind: .assign, title: "Assign", systemImage: "person.fill.badge.plus"),
+        BulkActionDescriptor(kind: .more, title: "More", systemImage: "ellipsis.circle"),
+    ]
+
+    /// Compact variant (narrow widths): the SAME actions, icon-only.
+    static let compactBulkActions: [BulkActionDescriptor] = [
+        BulkActionDescriptor(kind: .move, title: nil, systemImage: "arrow.left.arrow.right"),
+        BulkActionDescriptor(kind: .assign, title: nil, systemImage: "person.fill.badge.plus"),
+        BulkActionDescriptor(kind: .more, title: nil, systemImage: "ellipsis.circle"),
+    ]
+
+    /// VoiceOver text — deliberately IDENTICAL for both variants: the labels,
+    /// not visual titles, carry the meaning when icons stand alone.
+    static func bulkAccessibilityLabel(_ kind: BulkActionDescriptor.Kind, selectedCount: Int) -> String {
+        switch kind {
+        case .move:
+            return "Move \(selectedCount) selected tasks"
+        case .assign:
+            return "Assign \(selectedCount) selected tasks"
+        case .more:
+            return "More actions for \(selectedCount) selected tasks"
+        }
+    }
+}
+
+/// Issue #98: one bulk-action control SHARED by both responsive variants of
+/// the bottom bar (labeled when roomy, icon-only when narrow). Accessibility
+/// semantics are variant-independent by construction.
+struct KanbanBulkBarButtonItem: View {
+    let descriptor: KanbanSelectionChromePolicy.BulkActionDescriptor
+    let selectedCount: Int
+
+    var body: some View {
+        Group {
+            if let title = descriptor.title {
+                Label(title, systemImage: descriptor.systemImage)
+            } else {
+                Image(systemName: descriptor.systemImage)
+                    // Icon-only still needs a comfortable tap target even
+                    // though it draws less ink; 36pt keeps the compact row
+                    // single-line at phone widths while honoring most of the
+                    // HIG touch-target guidance.
+                    .padding(.horizontal, 4)
+                    .frame(minWidth: 36, minHeight: 36)
+                    .contentShape(Rectangle())
+            }
+        }
+        .accessibilityIdentifier(descriptor.accessibilityIdentifier)
+        .accessibilityLabel(
+            KanbanSelectionChromePolicy.bulkAccessibilityLabel(descriptor.kind, selectedCount: selectedCount)
+        )
+    }
+}
+
+/// Issue #98: the DEDICATED compact selection-mode header row. It REPLACES
+/// the ordinary board header entirely while selection is active, so the row
+/// never re-contests its width against the board picker / overflow /
+/// refresh / add controls. Cancel stays an idle-only action (`bulkBusy`
+/// keeps it disabled); the count remains ONE logical label.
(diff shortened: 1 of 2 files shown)


kaishi00:
## Summary
Fixes issue #98 by replacing the top row with a dedicated compact selection header when selection is active, and making the bottom bulk-action bar width-adaptive via ViewThatFits (labeled vs icon-only), extracting the responsive decisions into a testable policy enum with hosted-layout regression tests. Layout-only, view/policy extraction; business logic is preserved verbatim.

## Findings

### [minor] Extracted policy not fully wired into production — tested functions production never calls
**File:** Conduit/Views/KanbanViews.swift:110,1535; :122,1114
**Issue:** The new policy enum advertises a single source of truth, but two of its pinned behaviors don't actually flow through production:
- `KanbanSelectionChromePolicy.headerVariant(isSelectionActive:)` is unit-tested (`testSelectionModeReplacesRatherThanSharesTheHeaderRow`) but the production `header` uses its own inline `if isSelectionActive { KanbanSelectionHeaderBar } else { standardBoardHeader }` — the tested function is never invoked.
- `KanbanSelectionChromePolicy.selectedCountLabel(count:)` is used by the selection header, but the bottom `bulkActionBar` count (line 1045) still renders from the private `selectedCountLabel` computed property (line 1114) which duplicates the identical pluralization logic.

Consequently two of the advertised "pinned" regressions test functions production never (or no longer) calls, so a future drift in the inline branches would go uncaught — including the composed-bar test (`testComposedBulkBarFitsPhoneWidthWithoutWrappingForAllCounts`), which builds its mirror bar from the policy function rather than the live property.
**Suggestion:** Make `selectedCountLabel` delegate to `KanbanSelectionChromePolicy.selectedCountLabel(count:)` (delete the private copy), and render the header via `switch KanbanSelectionChromePolicy.headerVariant(isSelectionActive: isSelectionActive)` so the tested policy is the code path. Then the tests exercise production truth.

### [minor] `controlsEnabled` contains a redundant term that obscures the gating contract
**File:** Conduit/Views/KanbanViews.swift:1063
**Issue:** The cluster is created with `controlsEnabled: canRunBulk && !bulkBusy`, but `canRunBulk` already includes `!bulkBusy` internally (line 1120). The pre-#98 predicate `!canRunBulk || bulkBusy` reduces to `!canRunBulk`, so the extra term is dead. It reads as if `canRunBulk` were independent of `bulkBusy`, which is misleading for future maintainers and could lead to a subtle gating change if `canRunBulk` is ever refactored.
**Suggestion:** Pass `controlsEnabled: canRunBulk` and keep the comment that this is the pre-#98 per-control predicate.

### [nit] Dead `isSelectionActive` predicates inside `standardBoardHeader`
**File:** Conduit/Views/KanbanViews.swift:1589,1648
**Issue:** `standardBoardHeader` is only rendered when `!isSelectionActive` (header at line 1542), yet its board-picker menu and board-actions menu still carry `.disabled(bulkBusy || isSelectionActive)`. The `isSelectionActive` term can never be true in this view.
**Suggestion:** Drop `isSelectionActive` from these predicates, or note that they guard future reuse.

### [nit] Hosted text-scrape assertions are version-fragile by design
**File:** ConduitTests/KanbanSelectionLayoutTests.swift:218,293-298,346
**Issue:** `uiLabelTexts` depends on SwiftUI currently materializing `UILabel`s and on exact `text.count > 2`. This is documented in-repo, and the fallback fitting-height proofs are the real guards, but the `XCTFail` on a surviving "Move"/"Assign"/"More" title and the subheadline/body line-height magic numbers (test file lines 283, 310) can break on a future iOS without indicating a product regression.
**Suggestion:** Acceptable as-is given the documented coupling; if these ever flake, move the whole-word assertions behind the fitting-height proofs only.

### [nit] Icon-only tap target is 36pt, below the 44pt HIG guidance
**File:** Conduit/Views/KanbanViews.swift:277-282
**Issue:** The compact bulk-action buttons are `minWidth: 36 / minHeight: 36`, below Apple's 44pt minimum touch-target guidance, and with the tight 12pt internal spacing the effective tappable area is close to the edge-to-edge icon spacing.
**Suggestion:** The compromise is documented and justified for single-line fit on this bar; consider keeping 44pt on the `.more` menu and ensuring the large-content-size dynamic type case (which the tests do exercise for the header but not the cluster) still renders single-line.

## Verdict
APPROVE

No critical or major issues. The change is layout-only with good policy extraction and unusually thorough hosted-layout coverage; the findings above are maintainability/test-fidelity improvements, not correctness defects. No retain cycles, force unwraps, concurrency, or lifecycle issues were introduced.

<a href="https://opencode.ai/s/7lDJ30Sk">New%20session%20-%202026-08-27T10%3A09%3A13.630Z</a>
[opencode session](https://opencode.ai/s/7lDJ30Sk)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33061693602)

kaishi00:
## Summary
Fixes #98 by replacing the whole top row with a dedicated compact `KanbanSelectionHeaderBar` while selection is active, and making the bottom bulk-action bar width-adaptive via `ViewThatFits` (labeled vs. icon-only variants), with the responsive decisions extracted into a pure `KanbanSelectionChromePolicy` enum and pinned by hosted-layout + policy tests. Layout/policy extraction only; bulk staging, ownership, and gating semantics are preserved verbatim.

## Findings

### [minor] Bulk-bar count label bypasses the extracted policy (`selectedCountLabel` duplicated)
**File:** Conduit/Views/KanbanViews.swift:1116-1119 (used at 1047), policy at 122-124
**Issue:** The header's count is now policy-driven––`KanbanSelectionChromePolicy.selectedCountLabel(count:)` (line 122)––but the bottom `bulkActionBar` count (line 1047) still renders from the private `selectedCountLabel` computed property, which re-implements the identical pluralization. The regression test for the bottom bar (`testComposedBulkBarFitsPhoneWidthWithoutWrappingForAllCounts`) mirrors production geometry yet sources its label from the policy function (test line 325), not the live property. The bottom bar is the exact component the issue's symptom ("count text + busy indicator") describes, so a future pluralization/noun change in the policy would silently diverge from the bar and go untested.
**Suggestion:** Replace the private property with `KanbanSelectionChromePolicy.selectedCountLabel(count: selectedTaskIDs.count)` (single source of truth), and drop the private copy. Then the composed-bar test exercises production truth.

### [minor] Redundant `!bulkBusy` term in `controlsEnabled`
**File:** Conduit/Views/KanbanViews.swift:1065
**Issue:** `controlsEnabled: canRunBulk && !bulkBusy` is passed, but `canRunBulk` already includes `!bulkBusy` internally (line 1122), so the term is dead. The comment claims it mirrors the pre-#98 predicate (`!canRunBulk || bulkBusy`), which is true, but the extra term implies `canRunBulk` is independent of `bulkBusy`; if `canRunBulk` is ever refactored the redundant second term could silently mask a gating change.
**Suggestion:** Pass `controlsEnabled: canRunBulk` and keep the explanatory comment that this equals the pre-#98 per-control predicate.

### [minor] `bulkBusy` counter-active while busy but cancel policy tested at policy layer only
**File:** Conduit/Views/KanbanViews.swift:235,245; ConduitTests/KanbanSelectionLayoutTests.swift:363-371
**Issue:** The header Cancel relies on the belt-and-suspenders pattern (`.disabled(bulkBusy)` + `if allowsCancelExit(bulkBusy:)` around `onCancel()`). The busy-cancel gate is pinned only by a policy test (`testCancelExitIsIdleOnlyAndStaysBlockedWhileBulkBusy`) and a busy header that asserts geometry but passes a no-op `{}` cancel closure, so the once-fixed race (exit during an in-flight frozen-mutation) guards against the future by duplicating the policy check inline. Not a current defect.
**Suggestion:** Optional hardening: render the Cancel button directly via `Button(role:)`/style driven by `allowsCancelExit` and assert the hosted busy header's Cancel is disabled in a hosted composition test that calls the real closure.

### [nit] Dead `isSelectionActive` terms inside `standardBoardHeader`
**File:** Conduit/Views/KanbanViews.swift:1592, 1651, 1693
**Issue:** `standardBoardHeader` is only ever rendered when `!isSelectionActive` (the header switch at 1537 routes it otherwise), yet the board-picker menu (1592), board-actions menu (1651), and new-task button (1693) still carry `isSelectionActive` in their `.disabled(...)` predicates. The term can never be true in this view; it reads as a live guard that isn't.
**Suggestion:** Drop `isSelectionActive` from these predicates (or note they intentionally guard future reuse when selection chrome is later added beside them).

### [nit] Compact tap target 36pt below 44pt HIG
**File:** Conduit/Views/KanbanViews.swift:284-286
**Issue:** The icon-only variant pins `.frame(minWidth: 36, minHeight: 36)`, under Apple's 44pt minimum; combined with the tight 12pt HStack spacing, effective margins are small. Documented as a deliberate fit compromise.
**Suggestion:** Consider keeping 44pt on the `.more` menu (already carries the overflow), and note that the Dynamic-Type suite exercises the header but not the icon-only cluster.

### [nit] Hosted text-scrape assertions are OS-fragile by design
**File:** ConduitTests/KanbanSelectionLayoutTests.swift:110-112, 227, 301-305, 344-356
**Issue:** `uiLabelTexts` depends on SwiftUI materializing `UILabel`s and on exact `text.count > 2`; the "Move"/"Assign"/"More" whole-word `XCTFail` and the subheadline/body magic-number bounds (lines 202, 261-262, 291, 316-318) can break on a future iOS without a product regression. Documented in-review, and the fitting-height proofs and policy pins are the real guards.
**Suggestion:** Acceptable as-is given the documented coupling; if flake appears, move the whole-word assertions behind the fitting-height proofs only.

## Verdict
APPROVE

No critical or major issues. The layout-only extraction is clean: no force unwraps, no introduced retain cycles (all closures touch value-typed views / transient Button/Menu content; the test `HostBox` now retains its controller strongly), no concurrency or lifecycle regressions, and accessibility labels/identifiers are thoughtfully preserved (singular/plural noun fix included). The noteworthy residual is the duplicated `selectedCountLabel` pluralization that still bypasses the very policy the tests pin, plus a couple of dead/redundant predicate terms worth tidying.

<a href="https://opencode.ai/s/iPhFcH3O">New%20session%20-%202026-08-27T15%3A35%3A35.830Z</a>
[opencode session](https://opencode.ai/s/iPhFcH3O)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33088553133)

PR #105, opened 2026-08-28
feat(composer): optional hardware-keyboard Return-to-send in chat composer

kaishi00:
## Summary

Adds an opt-in **Chat → Keyboard → "Return key sends"** setting (default **off**) for hardware keyboards on iPad:

- **Setting ON:** plain hardware-keyboard **Return** submits through the exact same composer action path as the existing Send/action button (send / steer / interrupt — never stop). **Shift-Return** always inserts a newline.
- **Setting OFF:** behavior is byte-for-byte unchanged — Return inserts a newline; sending stays on the composer action button. Existing users see no change after updating (default false).

## Behavior matrix

| Composer state | Return (setting ON) | Shift-Return |
|---|---|---|
| idle + text/attachments | send (button's path) | newline |
| running + typed text | steer / interrupt (per busy-input mode) | newline |
| running, empty composer (stop-only) | newline — Return never acts as Stop | newline |
| synchronizing / reconnecting / unsupported | newline (gate closed) | newline |
| setting OFF, any state | newline (legacy) | newline |

## Implementation

- **`Conduit/Views/Components/ComposerReturnKey.swift` (new):** small pure decision layer — `decision(returnKeySends:shiftPressed:canSubmit:)` and `canSubmit(action:)` (send/steer/interrupt → true; stop/unavailable → false) — plus the preference key. Kept narrow on purpose: no composer state machine lives here.
- **`Conduit/Views/Components/ComposerPasteTextView.swift`:** `ImagePasteTextView.pressesBegan` is the only Return interception point. It classifies hardware Return presses (`UIPress.key.keyCode`: `keyboardReturnOrEnter`, `keyboardReturn`, `keypadEnter`) and consumes **only the Return press, and only when** the setting is on, the composer gate is open, **and** the composer's callback reports it acted — everything else forwards to `super`, preserving default newline behavior. Shift is detected from the press event, the key's modifiers, **and** direct tracking of physical Shift presses (hardens against Bluetooth keyboards whose chord modifiers can lag). Software keyboards never emit `UIPress` (they deliver `insertText("\n")` via the text-input protocol), so they are untouched by design — no `shouldChangeTextIn`, `returnKeyType`, or `submitLabel` hooks were used.
- **`Conduit/Views/Components/ComposerBar.swift`:** owns everything action-related. `@AppStorage(ComposerReturnKey.preferenceKey)` (default `false`), passes `canSubmitFromReturn: ComposerReturnKey.canSubmit(action: action)` computed from the existing composer action state, and `submitFromReturnKey()` re-checks the live gate before invoking the existing `submit()` — the identical path as the action button (haptics, draft collapse, `appState.submitComposer`). No send/attachment/draft/slash logic is duplicated in the text view.
- **`Conduit/Views/AuxiliaryViews.swift`:** new "Keyboard" section on the existing Chat settings page with the toggle and copy that states the hardware-keyboard scope explicitly.

## Persistence

Local, device-only preference: `UserDefaults` key `conduit.composerReturnKeySends` via `@AppStorage`. Default `false`, survives relaunch, takes effect without a reconnect, and is **never sent to the Hermes gateway**.

## Tests

New `ConduitTests/ComposerReturnKeyTests.swift` (16 tests): decision matrix (off+Return, on+Return, Shift-Return, non-submittable), action gate, press-consume/forward branch including the declined-callback fallback and missing-callback safety, modifier nil-classification, text view + representable legacy defaults, preference default-off, and persistence.

`ios_test` on an isolated Mac worktree at the same base commit (14d6603), all **0 uncovered**:

- `20260828-005725-8544` — ComposerReturnKeyTests + ComposerPasteTextViewTests + TurnStateTests: **passed**
- `20260828-005837-817b` — ComposerDraftStoreTests + ChatReturnSurfaceTests + HapticsTests + ChatTextSelectionTests + InteractionHapticsTests: **passed**
- `20260828-012649-3f92` (after review-gate fixes) — same three suites as the first run: **passed**

Full app + test targets build cleanly in every run.

## Review gate

Mandatory multi-model review (deepseek-v4-flash, step-3.7-flash, mimo-v2.5; consolidation on deepseek-v4-flash): consolidated **APPROVE** (2/3, one REQUEST CHANGES), **zero BLOCKERs**. The single dual-reviewer WARNING — whole-press-set consumption could drop chorded key presses, and Shift modifier reporting can lag on split-report Bluetooth keyboards — was fixed in 664b2d3: only the Return press is consumed and only when the composer acted, with physical Shift presses tracked directly. Reviewer suggestions S1 (keypad/alternate Enter treated as Return), S2 (declined-callback + shift-classification tests), and S3 (staleness comment, safe failure mode) are included.

## Caveats

- Software keyboard Return keeps inserting a newline even with the setting on; the settings copy states this explicitly.
- The gate freshness matches the action button (render-time state) and is re-checked live at press time; the stale-window failure mode is newline insertion, never an accidental send.
- Held/auto-repeated Return is inert after a submit: the composer clears synchronously, leaving a stop-only state the gate declines.

Image paste, drafts, attachments, slash suggestions, chat scrolling, session switching, and all backend behavior are untouched.

## Summary by CodeRabbit

* **New Features**
  * Added an optional setting to send messages by pressing Return on a hardware keyboard.
  * Return inserts a new line by default, or when Shift is held.
  * The setting is device-specific and available in Chat settings with explanatory accessibility guidance.

* **Bug Fixes**
  * Preserved normal newline behavior when message submission is unavailable or unsuccessful.

* **Tests**
  * Added coverage for Return-key submission, Shift-Return behavior, defaults, and fallback handling.

### DIFF
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -737,7 +737,12 @@ private struct ChatSettingsDetail: View {
                     ResponseBehaviorSettings(initialMode: busyInputMode, save: persistBusyInputMode)
                 }
             ),
-            trailingSection: AnyView(DeviceHapticsSettings())
+            trailingSection: AnyView(
+                VStack(spacing: 14) {
+                    ComposerReturnKeySettings()
+                    DeviceHapticsSettings()
+                }
+            )
         )
         .navigationTitle("Chat")
         .task { options = await loadOptions() }
@@ -752,6 +757,30 @@ private struct ChatSettingsDetail: View {
         .init(key: "display.memory_notifications", label: "Self-improvement updates", help: "Choose whether Conduit follows Hermes, always shows, or never shows maintenance updates.", control: .labeledOptions([(value: "default", label: "Use Hermes default"), (value: "on", label: "Always show"), (value: "off", label: "Never show")], defaultValue: "default")),
         .init(key: "agent.image_input_mode", label: "Image attachments", help: "How Hermes supplies images to a model.", control: .options(["auto", "native", "text"], defaultValue: "auto")),
     ]
+}
+
+/// Local, device-only composer input preference. Stored in UserDefaults via
+/// @AppStorage; never part of the Hermes profile configuration.
+private struct ComposerReturnKeySettings: View {
+    @AppStorage(ComposerReturnKey.preferenceKey) private var returnKeySends = false
+
+    var body: some View {
+        ConduitSettingsSection(
+            title: "Keyboard",
+            symbol: "keyboard",
+            tint: .conduitAura
+        ) {
+            Text("Press Return to send. Use Shift-Return for a new line.")
+                .font(.footnote)
+                .foregroundStyle(.secondary)
+            Text("Hardware keyboards only. The on-screen keyboard's Return key keeps inserting a new line.")
+                .font(.footnote)
+                .foregroundStyle(.secondary)
+            Toggle("Return key sends", isOn: $returnKeySends)
+                .tint(.conduitAccent)
+                .accessibilityHint("Applies to hardware keyboards only. The on-screen keyboard's Return key is unchanged.")
+        }
+    }
 }
 
 private struct DeviceHapticsSettings: View {
--- a/Conduit/Views/Components/ComposerBar.swift
+++ b/Conduit/Views/Components/ComposerBar.swift
@@ -32,6 +32,9 @@ struct ComposerBar: View {
     @State private var documentImportContext: AsyncAttachmentContext?
     @State private var attachmentGeneration: UInt64 = 0
     @State private var suppressNextTextChangeSuggestions = false
+    /// Local, device-only input preference. Defaults to off so existing
+    /// users keep Return inserting a newline after updating.
+    @AppStorage(ComposerReturnKey.preferenceKey) private var returnKeySends = false
     @Namespace private var glassNamespace
 
     struct AsyncAttachmentContext: Equatable {
@@ -332,7 +335,10 @@ struct ComposerBar: View {
                         onPastedImageError: { message in
                             handlePastedImageError(message, editorIdentity: currentEditorIdentity)
                         },
-                        editorIdentity: editorIdentity
+                        editorIdentity: editorIdentity,
+                        returnKeySends: returnKeySends,
+                        canSubmitFromReturn: ComposerReturnKey.canSubmit(action: action),
+                        onSubmitFromReturn: { submitFromReturnKey() }
                     )
                     .id(editorIdentity)
                     .padding(.horizontal, 5)
@@ -579,6 +585,16 @@ struct ComposerBar: View {
         .padding(.horizontal, 14)
         .padding(.top, 7)
         .padding(.bottom, 1)
+    }
+
+    /// Return-shortcut entry point. Invokes the exact same submission path
+    /// as the composer action button, but only for typed-message actions
+    /// (send/steer/interrupt): Return never acts as the stop-only control.
+    /// The gate is re-checked here so the existing composer action state —
+    /// not the text view — stays authoritative.
+    private func submitFromReturnKey() {
+        guard ComposerReturnKey.canSubmit(action: action) else { return }
+        submit()
     }
 
     private func submit() {
--- a/Conduit/Views/Components/ComposerBar.swift
+++ b/Conduit/Views/Components/ComposerBar.swift
@@ -337,6 +337,9 @@ struct ComposerBar: View {
                         },
                         editorIdentity: editorIdentity,
                         returnKeySends: returnKeySends,
+                        // May lag one render behind fast typing; the safe
+                        // failure mode is newline insertion, and
+                        // submitFromReturnKey() re-checks the live gate.
                         canSubmitFromReturn: ComposerReturnKey.canSubmit(action: action),
                         onSubmitFromReturn: { submitFromReturnKey() }
                     )
@@ -591,10 +594,14 @@ struct ComposerBar: View {
     /// as the composer action button, but only for typed-message actions
     /// (send/steer/interrupt): Return never acts as the stop-only control.
     /// The gate is re-checked here so the existing composer action state —
-    /// not the text view — stays authoritative.
-    private func submitFromReturnKey() {
-        guard ComposerReturnKey.canSubmit(action: action) else { return }
+    /// not the text view — stays authoritative. Reports whether the message
+    /// actually went out, so a declined shortcut press falls back to the
+    /// text view's default newline behavior instead of being swallowed.
+    @discardableResult
+    private func submitFromReturnKey() -> Bool {
+        guard ComposerReturnKey.canSubmit(action: action) else { return false }
         submit()
+        return true
     }
 
     private func submit() {
--- a/Conduit/Views/Components/ComposerPasteTextView.swift
+++ b/Conduit/Views/Components/ComposerPasteTextView.swift
@@ -193,8 +193,10 @@ final class ImagePasteTextView: UITextView {
     /// Ground-truth Shift state folded from the keyboard's own Shift press
     /// events. Some Bluetooth keyboards report chord modifiers with a lag on
     /// the event/key modifier surfaces, so the physical Shift press itself is
-    /// tracked as well.
-    private var hardwareShiftHeld = false
+    /// tracked as well. Left and right Shift are tracked independently so a
+    /// released key cannot clear a still-held second Shift.
+    private var heldShiftKeys: Set<UIKeyboardHIDUsage> = []
+    private var hardwareShiftHeld: Bool { !heldShiftKeys.isEmpty }
 
     override func pressesBegan(_ presses: Set<UIPress>, with event: UIPressesEvent?) {
         foldHardwareShift(presses, pressed: true)
@@ -204,8 +206,9 @@ final class ImagePasteTextView: UITextView {
                 || Self.shiftIsPressed(event: event, key: returnPress.key)
             // Consume only the Return press, and only when the composer
             // actually acted. Every other case (setting off, Shift-Return,
-            // non-submittable composer, declined callback) is forwarded with
-            // the remaining presses so default text behavior is preserved.
+            // marked text / IME composition, non-submittable composer,
+            // declined callback) is forwarded with the remaining presses so
+            // UIKit's normal text-input behavior is preserved.
             if handleReturnKeyPress(shiftPressed: shiftPressed) {
                 forwarding.remove(returnPress)
             }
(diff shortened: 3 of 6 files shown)


kaishi00:
## Summary

Adds an opt-in hardware-keyboard "Return sends" preference for the chat composer, with a pure decision layer, press-classification in the UIKit text view, gate re-checking in ComposerBar, and a Keyboard toggle on the Chat settings page.

## Findings

### [minor] Consumed Return press is still forwarded in `pressesEnded`; `pressesCancelled` is not handled
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:218
**Issue:** `pressesBegan` removes only the Return press from the forwarding set, but `pressesEnded` forwards the *entire* press set — including the consumed Return — to `super`. On iOS the text-input path can commit a key's text on key-up, so there is a risk that the cleared composer receives a stray newline immediately after a send, depending on how UIKit pairs the begin/end dispatch for a press whose `pressesBegan` never reached the input machinery. Additionally, `pressesCancelled` is not overridden, so `hardwareShiftHeld` can be left in a stale `true` state if a press is cancelled (interruption, keyboard disconnect), which would then mutate Shift-Return into a plain Return (a send).
**Suggestion:** Track the consumed `UIPress` identities (e.g., a `Set<ObjectIdentifier>` or the press objects) and filter them out of the sets passed to `super` in both `pressesEnded` and a new `pressesCancelled` override. Also reset `hardwareShiftHeld = false` in `pressesCancelled`. Worth an on-device confirmation that no newline lands after send with the feature on.

### [minor] `hardwareShiftHeld` single-bool tracking can be corrupted by chord/edge cases
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:224
**Issue:** The bool is folded from any Shift press in the set, so releasing one Shift while another is still held (left+right chord), a lagged press ordering where Return's `pressesBegan` arrives before Shift's on split-report Bluetooth keyboards, or a cancelled/replayed press can all make Shift detection resolve to "not held" — which, combined with the event/key modifier surfaces also not reporting it yet, turns an intended Shift-Return into an accidental send. The three-layer detection reduces but does not eliminate this.
**Suggestion:** This is an acknowledged tradeoff, but consider treating the case conservatively: if the Shift release for the tracked bool arrives *after* a Return press was already consumed as a submit, that press cannot be un-done anyway — so at minimum document it. A cheap hardening is to only set the bool on `pressesBegan` and to count held shift presses rather than a single flag.

### [nit] "Held/auto-repeated Return is inert after a submit" is overstated
**File:** Conduit/Views/Components/ComposerBar.swift:601
**Issue:** After a successful send the composer clears synchronously, so the gate becomes stop-only and `decision(...)` returns `.insertNewline` — which is *not* inert: the repeat key-downs are forwarded and insert newlines into the freshly cleared field if they arrive before `resignFirstResponder` lands (the press is only inert if the view is no longer the initial responder). The first-responder resign happens in `updateUIView`, which is a later render pass.
**Suggestion:** If auto-repeat is observed in the field, ignore held/repeated Return presses for a short window after a submit, or consume the Return press whenever the just-submitted state is stop-only rather than forwarding it as newline.

### [nit] Return-press consumption is keyed on "callback reported acted," not "send actually succeeded"
**File:** Conduit/Components/ComposerBar.swift:601
**Issue:** `submitFromReturnKey()` returns `true` after calling `submit()`, which only launches the async submit task. If the gateway rejects the message, the draft is restored but the Return press is already consumed (no newline). This exactly matches the action-button path (the button also doesn't insert a newline on failure), so it's consistent behavior — flagging only for awareness that the "acted" Bool means "initiated," not "delivered."

No retain cycles were found (the `onSubmitFromReturn` closure is cleaned up in `deactivate()`), the gate is safely double-checked at press time, decline paths fall back to newline, and the new toggle is properly labeled with an accessibility hint and clear hardware-only copy. Tests cover the decision matrix, gate, branch behavior, and persistence defaults.

## Verdict

APPROVE

<a href="https://opencode.ai/s/zP3n9jiI">New%20session%20-%202026-08-28T01%3A29%3A43.346Z</a>
[opencode session](https://opencode.ai/s/zP3n9jiI)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33132966742)

kaishi00:
**Hardening pass pushed** (9997852 + 24b621b), per follow-up review notes:

- **CI shards:** `ComposerReturnKeyTests` assigned to `unit-a` (next to `ComposerPasteTextViewTests`) in `scripts/test-shards.txt`; `scripts/check-test-shards.sh` passes locally (68 classes fully assigned) and the GitHub shard-coverage check is green.
- **IME safety:** `ComposerReturnKey.decision` now takes `hasMarkedText`; `ImagePasteTextView` feeds it `markedTextRange != nil`, so with marked text active the Return shortcut always declines and the press is forwarded to UIKit's normal multistage input. No manual commit/newline. Covered by `testSettingOnPlainReturnWithMarkedTextInsertsNewline` (policy) and `testMarkedTextForwardsReturnInsteadOfSubmitting` (view seam — a live IME session can't be synthesized in XCTest).
- **Cancelled press sequences:** `pressesCancelled` now releases tracked Shift state; L/R Shift tracked independently via `Set<UIKeyboardHIDUsage>` with a pure `updatedShiftState` transition (cancel = key-up). Covered by three new transition tests (cancel release, second-Shift survival, non-Shift key-code isolation).
- **Second-gate fixes:** Return classification now also scans `event.allPresses` for a currently-held Shift (keyboards delivering the Shift key-down in a later event than Return), and tracked Shift state resets on `resignFirstResponder` so an interrupted chord can't turn a later plain Return into Shift-Return.

Validation: focused run `20260828-022201-d5ab` (ComposerReturnKeyTests / ComposerPasteTextViewTests / TurnStateTests) passed; full unit suite `20260828-020445-171d` passed 67/67 suites, 0 uncovered, 0 hangs; local shard check passed; GitHub "Shard coverage check" job green on 24b621b.

Known residual hardware caveats (XCTest cannot synthesize UIPress/IME): physical-chord ordering on third-party Bluetooth keyboards, held-Return auto-repeat, and Full Keyboard Access routing are covered by the pure seams + code review but not on-device automation.

kaishi00:
## Summary
Adds an opt-in, device-local ("Chat → Keyboard → Return key sends", default off) hardware-keyboard Return-to-send shortcut for the chat composer, with a pure decision layer (`ComposerReturnKey`), press classification/consumption in the UIKit text view, a live-gated submit path in `ComposerBar`, and 16 unit tests.

## Findings

### [minor] Held/auto-repeated Return after a submit can insert stray newlines into the freshly cleared composer
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:201
**Issue:** The caveat "Held/auto-repeated Return is inert after a submit" is overstated. `submit()` clears the composer synchronously, so the very next auto-repeat key-down hits a stop-only/empty state where `decision(...)` returns `.insertNewline` and `pressesBegan` **forwards** the press; nothing in the code marks "just submitted." Until `resignFirstResponder` lands (it happens on the later render in `updateUIView`), a held Return keeps inserting `\n` into the empty field, which can then be sent accidentally later. This differs from `canSubmit=false` idle behavior only in that the user *just* sent, making the newlines unexpected.
**Suggestion:** Track a short post-submit window (e.g., timestamp the last consumed submit and ignore/consume Return key-downs within it), or consume the Return press whenever the live gate declines immediately after an acted submit rather than forwarding it as a newline.

### [minor] `submitFromReturnKey()` reports "acted" even when `submit()` declines to act
**File:** Conduit/Views/Components/ComposerBar.swift:601
**Issue:** `submitFromReturnKey()` returns `true` unconditionally after calling `submit()`, but `submit()` has an early-return guard (`trimmed.isEmpty && attachments.isEmpty`). Today that path is unreachable because `canSubmit(action:)` only admits `.send/.steer/.interrupt`, which all require text or attachments via `hasText` — so it is not a live bug. The contract is nonetheless fragile: a future change to `submit()`'s guard (e.g., rejecting slash-only text or empty-message policy) would silently swallow Return presses with no send and no newline, exactly the accidental-swallow the design tries to avoid.
**Suggestion:** Make `submit()` return `Bool` (or have `submitFromReturnKey` re-verify content) and return that verdict from `submitFromReturnKey`, so the consumed-press decision stays in lockstep with the actual act.

### [minor] `pressesEnded`/`pressesCancelled` forward the consumed Return press to `super`
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:222
**Issue:** `pressesBegan` filters the consumed Return press out of the forwarding set, but `pressesEnded` (and `pressesCancelled`) forward the original full set, including the consumed press, and `pressesChanged` is not overridden at all. UIKit commits text on key-down, so this is today harmless, but it contradicts the "consume only the Return press" contract and relies on `super` ignoring lifecycle events for a press it never began. On a UIKit refactor where text commits on key-up (the same concern raised for the previous revision), this becomes a stray-newline-after-send risk.
**Suggestion:** Track the consumed `UIPress` identities and filter them from the sets forwarded in `pressesEnded`/`pressesCancelled`, and mirror the filter in a `pressesChanged` override, for full consistency.

### [nit] Shift-detection lag on Bluetooth keyboards still fails in the unsafe direction
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:205
**Issue:** The three-layer detection (`hardwareShiftHeld`, event/key modifier flags, `event.allPresses`) reduces but cannot eliminate the lag case where a Bluetooth keyboard delivers Return's key-down before Shift's. In that window a Shift-Return is classified as plain Return and **sends**. This is the opposite of the PR's stated "safe failure mode is newline insertion," which only holds for gate staleness, not Shift classification. It is an acknowledged tradeoff, but worth explicit documentation so the owner accepts the residual accidental-send risk.
**Suggestion:** Document the residual case in the code (it may behave differently per-keyboard), or treat a Return key-down within a very short window of an unpaired Shift begin as not-held only after confirming the keyboard reports modifiers synchronously — at minimum, state the risk in the settings copy or review notes.

### [nit] `ComposerReturnKey` relies on `.keyboardReturn` as a distinct usage
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:312
**Issue:** `isReturnKeyPress` matches `.keyboardReturnOrEnter`, `.keyboardReturn`, and `.keypadEnter`. Some SDK versions only expose `.keyboardReturnOrEnter`/`.keypadEnter`; if `.keyboardReturn` is a newer alias this compiles only against the newer SDK and silently excludes that key code on older ones. Given the CI builds are green I assume it exists on the build SDK, but the overlap (`.keyboardReturn` may alias `.keyboardReturnOrEnter`) is worth a comment or de-duplication.
**Suggestion:** Verify `.keyboardReturn` is distinct or drop it if `.keyboardReturnOrEnter` already covers the main Return key.

No retain cycles (the `onSubmitFromReturn` closure is a SwiftUI struct capture cleared in `deactivate()`), no force unwraps, no new concurrency hazards (all mutation on MainActor via UIKit press callbacks), and the settings toggle is properly labeled with an accessibility hint and clear hardware-only copy.

## Verdict
APPROVE

<a href="https://opencode.ai/s/d0AK64Fo">New%20session%20-%202026-08-28T02%3A24%3A04.973Z</a>
[opencode session](https://opencode.ai/s/d0AK64Fo)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33135738644)

PR #108, opened 2026-08-29
fix(chat): drop Hermes Recent Summary compaction records at the history boundary

kaishi00:
## Summary

Follow-up to #86. Hermes introduced a third generation of context-compaction header, persisted as ordinary transcript records with conversational roles (often `user`):

```
[Recent Summary (d0, node 342)]
```

PR #86's normalizer recognized only the two older header formats, so these records passed through normalization, inherited the persisted role, and rendered as large user bubbles on session resume. This PR extends the existing detection in place — no new filtering path, no view-layer changes, no role reclassification.

## Root cause

`MessageNormalizer.hasCompactionSummaryPrefix` (Conduit/Services/HermesClient.swift) classified only `[CONTEXT COMPACTION ...` and `[CONTEXT SUMMARY]:` prefixes plus the `_compressed_summary` flag and the merged-record delimiter. The new header format was unrecognized format drift, not a removed fix — the PR #86 filtering was verified intact before this change.

## Fix

One anchor added to the existing bounded, start-anchored, case-insensitive prefix helper:

- leading whitespace is skipped by index, then only the first 32 characters are compared (lowercased), so a multi-megabyte summary is never copied or tail-scanned
- new anchor `[recent summary (` — the opening parenthesis is the invariant discriminator; depth/node numbering is arbitrary and never hard-coded
- role-independent: filtering happens in `visibleContentRemovingCompaction` before any `ChatMessage` exists, for conversational roles; the tool-output exemption is untouched
- the merged-delimiter path re-validates its retained prefix through the same helper, so double-compaction rows whose older prefix is a Recent Summary are dropped too

## Tests

9 new focused regression tests in `MessageNormalizerTests` (plus the existing PR #86 suite, unchanged and green):

- user-role Recent Summary omitted; assistant-role also omitted (role independence)
- leading whitespace accepted (`"   \n\t[Recent Summary ...]"`)
- arbitrary depth/node numbers (`(d1, node 17)`, `(d12, node 9001)`)
- case-insensitive header
- ordinary user discussion mentioning the header is NOT hidden (start-anchor)
- header quoted later inside a normal message is NOT hidden
- multi-megabyte payload dropped during normalization, never reaching Markdown/rendering
- merged record whose retained prefix (behind the `[PRIOR CONTEXT …]` wrapper) is a gen-3 summary is omitted after the delimiter split
- bounded-head helper test extended: positive for the new format + a negative pinning that `[Recent Summary` without the parenthesis anchor stays visible

## Verification

- `ios_test` run `20260829-010710-a143`: MessageNormalizerTests, AppStateChatResumeTests, SessionPresentationCacheTests, HermesClientTests — **passed**, 0 failures, 0 uncovered
- `ios_test` run `20260829-013445-c664`: MessageNormalizerTests re-run after the last test addition — **passed**
- Pre-push multi-model review gate (`workflows/multi-model-review.js`): **APPROVE 3/3** (deepseek-v4-flash, step-3.7-flash, mimo-v2.5; consolidated on deepseek-v4-flash) — 0 BLOCKER, 0 WARNING; the one adopted SUGGESTION (merged-delimiter gen-3 test) is included in this PR

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recognition of persisted Hermes compaction summaries across multiple header formats.
  * Added support for variations in capitalization, spacing, metadata values, and large summaries.
  * Prevented incorrectly removing ordinary content or summaries mentioned later in a message.

* **Tests**
  * Expanded coverage for summary detection and omission behavior across supported formats.

### DIFF
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -1664,8 +1664,8 @@ enum MessageNormalizer {
     private static let priorContextWrapperHeader =
         "[PRIOR CONTEXT — for reference only; not a new message]"
 
-    /// Hermes emits two generations of compaction headers, and the reload
-    /// path can drop the `_compressed_summary` flag while keeping the
+    /// Hermes emits multiple generations of compaction headers, and the
+    /// reload path can drop the `_compressed_summary` flag while keeping the
     /// recognizable text. Match case-insensitively like the other Hermes
     /// bracketed notices, anchored at the start of the content so ordinary
     /// discussion of "context compaction" is never mistaken for a summary.
@@ -1680,6 +1680,7 @@ enum MessageNormalizer {
         let head = content[headStart...].prefix(32).lowercased()
         return head.hasPrefix("[context compaction")
             || head.hasPrefix("[context summary]:")
+            || head.hasPrefix("[recent summary (")
     }
 
     /// Normalizes the gateway's native clarification event. Current Hermes
--- a/ConduitTests/MessageNormalizerTests.swift
+++ b/ConduitTests/MessageNormalizerTests.swift
@@ -1170,11 +1170,163 @@ final class MessageNormalizerTests: XCTestCase {
         XCTAssertTrue(MessageNormalizer.hasCompactionSummaryPrefix(
             "\n\t  [context summary]: legacy opening\n" + hugeTail
         ))
+        XCTAssertTrue(MessageNormalizer.hasCompactionSummaryPrefix(
+            "[Recent Summary (d0, node 342)]\n" + hugeTail
+        ))
+        // The Recent Summary anchor includes the opening parenthesis, so a
+        // different bracketed notice with the same words stays visible.
+        XCTAssertFalse(MessageNormalizer.hasCompactionSummaryPrefix(
+            "[Recent Summary quoted without the node anchor]?\n" + hugeTail
+        ))
         XCTAssertFalse(MessageNormalizer.hasCompactionSummaryPrefix(
             "Can you explain how context compaction works?\n" + hugeTail
         ))
         XCTAssertFalse(MessageNormalizer.hasCompactionSummaryPrefix("   \n\t"))
         XCTAssertFalse(MessageNormalizer.hasCompactionSummaryPrefix(""))
+    }
+
+    // MARK: - Hermes "Recent Summary" headers (third compaction generation)
+
+    func testPersistedRecentSummaryHeaderIsOmittedForUserRecord() {
+        let messages = MessageNormalizer.normalizeMessages([
+            .object([
+                "id": .number(101),
+                "role": .string("user"),
+                "content": .string("[Recent Summary (d0, node 342)]\nEarlier the user asked about the deploy pipeline.")
+            ])
+        ])
+
+        XCTAssertTrue(messages.isEmpty)
+    }
+
+    func testPersistedRecentSummaryHeaderIsOmittedRegardlessOfRole() {
+        let messages = MessageNormalizer.normalizeMessages([
+            .object([
+                "id": .number(102),
+                "role": .string("assistant"),
+                "content": .string("[Recent Summary (d0, node 342)]\nSummary of the assistant's prior turns.")
+            ])
+        ])
+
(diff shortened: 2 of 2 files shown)


kaishi00:
## Summary
Extends `MessageNormalizer.hasCompactionSummaryPrefix` (HermesClient.swift) with one new start-anchored, bounded-head anchor `[recent summary (` so Hermes' third-generation "Recent Summary (d0, node N)" compaction records are dropped at the history boundary, with 9 focused regression tests.

## Findings

### [minor] Gen-3 detection relies entirely on the parenthesized-form invariant
**File:** Conduit/Services/HermesClient.swift:1683
**Issue:** This header family has now drifted three times (`[CONTEXT COMPACTION ...` → `[CONTEXT SUMMARY]:` → `[Recent Summary (...)]`), and each generation was recognized only after the fact. The new anchor matches only the exact parenthesized form; a future variant such as `[Recent Summary]: ...` or `[RECENT SUMMARY] ...` (colon form, no paren) would silently escape filtering and reappear as large user bubbles.
**Suggestion:** Not a blocker — the tighter anchor is a deliberate, well-tested trade-off (bare `[Recent Summary` without the paren would false-positive on short user messages, and the header body uses a balanced paren that could be matched) — but consider a follow-up lint/test that asserts coverage for the em-dash/colon/spacing variants, and keep the anchor list in one place (it already is) so a gen-4 format is a one-line addition.

### [minor] Genuine user message beginning with the exact header is hidden
**File:** Conduit/Services/HermesClient.swift:1633
**Issue:** Any record whose content begins, after leading whitespace, with the literal `[Recent Summary (d0, node N)]` — including a user genuinely pasting a summary as their message — is dropped entirely. The start-anchored check has no content-length or role heuristic to distinguish a real quote from bookkeeping beyond the header itself. This is the same accepted trade-off as gen-1/gen-2 anchors, so it is not a regression.
**Suggestion:** Acceptable as-is; if this ever surfaces in the field, require the header to be immediately followed by a newline plus summary body (e.g. check that a depth/node caption follows) before classifying. No action needed for this PR.

### [nit] Ambiguous case normalization on the head
**File:** Conduit/Services/HermesClient.swift:1680
**Issue:** The head is `.lowercased()`, but whitespace within the header is not normalized, so `[Recent  Summary (d0...)]` (double space) or `[Recent Summary  (d0...)]` would not match. Hermes is not expected to emit these, so this is cosmetic, not a defect.
**Suggestion:** None required; if robustness is wanted, collapsable the head's internal run before comparing, but that adds a copy of the head for a format that doesn't occur.

### [nit] Leading-whitespace skip is not bounded for pathological all-whitespace payloads
**File:** Conduit/Services/HermesClient.swift:1677-1679
**Issue:** The `while content[headStart].isWhitespace` loop scans the file's entire leading whitespace run. The PR (and its bounded-head test) correctly guarantees the comparison/lowercasing is capped at 32 chars, but a megabyte of leading whitespace before the header would still be iterated byte-by-byte. Pre-existing from PR #86, not introduced here.
**Suggestion:** Skip, or cap the skip at, say, the first 256 whitespace chars (a compaction header begins within the first few bytes in practice) and bail early. Optional.

Notes on the positive side: the merged-delimiter re-validation at HermesClient.swift:1651 now covers gen-3 prefixes inside `[PRIOR CONTEXT …]` wrappers (covered by `testMergedRecentSummaryBehindWrapperNeverReachesChatMessages`); tool-result rows remain exempt (HermesClient.swift:1628); the bounded-head 32-char window safely covers the 17-char new anchor; the tests meaningfully pin both true-positive and false-negative cases (mid-message quotes, discussion, non-parenthesized header). No retain cycles, force unwraps, concurrency, or SwiftUI concerns are touched — this is a pure string-classification change.

## Verdict
APPROVE

<a href="https://opencode.ai/s/3aDJn2FJ">New%20session%20-%202026-08-29T01%3A37%3A24.117Z</a>
[opencode session](https://opencode.ai/s/3aDJn2FJ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33226844317)

PR #109, opened 2026-08-29
fix(auth): carry password session cookies into ticket request

TechtronicsGroup:
## Summary
- parse password-login Set-Cookie headers immediately instead of racing iOS cookie persistence
- disable URLSession automatic cookie fallback so stale shared sessions cannot authenticate a new login
- let Foundation enforce cookie Domain, Path, Secure, and expiry rules before the ticket request
- keep only cookies returned by the current login

## Root cause
Hermes accepted the username/password, but Conduit immediately requested `/api/auth/ws-ticket` without the newly returned session cookie. The server therefore returned 401 `Unauthorized`, producing the logout loop and “Could not get session ticket” error.

## Tests
- RED: login to ticket regression reproduced `ticketFailed("Unauthorized")`
- RED: path-scoped cookie was rejected when leaked to `/api/auth/ws-ticket`
- RED: native auth session required automatic cookie attachment to be disabled
- GREEN: `NativeAuthClientTests` 8/8
- GREEN: CI `unit-d` lane 183/183
- GREEN: iPad simulator build
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication reliability by preserving session cookies from password login.
  * Ensured only valid, relevant cookies are sent when requesting WebSocket access tickets.
  * Prevented unrelated shared cookies from being attached to authentication requests.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2155,11 +2155,12 @@ final class AppState: ObservableObject {
         }
 
         do {
-            let ticket = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
+            let authenticatedConnection = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
                 username: credentials.username,
                 password: credentials.password
             )
-            await connect(with: HermesConnection(baseUrl: credentials.baseURL, ticket: ticket), profile: activeProfile)
+            authenticatedConnection.commitCookies()
+            await connect(with: HermesConnection(baseUrl: credentials.baseURL, ticket: authenticatedConnection.ticket), profile: activeProfile)
         } catch {
             // A rejected saved password falls back to the native login screen
             // without erasing it, allowing the user to correct the account.
@@ -3841,16 +3842,17 @@ final class AppState: ObservableObject {
                 if let credentials = KeychainHelper.loadCredentials(),
                    credentials.baseURL.trimmingCharacters(in: CharacterSet(charactersIn: "/")) == savedConnection.baseUrl.trimmingCharacters(in: CharacterSet(charactersIn: "/")) {
                     do {
-                        let ticket = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
+                        let authenticatedConnection = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
                             username: credentials.username,
                             password: credentials.password
                         )
                         guard refreshTransportContinuation() else { return }
+                        authenticatedConnection.commitCookies()
                         // URLSession and WebKit have separate cookie stores.
                         // Reload the bridge so it receives the fresh session.
                         dashboardTicketBridge?.reload()
                         await connect(
-                            with: HermesConnection(baseUrl: credentials.baseURL, ticket: ticket),
+                            with: HermesConnection(baseUrl: credentials.baseURL, ticket: authenticatedConnection.ticket),
                             profile: activeProfile,
                             syncPurpose: continuationPurpose,
                             cancelsResumeRestoration: false,
--- a/Conduit/Services/NativeAuthClient.swift
+++ b/Conduit/Services/NativeAuthClient.swift
@@ -36,13 +36,29 @@ enum AuthClientError: LocalizedError {
     }
 }
 
-/// URLSession-based Hermes dashboard authentication. Its session uses the
-/// shared cookie storage so the authenticated cookie can be copied into
-/// DashboardTicketBridge's WebKit cookie store after sign-in.
+/// A fully validated native authentication transaction. Cookies remain private
+/// to the transaction until its caller accepts the ticket and commits them.
+struct NativeAuthConnection {
+    let ticket: String
+    fileprivate let cookies: [HTTPCookie]
+
+    /// Publishes this successful transaction for DashboardTicketBridge/WebKit.
+    /// Calling this more than once is harmless; callers should commit only the
+    /// connection they are about to make active.
+    func commitCookies() {
+        NativeAuthCookiePolicy.persist(cookies)
+    }
+}
+
+/// URLSession-based Hermes dashboard authentication. Automatic URLSession
+/// cookie handling is disabled: each login transaction captures its own
+/// response cookies, scopes them to the exact ticket URL, and returns them for
+/// explicit commit only after a valid ticket has been received.
 struct NativeAuthClient {
     let baseURL: String
     let cloudflareAccess: CloudflareAccessCredentials?
     private let session: URLSession
+    private let redirectDelegate: SecureRedirectDelegate
 
     init(
         baseURL: String,
@@ -55,14 +71,19 @@ struct NativeAuthClient {
         let configuration = sessionConfiguration ?? URLSessionConfiguration.default
         configuration.httpCookieAcceptPolicy = .always
         configuration.httpCookieStorage = HTTPCookieStorage.shared
-        configuration.httpShouldSetCookies = true
-        self.session = URLSession(configuration: configuration, delegate: SecureRedirectDelegate(), delegateQueue: nil)
+        // Every Cookie header is built from this login's captured cookies.
+        // Automatic handling would let a stale shared-jar cookie authenticate
+        // when the current login returned no cookie applicable to the route.
+        configuration.httpShouldSetCookies = false
+        let redirectDelegate = SecureRedirectDelegate()
+        self.redirectDelegate = redirectDelegate
+        self.session = URLSession(configuration: configuration, delegate: redirectDelegate, delegateQueue: nil)
     }
 
     func authProviders() async throws -> [[String: Any]] {
         let request = try request(path: "/api/auth/providers")
-        let (data, response) = try await session.data(for: request)
-        guard let http = response as? HTTPURLResponse else {
+        let result = try await perform(request)
+        guard let http = result.response as? HTTPURLResponse else {
             throw AuthClientError.providerDiscoveryFailed("No response")
         }
         switch http.statusCode {
@@ -72,16 +93,16 @@ struct NativeAuthClient {
             break
         }
         guard (200...299).contains(http.statusCode) else {
-            throw AuthClientError.providerDiscoveryFailed(parseError(data) ?? "HTTP \(http.statusCode)")
+            throw AuthClientError.providerDiscoveryFailed(parseError(result.data) ?? "HTTP \(http.statusCode)")
         }
 
-        if let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any] {
+        if let json = try? JSONSerialization.jsonObject(with: result.data) as? [String: Any] {
             return json["providers"] as? [[String: Any]] ?? []
         }
         return []
     }
 
-    func login(username: String, password: String) async throws {
+    func login(username: String, password: String) async throws -> [HTTPCookie] {
         var request = try request(path: "/auth/password-login")
         request.httpMethod = "POST"
         request.setValue("application/json", forHTTPHeaderField: "Content-Type")
@@ -91,36 +112,68 @@ struct NativeAuthClient {
             "password": password
         ])
 
-        let (data, response) = try await session.data(for: request)
-        guard let http = response as? HTTPURLResponse else {
+        let result = try await perform(request)
+        guard let http = result.response as? HTTPURLResponse else {
             throw AuthClientError.loginFailed("No response")
         }
         guard (200...299).contains(http.statusCode) else {
-            throw AuthClientError.loginFailed(parseError(data) ?? "HTTP \(http.statusCode)")
+            throw AuthClientError.loginFailed(parseError(result.data) ?? "HTTP \(http.statusCode)")
         }
+        guard http.url != nil else {
+            throw AuthClientError.loginFailed("Response URL missing")
+        }
+
+        // Redirect responses may set the session before the final JSON landing.
+        // The delegate captures those headers per URLSession task, and the final
+        // response is appended last so a later rotation wins by cookie identity.
+        return NativeAuthCookiePolicy.merged(
+            result.redirectCookies + NativeAuthCookiePolicy.acceptedCookies(from: http)
+        )
     }
 
-    func mintWsTicket() async throws -> String {
+    func mintWsTicket(authenticatedCookies: [HTTPCookie]) async throws -> NativeAuthConnection {
         var request = try request(path: "/api/auth/ws-ticket")
         request.httpMethod = "POST"
-        let (data, response) = try await session.data(for: request)
-        guard let http = response as? HTTPURLResponse else {
+        guard let ticketURL = request.url else {
+            throw AuthClientError.invalidURL
+        }
+
+        // Scope only this login's cookies. No shared-jar read occurs here, so a
+        // stale or concurrent session can never replace an empty/missing match.
+        for (name, value) in NativeAuthCookiePolicy.headerFields(
+            for: authenticatedCookies,
+            url: ticketURL
+        ) {
+            request.setValue(value, forHTTPHeaderField: name)
+        }
+
+        let result = try await perform(request)
+        guard let http = result.response as? HTTPURLResponse else {
             throw AuthClientError.ticketFailed("No response")
         }
         guard (200...299).contains(http.statusCode) else {
-            throw AuthClientError.ticketFailed(parseError(data) ?? "HTTP \(http.statusCode)")
+            throw AuthClientError.ticketFailed(parseError(result.data) ?? "HTTP \(http.statusCode)")
         }
-        guard let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any],
+        guard let json = try? JSONSerialization.jsonObject(with: result.data) as? [String: Any],
               let ticket = json["ticket"] as? String,
               !ticket.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
             throw AuthClientError.ticketFailed("No ticket in response")
         }
-        return ticket
+
+        // A deployment may rotate its session while minting the ticket. Keep
+        // that rotation transaction-local; the caller commits it only when this
+        // ticket is accepted as the active connection.
+        let committedCookies = NativeAuthCookiePolicy.merged(
+            authenticatedCookies
+                + result.redirectCookies
+                + NativeAuthCookiePolicy.acceptedCookies(from: http)
+        )
+        return NativeAuthConnection(ticket: ticket, cookies: committedCookies)
     }
 
-    func connect(username: String, password: String) async throws -> String {
-        try await login(username: username, password: password)
-        return try await mintWsTicket()
+    func connect(username: String, password: String) async throws -> NativeAuthConnection {
+        let authenticatedCookies = try await login(username: username, password: password)
+        return try await mintWsTicket(authenticatedCookies: authenticatedCookies)
     }
 
     private func request(path: String) throws -> URLRequest {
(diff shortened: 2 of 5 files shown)


kaishi00:
## Summary
Fixes the logout loop where Conduit's password-login session cookie was not attached to the subsequent `/api/auth/ws-ticket` request, by parsing Set-Cookie headers immediately, inserting them into the shared jar, disabling URLSession automatic cookie handling, and attaching only the current login's cookies to the ticket request.

## Findings

### [severity: major] Fix's production path is untested — custom URLProtocol bypasses URLSession cookie machinery
**File:** ConduitTests/NativeAuthClientTests.swift:152 (`NativeAuthURLProtocol`), Conduit/Services/NativeAuthClient.swift:61
**Issue:** Every test routes requests through a custom `URLProtocol`, which short-circuits URLSession's cookie layer entirely — neither Set-Cookie storage nor request-side cookie attachment happen through the real HTTP stack. So the tests validate only the *manual* header logic, but cannot prove that `httpShouldSetCookies = false` actually suppresses URLSession from auto-attaching applicable stored cookies from the shared jar on the production transport. If that flag turns out not to suppress request-side attachment on a given iOS/macOS version (its documented contract is about whether the session "sets cookies", and historical NSURLRequest behavior around `httpShouldSetCookies` is subtle), the stale shared-jar cookie still rides along with the manually attached fresh one — the exact bug this PR claims to eliminate.
**Suggestion:** Add an integration test that runs against the real stack without the custom protocol, or wire cookie attachment entirely through code (attach only the manually computed header and do not depend on URLSession's flag to suppress jar cookies). At minimum, assert the intended final header (`hermes_session_at=test-access-token`) with no duplicates/stale values on a path that exercises the real transport.

### [severity: major] Set-Cookie on intermediate redirect responses is silently dropped, and empty results are swallowed
**File:** Conduit/Services/NativeAuthClient.swift:104
**Issue:** `guard let responseURL = http.url else { return [] }` reads only the *final* response of the login request. If `/auth/password-login` follows a redirect (which many auth flows use to set the session cookie), the Set-Cookie from the intermediate 3xx is never parsed, `login` returns `[]`, the ticket request gets no Cookie header, and the user sees the same `401`/retry loop the PR is meant to fix. Returning `[]` silently also obscures a genuine no-URL failure behind a generic downstream `ticketFailed("Unauthorized")`.
**Suggestion:** Throw `AuthClientError.loginFailed` when `http.url` is nil, and capture cookies across the redirect chain (e.g., accumulate parse of every redirect response via the task delegate, or re-read the jar scoped to the login URL after the request completes).

### [severity: minor] Cookie intersection round-trips through the jar and can silently miss due to domain normalization
**File:** Conduit/Services/NativeAuthClient.swift:129-136
**Issue:** The filter compares `captured.domain == candidate.domain` between a freshly parsed `HTTPCookie` (from `HTTPCookie.cookies(withResponseHeaderFields:for:)`) and one re-read from the shared jar. Foundation normalizes cookie domains (leading-dot vs host-only) inconsistently across OS versions; any divergence yields an empty intersection → no cookies attached → `401`, in a hard-to-reproduce way. The jar round-trip is redundant given login already holds the parsed cookies.
**Suggestion:** Build the header directly from the parsed `authenticatedCookies`, filtered against the ticket URL yourself (name/path/domain/secure/expiry), and skip the jar re-read. This also removes the O(n·m) `contains` scan.

### [severity: minor] Tests mutate global shared state and break under parallel test execution
**File:** ConduitTests/NativeAuthClientTests.swift:235-258, 6-9
**Issue:** The suite reads/writes `HTTPCookieStorage.shared` and static `requestRecords`/`responseRecords`. If parallel test execution is enabled (Xcode test plans), the two fixture hosts (200/201) share one jar keyed only by domain, and `setUp`'s `reset()` cannot race-proof overlapping methods — leading to flaky cross-test contamination (e.g., `.200` test failing because `.201` wiped cookies mid-run).
**Suggestion:** Enforce serial execution (test plan ordering) or isolate state — e.g., key cookie setup/reset on unique test names and remove it in both `setUp` and `tearDown`.

### [severity: minor] `XCTFail` falls through instead of returning
**File:** ConduitTests/NativeAuthClientTests.swift:50-51
**Issue:** In `testTicketRequestDoesNotFallBackToUnrelatedSharedCookie`, after `XCTFail` the test continues to the trailing `XCTAssertNil`, so a false "success" path still reaches the assertion (which happens to pass), weakening the failure signal.
**Suggestion:** Use `return XCTFail(...)`.

### [severity: nit] Header parsing reduces duplicate header entries to a single value
**File:** Conduit/Services/NativeAuthClient.swift:105-108
**Issue:** `reduce` collapses any repeated header names into one `String`; and server response splitting cookies by comma depends on values having no embedded commas. Session tokens rarely contain commas, so this is low risk, but a value with a comma would split the cookie parse.
**Suggestion:** Prefer no manual reduce — use the response's cookie storage or iterate `allHeaderFields` preserving each entry.

### [severity: nit] Disabling `httpShouldSetCookies` also disables storing cookies from non-login responses on this session
**File:** Conduit/Services/NativeAuthClient.swift:61
**Issue:** The flag affects the whole session, including the `/api/auth/ws-ticket` response and `authProviders`. If the server rotates the session cookie *on the ticket call*, the rotated cookie will no longer be persisted for downstream WebKit/websocket handshakes as the previous design relied on it to be.
**Suggestion:** Confirm the server does not rotate cookies in the ticket response; if it does, parse and store the ticket response's Set-Cookie explicitly as well.

## Verdict
REQUEST_CHANGES

The manual cookie-carrying logic and its unit-test intent are sound, but the core claim — that stale shared-jar cookies can no longer authenticate a new login — is only validated through a custom `URLProtocol` that bypasses the very URLSession cookie machinery at issue, and the redirect/empty-cookie edge case can silently reintroduce the original bug.

<a href="https://opencode.ai/s/YcoOTbC3">New%20session%20-%202026-08-29T02%3A05%3A53.412Z</a>
[opencode session](https://opencode.ai/s/YcoOTbC3)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33228033730)

TechtronicsGroup:
Addressed the review findings with a stronger transactional design:

- Added real HTTP integration tests using `NWListener` plus production `URLSession`—no custom `URLProtocol`—with a pre-seeded stale applicable shared cookie. The server asserts the emitted ticket request contains exactly the current login cookie.
- Capture `Set-Cookie` from every same-origin redirect response per URLSession task and explicitly attach accepted task-local cookies to the redirect landing. The landing now rejects the request unless it received that redirect cookie.
- Removed the shared-cookie-jar round trip from ticket header construction. Domain, path, Secure, and expiry applicability are evaluated directly against the immutable cookies captured by that login.
- Reject parsed cookies unless their canonical domain exactly matches the response host. This prevents a dashboard response from poisoning a sibling, parent, public-suffix, or unrelated origin in the shared store.
- Canonicalize leading-dot domains in cookie identity, with a real redirect-to-final rotation test proving only the final cookie reaches the ticket endpoint.
- Authentication now returns an uncommitted `NativeAuthConnection`. Login and ticket-response cookies are published only when the caller accepts the valid ticket and calls `commitCookies()`. Failed, cancelled, superseded, or invalid-ticket transactions publish nothing.
- Automatic reconnect checks its generation before committing cookies, so a superseded recovery cannot replace the active browser session.
- Added a deterministically overlapping concurrent-login test. A barrier holds both login responses until both requests are in flight; alpha and beta are asserted individually, not as an unordered set. The test also proves no cookie is shared before commit and that the caller's last explicit commit owns the shared/WebKit session.
- Mutation-tested that concurrency assertion by deliberately swapping alpha/beta ticket responses; the test turned red as required.
- Added real duplicate `Set-Cookie` coverage, including path-scoped, Secure-on-HTTP, expired, and foreign-domain cookies.
- Added invalid-ticket-body coverage proving neither login nor ticket rotation cookies are published.
- Added request/resource timeouts and deterministic cancellation of active loopback connections during teardown.
- Added cleanup in both `setUp` and `tearDown`, fixed the `XCTFail` fallthrough, and assigned the integration class to `unit-d`.

Local verification on the final staged candidate:

- Expanded native-auth unit and real-HTTP integration suites: passed
- Deterministic concurrent-login test repeated successfully after its mutation RED check
- Test-shard validation: passed
- iPhone simulator compilation/test execution: passed
- `git diff --check`: passed

The prior full `unit-d` lane passed 186/186 before the final transactional hardening; GitHub CI will now rerun the lane against this exact pushed commit, and I will treat that live result as authoritative.

kaishi00:
## Summary

This PR fixes the native-auth logout loop by capturing password-login `Set-Cookie` headers per URLSession task, attaching only the current login's cookies (scoped by domain/path/Secure/expiry) to the `/api/auth/ws-ticket` request, disabling URLSession automatic cookie handling so stale shared-jar cookies can't authenticate, and deferring all cookie publication until the caller accepts the ticket (`commitCookies()`). It adds a custom-`URLProtocol` unit suite plus a real-transport (`NWListener`) integration suite.

## Findings

### [severity: minor] `allHeaderFields` collapsed via `String(describing:)` is fragile for duplicate `Set-Cookie`
**File:** Conduit/Services/NativeAuthClient.swift:231-238
**Issue:** The `reduce` collapses every header name to its last `String(describing:)` value. On some Foundation versions `HTTPURLResponse.allHeaderFields` may surface duplicate header names as an array (or combine them in an OS-version-dependent canonical form); if the value is an array, `String(describing:)` yields `["a=1; Path=/", "a=2; Path=/"]`, which a cookie parser cannot interpret and silently drops cookies. The whole fix depends on this parse (both the login session cookie and the ticket-response rotation cookie), so a silently empty parse would reintroduce the original 401 loop on an OS where the shape differs. Tests pass on the current OS but do not pin that this canonical-form assumption holds everywhere.
**Suggestion:** Build the fields preserving entries — e.g. append `Set-Cookie` values when the key already exists (join with `\n` or pass an array), or read the raw `Set-Cookie` header directly — rather than relying on the collapsed dictionary + `String(describing:)`.

### [severity: minor] Same-origin redirect handling for `authProviders()` silently changed semantics
**File:** Conduit/Services/NativeAuthClient.swift:89-94, SecureRedirectDelegate:365-401
**Issue:** Previously `authProviders()` returned `[]` for *any* 301/302/303/307/308. Now `SecureRedirectDelegate` follows same-origin redirects and only returns `[]` when the flow ends on a redirect (cross-origin cancel) or another 3xx. A dashboard that serves `/api/auth/providers` behind a same-origin redirect whose landing returns 200 + `providers` JSON will now be treated as native-password-capable where it previously fell back to the WebView flow. This is a behavior change not called out in the PR.
**Suggestion:** Confirm this is intentional; if the redirect-to-landing provider path should stay a WebView fallback, short-circuit in `authProviders` (e.g., have the delegate report when the logged-in request was redirected) rather than re-executing the landing.

### [severity: minor] Tests mutate `HTTPCookieStorage.shared` and share it across test classes
**File:** ConduitTests/NativeAuthClientIntegrationTests.swift:6-15, 298-306; ConduitTests/NativeAuthClientTests.swift:240-251
**Issue:** Both suites read/write the global `HTTPCookieStorage.shared` (fixture IPs, `127.0.0.1`, `localhost`, `conduit-auth-poison.invalid`). `setUp`/`tearDown` reset within a class, but if the runner ever parallelizes the two classes, one class's tearDown can wipe a cookie another class's in-flight test just seeded, causing cross-test flakiness (e.g., a concurrent-login test losing a committed session mid-assertion).
**Suggestion:** Keep the reset in both `setUp` and `tearDown` (already done) and additionally isolate by unique cookie names per case if parallelization is ever enabled; documenting the serial-execution requirement is acceptable.

### [severity: nit] Dead `guard let ticketURL` in `mintWsTicket`
**File:** Conduit/Services/NativeAuthClient.swift:137-139
**Issue:** `request(path:)` already throws `AuthClientError.invalidURL` when the URL can't be built and returns a `URLRequest` with a non-nil `url`, so `guard let ticketURL = request.url else { throw .invalidURL }` is unreachable dead code.
**Suggestion:** Remove the redundant guard to avoid implying the failure is possible.

### [severity: nit] Cookie publication depends on every caller remembering `commitCookies()`
**File:** Conduit/Services/NativeAuthClient.swift:48-50; AppState.swift:2162, 3850; LoginView.swift:253
**Issue:** `NativeAuthConnection` keeps cookies `fileprivate` and relies on each of the three call sites to invoke `commitCookies()`. A future caller that forgets it will silently fail to publish the session to the WebKit bridge (the exact symptom this PR fixes), with no compile-time enforcement. It's documented, but the coupling is easy to miss.
**Suggestion:** Consider making the connection's cookies be committed from a single funnel (e.g., `AppState` routes all native connections through one helper), or add a DEBUG assertion/log when a `NativeAuthConnection` is deallocated without committing.

## Verdict

COMMENT

The core bug is convincingly fixed: cookies are captured per task, scoped exactly to the ticket route, and — crucially — the "stale shared-jar cookie can't authenticate" claim, which the custom `URLProtocol` unit tests cannot prove, is now validated by real-transport integration tests that seed a stale applicable cookie and assert the emitted header. The transaction model with deferred `commitCookies()` is sound and the concurrency test is solid. Remaining items are robustness and semantics (header-collapse fragility, `authProviders` redirect behavior change) rather than blockers.

<a href="https://opencode.ai/s/TbJ0ibEz">New%20session%20-%202026-08-29T02%3A37%3A43.121Z</a>
[opencode session](https://opencode.ai/s/TbJ0ibEz)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33229381025)

kaishi00:
## Summary

Fixes the native-auth logout loop by having `NativeAuthClient` capture password-login `Set-Cookie` headers per URLSession task (including same-origin redirects), attach only the current login's cookies to `/api/auth/ws-ticket`, disable URLSession automatic cookie handling, and defer all publication to an explicit `NativeAuthConnection.commitCookies()` after the ticket is validated.

## Findings

### [severity: minor] Load-bearing Set-Cookie parse depends on an OS-canonicalized, collapsed header dictionary
**File:** Conduit/Services/NativeAuthClient.swift:231-238
**Issue:** `allHeaderFields.reduce` collapses each header name to its last `String(describing:)` value, and then hands that to `HTTPCookie.cookies(withResponseHeaderFields:)`. On Foundation versions that surface duplicate `Set-Cookie` values as an array, `String(describing:)` yields something like `["a=1; Path=/", "a=2; Path=/"]`, which the cookie parser cannot interpret and silently drops. Because this single parse feeds both the login session cookie and the ticket-response rotation cookie, a silently empty parse would recreate the exact 401 loop this PR fixes. The real-transport tests pin the current OS's canonical form, but not others.
**Suggestion:** Preserve each `Set-Cookie` entry (keep first occurrence and append duplicates, or iterate the raw header field values directly) instead of relying on the collapsed dictionary plus `String(describing:)`.

### [severity: minor] `deleteCookie` is called on cookies never added to the storage
**File:** Conduit/Services/NativeAuthClient.swift:258-266
**Issue:** `persist` calls `HTTPCookieStorage.shared.deleteCookie(cookie)` for an expired cookie that was freshly parsed and never `setCookie`-ed into the jar. On some Foundation versions, deleting a non-persisted cookie can raise an exception rather than being a no-op; the integration suite exercises this path via the `expired=...` cookie but only on one OS.
**Suggestion:** Skip expired cookies entirely (`expired` → `return`) instead of deleting, or guard with a membership check first.

### [severity: minor] Redeemability of ticket URL guard is dead code
**File:** Conduit/Services/NativeAuthClient.swift:137-139
**Issue:** `request(path:)` already returns a `URLRequest` with a non-nil `url` (it throws `invalidURL` otherwise), so `guard let ticketURL = request.url else { throw .invalidURL }` is unreachable.
**Suggestion:** Remove the redundant guard and keep `request.url` for the cookie-scoping lookup.

### [severity: minor] Cancellation surfaces `URLError(.cancelled)` instead of `CancellationError`
**File:** Conduit/Services/NativeAuthClient.swift:189-214
**Issue:** `perform` resumes the continuation with the URLError delivered by the cancelled `dataTask` rather than mapping cancellation to `CancellationError`. Callers distinguish cancellation specifically (e.g. `LoginView.swift:255` catches `is CancellationError { return }`), so a user-initiated cancel of the login now renders as an error message instead of a silent return.
**Suggestion:** In the completion handler, map `error` to `CancellationError` when the task was cancelled (or check `Task.isCancelled`/the task's `cancellationError`).

### [severity: minor] "Path-scoped cookie rejected" regression tests do not exercise the custom path-boundary logic
**File:** ConduitTests/NativeAuthClientTests.swift:294,314; ConduitTests/NativeAuthClientIntegrationTests.swift:25
**Issue:** The fixtures use `Path=/auth`, which never prefix-matches `/api/auth/ws-ticket` under any matcher (naive or RFC). The hand-rolled boundary logic in `applicableCookies` (NativeAuthClient.swift:300-309) — the one place a naive `hasPrefix` comparison would wrongly leak a `Path=/api` cookie onto `/api/auth/ws-ticket` — is therefore never pinned by any test, despite the PR describing it as a RED-covered scenario.
**Suggestion:** Add a fixture whose cookie `Path=/api` (or `/api/auth/ws-ticket-extra`) is the boundary case, asserting it must not be attached.

### [severity: nit] Global shared state is mutated by both test suites
**File:** ConduitTests/NativeAuthClientTests.swift:240-263; ConduitTests/NativeAuthClientIntegrationTests.swift:278-306
**Issue:** Both suites read/write `HTTPCookieStorage.shared` and static fixture records, cleaned in `setUp`/`tearDown`. Fixture domains are disjoint today, but any future parallelization (test plans) or a shared-domain fixture could produce cross-test contamination; the suites currently depend on serial execution that is not enforced.
**Suggestion:** Keep the bidirectional cleanup (already present) and document the serial-execution requirement, or isolate per-case cookie names.

### [severity: nit] Cookie publication is a manual, easy-to-miss contract
**File:** Conduit/Services/NativeAuthClient.swift:41-51
**Issue:** `NativeAuthConnection` keeps cookies `fileprivate` and relies on all three call sites remembering `commitCookies()`. A future call site that forgets it silently produces the exact "logged in but unauthenticated" symptom this PR fixes, with no compile-time enforcement.
**Suggestion:** Route all commits through a single funnel (e.g., a helper in `AppState`), or add a DEBUG assertion on `NativeAuthConnection` deinit if it was never committed.

## Verdict

COMMENT

The two major findings from the earlier review are addressed: the "stale shared-jar cookie cannot authenticate" claim is now proven on the real transport (`NativeAuthURLProtocol` bypasses URLSession's cookie machinery), and empty/missing URLs now throw instead of returning `[]`. The transactional `commitCookies()` model, redirect-chain capture, and deterministic concurrent-login test are sound. Remaining items are robustness and coverage gaps (header-collapse fragility, delete-of-unstored-cookie, dead guard, cancellation error taxonomy) rather than blockers.

<a href="https://opencode.ai/s/KeJqJO1Y">New%20session%20-%202026-08-29T02%3A45%3A08.642Z</a>
[opencode session](https://opencode.ai/s/KeJqJO1Y)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33229672484)

kaishi00:
One edge case I’d like to see covered before merge: **base URLs with a path prefix**.

`SecureRedirectDelegate` currently special-cases password-login redirects using:

```swift
task.originalRequest?.url?.path == "/auth/password-login"
```

That works for a root-mounted gateway, but if Conduit is configured against something like:

```text
https://example.com/hermes
```

the actual login path becomes:

```text
/hermes/auth/password-login
```

and this check will fail. The redirect itself will still proceed, but the task-local cookies captured from the login response won’t be explicitly attached to the redirect landing, which could recreate the missing-session-cookie failure for path-prefixed deployments.

I’d suggest either comparing against the login URL derived from the normalized `baseURL`, rather than a hard-coded absolute path, or adding a real-transport integration case using a prefixed route such as:

```text
/hermes/auth/password-login
→ /hermes/auth/password-complete
→ /hermes/api/auth/ws-ticket
```

Otherwise, the updated transactional cookie design looks solid to me; I don’t consider the remaining review notes blockers.

kaishi00:
Superseded by #113, which has been merged to main (merge commit c4e583c).

Thank you @TechtronicsGroup / Christopher Borell — the transactional native-auth cookie architecture in this PR was the right design, and it was **adopted, not rewritten**: the head commit (ca2fd22) was cherry-picked onto a first-party branch from current main as commit 65e803d with author attribution preserved, so your work carries into the merged history.

#113 finished the remaining work on top: the path-prefixed base-URL blocker (SecureRedirectDelegate now compares against the exact endpoint derived from the normalized base URL instead of the hard-coded /auth/password-login path), expired-cookie identity-scoped deletion at commit, cancellation mapped to CancellationError, duplicate Set-Cookie parsing hardening, dead-guard removal, corrected cookie path-boundary tests, and extended real-transport coverage — all preserving this PR's transactional commitCookies() model. Full write-up and verification results are in the #113 description.

Closing as superseded.

PR #110, opened 2026-08-29
fix(resume): bound WS message size and prefer compact session.resume (#106)

### LINKED ISSUE #106: [BUG] Frequent "Hermes connection closed." + reconnect loop on large sessions — WS close 1009 (message too big) from iOS URLSessionWebSocketTask
## Bug Description

On iOS, Conduit frequently shows **"Hermes connection closed."** while the network is fine — and the New Chat icon keeps blinking, indicating the gateway connection is flapping. The app enters a **reconnect loop**: it connects, exchanges 1-2 messages, gets disconnected (WS close code **1009 / Message Too Big**), reconnects, and repeats every 2-7 seconds.

Notably, **only some sessions** trigger this — small sessions work fine; large sessions (many messages) reliably reproduce it.

### 1. Server-side WS log (gui.log)

```
2026-08-28 12:01:20,720 INFO tui_gateway.ws: ws accepted peer=61.171.230.205:0
2026-08-28 12:01:21,499 INFO tui_gateway.ws: ws closed peer=61.171.230.205:0 reason=client_disconnect(code=1009,reason=) messages=1 ...
2026-08-28 12:01:22,649 INFO tui_gateway.ws: ws accepted peer=61.171.230.205:0
2026-08-28 12:01:26,278 INFO tui_gateway.ws: ws closed peer=61.171.230.205:0 reason=client_disconnect(code=1009,reason=) messages=2 ...
```

The server never sends close code 1009 (it configures `ws_max_size=384MB`); the 1009 close frame **comes from the iOS client** (URLSessionWebSocketTask).

### 2. Reverse-proxy access log (nginx, Conduit UA)

```
12:01:19 GET /api/ws?ticket=...  101 407744
12:01:21 GET /api/ws?ticket=...  101 131853
12:01:26 GET /api/ws?ticket=...  101 614278
12:01:37 GET /api/ws?ticket=...  101 169923
12:01:45 GET /api/ws?ticket=...  101 290694
12:01:52 GET /api/ws?ticket=...  101 290694   <- identical 290694-byte response re-fetched
12:01:56 GET /api/ws?ticket=...  101 290694   <- again
12:02:09 GET /api/ws?ticket=...  101 290694   <- again
```

Every 2-7 seconds a new WS connection transfers 130-680 KB then dies. The repeated identical 290,694-byte frame is a large fixed response (e.g. `session.resume` returning the full transcript of a big session, or a large `session.list`).

### 3. Root cause in client code

`Conduit/Services/HermesClient.swift` -> `openSession(_:)` calls `session.resume` with:

```swift
let result = try await rpc("session.resume", params: [
    "session_id": sessionId,
    "cols": 96,
    "source": "desktop"
])
```

- **No `maximumMessageSize` is set** on the `URLSessionWebSocketTask`. iOS's default receive buffer limit causes `receive()` to fail once a single message exceeds it, and URLSessionWebSocketTask then closes the connection with **code 1009**.
- **No `omit_messages` / `defer_history` param** is sent, so the server returns the **full transcript** of the session in one WS frame. Large sessions -> single frame > iOS limit -> 1009 -> reconnect -> repeat.
- The server explicitly supports the correct pattern (see Hermes `tui_gateway/methods_session.py`): *"Desktop hydrates persisted transcripts through the authenticated REST route in parallel. Suppress the duplicate WebSocket transcript only when the caller explicitly requests it"* - i.e. Desktop passes `omit_messages: true` and pulls history over REST instead.

## Steps to Reproduce

1. Have a session with a large number of messages (e.g. 300+ messages, a few hundred KB of transcript).
2. Open that session in Conduit on iOS.
3. The app connects -> sends the first RPC (`session.resume`) -> the large response exceeds the WS message limit -> connection closes with 1009 -> "Hermes connection closed." -> auto-reconnect loop (icon blinking).
4. Small sessions do not reproduce.

## Expected Behavior

- Large-session `session.resume` responses should not kill the connection.
- Either set `maximumMessageSize` on the WebSocket task (e.g. to a few MB), or send `omit_messages: true` / `defer_history: true` on `session.resume` and hydrate the transcript over REST (like the desktop client), or paginate large transcripts.
- On a 1009 / oversized-frame close, the app should not enter a tight reconnect loop; it should fall back gracefully.

## Environment

- Hermes Conduit iOS (App Store version, v140 build per UA string)
- Hermes Agent v0.20.5 (dashboard 9118/9119 behind nginx)
- iPhone / iOS (WS via `URLSessionWebSocketTask`)

## Related

- #35 "Failed to restore this conversation: session not found" (resume failure handling)
- #52 "App restores an older conversation after backgrounding" (session restore logic)
- #99 Chat caching proposal (large transcript reload on every return)

kaishi00:
Fixes #106.

## Root cause

Two compounding problems on large sessions:

1. `URLSessionWebSocketTask` closed the connection with **1009 / message too big** as soon as a single incoming frame exceeded its platform-default (~1 MiB) `maximumMessageSize` — `URLSessionWebSocketTransport.makeSocket` never raised it.
2. `session.resume` shipped the **entire persisted transcript** in one frame, so opening a deep conversation reliably exceeded that limit → disconnect → tight reconnect loop (the blinking New Chat icon from the issue's gateway logs).

Upstream contract verified in NousResearch/hermes-agent: the gateway reads `omit_messages` via loose `params.get(...)` (`"messages": [] if omit_messages else …`), and Hermes Desktop sends `cols: 96, omit_messages: true`, hydrating the transcript over the authenticated REST route with the live projection rendered on top.

## Changes

- **Transport bound (defense-in-depth, kept unconditionally):** the production socket gets an explicit bounded `maximumMessageSize` of **4 MiB**. Not unbounded; transcript-sized payloads stay off the socket entirely via compact resume.
- **Compact resume:** `HermesClient.openSession` sends `omit_messages: true`. `AppState.reconcile` selects compact only when a **ready** history source exists (dashboard bridge `isReady`, or the test seam); a cold bridge or no bridge performs exactly one legacy resume, identical to pre-change behavior.
- **Hydration through the existing path:** the durable transcript still comes from the dashboard `/api/sessions/{id}/messages` route, parsed by the same production `MessageNormalizer` path; merge semantics (presentation cache, pending decision restoration, inflight projection via the streaming bubble, buffered-event dedup, YOLO/approvals/aliases, profile-switch ownership guards) are untouched.
- **Narrow compatibility fallback:** when the history source is *structurally unavailable* (404/410/501 — e.g. a gateway predating the endpoint), *not ready yet* (`notReady`), or resolves rows to a **foreign session identity**, Conduit performs exactly **one** legacy full-transcript re-resume (`openSessionLegacy`, no `omit_messages`). Older gateways also ignore the unknown flag outright (loose `params` reading upstream), a second safety net.
- **Failures surface:** auth (`signInRequired`), server errors, network failures, and malformed payloads **fail the resume loudly** instead of silently degrading — no broad fallback concealment.
- `DashboardTicketBridgeError` gained a typed `.http(status:detail:)` case so the unsupported-endpoint condition is recognized without string matching; `isReady` became `private(set)`.

## Compatibility notes

- Compact resume is **readiness-gated** rather than unconditional: gateways without the history endpoint (and cold bridges) take the legacy single-resume path, so their behavior is byte-identical to today's.
- Disclosed assumption: old gateways are assumed to ignore unknown `session.resume` params — verified against upstream handler sources (`params: dict` + `.get`). A hypothetical strictly-validating old gateway would fail loudly (JSON-RPC error surfaced), not loop.

## Known window (reviewed, accepted)

The legacy fallback required for gateways without a history source carries the transcript in the RPC response, whose size is bounded only by the 4 MiB socket limit — same as pre-change builds. This is the issue-sanctioned "one legacy resume" and cannot be avoided without violating the fallback requirement; it is documented in code and covered by a deterministic >4 MiB fallback regression test.

## Tests

New `CompactResumeTranscriptTests` (8): compact request + real-normalizer hydration, running-turn inflight preservation without duplication, deterministic large-session (>1 MiB old default) compact resume, unavailable→legacy fallback (asserts exactly one legacy re-resume), **>4 MiB legacy fallback hydration**, auth-failure and malformed-payload surfacing without fallback, and unavailable-source classification. `HermesClientTests` +3: 4 MiB transport limit on the real task, compact request params, legacy request params. `ChatResumeLifecycleOperations.openSession` seam gained the flavor flag (56 closures updated).

## Validation

- Focused suites + all session-resume/reconciliation suites: **pass**
- **Full Conduit suite via `ios_test`: 69/69 units, 9/9 batches, no hangs, no uncovered suites**
- Release-configuration simulator build: **BUILD SUCCEEDED**
- Pre-push multi-model review gate: **APPROVE WITH NITS, 0 blockers** (2 findings adjudicated above, nits applied)

🤖 Generated with Claude Code

---

## Follow-up (7433f07): review-findings hardening

Closes the remaining reliability gaps from the first review round without changing the architecture.

**Cold bridge readiness.** A bridge that exists but is not ready yet no longer selects the legacy full-transcript path. Compact resume always begins whenever a history source exists; `requestJSON`'s existing bounded readiness poll (30 × 100 ms) waits *inside the concurrent transcript fetch*, so the WebSocket resume is never delayed by the wait. A bridge that becomes usable within the window hydrates normally (no legacy resume); one that does not degrades to exactly one legacy resume. A >4 MiB session on a modern gateway therefore no longer immediately goes out as a full-transcript request that cannot fit the socket bound.

**History failure routing.** Exactly one legacy resume for: 404/410 (structural endpoint gap), 408/429/5xx (transient HTTP), status 0 (WebKit-level network/abort failures, including the 24 MiB response cap), and bridge readiness/deadline timeouts (`.notReady`). Surfaced, never concealed: 401/403 (`.signInRequired`), malformed successful payloads, and any unclassified error. Classification is centralized in `AppState.historySourceIsUnavailable(_:)` and applies identically to the bridge fetch and the test seam.

**Fallback is exactly once.** Structurally: one compact `openChatResumeSession` call, then a straight-line `switch` with a single legacy call site and a throwing failure path that never re-enters; every fallback test asserts `[true, false]` (and no test can observe a third call).

**Legacy resume timeout.** Dedicated `legacyResumeTimeout = 60 s` via `resumeTimeout(omitMessages:)` (compact keeps the ordinary 30 s budget), pinned by `testLegacyResumeUsesDedicatedTimeoutBudget`.

**>4 MiB test corrected.** The former `testLegacyFallbackHydratesSessionExceedingTransportBound` injected messages through the AppState seam — bypassing HermesClient, JSON-RPC, and URLSession — while implying transport coverage. Renamed to `testLegacyFallbackMergeHandlesLargeTranscriptAtAppStateLevel` with explicit scope documentation: it now asserts only AppState merge behavior, and states that a legacy WebSocket response > 4 MiB cannot traverse the socket (the supported large-transcript path is compact resume + REST hydration). The real-transport 4 MiB test and the compact large-session regression remain.

**Tests added/updated:** transient-failure fallback loop (429/500/503/0/`notReady`), structural-gap loop (404/410/501), delayed-hydration success avoids legacy resume, cold-bridge compact-first via the real `connect()` flow, timeout budget pin, expanded classification matrix (incl. 408), and the updated cold-bridge expectation in the existing viewport-cancellation connect test.

**Validation:** focused suites pass; full Conduit suite **69/69 units, 9/9 batches** via `ios_test`; `plan-tests.py validate` OK; pre-push multi-model review gate on the full two-commit diff: **APPROVE WITH NITS, 0 blockers (3/3 reviewers)** — warnings addressed (timeout test pin, 408 classification), nits applied (isReady widening reverted, doc corrections).

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2664,14 +2664,17 @@ final class AppState: ObservableObject {
             // response; the HTTP endpoint reads the timestamped rows from
             // state.db and hydrates the conversation instead.
             let bridge = dashboardTicketBridge
-            // Request the compact projection only when a history source can
-            // hydrate the transcript without a second round trip: a dashboard
-            // bridge that is ready to serve requests, or a test seam standing
-            // in for one. A cold bridge resumes with the transcript carried in
-            // the RPC response, exactly as pre-compact builds did — and so do
-            // the legacy fallbacks below (gateway without the history route),
-            // whose response size is bounded only by the socket limit.
-            let compactResume = chatResumeLifecycleOperations.persistedTranscript != nil || bridge?.isReady == true
+            // Prefer the compact projection whenever a history source exists
+            // to hydrate the transcript — the dashboard bridge, or a test seam
+            // standing in for one. A cold bridge does not change the flavor:
+            // requestJSON's bounded readiness poll (30 × 100 ms) waits for the
+            // page inside the concurrent transcript fetch, so the compact
+            // resume is never delayed by that wait, and a bridge still not
+            // usable afterwards degrades to the single legacy resume. The
+            // legacy resume — cold bridge, gateway without the history route —
+            // carries the transcript in the RPC response, whose size is
+            // bounded by the socket limit (the pre-compact behavior).
+            let compactResume = chatResumeLifecycleOperations.persistedTranscript != nil || bridge != nil
             async let resumedSession = openChatResumeSession(
                 sessionId,
                 using: client,
@@ -2680,8 +2683,7 @@ final class AppState: ObservableObject {
             async let transcriptOutcome = persistedTranscriptOutcome(
                 sessionId: sessionId,
                 profile: profile,
-                using: bridge,
-                enabled: compactResume
+                using: bridge
             )
 
             var result = try await resumedSession
@@ -6088,12 +6090,8 @@ final class AppState: ObservableObject {
     private func persistedTranscriptOutcome(
         sessionId: String,
         profile: String,
-        using bridge: DashboardTicketBridge?,
-        enabled: Bool
+        using bridge: DashboardTicketBridge?
     ) async -> PersistedTranscriptOutcome {
-        // The caller passes `enabled` false when the resume already carries
-        // the transcript (no hydration needed), so no history request runs.
-        guard enabled else { return .unavailable }
         let fetchOutcome: PersistedTranscriptFetchOutcome
         if let persistedTranscript = chatResumeLifecycleOperations.persistedTranscript {
             fetchOutcome = await persistedTranscript(sessionId, profile)
@@ -6104,21 +6102,13 @@ final class AppState: ObservableObject {
                     path: sessionMessagesPath(sessionId: sessionId, profile: profile)
                 ))
             } catch {
-                // Older gateways can omit the endpoint, and a freshly created
-                // (or reloading) bridge is not ready to serve requests yet.
-                // Both leave the resume without a usable history source, so
-                // the legacy resume that carries the transcript remains a
-                // safe fallback in those cases; unrelated failures must
-                // surface instead. Note the bridge caps REST response size
-                // (DataURLLimits.maxJSONResponseBytes), so an oversized
-                // transcript also surfaces here rather than silently
+                // requestJSON's readiness poll is the bounded wait for a cold
+                // or reloading bridge; whatever it raises is classified with
+                // the seam-sourced failures below. Note the bridge caps REST
+                // response size (DataURLLimits.maxJSONResponseBytes), so an
+                // oversized transcript also lands here rather than silently
                 // degrading the resume.
-                if Self.historySourceIsUnavailable(error) {
-                    sessionCatalogLog.debug("Persisted transcript endpoint unavailable for \(sessionId, privacy: .public)")
-                    return .unavailable
-                }
-                sessionCatalogLog.debug("Persisted transcript unavailable for \(sessionId, privacy: .public): \(error.localizedDescription, privacy: .public)")
-                return .failed(error)
+                fetchOutcome = .failed(error)
             }
         }
 
@@ -6126,6 +6116,17 @@ final class AppState: ObservableObject {
         case .unavailable:
             return .unavailable
         case .failed(let error):
+            // One classification for both the bridge fetch and the seam:
+            // structural endpoint gaps, a bridge that never became ready
+            // inside its bounded poll, and transient transport trouble
+            // (429, 5xx, status-0 WebKit/network failures, request timeouts)
+            // degrade to the single legacy resume; authentication and every
+            // other failure must surface instead of silently degrading.
+            if Self.historySourceIsUnavailable(error) {
+                sessionCatalogLog.debug("Persisted transcript unavailable for \(sessionId, privacy: .public): \(error.localizedDescription, privacy: .public)")
+                return .unavailable
+            }
+            sessionCatalogLog.debug("Persisted transcript failed for \(sessionId, privacy: .public): \(error.localizedDescription, privacy: .public)")
             return .failed(error)
         case .payload(let response):
             // The dashboard server returns `messages`; the API-server variant
@@ -6146,17 +6147,29 @@ final class AppState: ObservableObject {
         }
     }
 
-    /// A history source that is structurally absent (404/410/501 from the
-    /// history route) or not currently usable (the bridge's WebKit page is
-    /// still loading or reloading) cannot hydrate the transcript right now,
-    /// so the legacy resume is the correct fallback. Every other failure is
-    /// transient or environmental (network, auth, server error) and must
-    /// propagate instead of silently degrading the resume.
+    /// Classifies a history-fetch failure for the resume fallback. Everything
+    /// that leaves the resume without a usable transcript *right now*
+    /// degrades to the single legacy resume:
+    ///
+    ///  - structural endpoint gaps — 404/410 (the gateway predates the
+    ///    messages route);
+    ///  - a bridge that did not become ready within its bounded readiness
+    ///    poll, or a request that outlived its deadline (both funnel into
+    ///    `.notReady` by the bridge's error contract);
+    ///  - transient transport trouble — HTTP 408/429, any 5xx (which
+    ///    subsumes 501), and status 0 (WebKit-level network/abort failures,
+    ///    including response caps).
+    ///
+    /// Client-originated request/bridge timeouts always arrive through one of
+    /// the status-0/`.notReady` funnels above; other authentication (401/403 →
+    /// `.signInRequired`) and every unlisted error is a real failure that must
+    /// propagate instead of silently degrading the resume into a fallback.
     nonisolated static func historySourceIsUnavailable(_ error: Error) -> Bool {
         if case DashboardTicketBridgeError.notReady = error { return true }
         guard case let DashboardTicketBridgeError.http(status, _) = error else { return false }
         switch status {
-        case 404, 410, 501: return true
+        case 0, 404, 408, 410, 429: return true
+        case 500...599: return true
         default: return false
         }
     }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -30,7 +30,8 @@ struct ChatResumeLifecycleOperations {
     var connectClient: (@MainActor (HermesClient) async throws -> Void)?
     var loadCatalog: (@MainActor (HermesClient, Bool) async throws -> [SessionSummary])?
     var mintTicket: (@MainActor (String) async throws -> String)?
-    var openSession: (@MainActor (HermesClient, String) async throws -> SessionResumeResult)?
+    var openSession: (@MainActor (HermesClient, String, Bool) async throws -> SessionResumeResult)?
+    var persistedTranscript: (@MainActor (String, String) async -> PersistedTranscriptFetchOutcome)?
     var branchSession: (@MainActor (
         HermesClient,
         String,
@@ -66,7 +67,8 @@ struct ChatResumeLifecycleOperations {
         connectClient: (@MainActor (HermesClient) async throws -> Void)? = nil,
         loadCatalog: (@MainActor (HermesClient, Bool) async throws -> [SessionSummary])? = nil,
         mintTicket: (@MainActor (String) async throws -> String)? = nil,
-        openSession: (@MainActor (HermesClient, String) async throws -> SessionResumeResult)? = nil,
+        openSession: (@MainActor (HermesClient, String, Bool) async throws -> SessionResumeResult)? = nil,
+        persistedTranscript: (@MainActor (String, String) async -> PersistedTranscriptFetchOutcome)? = nil,
         branchSession: (@MainActor (
             HermesClient,
             String,
@@ -102,6 +104,7 @@ struct ChatResumeLifecycleOperations {
         self.loadCatalog = loadCatalog
         self.mintTicket = mintTicket
         self.openSession = openSession
+        self.persistedTranscript = persistedTranscript
         self.branchSession = branchSession
         self.setSessionTitle = setSessionTitle
         self.refreshContext = refreshContext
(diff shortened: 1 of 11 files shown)


kaishi00:
## Summary
This PR fixes #106 (WebSocket 1009 disconnect loop on large-session resume) by raising the production socket's `maximumMessageSize` to 4 MiB and making `HermesClient.openSession` request a compact `session.resume` (`omit_messages: true`), hydrating the durable transcript from the dashboard `/api/sessions/{id}/messages` REST route with a classified fallback to the legacy full-transcript resume when that source is structurally unavailable.

## Findings

### [severity: major] Transient history-endpoint failures now hard-fail an otherwise-succeeded resume
**File:** Conduit/Services/AppState.swift:2719-2722
**Issue:** The legacy fallback is only taken for `.unavailable` (404/410/501/`notReady`). Every other history failure — including transient gateway 5xx/429, the 12s JS-abort path (`status: 0`), and plain network errors — is re-thrown after `resumeSession` already succeeded. Pre-change, `dashboardSessionTranscript` swallowed every error (`catch { return nil }`) and the resume continued off the RPC transcript. So a parallel aux-channel blip (gateway restarting, flaky cellular while the WS is healthy) now converts a working resume into a hard "Failed to restore this conversation" + `.reconnecting`. This is deliberate per the PR notes, but it changes failure semantics on the very path being made more reliable, and the classification set is broader than "auth/malformed".
**Suggestion:** For transient classes (5xx/429/`status: 0`/timeout), perform exactly one bounded legacy re-resume before surfacing; keep the hard fail for 401/403 and malformed payloads only. This preserves "failures surface" for genuine auth/contract breaks while not turning an intermittent HTTP bump into a resume-blocker.

### [severity: moderate] Compact resume is readiness-gated, so the issue's coldest path is unprotected by compaction
**File:** Conduit/Services/AppState.swift:2674, 2710-2718
**Issue:** `compactResume` requires `bridge?.isReady == true` at decision time. On a cold start — precisely the scenario in #106 (first resume after launch while the WKWebView has not loaded) — the bridge is typically not ready, so the full transcript still travels in the RPC frame, protected only by the 4 MiB raise. Sessions >4 MiB can still reproduce the 1009 disconnect on first resume against new gateways, and no retry-into-compact happens once the bridge warms (only reconnects self-heal). The PR's "known window" covers gateways *without* the endpoint, but the cold-bridge window produces the identical path even on gateways that have it.
**Suggestion:** On a cold bridge, issue the compact resume and wait for readiness with a bounded deadline rather than immediately downgrading to legacy, or re-dispatch one compact reconciliation when the bridge reports ready if the legacy resume did not carry a transcript.

### [severity: minor] Legacy resume inherits the 30s RPC timeout
**File:** Conduit/Services/HermesClient.swift:789-798 (and `requestTimeout` at :426)
**Issue:** `resumeSession` uses the default `requestTimeout` (30s). A 4 MiB legacy fallback frame on a slow link can exceed 30s, timing out and printing "Request timed out: session.resume", defeating the fallback.
**Suggestion:** Pass an explicit higher timeout (e.g. `timeout: 60`) for the legacy variant.

### [severity: minor] `.http(status:detail:)` drops the status code from the user-facing message
**File:** Conduit/Services/DashboardTicketBridge.swift:139-141
**Issue:** `errorDescription` for `.http` returns only `detail`. Callers/surfaces that previously saw `"Dashboard request failed (500)."` now see just the detail string, losing the status context in diagnostics the user might report.
**Suggestion:** Include the status in the description, e.g. `"Dashboard request failed (\(status)). \(detail)".

### [severity: minor] Foreign-session identity discards a healthy compact snapshot and re-runs the full resume
**File:** Conduit/Services/AppState.swift:2704-2708
**Issue:** When the REST rows resolve to a foreign identity, the already-fetched compact `result` (including its live-projection `snapshot`) is discarded and a fresh legacy `session.resume` is issued, on a socket that may be mid-heavy-load. Rare, but wasted work plus a window where the fresher snapshot is dropped.
**Suggestion:** This is acceptable; at most, avoid the second RPC by noting the mismatch only disables *hydration*, not the snapshot — the compact result with the carried-transcript path could be merged rather than re-resumed.

### [severity: nit] `transcriptMatchesSession` is evaluated twice per reconcile
**File:** Conduit/Services/AppState.swift:2697 and 2749-2755
**Issue:** The same predicate is computed once for the hydration gate and again for `shouldUsePersistedTranscript`. Harmless but redundant and a maintenance hazard if the rule changes.
**Suggestion:** Reuse the first result.

## Verdict
REQUEST_CHANGES — primarily for Finding 1: a genuine, plausible behavioral regression (transient aux-channel failure kills a resume that previously succeeded) on the exact path this PR aims to make reliable. Finding 2 is a documented-but-avoidable residual gap for the issue's own cold-start scenario. The transport bound, compact projection, compatibility masking for old gateways that ignore `omit_messages` (correctly handled via `resumeCarriedTranscript`), and test coverage (incl. the >4 MiB deterministic fixture) are all well executed.

<a href="https://opencode.ai/s/KkPN7LJE">New%20session%20-%202026-08-29T12%3A12%3A20.129Z</a>
[opencode session](https://opencode.ai/s/KkPN7LJE)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33251872826)

kaishi00:
## Summary

Fixes #106 (WebSocket 1009 disconnect/reconnect loop on large-session resume) by installing an explicit 4 MiB `maximumMessageSize` on the production socket and making `HermesClient.openSession` request a compact `session.resume` (`omit_messages: true`), hydrating the durable transcript from the dashboard `/api/sessions/{id}/messages` REST route with a classified single-`openSessionLegacy` fallback (structural gap, cold bridge, transient HTTP) and loud surfacing for auth/malformed failures. The follow-up commit (7433f07) correctly addresses the two main findings from the first review round (transient-failure fallback and cold-bridge compact-first) plus the legacy 60s timeout.

## Findings

### [severity: moderate] Cold or slow bridge delays restore by the full 3s readiness poll before deciding on legacy
**File:** Conduit/Services/AppState.swift:2667-2725 (with DashboardTicketBridge.swift:433-435)
**Issue:** The follow-up commit made cold bridges start with the compact resume, but the *presentation* result is still held until the transcript fetch completes, and `requestJSON`'s bounded readiness poll (30 × 100 ms) runs to exhaustion before `.notReady` is produced and the legacy resume starts. Pre-change, the legacy resume ran concurrently with the REST fetch, so a cold- or slow-loading bridge (first launch, `webViewWebContentProcessDidTerminate`, a page that takes >3s) now restores up to `~3s + legacy RPC` later than before, and shows the reconnecting/busy state during that window. If the bridge never becomes ready, the flow pays the full poll then still takes the legacy path — the "cold start" window of #106 is narrowed but not removed for sessions >4 MiB on modern gateways whose bridge loads late.
**Suggestion:** Short-circuit the poll when the bridge is known-dead (`isLoadFailed`) or known-cold at decision time (return `.unavailable` immediately), or start the bounded readiness wait concurrently with the resume decision instead of gating the legacy resume on poll completion.

### [severity: minor] 200-OK history payload missing/null `messages` key now hard-fails an otherwise-successful resume
**File:** Conduit/Services/AppState.swift:6134-6139
**Issue:** If the history endpoint returns a 200 with the array key absent or `null` (an empty transcript serialized as `{"session_id": …}` without an array, or a loose gateway variant), `rawMessages` is nil and the fetch returns `.failed(HermesError.invalidResponse)`. That error is not classified by `historySourceIsUnavailable` and is re-thrown after the compact `session.resume` already succeeded, converting a working resume into `turnState = .reconnecting` + "Failed to restore this conversation" (AppState.swift:2923-2940) and, on the next reconnect, repeating. Pre-change, a nil row list silently fell back to the RPC-carried transcript. CodeRabbit flagged the same spot and it remains unaddressed.
**Suggestion:** Treat an absent or `null` array as `.unavailable` (degrade to exactly one legacy resume, matching the 404/410 handling), and only surface (`.failed`) when the shape is a present-but-non-array value — genuine malformation. Consider the same for unclassified 4xx (e.g. 400), which similarly hard-fails repeatedly on reconnect.

### [severity: nit] `.http(status:detail:)` drops the HTTP status from the user-facing error
**File:** Conduit/Services/DashboardTicketBridge.swift:141-144, 809
**Issue:** `errorDescription` for `.http` returns only `detail`. When the JS page supplies a bare `error` string, the status code (frequently what a user pastes into a bug report) is lost; pre-change the default `"Dashboard request failed (\(status))."` carried it.
**Suggestion:** Include the status, e.g. `"Dashboard request failed (\(status)). \(detail)"`.

### [severity: nit] `transcriptMatchesSession` is computed twice per reconcile
**File:** Conduit/Services/AppState.swift:2699-2703 and 2751-2756
**Issue:** The identity predicate is evaluated once for the hydration gate and again for `shouldUsePersistedTranscript`; redundant and a maintenance hazard if the rule changes.
**Suggestion:** Carry the first result when hydrating (or compute once after `transcript` is finalized).

### [severity: nit] Oversized legacy frame is indistinguishable from any other disconnect
**File:** Conduit/Services/HermesClient.swift:572-594
**Issue:** In the accepted "known window" (gateway without the history endpoint + session >4 MiB), `receive()` fails on the 4 MiB bound and the receive loop treats it exactly like any error — `onDisconnected` → reconnect → repeat — with no log distinguishing a 1009 message-too-big from a network drop.
**Suggestion:** Log `socket.closeCode` (or the URLError `NSURLErrorCannotDecodeContentData` variant) on receive failure so the residual window is diagnosable from gateway logs.

### [severity: nit] `requestJSON` error-contract docstring omits the new `.http` case
**File:** Conduit/Services/DashboardTicketBridge.swift:420-425
**Issue:** The doc lists only `.notReady`, `.signInRequired`, and `.requestFailed`; non-auth HTTP/status-0 failures now throw `.http(status:detail:)`.
**Suggestion:** Add the new case to the contract documentation (also satisfies the flagged docstring-coverage check).

## Verdict

COMMENT

The architecture is sound and the two substantive findings from the first review round are genuinely fixed (transient failures degrade to exactly one legacy resume via centralized classification; cold bridges start compact; legacy resume gets a 60s budget; the >4 MiB test was honestly re-scoped). The remaining items are behavioral edge cases rather than defects: the worth-flagging moderate is the restored-latency tradeoff for cold bridges, and the worth-fixing minor is the missing/`null` messages-key hard fail — both are bounded by the single-legacy-resume design and the deliberate "failures surface" contract. No retain cycles, force unwraps, Swift 6 concurrency violations, or SwiftUI lifecycle issues were introduced (production closures continue using `[weak self]`; no new force-unwrap/`as!` sites).

<a href="https://opencode.ai/s/Nnc243YU">New%20session%20-%202026-08-29T13%3A16%3A31.465Z</a>
[opencode session](https://opencode.ai/s/Nnc243YU)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33254333858)

PR #111, opened 2026-08-29
feat(chat): add chat-only text size preference (#85)

### LINKED ISSUE #85: Is there any font-size config?
Hi. Thanks for the app, it's great. 
Is there any font-size config for the chat? I couldn't find it

kaishi00:
## Chat-only font size preference (issue #85)

Closes #85.

Adds a five-position stepped **chat text size** preference (Smallest · Smaller · Default · Larger · Largest) that reflows readable transcript content independently from the rest of the UI. Live update, no Save button, local device-only — never synced to Hermes, profiles, or the gateway.

### Design

**`ChatTypography` is the sole authority for transcript-content font resolution.** The selected size is a first-class rendering input that **augments** iOS Dynamic Type (`system Dynamic Type size × Conduit chat scale`), so `Default` renders exactly like today's transcript and accessibility text sizing keeps working at every position. Semantic roles (body, headings by level, quote, table header/body, thinking, block code, source cards) map to the same Dynamic Type styles the renderer used before — hierarchy and monospace/code treatment preserved.

### Coverage

- **Markdown paths:** settled user/assistant/partial/system Markdown, streaming (stable chunks + live tail converge), large documents (collapsed preview, expanded flow chunks, specialized callouts/columns), tables (ordinary + paged), block code, large/sliced code, and the source cards around Mermaid/KaTeX previews. The WebKit diagram/formula previews intentionally keep their internal size (documented in `ChatTypography.swift`).
- **Typography-sensitive caches all include the chat-size identity:** `MarkdownRenderCache` key, `MarkdownTableLayout` measurement key, `LargeFlowChunkBox` memo, `LargeCodeSliceIdentity` (highlight task identity). `SyntaxHighlighter` needs no key — it bakes colors only; fonts are applied at bridge time.
- **Equatable settled-row gates:** `UserMessageContent`, `SettledAssistantMessageContent`, and `SettledThinkingCardContent` carry an explicit `chatTextSize` input — a preference change re-opens the gate; ordinary streaming publishes keep the existing isolation.
- **Chrome untouched:** timestamps, identity labels, copy/branch/read-aloud buttons, navigation, Settings, composer, tool/status cards, selection handles.

### Settings

Settings → Chat gains a live five-position slider (device-only `@AppStorage`, same pattern as the Return-key preference): A–A scale labels, selected-value caption, `accessibilityLabel`/`accessibilityValue`/`accessibilityHint`.

### Tests

New `ChatTypographyTests` (14 tests): preference default/round-trip/invalid fallback, scale spec, `.default` ≡ legacy fonts, monotonic ordering, Dynamic Type augmentation via trait collections, hierarchy/monospace/traits at every size, render-cache size separation + return-hit, streaming/settled convergence, formatter run scaling, selection content invariance, flow-chunk memo invalidation, table measurement reflow, large-code-slice identity, all three Equatable gates, and a **view-level integration test** pinning the ordinary `MarkdownTable` → `MarkdownTableLayout` chat-size wiring (mounted column widths must grow with the selected size). `SettledMessageIsolationTests` adapted for the new gate parameter.

### Validation

- Full `ios_test` suite: **70/70 units passed** (69 pre-existing + new suite), zero failures, zero uncovered.
- Multi-model review gate ran on the branch diff; the one verified BLOCKER (ordinary `MarkdownTable` missing the measurement wiring — caught by review, missed by direct-call tests) and the list-marker Dynamic Type warning are fixed in this commit, plus the flagged slider-knob/test-newline nits.

## Summary by CodeRabbit

- **New Features**
  - Added a device-only Chat text size setting with five size options.
  - Changes apply immediately without saving or profile synchronization.
  - Conversation text, headings, tables, quotes, lists, code blocks, and thinking content now follow the selected size.
  - Chat text sizing works alongside accessibility text-size settings.

- **Bug Fixes**
  - Preserved browsing position when text resizing reflows chat content.

- **Tests**
  - Added coverage for persistence, rendering, scaling, reflow, and cache updates.

### DIFF
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -739,6 +739,7 @@ private struct ChatSettingsDetail: View {
             ),
             trailingSection: AnyView(
                 VStack(spacing: 14) {
+                    ChatTextSizeSettings()
                     ComposerReturnKeySettings()
                     DeviceHapticsSettings()
                 }
@@ -757,6 +758,61 @@ private struct ChatSettingsDetail: View {
         .init(key: "display.memory_notifications", label: "Self-improvement updates", help: "Choose whether Conduit follows Hermes, always shows, or never shows maintenance updates.", control: .labeledOptions([(value: "default", label: "Use Hermes default"), (value: "on", label: "Always show"), (value: "off", label: "Never show")], defaultValue: "default")),
         .init(key: "agent.image_input_mode", label: "Image attachments", help: "How Hermes supplies images to a model.", control: .options(["auto", "native", "text"], defaultValue: "auto")),
     ]
+}
+
+/// Local, device-only chat text-size preference (issue #85). Stored in
+/// UserDefaults via @AppStorage with the same lifetime as
+/// ComposerReturnKey — never part of the Hermes profile configuration and
+/// never synchronized anywhere. A five-position stepped slider: the change
+/// is live (the visible transcript re-renders as the slider moves), there
+/// is no Save button, and `Default` keeps today's appearance.
+private struct ChatTextSizeSettings: View {
+    @AppStorage(ChatTypography.preferenceKey) private var chatTextSizeRaw = ChatTypography.defaultSize.rawValue
+
+    private var selected: ChatTextSize {
+        ChatTypography.resolve(rawValue: chatTextSizeRaw)
+    }
+
+    var body: some View {
+        ConduitSettingsSection(
+            title: "Chat text size",
+            symbol: "textformat.size",
+            tint: .conduitAura
+        ) {
+            Text("Readable conversation content only — messages, lists, tables, and code. Buttons, timestamps, and the rest of the interface keep their size, and iOS Dynamic Type still applies on top.")
+                .font(.footnote)
+                .foregroundStyle(.secondary)
+            VStack(alignment: .leading, spacing: 8) {
+                HStack {
+                    Text("A")
+                        .font(.system(size: 12))
+                        .foregroundStyle(.secondary)
+                        .accessibilityHidden(true)
+                    Spacer()
+                    Text("A")
+                        .font(.system(size: 22, weight: .medium))
+                        .foregroundStyle(.secondary)
+                        .accessibilityHidden(true)
+                }
+                Slider(
+                    value: Binding(
+                        get: { Double(selected.rawValue) },
+                        set: { chatTextSizeRaw = Int($0.rounded()) }
+                    ),
+                    in: 0...Double(ChatTextSize.allCases.count - 1),
+                    step: 1
+                )
+                .tint(.conduitAccent)
+                .accessibilityLabel("Chat text size")
+                .accessibilityValue(selected.displayName)
+                .accessibilityHint("Five steps from smallest to largest. Applies to conversation text immediately.")
+                Text(selected.displayName)
+                    .font(.subheadline.weight(.medium))
+                    .foregroundStyle(.secondary)
+                    .frame(maxWidth: .infinity, alignment: .center)
+            }
+        }
+    }
 }
 
 /// Local, device-only composer input preference. Stored in UserDefaults via
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -1409,6 +1409,11 @@ struct SystemBubble: View {
     }
 }
 
+/// Review activity summary. Intentionally OUTSIDE the chat text-size
+/// preference (issue #85): these status/activity cards — like tool cards —
+/// are not ordinary transcript Markdown and keep their fixed typography.
+/// (Of the system-message surfaces, only `SystemBubble`'s Markdown body
+/// follows `ChatTypography`.)
 private struct ReviewSummaryCard: View {
     let activity: ReviewActivity
     let timestamp: String
@@ -1489,6 +1494,9 @@ private struct ReviewSummaryCard: View {
     }
 }
 
+/// Model-change summary. Intentionally OUTSIDE the chat text-size
+/// preference (issue #85) — a status/activity card with fixed typography;
+/// see ReviewSummaryCard.
 private struct ModelChangeSummaryCard: View {
     let model: String
     let provider: String
--- a/Conduit/Views/ChatView.swift
+++ b/Conduit/Views/ChatView.swift
@@ -96,6 +96,18 @@ struct ChatView: View {
         viewport.renderedScrollScope
     }
 
+    /// The chat-only text-size preference (issue #85). Local, device-only
+    /// @AppStorage like ComposerReturnKey — never synced to Hermes or the
+    /// profile. Injected once here so every transcript Markdown path
+    /// (settled, streaming, large-document, tables, code) resolves the same
+    /// typography, while non-chat subtrees keep the `.default` environment
+    /// value and today's appearance.
+    @AppStorage(ChatTypography.preferenceKey) private var chatTextSizeRaw = ChatTypography.defaultSize.rawValue
+
+    private var chatTextSize: ChatTextSize {
+        ChatTypography.resolve(rawValue: chatTextSizeRaw)
+    }
+
     var body: some View {
         VStack(spacing: 0) {
             // Message list
@@ -115,6 +127,7 @@ struct ChatView: View {
             // Composer + control bar
             ComposerBar()
         }
+        .environment(\.chatTextSize, chatTextSize)
         .background(Color.clear)
         .overlay(alignment: .top) {
             if appState.isOpeningNotificationSession {
@@ -881,6 +894,7 @@ struct UserBubble: View {
     let message: ChatMessage
     let gatewayResolver: GatewayMediaDataURLResolver?
     @Environment(\.sizeCategory) private var sizeCategory
+    @Environment(\.chatTextSize) private var chatTextSize
 
     var body: some View {
         HStack {
@@ -889,7 +903,8 @@ struct UserBubble: View {
                 UserMessageContent(
                     message: message,
                     gatewayResolver: gatewayResolver,
-                    sizeCategory: sizeCategory
+                    sizeCategory: sizeCategory,
+                    chatTextSize: chatTextSize
                 )
                     .equatable()
 
@@ -903,17 +918,23 @@ struct UserBubble: View {
 /// Settled user-row content: Markdown plus attachment previews. Equatable
 /// over the message and the stable gateway resolver so streaming publishes
 /// cannot re-evaluate it (same contract as
-/// SettledAssistantMessageContent).
-private struct UserMessageContent: View, Equatable {
+/// SettledAssistantMessageContent). Internal (not private) so the gate
+/// contract is directly testable.
+struct UserMessageContent: View, Equatable {
     let message: ChatMessage
     let gatewayResolver: GatewayMediaDataURLResolver?
     /// Explicit Dynamic Type input — see SettledAssistantMessageContent.
     let sizeCategory: ContentSizeCategory
+    /// Explicit chat text-size input (issue #85): a preference change must
+    /// re-open the gate and re-render, while ordinary streaming publishes —
+    /// which leave both explicit inputs equal — keep the isolation.
+    let chatTextSize: ChatTextSize
 
     static func == (lhs: Self, rhs: Self) -> Bool {
         lhs.message == rhs.message
             && lhs.gatewayResolver === rhs.gatewayResolver
             && lhs.sizeCategory == rhs.sizeCategory
+            && lhs.chatTextSize == rhs.chatTextSize
     }
 
     var body: some View {
@@ -1162,6 +1183,11 @@ struct SettledAssistantMessageContent: View, Equatable {
     /// UIKit trait collections, layout direction through TextKit — and do
     /// not require body re-evaluation.
     let sizeCategory: ContentSizeCategory
+    /// Explicit chat text-size input (issue #85): equality describes every
+    /// typography input, so a preference change re-opens the gate and
+    /// re-renders (the render cache keys on the chat-size identity), while
+    /// unrelated streaming publishes — equal inputs — stay gated.
+    let chatTextSize: ChatTextSize
 
     static func == (lhs: Self, rhs: Self) -> Bool {
         lhs.message == rhs.message
@@ -1169,6 +1195,7 @@ struct SettledAssistantMessageContent: View, Equatable {
             && lhs.avatarURL == rhs.avatarURL
             && lhs.gatewayResolver === rhs.gatewayResolver
             && lhs.sizeCategory == rhs.sizeCategory
+            && lhs.chatTextSize == rhs.chatTextSize
     }
 
     var body: some View {
@@ -1308,6 +1335,7 @@ struct AssistantBubble: View {
     let gatewayResolver: GatewayMediaDataURLResolver?
     @EnvironmentObject var appState: AppState
     @Environment(\.sizeCategory) private var sizeCategory
+    @Environment(\.chatTextSize) private var chatTextSize
 
     var body: some View {
         VStack(alignment: .leading, spacing: 6) {
@@ -1316,7 +1344,8 @@ struct AssistantBubble: View {
                 displayName: appState.profileDisplayName(appState.activeProfile),
                 avatarURL: appState.profileAvatarURL(for: appState.activeProfile),
                 gatewayResolver: gatewayResolver,
-                sizeCategory: sizeCategory
+                sizeCategory: sizeCategory,
+                chatTextSize: chatTextSize
             )
             .equatable()
 
@@ -1501,13 +1530,16 @@ private struct ModelChangeSummaryCard: View {
 
 /// Settled reasoning-row content, gated the same way as assistant/user
 /// content: identical presentation inputs skip body evaluation during
-/// unrelated AppState publishes.
-private struct SettledThinkingCardContent: View, Equatable {
+/// unrelated AppState publishes. Internal (not private) so the gate
+/// contract is directly testable.
+struct SettledThinkingCardContent: View, Equatable {
     let message: ChatMessage
     let displayName: String
     let avatarURL: URL?
     /// Explicit Dynamic Type input — see SettledAssistantMessageContent.
     let sizeCategory: ContentSizeCategory
+    /// Explicit chat text-size input — see SettledAssistantMessageContent.
+    let chatTextSize: ChatTextSize
     @State private var expanded = false
 
     static func == (lhs: Self, rhs: Self) -> Bool {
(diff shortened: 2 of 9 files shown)


kaishi00:
## Summary
This PR adds a five-position, device-only chat text-size preference that scales readable transcript content (Markdown, tables, code, thinking blocks) on top of iOS Dynamic Type via a new `ChatTypography` authority, with live updates wired through a new `@AppStorage` environment value and size-aware cache identities.

## Findings

### [severity: minor] Review/model-change system cards don't scale while SystemBubble does
**File:** Conduit/Views/ChatView.swift:1438, 1463, 1512
**Issue:** `ReviewSummaryCard` and `ModelChangeSummaryCard` render readable message content (summary/detail text) with raw `UIFont.preferredFont(...)` calls, so they ignore the new chat text size. Meanwhile plain system messages (`SystemBubble`, line 1393) render through `MarkdownText` and *do* scale. This is an inconsistency inside the same `system` message role that users will notice if they set a non-default size.
**Suggestion:** Route these two cards' fonts through `ChatTypography` (a `.body`/`.subheadline` role at `chatTextSize`), or explicitly document them as excluded chrome alongside tool/status cards.

### [severity: minor] `LargeCodeSliceView` shows stale-size highlighted code after a chat-size change
**File:** Conduit/Views/Components/MarkdownLargeDocumentView.swift:993, 1010-1051
**Issue:** `highlighted` is `@State` and caches an `NSAttributedString` with fonts baked in at bridge time. When `chatTextSize` changes, `sliceIdentity` changes so `.task(id:)` restarts the highlight pass, but until that async pass completes the view keeps rendering the old attributed string (whose every run carries a font attribute, so the new `font:` param has no effect). During the window between a slider step and pass completion — visible when dragging through all five positions — the code block renders at the previous size while surrounding text updates immediately.
**Suggestion:** Clear `highlighted = nil` on identity changes (e.g. `.onChange(of: chatTextSize) { highlighted = nil }` or reset inside the `.task(id:)` when `sliceIdentity != previousIdentity`), so the plain-text fallback renders at the new size until the highlight lands.

### [severity: nit] Dead public API in `ChatTypography`
**File:** Conduit/Views/Components/ChatTypography.swift:161-165
**Issue:** `ChatTypography.pointSize(for:chatSize:)` is never called anywhere in the app or tests (verified by search); `scaledFont` is only used internally by `font(for:)`. Unused API invites divergent call sites later.
**Suggestion:** Remove `pointSize(for:)` (or fold the comment into `scaledFont`) unless a call site is planned.

### [severity: nit] Integration test relies on a dispatch-queue settle that may flake
**File:** ConduitTests/ChatTypographyTests.swift:518-564
**Issue:** `testOrdinaryTableViewWiresChatSizeIntoColumnWidths` mounts a `UIWindow`/`UIHostingController` and settles layout with a bare `DispatchQueue.main.async` fulfillment before re-laying-out. SwiftUI's `GeometryReader`-driven `availableWidth` update can need more than one run-loop pass; under load this can assert on pre-settle widths and flake.
**Suggestion:** Poll the mounted `TextViews`' nonzero/settled widths with a deadline (e.g. repeat layout + `wait` until widths stop changing) instead of a single settle hop.

## Verdict
APPROVE

No critical or major issues found. The design is sound: the pre-existing `@MainActor` isolation on the four UIApplication-touching cache/storage paths is preserved, all typography-sensitive caches key on the new `cacheIdentity`, the three Equatable gates (plus the streaming/tail paths) consistently thread `chatTextSize`, and no new force-unwraps, retain cycles, or unbounded resources were introduced. `Default` reproducibly equals the legacy fonts, and the `@AppStorage` pattern matches the existing `ComposerReturnKey` precedent.

<a href="https://opencode.ai/s/xRE6SA1i">New%20session%20-%202026-08-29T21%3A24%3A32.200Z</a>
[opencode session](https://opencode.ai/s/xRE6SA1i)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33275947118)

kaishi00:
## Follow-up pass pushed (`79a39f3`) — addresses the remaining review findings

**1. Stale highlighted large-code typography — fixed.**
`LargeCodeSliceView` no longer keeps the previous highlighted attributed string mounted across a typography identity change. State moved into `LargeCodeSliceHighlightState`, a **value-type struct in `@State`** (the gate's runtime-verified blocker: a reference type mutated in place is invisible to SwiftUI — swap-in and invalidation never rendered). On any `sliceIdentity` change (chat size, Dynamic Type, source), `.onChange` clears the stored result synchronously, so the slice falls back to plain text at the **new** font immediately; when the off-main highlight pass completes, `syncAndStore` records it in a single state write. The live `sliceIdentity == passIdentity` stale guard and `.task(id:)` cancellation are preserved.

**2. Mounted table integration test hardened.** The single `DispatchQueue.main.async` hop is replaced by a bounded settle loop (≤10 layout passes × 50 ms, requiring usable + two-consecutive-pass stable widths, loud `XCTUnwrap` failure otherwise). The mounted-widths-change-with-chat-size assertion is unchanged.

**3. `ChatTypography.pointSize(for:chatSize:)` removed** — verified zero remaining callers (markers now resolve a full `UIFont`).

**4. Status-card scope documented.** `ReviewSummaryCard` / `ModelChangeSummaryCard` carry comments pinning them as status/activity cards intentionally outside `ChatTypography` (their typography is unchanged); `SystemBubble`'s Markdown body keeps following the preference.

**5. Added regression coverage**
- `testHighlightStateInvalidatesDisplayedResultOnIdentityChange` — identity-change invalidation drops the stored result; unchanged identity is a no-op (stale passes are refused by the view's live identity guard, which is preserved).
- `testLargeCodeSliceTypographyFollowsIdentityChange` (hosted, view-level) — mounts `LargeCodeBlockView` with a >32 KB source, waits for tokenized slices, flips the environment chat size Default → Largest, and asserts **no mounted slice keeps the previous size's fonts** and everything renders at the new size. This is the test that fails on the in-place-mutation blocker the second gate caught.
- `testChatTextSizeReflowDoesNotScrollWhileBrowsing` (viewport controller) — a typography reflow reaches the viewport only as layout-metric ticks; ticks while browsing away from bottom issue no scroll commands and schedule no effective correction, so a long conversation never gets thrown to the bottom by a size change.
- `testSelectionGeometryTracksTypographyReflow` (TextKit) — after a font-size reflow the selection stays expressible and caret geometry tracks the reflowed layout (non-degenerate, in-bounds, moved off the old geometry).

**Validation:** focused suites (`ChatTypographyTests`, `ChatViewportControllerTests`, `ChatTextSelectionTests`, `MarkdownLargeDocumentTests`) and the **full suite: 70/70 units, 0 failures** — both re-run on the Mac after every change in this pass. Multi-model review gate re-run twice on the incremental diffs; final verdict APPROVE (2–1, dissenting concerns adjudicated as non-blocking hardening).

### Manual QA checklist (simulator/device, before merge)

1. Normal paragraph Markdown — slide through all five positions.
2. Headings / lists / quotes — hierarchy intact at each position.
3. Ordinary code block — font follows the slider; Copy unaffected.
4. **Large syntax-highlighted code (expand "Show all lines")** — move across positions: no period where highlighted code remains at the previous size (plain text at the new size may flash while re-highlighting — that is the intended fallback).
5. Wide table and a large/paged table — widths re-measure, no clipping or stale widths.
6. Streaming response — change size mid-stream: stable chunks and the live tail converge on one size.
7. Long conversation scrolled away from bottom — change size: view must not jump to bottom/top.
8. Cross-block text selection before/after a size change — selection rects and copy stay correct.

kaishi00:
## Summary
Adds a five-position, device-only chat text-size preference that scales transcript-readable typography (Markdown, tables, code, thinking) on top of iOS Dynamic Type via a new `ChatTypography` authority, wired through `@AppStorage`, an injected environment value, size-aware cache identities, and re-opened Equatable gates.

## Findings

### [severity: minor] `cacheIdentity` is the enum position, so re-tuning the scale factors cannot invalidate baked-font caches
**File:** Conduit/Views/Components/ChatTypography.swift:73
**Issue:** `cacheIdentity` is `String(rawValue)`, i.e. the *position*, while everything downstream stores resources whose font sizes bake the *scale factor*: `MarkdownRenderCache` (MarkdownText.swift:290-300), `MarkdownTableLayout` (MarkdownText.swift:1204-1214), `LargeFlowChunkBox` (MarkdownLargeDocumentView.swift:663-686), and `LargeCodeSliceHighlightState` (MarkdownLargeDocumentView.swift:995-1021). The doc comment on lines 69-72 claims "re-tuning the scale factors later needs no cache migration" — but a scale change with the same position leaves every baked-font cache (and the mounted highlighted slices, whose `sliceIdentity` also keys only on position) serving stale-size output. This is the same class of staleness bug the PR just fixed for chat-size changes.
**Suggestion:** Fold the scale factor (or a `ChatTypography` version token) into `cacheIdentity` and `LargeCodeSliceIdentity`, e.g. `String(rawValue)` → `"\(rawValue)-\(scale)"`, so a future re-tune bumps every cache identity.

### [severity: minor] Optional `chatTextSize` defaults invite a silent-ignore regression at future call sites
**File:** Conduit/Views/Components/MarkdownText.swift:390, 1200 (and MarkdownLargeDocumentView.swift:667)
**Issue:** `MarkdownSelectionFormatter.attributedText`, `MarkdownTableLayout.columnWidths`, and `LargeFlowChunkBox.attributedText` all declare `chatTextSize: ChatTextSize = .default`. A call site that omits the argument compiles fine and quietly renders at `Default` — the exact kind of missed-wiring bug this PR's review gate flagged as the BLOCKER on the ordinary `MarkdownTable` (a width/measure mismatch when the cell renders one size and the layout measures another). The optional snapshots (`LargeCodeSliceIdentity.identity`) are correctly required, so the split is inconsistent.
**Suggestion:** Make `chatTextSize` a required parameter on these three APIs so a missed argument is a compile error, or keep the default only where a caller provably cannot supply it.

### [severity: minor] `ReviewSummaryCard`/`ModelChangeSummaryCard` reproduce unscaled, while `SystemBubble` scales
**File:** Conduit/Views/ChatView.swift:1413-1540
**Issue:** Now documented as intentional (status/activity cards), but the visual result is still user-visible inconsistency inside a single `system` role: `SystemBubble`'s Markdown body scales with the preference, while the summary/detail text of review and model-change cards renders at fixed Dynamic Type. The Settings copy says the preference covers "messages, lists, tables, and code," which arguably includes these summaries. Users at non-default sizes will see the mismatch.
**Suggestion:** Either route the cards through `ChatTypography` (`.body`/`.subheadline` at `chatTextSize`), or call them out in the Settings explanatory copy as excluded like tool/status cards.

### [severity: nit] Per-step transcript-wide re-render during slider drag
**File:** Conduit/Views/AuxiliaryViews.swift:790-806
**Issue:** Each slider step invalidates every typography-sensitive cache identity and re-loads the whole visible (and settled) transcript at the new size. That is the intended live-update behavior and the viewport controller prevents scroll jumps, but on very long transcripts moving across all five positions can hitch the main thread per transition.
**Suggestion:** Acceptable as designed; if it ever shows up, throttle the `@AppStorage` writes to the last position on a drag end, or batch per-transition cache rebuilds.

## Verdict
APPROVE

No critical or major issues. The design is sound: `@MainActor` isolation is preserved on all four `UIApplication`-touching cache paths; the previously flagged stale-highlighted-slice bug is properly fixed via a value-type `@State` with synchronous invalidation, single-write convergence, cancellation, and a live-identity guard; all typography-sensitive caches and the three Equatable gates thread `chatTextSize` consistently; `.default` verifiably reproduces legacy fonts; Dynamic Type augmentation is preserved; and accessibility (slider label/value/hint, hidden decorative glyphs) is handled. No new force unwraps, retain cycles, or unbounded resources. Findings above are hardening/minor items.

<a href="https://opencode.ai/s/70LELnKW">New%20session%20-%202026-08-30T00%3A12%3A17.701Z</a>
[opencode session](https://opencode.ai/s/70LELnKW)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33282714014)

kaishi00:
## Final correctness patch pushed (`fccf3db`) — render-time identity guard for large-code slices

Implements the remaining edge case from the review: between an identity change and `onChange` delivery there is one body evaluation where `highlightState.highlighted` still carries the previous identity. `LargeCodeSliceView` now renders the highlighted attributed string **only** via `highlightState.highlightedText(for: sliceIdentity)` — an exact identity match — so any mismatch routes that evaluation to the plain-text path at the current font. The invariant is now:

> a highlighted attributed string is mounted only while its recorded identity exactly matches the live `sliceIdentity`.

Layered safeguards for the slice highlight are: (1) this render-time guard, (2) synchronous `.onChange(of: sliceIdentity)` invalidation, (3) `.task(id: sliceIdentity)` cancellation, (4) the live `sliceIdentity == passIdentity` stale-pass guard before recording.

New focused test `testHighlightedResultIsOnlyEligibleForItsOwnIdentity` pins the eligibility invariant directly (stored A / live B → not eligible; stored B / live B → eligible; pristine state → nothing eligible). Also folded in the gate's cheap suggestions: stable `.string.isEmpty`-style accessor replaced by `.characters.isEmpty` for the defensive empty-result check, dead `isCurrent(for:)` removed, the browsing-mode viewport test now also pins that a relatch correction is never scheduled while browsing, and comment/wording accuracy fixes.

**Validation:** affected suites (`ChatTypographyTests`, `ChatViewportControllerTests`, `ChatTextSelectionTests`, `MarkdownLargeDocumentTests`) and the full suite (**70/70 units, 0 failures**) re-run on the Mac after these edits. Multi-model review gate on this patch: **APPROVE, unanimous, no blockers.**

The manual QA checklist in the previous comment still applies before merge — especially item 4 (large highlighted code across slider positions must never show previous-size highlighted text).

kaishi00:
## Summary
This PR adds a five-position, device-only chat text-size preference (issue #85) that scales transcript-readable Markdown typography (body, headings, lists, quotes, tables, code, thinking) on top of iOS Dynamic Type via a new `ChatTypography` authority, wired through `@AppStorage`, an injected `@Environment` value, size-aware cache identities, and re-opened Equatable gates, with a render-time identity guard for large-code highlight slices.

## Findings

### [severity: minor] Optional `chatTextSize = .default` defaults invite the exact silent-ignore regression this PR's gate caught
**File:** Conduit/Views/Components/MarkdownText.swift:390, Conduit/Views/Components/MarkdownText.swift:1200, Conduit/Views/Components/MarkdownLargeDocumentView.swift:667
**Issue:** `MarkdownSelectionFormatter.attributedText`, `MarkdownTableLayout.columnWidths`, and `LargeFlowChunkBox.attributedText` all default their new `chatTextSize` param to `.default`. A future call site that omits the argument compiles cleanly and silently renders at `Default` — the same class of missed-wiring bug the review gate flagged as the BLOCKER on the ordinary `MarkdownTable` (cell renders at one size while layout measures at another). The large-code slice identity side is correctly required (`LargeCodeSliceIdentity.identity`), so the split is inconsistent.
**Suggestion:** Make `chatTextSize` a required parameter on these three APIs so a missed argument is a compile error, or keep the default only where a caller provably cannot supply it.

### [severity: minor] System-role typography is inconsistent inside one message role
**File:** Conduit/Views/ChatView.swift:1413-1540
**Issue:** `SystemBubble`'s Markdown body scales with the preference (render:1413 through `MarkdownText`), while `ReviewSummaryCard` (1443, 1468) and `ModelChangeSummaryCard` (1520, 1834, 1981) render their summary/detail text at fixed Dynamic Type. The exclusion is now explicitly documented as intentional (status/activity cards), but it remains a user-visible mismatch within the same `system` role, and the Settings copy ("messages, lists, tables, and code") arguably covers these summaries.
**Suggestion:** Either route the two cards through `ChatTypography` (`.body`/`.subheadline` at `chatTextSize`), or call the exclusion out in the Settings explanatory copy alongside tool/status cards so the boundary matches the label.

### [severity: nit] `cacheIdentity` is the position, so the "no cache migration" claim over-promises for future scale re-tunes
**File:** Conduit/Views/Components/ChatTypography.swift:73
**Issue:** `cacheIdentity` is `String(rawValue)` — the position — while every downstream cache (`MarkdownRenderCache`, `MarkdownTableLayout` key, `LargeFlowChunkBox`, `LargeCodeSliceHighlightState`) bakes the *scale factor* into stored resources. Practically inert today (scale is a compile-time constant and the caches are in-memory per-process), but the doc comment at line 69-72 asserts a scale re-tune "needs no cache migration," which would not hold for any runtime-tunable scale.
**Suggestion:** Fold the scale factor or a `ChatTypography` version token into `cacheIdentity` (and `LargeCodeSliceIdentity`), e.g. `"\(rawValue)-\(scale)"`, making the invariant robust rather than incidentally correct.

### [severity: nit] Hosted highlight test uses fixed-duration `pump` sleeps
**File:** ConduitTests/ChatTypographyTests.swift:522-529 (and uses at 539, 557)
**Issue:** `pump(4.0)` always waits the full 8 seconds total and can assert on pre-settle state under CI load if the highlight pass finishes late; if it finishes early the test still burns the full budget. The table test at 768-786 already uses a bounded condition-based settle loop.
**Suggestion:** Replace the fixed sleeps with a bounded poll loop that breaks as soon as the "tokenized slices mounted" / "all slices at new size" conditions hold.

### [severity: nit] Per-step transcript-wide re-render while dragging the slider
**File:** Conduit/Views/AuxiliaryViews.swift:784-806
**Issue:** Each slider step re-opens every Equatable gate and invalidates every typography-sensitive cache identity, re-rendering the whole visible+settled transcript at the new size. This is intended live-update behavior (and the viewport controller prevents scroll jumps), but on very long transcripts dragging across all five positions can hitch the main thread per transition.
**Suggestion:** Acceptable as designed; if it ever shows up, debounce the `@AppStorage` write to the slider's final position on drag end, or batch per-transition cache rebuilds.

## Verdict
APPROVE

No critical or major issues found. The core invariants hold: `@MainActor` isolation is preserved on all four `UIApplication`-touching cache paths; every typography-sensitive cache and the three Equatable gates consistently thread the chat-size identity (including the ordinary `MarkdownTable` wiring that was the earlier blocker); the large-code highlight staleness is properly closed by the value-type `@State` plus synchronous `.onChange` invalidation, `.task(id:)` cancellation, the live `passIdentity` stale guard, and the render-time `highlightedText(for:)` eligibility check; `.default` byte-for-byte reproduces the legacy fonts and Dynamic Type augmentation still responds at every position. No new retain cycles, force unwraps, or data races were introduced, and the slider exposes label/value/hint accessibility. Findings above are hardening/minor items; the fixed-duration pump and the optional-parameter defaults are the most worth acting on.

<a href="https://opencode.ai/s/yChBkQ4p">New%20session%20-%202026-08-30T00%3A43%3A58.607Z</a>
[opencode session](https://opencode.ai/s/yChBkQ4p)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33283961426)

PR #112, opened 2026-08-30
fix(chat): honor Hermes display projection on persisted transcript rows

kaishi00:
## Summary

Persisted Hermes rows can ride as `role=user` for provider-history semantics while `display_kind` / `display_content` / `display_metadata` tell clients how to actually present them. Conduit mapped the physical role straight onto the UI role and ignored those fields, so synthetic rows — model switches, auto-continues, hidden compaction handoffs, delegation completions — leaked as human user bubbles after a session reload. The leak became prominent on the compact-resume path because `GET /api/sessions/{id}/messages` returns the raw persisted rows, unlike `session.resume`, whose server-side projection filters them.

Conduit-only fix (no `NousResearch/hermes-agent` changes), implemented at the shared normalization boundary so both the REST hydration and the legacy `session.resume` transcript enforce the same contract. Semantics mirror Hermes Desktop/TUI hydration, verified against the installed v0.20.6 upstream source.

## Contract handling

- **`display_kind=hidden`** — row dropped before any content work, regardless of physical role; multi-megabyte model-facing scaffolding is never scanned or copied.
- **`display_content`** — field PRESENCE is the authority boundary: any existing value (string, explicit empty string, JSON null, odd scalar, array, object) resolves via the existing coercion and is authoritative, so a malformed projection can never fall back and expose the physical carrier. The carrier is never extracted when a projection exists, and the legacy compaction filters never re-judge projected text.
- **Known synthetic kinds** — forced system/timeline presentation, never a user bubble:
  - `model_switch` → existing model-change card when the persisted marker matches, canned "Model changed" otherwise; an explicit `display_content` outranks the card.
  - `personality_switch` → "Personality changed" (the persona prompt embedded in the marker is never rendered).
  - `auto_continue` → "Resumed interrupted turn".
  - `async_delegation_complete` → "N background agent(s) finished" from `display_metadata.task_count`; "Background agent work finished" when metadata is missing/malformed. Metadata arrives as object, JSON text, array, or garbage — none can fail normalization.
  - `internal_notification` → the operational notice text, with a `[System: …]` wrapper stripped.
- **Unknown kinds** — ordinary normalization, conservatively preserved (no deletion, no reinterpretation).
- **Tool rows** — legacy handling preserved; only `hidden` drops them, matching the gateway's own resume projection.
- **Legacy #86/#108 compaction filters** — untouched and still applied to every row lacking display metadata (ordering guarantees they can never discard authentic `display_content`).

## Model / UI

`ChatMessage` gains one optional field, `displayKind: String?`, for projection provenance; `SystemBubble` styles those rows as runtime notices. Everything else is reduced to final display role/content during normalization — no parallel message model.

## Out of scope (deliberately unchanged)

Rows persisted without display metadata (e.g. the forensically-tracked updater handoff, `role=user, display_kind=NULL`) remain as-is: without trustworthy metadata they are indistinguishable from genuine user pastes, and prose heuristics are out of scope. Those need the upstream tagging fix.

## Tests

- 26 new `MessageNormalizerTests` covering every contract rule plus precedence/edge pins (hidden × role × display_content, carrier-vs-projection ordering, compaction-flag vs projection, JSON-string/malformed metadata, unknown kinds, tool-row safety).
- `CompactResumeTranscriptTests.testCompactResumeAppliesDisplayProjectionToSyntheticRows` — realistic REST history payload driven through the actual `AppState.openSession` compact hydration: the exact reload path that exposed the bug.
- `HermesClientTests.testLegacyResumeAppliesDisplayProjectionToPersistedRows` — same contract through the real `session.resume` RPC path (REST/legacy parity).
- Full suite green: **1,366 ConduitTests + 6 ConduitUITests, 0 failures**. Required suites: MessageNormalizerTests 96 (26 new), CompactResumeTranscriptTests 13 (+1), HermesClientTests 11 (+1), AppStateChatResumeTests 83, SessionPresentationCacheTests 43.

## Review

Multi-model review cycle (3 reviewers) run before push. First-round findings — projected-text prefix guard, review-activity carrier extraction, model_switch projection precedence, non-positive `task_count` — were fixed and re-reviewed; the final round returned unanimous APPROVE, and its remaining suggestions are applied in this commit.

A follow-up hardening pass (a79c24b) tightened three edge semantics per external review: `task_count` converts via `doubleValue` + `Int(exactly:)` with a `> 0` requirement (1e100/NaN/fractional degrade to the generic notice, never trap); `display_content` presence is the authority boundary (explicit null/bool/number resolve to authoritative empty — the physical carrier can never resurface); and all five synthetic timeline kinds render an explicit empty projection verbatim instead of trimming it back into absence. The 3-reviewer round on that pass caught one blocker (the internal_notification arm still substituted a canned label for an empty projection), which was fixed as prescribed and confirmation-approved.

## Summary by CodeRabbit

* **New Features**
  * Improved transcript rendering using persisted display projections.
  * Hidden timeline entries are now omitted automatically.
  * Synthetic events, such as resumed turns, model changes, and delegation completion, appear as clear system timeline notices.
  * Displayed content now takes precedence over underlying stored payloads when available.
  * Runtime notices receive consistent system-message presentation.

* **Bug Fixes**
  * Improved handling of malformed, unknown, and non-text display content while preserving usable transcript data.

### DIFF
--- a/Conduit/Models/Models.swift
+++ b/Conduit/Models/Models.swift
@@ -171,6 +171,12 @@ struct ChatMessage: Identifiable, Equatable {
     var approval: ApprovalActivity?
     var attachments: [Attachment]?
     var code: String?
+    /// Canonical Hermes `display_kind` when the persisted row arrived with
+    /// explicit synthetic-timeline projection (`model_switch`, `auto_continue`,
+    /// …). Nil for ordinary rows. Normalization already reduced the row to its
+    /// final display role/content; this only lets the timeline UI style the
+    /// notice without re-deriving it from text.
+    let displayKind: String?
 
     // Non-codable because it contains closures in some uses; serialization
     // is handled by the gateway, not by us. We construct these from RPC results.
@@ -188,7 +194,8 @@ struct ChatMessage: Identifiable, Equatable {
         clarify: ClarifyActivity? = nil,
         approval: ApprovalActivity? = nil,
         attachments: [Attachment]? = nil,
-        code: String? = nil
+        code: String? = nil,
+        displayKind: String? = nil
     ) {
         self.id = id
         self.role = role
@@ -203,6 +210,7 @@ struct ChatMessage: Identifiable, Equatable {
         self.approval = approval
         self.attachments = attachments
         self.code = code
+        self.displayKind = displayKind
     }
 }
 
--- a/Conduit/Services/HermesClient.swift
+++ b/Conduit/Services/HermesClient.swift
@@ -1700,20 +1700,16 @@ enum MessageNormalizer {
     }
 
     /// `display_content` is the authoritative visible payload whenever the
-    /// field exists — including an explicit empty string, which must never
-    /// fall back to the physical carrier. A JSON null is treated as absent:
-    /// no current Hermes write path emits one, and a broken projection should
-    /// not silently hide a genuine message. Non-textual scalars (a stray
-    /// bool/number) are malformed rather than an intentional empty
-    /// projection, so they degrade to the physical carrier too.
+    /// field exists — including an explicit empty string, JSON null, or an
+    /// odd scalar, all of which resolve to authoritative empty text rather
+    /// than a fallback. Field presence is the authority boundary: once the
+    /// field exists, the physical carrier must never resurface, because a
+    /// malformed projection falling back to physical content could expose
+    /// model-facing internal rows. Nil return therefore means exactly
+    /// "field absent".
     private static func projectedDisplayContent(in obj: [String: AnyCodable]) -> String? {
         guard let value = obj["display_content"] else { return nil }
-        switch value {
-        case .string, .array, .object:
-            return extractContent(value)
-        default:
-            return nil
-        }
+        return extractContent(value)
     }
 
     /// `display_metadata` may arrive as a JSON object (current servers) or as
@@ -1732,9 +1728,13 @@ enum MessageNormalizer {
     }
 
     /// Upstream persists `task_count` on `async_delegation_complete` rows so
-    /// clients can phrase the completion notice.
+    /// clients can phrase the completion notice. The count goes through
+    /// `doubleValue` + `Int(exactly:)` so non-representable numbers (1e100,
+    /// NaN, fractional) degrade to the generic notice instead of trapping
+    /// normalization — malformed metadata must never fail hydration.
     private static func delegationCompleteNotice(metadata: AnyCodable?) -> String {
-        guard let taskCount = displayMetadataObject(metadata)?["task_count"]?.intValue,
+        guard let count = displayMetadataObject(metadata)?["task_count"]?.doubleValue,
+              let taskCount = Int(exactly: count),
               taskCount > 0 else {
             return "Background agent work finished"
         }
@@ -1753,12 +1753,11 @@ enum MessageNormalizer {
         systemNotice: String?,
         rawVisible: String
     ) -> String {
-        let projectedNotice: String? = {
-            guard let projected, !projected.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty else {
-                return nil
-            }
-            return projected
-        }()
+        // `projected` is non-nil only when the `display_content` field
+        // exists, and its value — including an explicit empty string — is
+        // authoritative: trimming an empty projection back into absence
+        // would fall through to canned labels or the physical carrier,
+        // which must never resurface.
         switch kind {
         case .hidden:
             // Defense in depth: hidden rows are dropped before content work.
@@ -1770,8 +1769,8 @@ enum MessageNormalizer {
             // model-change card detection and keeps that presentation. A
             // marker the card regex cannot parse is still model scaffold, so
             // it degrades to the canned label rather than rendering raw.
-            if let projectedNotice {
-                return projectedNotice
+            if let projected {
+                return projected
             }
             if let modelChange {
                 return "[Model has been changed to \(modelChange.provider)/\(modelChange.model)]"
@@ -1780,16 +1779,22 @@ enum MessageNormalizer {
         case .personalitySwitch:
             // The physical marker embeds the full persona prompt; surface a
             // canned label instead, like Hermes Desktop does.
-            return projectedNotice ?? "Personality changed"
+            return projected ?? "Personality changed"
         case .autoContinue:
-            return projectedNotice ?? "Resumed interrupted turn"
+            return projected ?? "Resumed interrupted turn"
         case .asyncDelegationComplete:
-            return projectedNotice ?? delegationCompleteNotice(metadata: metadata)
+            return projected ?? delegationCompleteNotice(metadata: metadata)
         case .internalNotification:
             // The row content is the operational notice itself (wake events,
-            // background completions); show it as a system notice, with a
-            // `[System: …]` wrapper stripped when present.
-            let visible = projectedNotice ?? systemNotice ?? rawVisible
+            // background completions); show it as a system notice. An
+            // explicit projection — including an empty one — is authoritative
+            // and rendered verbatim; the canned label is only for rows that
+            // have no projection at all, and the `[System: …]` strip applies
+            // only to that no-projection fallback.
+            if let projected {
+                return projected
+            }
+            let visible = systemNotice ?? rawVisible
             return visible.isEmpty ? "System notification" : visible
         }
     }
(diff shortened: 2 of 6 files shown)


kaishi00:
## Summary
This PR makes Conduit's message normalization honor Hermes' persisted `display_kind`/`display_content`/`display_metadata` contract so synthetic rows (model switches, auto-continues, hidden scaffolding, delegation completions, internal notices) never leak as human `user` bubbles after a session reload. The handling lives at the shared `MessageNormalizer` boundary so both the REST history hydration (`AppState.openSession` compact path) and the legacy `session.resume` transcript enforce the same rules.

## Findings

### [minor] Case-sensitive `display_kind` matching defeats the fix on casing drift
**File:** Conduit/Services/HermesClient.swift:1503
**Issue:** `HermesDisplayKind(rawValue:)` matches case-sensitively. A hidden row carrying `"Hidden"`, `"HIDDEN"`, or any other casing is treated as an *unknown* kind, so it survives as an ordinary row and its multi-megabyte model-facing scaffold can resurface as a user bubble — the exact leak this PR exists to fix. The test `testDisplayKindMatchingToleratesSurroundingWhitespaceOnly` deliberately pins this behavior as "conservative," but it is the riskiest failure mode in the change (hiding is destructive, and a glaring wrong-case hidden row silently de-remediates).
**Suggestion:** Match case-insensitively (lowercased input before `rawValue:`) for the synthetic kinds, or at minimum for `.hidden`, and update the pinning test to document that intent.

### [minor] `display_content` object/array projections produce odd visible text
**File:** Conduit/Services/HermesClient.swift:1712
**Issue:** `projectedDisplayContent` treats `.object`/`.array` as authoritative but funnels them through `extractContent`, which pretty-prints an object lacking `text`/`content` keys (so `{}` becomes the literal string `"{}"`) and joins array members with no separator (so `["a","b"]` becomes `"ab"`). This is inconsistent with the documented rule that only textual projections are authoritative, and the doc comment ("JSON null and non-textual scalars are treated as absent/malformed") doesn't cover empty-object/array shapes.
**Suggestion:** Restrict authoritative projections to `.string` (and possibly monomorphic arrays of strings with a separator), and treat `.object` without a `text`/`content` member and `.array` as malformed → physical carrier.

### [minor] `task_count` as a JSON string fails to phrase the delegation notice
**File:** Conduit/Services/HermesClient.swift:1737
**Issue:** `delegationCompleteNotice` reads `task_count` through `.intValue`, which only returns a value for `.number`. A JSON-string count (e.g. `"2"`) degrades to the generic "Background agent work finished" label instead of "2 background agents finished". The JSON-text metadata path is already supported here, so a numeric-string `task_count` from an older backend is plausible.
**Suggestion:** Also decode `.stringValue.flatMap(Int.init)` before falling back to the generic label.

### [minor] Dead `.hidden` branch in `timelineNoticeContent`
**File:** Conduit/Services/HermesClient.swift:1763
**Issue:** `case .hidden` returns `"Hidden system event"`, but `.hidden` rows are dropped via `continue` at HermesClient.swift:1507 before any content work, so this branch is unreachable.
**Suggestion:** Either remove the branch or, if kept as defense-in-depth, make it `fatalError`-free but reachable-only-on-invariant-break — a comment clarifying intent is enough.

### [nit] Legacy-parity test doesn't exercise real gateway filtering
**File:** ConduitTests/HermesClientTests.swift:308
**Issue:** The PR body says the gateway's own `session.resume` projection already filters these rows; the test feeds *unfiltered* rows with `display_kind` through `openSessionLegacy`. The test is valid defense-in-depth but doesn't verify the claimed REST/legacy parity against the real gateway behavior.
**Suggestion:** Keep the test, but consider framing it as "conduit-side defense in depth" in the comment rather than parity verification.

## Verdict
APPROVE

The logic is sound and carefully ordered (hidden check precedes content scanning; projections bypass the legacy compaction filters; unknown kinds are preserved; tool rows keep legacy handling; timeline rows skip review/carrier extraction). No retain cycles, force unwraps, or concurrency issues were introduced — normalization stays a pure function with no shared mutable state. The three minor findings above are edge-case robustness gaps, not correctness blockers.

<a href="https://opencode.ai/s/niDasR3d">New%20session%20-%202026-08-30T14%3A14%3A30.693Z</a>
[opencode session](https://opencode.ai/s/niDasR3d)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33316266867)

kaishi00:
## Summary
This PR makes Conduit's message normalization honor Hermes' persisted `display_kind`/`display_content`/`display_metadata` contract so synthetic rows (model switches, auto-continues, hidden scaffolding, delegation completions, internal notices) never leak as human `user` bubbles after a session reload. The handling lives at the shared `MessageNormalizer` boundary so both REST history hydration and the legacy `session.resume` transcript enforce the same rules.

## Findings

### [minor] Case-sensitive `display_kind` matching defeats the fix on casing drift
**File:** Conduit/Services/HermesClient.swift:1503
**Issue:** `HermesDisplayKind(rawValue:)` (enum at HermesClient.swift:1260) matches case-sensitively after whitespace trimming. A hidden row carrying `"Hidden"`, `"HIDDEN"`, or any other casing is treated as an *unknown* kind, so it survives as an ordinary row and its multi-megabyte model-facing scaffold can resurface as a user bubble — the exact leak this PR exists to fix. The test `testDisplayKindMatchingToleratesSurroundingWhitespaceOnly` deliberately pins only whitespace tolerance, but casing drift is the riskiest failure mode in the change (hiding is destructive, and a wrong-case hidden row silently de-remediates).
**Suggestion:** Match case-insensitively (lowercased input before `rawValue:`), at minimum for `.hidden`, and update the pinning test to document that intent.

### [minor] `task_count` as a JSON string fails to phrase the delegation notice
**File:** Conduit/Services/HermesClient.swift:1736
**Issue:** `delegationCompleteNotice` reads `task_count` via `doubleValue`, which only returns a value for `.number`. A JSON-string count (e.g. `"2"`) degrades to the generic "Background agent work finished" label instead of "2 background agents finished". The JSON-text metadata path is already supported here, so a numeric-string `task_count` from an older/multi-encoding backend is plausible.
**Suggestion:** Also decode `stringValue.flatMap(Int.init)` before falling back to the generic label.

### [nit] Dead `.hidden` branch in `timelineNoticeContent`
**File:** Conduit/Services/HermesClient.swift:1762
**Issue:** `case .hidden` returns `"Hidden system event"`, but `.hidden` rows are dropped via `continue` at HermesClient.swift:1507 before any content work, so this branch is unreachable. It behaves inconsistently with the outer drop contract if the invariant ever breaks (it renders text instead of dropping).
**Suggestion:** Acceptable as defense-in-depth, but the surrounding comment already documents this; consider tightening the comment to state the desired behavior on invariant break (drop, not render).

### [nit] Object/array `display_content` projections can produce litter text
**File:** Conduit/Services/HermesClient.swift:1710 (via `extractContent` at 2151)
**Issue:** `projectedDisplayContent` treats `.object`/`.array` as authoritative but funnels them through `extractContent`, which pretty-prints an object lacking `text`/`content` keys (so `{}` becomes the literal string `"{}"`) and joins array members with no separator (so `["a","b"]` becomes `"ab"`). This is inconsistent with the doc comment ("JSON null and non-textual scalars are treated as absent/malformed"), which doesn't cover empty-object/array shapes surfacing odd visible text.
**Suggestion:** Restrict authoritative projections to `.string` (and possibly monomorphic arrays of strings with a separator) and treat odd `.object`/`.array` shapes as malformed → physical carrier, or explicitly document the "presence is authority including object/array" intent in the docstring.

## Verdict
APPROVE

The logic is sound and carefully ordered: hidden rows drop before any content scanning, projections bypass the legacy compaction filters (ordering guarantees authentic `display_content` can never be discarded), unknown kinds are preserved, and tool rows keep legacy handling while timeline rows skip review/carrier extraction. The `Int(exactly:)` + `> 0` conversion cleanly closes the `1e100`/fractional trap, and empty projections are preserved verbatim across all five synthetic kinds (the previously flagged `internal_notification` blocker is correctly fixed). No retain cycles, force unwraps, or concurrency issues were introduced — normalization stays a pure static function with no shared mutable state. The findings above are edge-case robustness gaps, not correctness blockers.

<a href="https://opencode.ai/s/dc3ayJIA">New%20session%20-%202026-08-30T15%3A17%3A52.812Z</a>
[opencode session](https://opencode.ai/s/dc3ayJIA)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33319188694)

PR #113, opened 2026-08-30
fix(auth): finalize transactional native session cookies

kaishi00:
## Summary

First-party replacement for #109 (supersedes it). The contributor's implementation from @TechtronicsGroup / PR #109 was **adopted, not rewritten**: commit 1 of this branch is a straight cherry-pick of PR #109's head (`ca2fd22`) with author attribution preserved. Commits 2–3 finish the remaining correctness/hardening work on top of current `main`.

**The transactional `commitCookies()` architecture from PR #109 is preserved exactly** — no shared-cookie-jar-driven authentication, no automatic URLSession cookie attachment, no cross-origin cookie forwarding, and failed/cancelled/invalid/superseded transactions still publish nothing. The reconnect generation check before cookie commit is untouched.

## The blocker this fixes: path-prefixed base URLs

`SecureRedirectDelegate` detected the password-login task with a hard-coded absolute-path comparison:

```swift
task.originalRequest?.url?.path == "/auth/password-login"
```

Conduit supports base URLs with path prefixes (`https://example.com/hermes`), which produce `/hermes/auth/password-login` and `/hermes/api/auth/ws-ticket`. For those deployments the comparison failed: the redirect was still *followed*, but because URLSession automatic cookie handling is deliberately disabled in this design, the session cookie captured from the login response was never attached to the redirect landing — recreating exactly the login/logout loop this PR fixes.

The delegate now receives the **exact expected password-login endpoint derived from the normalized configured base URL** (`normalizedBaseURL + /auth/password-login`) when `NativeAuthClient` is constructed, and compares the task's original request against that endpoint (origin + path). No `hasSuffix` heuristics. All prior redirect security is intact: allowed transports only, same-origin only, no cross-origin cookie forwarding, only the configured password-login flow receives task-local login cookies, other request types keep their original explicit headers.

Covered by a new real `NWListener` + production `URLSession` test with a prefixed base URL: `http://127.0.0.1:<port>/hermes` → `POST /hermes/auth/password-login` (302 + Set-Cookie) → `/hermes/auth/password-complete` (receives the cookie despite disabled automatic handling) → `/hermes/api/auth/ws-ticket` (receives exactly the current transaction's cookie, not a pre-seeded stale shared one) → ticket returned; nothing is published before `commitCookies()`, and committing supersedes the stale shared cookie. The existing root-mounted redirect test remains.

## Additional hardening

- **Cancellation taxonomy**: cancelling a native-auth operation cancels the underlying `URLSessionTask`, which reports `URLError(.cancelled)`; `perform(_:)` now maps that back to `CancellationError()` so callers' existing `catch is CancellationError` handling stays silent (login cancellation must not render a login failure). Unrelated URL errors pass through untouched. `AppState.restoreSavedCredentials` also now treats `CancellationError` as a silent fallback to the login screen.
- **Expired parsed cookies**: an expired `Set-Cookie` is the server retiring that cookie identity, so `NativeAuthCookiePolicy.persist` deletes the stored cookie with the same canonical identity (name + canonical domain + path) from the shared jar — via the actual stored cookie objects — and never inserts the expired parse itself. Deletion is scoped to the exact identity: same-name cookies under other paths and unrelated cookies survive. Pinned end-to-end (a ticket response that retires `hermes_session_at` removes the stale session at commit, publishes nothing itself, and leaves other cookies intact) plus a direct unit test for canonical (dotted-domain) matching and normal non-expired rotation.
- **Duplicate `Set-Cookie` parsing**: `acceptedCookies(from:)` collects `Set-Cookie` values case-insensitively and parses each header value individually, so array-style duplicate header values are handled and comma-joining can never corrupt `Expires` dates. Foundation's `HTTPCookie` parser remains the authority for cookie syntax; response order is preserved so later rotations win through `merged(...)`. Covered by unit tests for both the array-style and comma-joined representations plus the existing real-transport duplicate test.
- **Explicit storage isolation**: the auth session now uses `httpCookieAcceptPolicy = .never` plus an inert private cookie store (previously `.always` + shared, which relied on undocumented CFNetwork coupling treating `httpShouldSetCookies = false` as disabling storage too). Response cookies can therefore never be auto-published by URLSession; the shared jar is written only by `NativeAuthConnection.commitCookies()`.
- **Diagnosability**: `connect()` fails with a descriptive ticket error when a login response yields no host-scoped cookie (exact-host acceptance intentionally drops domain-scoped/foreign cookies), instead of an opaque ticket 401 downstream.
- **Dead code**: removed the redundant `guard let ticketURL = request.url else { throw .invalidURL }` — request construction already guarantees a URL or throws — via a shared `endpointURL` helper.

## Cookie path-boundary semantics (review finding, corrected)

A review comment suggested `Path=/api` should *not* match `/api/auth/ws-ticket`; that is incorrect — under RFC 6265 §5.1.4 path matching, `/api` **is** a legitimate parent and must match. The boundary the custom matcher actually protects is: cookie `Path=/api/auth/ws` vs request `/api/auth/ws-ticket` — a naive `hasPrefix` would accept it, but the character after the cookie path is `-`, so it must be (and is) rejected. Both cases are pinned by real-transport tests, alongside a positive parent-path case.

## Review finding investigated, documented as not reproduced

The claim that PR #109 changed same-origin redirect semantics in `authProviders()`: pre-PR `main` already used a `URLSession` with `SecureRedirectDelegate` that followed allowed same-origin redirects, so intermediate 3xx responses were consumed before `authProviders()` saw them; the 3xx→`[]` fallback branch was reachable only for *blocked* (cross-origin) redirects. The adopted PR's delegate behaves identically for non-password-login tasks (`completionHandler(request)`), and the password-login branch does not apply to `/api/auth/providers`. A new real-transport test pins the unchanged semantics: a same-origin redirect on `/api/auth/providers` is followed and the final landing is parsed (the pre-existing cross-origin-fallback test still covers the blocked case). **No behavior change was required or made.**

## Non-goals honored

`NativeAuthConnection`, transactional `commitCookies()`, `DashboardTicketBridge`, `ConnectionURLPolicy`, credential storage, Face ID behavior, Hermes protocol behavior, and WebView login behavior are unchanged; no custom cookie parser was added; test isolation continues to use the shared cookie jar with disjoint fixture domains scrubbed in setUp/tearDown.

## Verification

Run on the Mac build host (Xcode 26.6, iPhone 17 Pro simulator) against this branch — not old PR #109 CI:

- Focused: `ConduitTests/NativeAuthClientTests` (13 tests) + `ConduitTests/NativeAuthClientIntegrationTests` (10 tests) — **23/23 passed, 0 failures**, covering: root-mounted password login, same-origin login redirect, path-prefixed `/hermes/...` login redirect, stale shared cookie cannot authenticate, concurrent login isolation, invalid ticket publishes no cookies, ticket-response cookie rotation and expired-identity deletion at commit, duplicate `Set-Cookie` (combined-string and array representations), Secure cookie rejection over HTTP, foreign-domain cookie rejection, expiry handling, cookie path-boundary semantics, provider-discovery same-origin redirect, and cancellation → `CancellationError`.
- Full `xcodebuild test`: **ConduitTests 1387/1387 passed, 0 failures.** The `ConduitUITests` runner failed to *initialize* on the local Mac simulator ("Timed out waiting for AX loaded notification") on repeated attempts — a simulator AX infrastructure stall during runner bootstrap, before any UI test executed; this branch changes no UI code, so it is reported as an infrastructure stall, not a failure. The CI UI lane (with its retry/recovery runner scripts) covers it on GitHub runners.
- `git diff --check` clean; no `project.yml`/test-membership changes (no new files), so regeneration is a no-op.

Once this PR is green and reviewed, #109 can be closed as superseded. Credit to @TechtronicsGroup / PR #109 for the original transactional cookie architecture.

🤖 Generated with [ZCode](https://zcode.ai)

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2180,11 +2180,12 @@ final class AppState: ObservableObject {
         }
 
         do {
-            let ticket = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
+            let authenticatedConnection = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
                 username: credentials.username,
                 password: credentials.password
             )
-            await connect(with: HermesConnection(baseUrl: credentials.baseURL, ticket: ticket), profile: activeProfile)
+            authenticatedConnection.commitCookies()
+            await connect(with: HermesConnection(baseUrl: credentials.baseURL, ticket: authenticatedConnection.ticket), profile: activeProfile)
         } catch {
             // A rejected saved password falls back to the native login screen
             // without erasing it, allowing the user to correct the account.
@@ -3926,16 +3927,17 @@ final class AppState: ObservableObject {
                 if let credentials = KeychainHelper.loadCredentials(),
                    credentials.baseURL.trimmingCharacters(in: CharacterSet(charactersIn: "/")) == savedConnection.baseUrl.trimmingCharacters(in: CharacterSet(charactersIn: "/")) {
                     do {
-                        let ticket = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
+                        let authenticatedConnection = try await NativeAuthClient(baseURL: credentials.baseURL, cloudflareAccess: KeychainHelper.loadCloudflareAccess(for: credentials.baseURL)).connect(
                             username: credentials.username,
                             password: credentials.password
                         )
                         guard refreshTransportContinuation() else { return }
+                        authenticatedConnection.commitCookies()
                         // URLSession and WebKit have separate cookie stores.
                         // Reload the bridge so it receives the fresh session.
                         dashboardTicketBridge?.reload()
                         await connect(
-                            with: HermesConnection(baseUrl: credentials.baseURL, ticket: ticket),
+                            with: HermesConnection(baseUrl: credentials.baseURL, ticket: authenticatedConnection.ticket),
                             profile: activeProfile,
                             syncPurpose: continuationPurpose,
                             cancelsResumeRestoration: false,
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2186,6 +2186,10 @@ final class AppState: ObservableObject {
             )
             authenticatedConnection.commitCookies()
             await connect(with: HermesConnection(baseUrl: credentials.baseURL, ticket: authenticatedConnection.ticket), profile: activeProfile)
+        } catch is CancellationError {
+            // Intentional cancellation (e.g. a superseded connect) is not a
+            // login failure; fall back to the login screen silently.
+            showLogin = true
         } catch {
             // A rejected saved password falls back to the native login screen
             // without erasing it, allowing the user to correct the account.
--- a/Conduit/Services/NativeAuthClient.swift
+++ b/Conduit/Services/NativeAuthClient.swift
@@ -36,13 +36,29 @@ enum AuthClientError: LocalizedError {
     }
 }
 
-/// URLSession-based Hermes dashboard authentication. Its session uses the
-/// shared cookie storage so the authenticated cookie can be copied into
-/// DashboardTicketBridge's WebKit cookie store after sign-in.
+/// A fully validated native authentication transaction. Cookies remain private
+/// to the transaction until its caller accepts the ticket and commits them.
+struct NativeAuthConnection {
+    let ticket: String
+    fileprivate let cookies: [HTTPCookie]
+
+    /// Publishes this successful transaction for DashboardTicketBridge/WebKit.
+    /// Calling this more than once is harmless; callers should commit only the
+    /// connection they are about to make active.
+    func commitCookies() {
+        NativeAuthCookiePolicy.persist(cookies)
+    }
+}
+
+/// URLSession-based Hermes dashboard authentication. Automatic URLSession
+/// cookie handling is disabled: each login transaction captures its own
+/// response cookies, scopes them to the exact ticket URL, and returns them for
+/// explicit commit only after a valid ticket has been received.
 struct NativeAuthClient {
     let baseURL: String
     let cloudflareAccess: CloudflareAccessCredentials?
     private let session: URLSession
+    private let redirectDelegate: SecureRedirectDelegate
 
     init(
         baseURL: String,
@@ -55,14 +71,19 @@ struct NativeAuthClient {
         let configuration = sessionConfiguration ?? URLSessionConfiguration.default
         configuration.httpCookieAcceptPolicy = .always
         configuration.httpCookieStorage = HTTPCookieStorage.shared
-        configuration.httpShouldSetCookies = true
-        self.session = URLSession(configuration: configuration, delegate: SecureRedirectDelegate(), delegateQueue: nil)
+        // Every Cookie header is built from this login's captured cookies.
+        // Automatic handling would let a stale shared-jar cookie authenticate
+        // when the current login returned no cookie applicable to the route.
+        configuration.httpShouldSetCookies = false
+        let redirectDelegate = SecureRedirectDelegate()
+        self.redirectDelegate = redirectDelegate
+        self.session = URLSession(configuration: configuration, delegate: redirectDelegate, delegateQueue: nil)
     }
 
     func authProviders() async throws -> [[String: Any]] {
         let request = try request(path: "/api/auth/providers")
-        let (data, response) = try await session.data(for: request)
-        guard let http = response as? HTTPURLResponse else {
+        let result = try await perform(request)
+        guard let http = result.response as? HTTPURLResponse else {
             throw AuthClientError.providerDiscoveryFailed("No response")
         }
         switch http.statusCode {
@@ -72,16 +93,16 @@ struct NativeAuthClient {
             break
         }
         guard (200...299).contains(http.statusCode) else {
-            throw AuthClientError.providerDiscoveryFailed(parseError(data) ?? "HTTP \(http.statusCode)")
+            throw AuthClientError.providerDiscoveryFailed(parseError(result.data) ?? "HTTP \(http.statusCode)")
         }
 
-        if let json = try? JSONSerialization.jsonObject(with: data) as? [String: Any] {
+        if let json = try? JSONSerialization.jsonObject(with: result.data) as? [String: Any] {
             return json["providers"] as? [[String: Any]] ?? []
         }
         return []
     }
 
-    func login(username: String, password: String) async throws {
+    func login(username: String, password: String) async throws -> [HTTPCookie] {
         var request = try request(path: "/auth/password-login")
         request.httpMethod = "POST"
         request.setValue("application/json", forHTTPHeaderField: "Content-Type")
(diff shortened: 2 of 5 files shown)


kaishi00:
## Summary

Adopts PR #109's transactional native-auth cookie capture and finalizes it: login and ticket responses are parsed transaction-locally (URLSession auto-cookie handling fully disabled), cookies are committed to the shared jar only via `NativeAuthConnection.commitCookies()` after a valid ticket, `SecureRedirectDelegate` now derives the password-login endpoint from the normalized base URL (fixing path-prefixed deployments), and cancellation is remapped to `CancellationError`. Extensive real-transport and unit tests added.

## Findings

### [minor] Cookies are committed before the connection is confirmed active
**File:** Conduit/Views/LoginView.swift:253, Conduit/Services/AppState.swift:2187,3939
**Issue:** `commitCookies()` runs before the wrapped `appState.connect(with:)`/`HermesClient` connect actually succeeds. If the enclosing `Task` is cancelled or the subsequent connect fails after the commit, a valid session cookie has already been published into the shared jar with no active Hermes connection consuming it. The previous anonymous "commit succeeded, nothing used it" flow at least discarded cookies, so this is a small regression path (stale-but-valid cookie left for the WebKit bridge on a failed/cancelled connect).
**Suggestion:** Commit after the connection is established — pass the `NativeAuthConnection` into `appState.connect(...)` and have the connection path commit it inside the success block, or defer `commitCookies()` until after `connect(with:)` returns.

### [minor] Non-Sendable state crosses concurrency boundaries (Swift 6 debt)
**File:** Conduit/Services/NativeAuthClient.swift:41,386,392
**Issue:** `NativeAuthConnection` and `NativeAuthHTTPResult` wrap `[HTTPCookie]` (non-Sendable) and are returned from async functions; the `dataTask` completion handler on a background queue captures the `CheckedContinuation` and calls into the non-Sendable `redirectDelegate`. `URLSessionTaskHolder` is `@unchecked Sendable`. This compiles under SWIFT_VERSION 5.9, but under complete Swift 6 strict concurrency this file will fail to compile (or produce hard errors) without `@unchecked` / isolation work.
**Suggestion:** Fine to merge in Swift 5.9 mode, but add a tracking issue and plan to isolate `HTTPCookie` access (e.g. `@Sendable` wrappers or moving parse/merge onto the client's own serial queue) before the project's Swift 6 migration.

### [minor] Ticketed app candidates: cookies from a *succeeded-then-cancelled* transaction may still be committed
**File:** Conduit/Services/NativeAuthClient.swift:211-248, Conduit/Views/LoginView.swift:241-254
**Issue:** `withTaskCancellationHandler` only cancels the underlying `URLSessionTask`; it does not make the operation throw if the continuation has already resumed with a successful result. If the login completes and the caller's `Task` is cancelled just before/after, `connect` can return a valid `NativeAuthConnection` and `commitCookies()` still runs in a cancelled task, publishing the cookie. Callers should check `Task.isCancelled` before committing.
**Suggestion:** In `commitCookies()` callers (or inside `NativeAuthConnection.commitCookies` via a `Task`-context check) guard with `Task.isCancelled` before publishing, or have `perform(_:)` discard the result when the wrapping task is already cancelled.

### [minor] `Max-Age` cookies bypass the expiry check in `persist`/`applicableCookies`
**File:** Conduit/Services/NativeAuthClient.swift:320,353
**Issue:** `expiresDate` is `nil` for `Max-Age` cookies (Foundation converts Max-Age to an absolute expiry only when installing into a cookie storage; the parsed `HTTPCookie` here typically has no `expiresDate`). Such cookies are treated as non-expiring by both the pre-send filter and `persist` and are committed as if session-persistent; a short-lived rotation could then be sent from the shared jar far beyond its intended lifetime. The PR's tests only cover `Expires=`-style expiry.
**Suggestion:** If the cookie carries a `Max-Age` (or its `isSessionOnly` is false but `expiresDate` is nil), compute the relative expiry from parse time and apply the same skip/filter logic, or document this as an accepted limitation.

### [minor] Non-password-login redirects do not re-attach the explicit Cookie header
**File:** Conduit/Services/NativeAuthClient.swift:453-455,140-152
**Issue:** `isPasswordLoginTask` gates cookie re-attachment, so a same-origin redirect on `/api/auth/ws-ticket` (e.g. trailing-slash or HTTPS-upgrade) relies entirely on URLSession forwarding the manually-set `Cookie` header from the original request. If the destination path does not match the committed cookie's path (or CFNetwork strips it), the landing request 401s. This branch has no test coverage.
**Suggestion:** Consider attaching `headerFields(for: taskCookies, url: destination)` for the ws-ticket task too (it shares the same per-task captured/committed cookie set), and add a real-transport redirect test on the ticket endpoint.

### [nit] Mutates the caller-supplied `sessionConfiguration`
**File:** Conduit/Services/NativeAuthClient.swift:71-78
**Issue:** `init` unconditionally mutates the passed-in configuration (`httpCookieAcceptPolicy`, `httpCookieStorage`, `httpShouldSetCookies`). A caller reusing one configuration object across multiple `NativeAuthClient`s (or other sessions) will have their configuration silently changed. Pre-existing pattern from before this PR, but worth noting.
**Suggestion:** Copy the supplied configuration (`sessionConfiguration?.copy()`) before mutating, or document that the initializer takes ownership and mutates it.

### [nit] Intentional behavior: strict exact-host cookie acceptance
**File:** Conduit/Services/NativeAuthClient.swift:325-343,180-187
**Issue:** `accepts` requires `canonicalDomain == responseHost`, so any legitimately domain-scoped session cookie (e.g. `Domain=.example.com` from a dashboard served on a subdomain) causes `connect()` to throw "Login succeeded but no host-scoped session cookie was accepted", with a user-facing message. This is a deliberate hardening trade-off, but it is a real compatibility constraint for some Starlette/FastAPI deployments that default to a `Domain` attribute.
**Suggestion:** Document this constraint in README/support notes so deployments know native password login requires host-only, exact-host cookies; the current behavior (fail with a descriptive message) is the right call over silently proceeding.

## Verdict

APPROVE

The transactional cookie design is sound, the concurrency story (per-task cookie buffering under a lock, task-holder cancellation) is correct, and all production `NativeAuthClient` call sites commit cookies. No force-unwraps introduced in production code, no retain cycles, and no SwiftUI/accessibility regressions (LoginView changes are limited to a rename and a `commitCookies()` call). Findings above are minor/hardening items, not blockers.

<a href="https://opencode.ai/s/1VwyEPkh">New%20session%20-%202026-08-30T16%3A26%3A15.036Z</a>
[opencode session](https://opencode.ai/s/1VwyEPkh)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33322395302)

kaishi00:
## Summary

This PR adopts PR #109's transactional native-auth cookie architecture, disables automatic URLSession cookie handling (`.never` + inert private store), commits cookies to the shared jar only via `NativeAuthConnection.commitCookies()` after a valid ws-ticket, fixes `SecureRedirectDelegate` endpoint detection for path-prefixed base URLs, remaps `URLError(.cancelled)` to `CancellationError`, adds expired-cookie deletion by canonical identity, and pins it all with 23 focused tests. The design is sound; the previous reviewer's findings were largely addressed (storage isolation, fail-fast on missing host cookie, remote redirect-cookie tests). Remaining items are minor/hardening.

## Findings

### [minor] Cookies are committed before the connection is confirmed active (unaddressed from prior review)
**File:** Conduit/Views/LoginView.swift:253, Conduit/Services/AppState.swift:2187,3939
**Issue:** `commitCookies()` runs before the wrapped `appState.connect(with:)`/`HermesConnection` connect succeeds. If `connect(with:)` bails early (e.g. the `automaticChatResumeWorkIsCurrent` guard at AppState.swift:1949) or a downstream connect fails, a valid session cookie is already published into the shared jar with no active Hermes connection and no cleanup, leaving it for `DashboardTicketBridge` to copy into WebKit on a later reload. This is a bounded regression window versus discarding on failure.
**Suggestion:** Defer `commitCookies()` until after the connection path reports success, or pass the `NativeAuthConnection` into `connect(...)` so the commit is part of the connect success path.

### [minor] A succeeded-then-cancelled transaction can still publish cookies (unaddressed from prior review)
**File:** Conduit/Services/NativeAuthClient.swift:211-248
**Issue:** `withTaskCancellationHandler` cancels the underlying `URLSessionTask`, but `withCheckedThrowingContinuation` only throws if the task reports `.cancelled`. If the login/ticket completes concurrently with the wrapping task's cancellation, `resume(returning:)` wins and `connect()` returns a value; callers then run `commitCookies()` inside a cancelled task. The `onCancel` handler also fires even if the operation already completed, per `withTaskCancellationHandler` semantics.
**Suggestion:** Guard the commit sites (or `NativeAuthConnection.commitCookies`) with `Task.isCancelled`, or have `perform(_:)` discard the result when the wrapping task is already cancelled.

### [minor] `Max-Age`-only cookies bypass the expiry filter
**File:** Conduit/Services/NativeAuthClient.swift:318,362
**Issue:** A `Set-Cookie` with `Max-Age` (no `Expires`) parsed by `HTTPCookie.cookies(withResponseHeaderFields:)` can have `expiresDate == nil` while `isSessionOnly == false`. Both the pre-send filter in `applicableCookies` and `persist` treat those as non-expiring, so a short-lived rotation is never skipped/filtered in the transaction layer (the shared jar ultimately applies Max-Age, so the persisted state is correct; only the transaction-local window is wrong). The PR's tests only cover `Expires=`-style expiry.
**Suggestion:** If `expiresDate` is nil but `isSessionOnly` is false, compute relative expiry from parse time and apply the same skip/filter logic, or document it as an accepted limitation.

### [minor] Non-password-login redirects don't re-attach the explicit Cookie header
**File:** Conduit/Services/NativeAuthClient.swift:462-465
**Issue:** `isPasswordLoginTask` gates cookie re-attachment, so a same-origin redirect on `/api/auth/ws-ticket` (trailing-slash, path normalization, or cached 308) relies entirely on CFNetwork forwarding the manually-set `Cookie` header from the original request. If the destination path doesn't match the committed cookie's `Path` (or the header is stripped), the landing 401s. No test covers this path.
**Suggestion:** Re-attach `headerFields(for: taskCookies, url: destination)` for the ticket task too (it shares the same per-task captured/committed cookie set) and add a real-transport redirect test on the ticket endpoint.

### [nit] `seedSharedCookie` fails silently, letting a test precondition vanish
**File:** ConduitTests/NativeAuthClientTests.swift:400-409
**Issue:** Unlike `seedLoopbackCookie` (NativeAuthClientIntegrationTests.swift:547-557), `seedSharedCookie` returns without `XCTFail` when `URL(string:)`/`HTTPCookie(properties:)` fail. `testTicketRequestDoesNotFallBackToUnrelatedSharedCookie` then passes trivially because "no Cookie header" holds even without the stale shared cookie.
**Suggestion:** Mirror `seedLoopbackCookie` and call `XCTFail` when the fixture cookie can't be created.

### [nit] `init` mutates the caller-supplied `sessionConfiguration`
**File:** Conduit/Services/NativeAuthClient.swift:71-78
**Issue:** The initializer unconditionally mutates `httpCookieAcceptPolicy`, `httpCookieStorage`, and `httpShouldSetCookies` on the passed-in configuration. Production uses `URLSessionConfiguration.default` (a fresh instance per access), so this is safe there; only callers that reuse a configuration object across sessions are affected. Pre-existing pattern.
**Suggestion:** `configuration.copy()` before mutating, or document that the initializer takes ownership.

### [minor] Swift 6 strict-concurrency debt (informational, not a blocker)
**File:** Conduit/Services/NativeAuthClient.swift:41-51,395-399,401-429
**Issue:** `NativeAuthConnection`/`NativeAuthHTTPResult` move non-Sendable `[HTTPCookie]` across async boundaries, and `URLSessionTaskHolder` and the delegate need `@unchecked Sendable` + `NSLock`. This compiles under Swift 5.9 but will fail under complete strict concurrency. Lock discipline is correct; this is future-readiness work, not a current defect.
**Suggestion:** Track as an issue; isolate `HTTPCookie` access (Sendable wrappers or a serial queue) before the Swift 6 migration.

## Verdict

APPROVE

No critical or major issues. No retain cycles, no force-unwraps in production code, no SwiftUI/accessibility regressions, and the lock-protected per-task cookie buffering is correct under concurrency. The findings above are hardening items (mostly carried over from the prior review and still unaddressed); the highest-value follow-ups are the commit-before-connect ordering and the cancelled-task commit guard.

<a href="https://opencode.ai/s/aD4rpi7v">New%20session%20-%202026-08-30T16%3A46%3A26.509Z</a>
[opencode session](https://opencode.ai/s/aD4rpi7v)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33323338525)

PR #115, opened 2026-08-30
fix(voice): align STT provider identity, gating, and timeout with Hermes upstream

kaishi00:
## Summary

Fixes the confirmed voice/STT parity bug where Conduit showed **OpenAI STT as "Need Auth" and disabled the microphone** even though the connected Hermes runtime would transcribe successfully on the selected direct-OpenAI route.

### Root cause

Conduit's provider-readiness parser collapsed the upstream picker rows `"Nous Subscription"` and `"OpenAI"` into a single `openai` identity:

```swift
case "nous subscription", "openai": return "openai"
```

then attached readiness with a first-match `readiness.first { $0.id == id && $0.kind == kind }`. Because Hermes lists the Nous Subscription row first, the direct-OpenAI selection inherited the managed row's `needs_auth` status, and since `supportsTranscription` required the selected row to be `ready`, the composer microphone was disabled — the Reddit-reported regression.

### Upstream contract (verified against upstream main, 2026-08-30)

- Current Hermes writes **`stt.provider = "nous"`** for the managed "Nous Subscription" selection (serviced by the OpenAI-compatible implementation through the managed gateway, strictly — a local `OPENAI_API_KEY` must not override it) and **`stt.provider = "openai"`** for the direct BYOK row. Older gateways express the managed route as `openai` + `use_gateway: true`.
- STT toolset rows (`GET /api/tools/toolsets/stt/config`) carry only a display `name` — no `stt_provider` field. TTS rows add `tts_provider`, but the managed TTS row's vendor field also says `openai`, so the managed flags must decide identity.
- Upstream Desktop gates dictation on `stt.enabled != false` alone; provider/auth failures surface from the real transcription attempt. Readiness metadata is diagnostic.
- Desktop's transcription request timeout is `min(600_000ms, max(180_000ms, ceil(dataUrl.length × 0.1)))`.

## Changes

1. **Distinct provider identity** (`HermesVoiceConfigurationService`): readiness rows normalize to the provider IDs Hermes itself writes (`local`, `nous`, `openai`, `groq`, `xai`, `elevenlabs`, `deepinfra`), with managed flags outranking vendor fields. Both rows stay visible and correctly labeled in Settings ("Nous Subscription" vs "OpenAI"); Mistral STT is deliberately **not** re-exposed (upstream package quarantine).
2. **Readiness is diagnostic, not an authorization gate**: `supportsTranscription = sttEndpointAvailable && stt.enabled`, matching upstream Desktop. The mic is never disabled by `needs_auth`/`needs_keys`/`unknown` rows; real provider/auth errors surface from the actual `/api/audio/transcribe` call. `stt.enabled: false` remains authoritative. TTS readiness gating is unchanged.
3. **Gateway-owned provider selection**: row-backed selections are submitted by display name to `PUT /api/tools/toolsets/{stt,tts}/provider` (the upstream Desktop client contract, present in both gateway generations), so the server owns the row→config mapping. Managed "Nous Subscription" rows fail loud instead of ever writing `nous` raw; vendor rows fall back to the legacy direct config write. A `needs_nous_auth` response is surfaced as a notice, never a silent failure.
4. **Transcription timeout**: ported Desktop's policy exactly — `min(600s, max(180s, ceil(0.1ms per data-URL character)))` — replacing the fixed 90s ceiling that truncated legitimate remote-provider transcriptions. Cancellation is preserved.

## Deliberately out of scope

- **`GET /api/audio/voice-config` not consumed.** It resolves the runtime route (including credentials) for *client-direct* voice. Conduit always uses the relay endpoints, and `voice-config`'s own contract makes relay the floor, not an error — so its verdict adds nothing to capability. Parsing a credentials-bearing payload for diagnostics-only gains would expand the key-handling surface Conduit deliberately avoids (no credential persistence, no key logging). Absence on older gateways is also irrelevant since the endpoint is not required for the fix.
- **No TTS behavior change**: TTS readiness gating, providers, and the speak path are untouched (the split TTS rows now only carry correct, distinct diagnostics).
- **No voice UI redesign, no state-machine changes, no profile/session changes.**

## Tests

22 new focused tests (suite grew 1387 → 1409, all passing):

- **Reddit reproduction**: `testNousSubscriptionAndDirectOpenAIRemainDistinctProviders` / `testDirectOpenAIIsNotDisabledBecauseNousSubscriptionNeedsAuth` — exact fixture from the report; OpenAI must receive its own `ready` row and transcription stays available.
- `testNeedsKeysReadinessStaysDiagnosticAndDoesNotDisableTranscription`, `testDisabledSTTConfigStillDisablesHermesTranscription` (the `stt.enabled` authority is preserved).
- `testCurrentUpstreamPickerCatalogMapsToCanonicalProviderIDs` (full current catalog + managed-flag precedence over the vendor field, STT and TTS), `testManagedNousFieldEditorsTargetVendorConfigKeys` (managed route shares `stt.openai.*`; ElevenLabs uses `model_id`).
- Save-path contracts via mock requester: row-name submission to the toolset provider endpoint (STT + TTS), schema-only legacy fallback, managed fail-loud without the endpoint, vendor-row fallback, `needs_nous_auth` notice.
- `HermesVoiceGatewayTimeoutTests`: upstream constants, floor/scale/cap, monotonicity, no-trap extremes.
- `AppStateVoiceCapabilityTests`: ready snapshot keeps `canStartVoiceConversation`, `stt.enabled=false` still blocks, Apple on-device mode ignores Hermes STT readiness (both directions of independence, including `permissionRequired` vs `permissionDenied`).

Also updated two pre-existing tests that asserted the removed readiness-gating behavior.

## Verification

- Focused voice suites (config service, gateway timeout, AppState capability, controller, audio session): **green**.
- Full `ConduitTests` on the Mac build host (iPhone 17 Pro simulator): **1409 executed, 0 failures**.
- `git diff --check` clean. No repo lint tooling exists to run.

## Review gate

Three-model review cycle (DeepSeek V4 Flash / Step 3.7 Flash / MiMo V2.5): **3× APPROVE, zero blockers**. Applied endorsed suggestions: vendor-row legacy fallback for endpoint failure, `error`-key check on the provider PUT response, TTS display-name parity, `utf8.count` for the payload length, dead-switch cleanup, and the additional save-path/timeout/Apple-mode tests.

## Known upstream inconsistency (outside Conduit's control)

On older gateways the managed route's stored config value is `openai` + gateway intent, so after picking "Nous Subscription" the selected-provider highlight reflects the stored `openai` value even though the Nous row reports active. The write itself is correct on both generations; the display is honest about what the config holds.

## Summary by CodeRabbit

* **New Features**
  * Improved voice provider setup with clearer provider names and expanded transcription provider options.
  * Added support for managed provider configurations and safer provider selection handling.
  * Transcription requests now allow more time for longer recordings, improving reliability.
* **Bug Fixes**
  * Voice availability now reflects actual transcription settings and live capability checks instead of provider readiness indicators alone.
  * Apple on-device transcription availability is evaluated independently using its own permissions and device support.
* **Tests**
  * Added coverage for voice capability gating, provider configuration, and transcription timeout behavior.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -8534,16 +8534,22 @@ final class AppState: ObservableObject {
     /// aloud / voice conversation mutual exclusion can be exercised with
     /// mock controllers. `readAloudGatewayBridge` is set to the same bridge
     /// so a mock read aloud gateway installed on the controller counts as
-    /// current and is not replaced by a real one at tap time.
+    /// current and is not replaced by a real one at tap time. Pass
+    /// `transcriptionMode` / `appleSpeechAvailability` to pin the route the
+    /// capability checks consult; nil leaves the current value.
     func installVoiceCapabilityStateForTesting(
         bridge: DashboardTicketBridge,
         snapshot: VoiceCapabilitySnapshot,
-        isVoiceEnabled: Bool
+        isVoiceEnabled: Bool,
+        transcriptionMode: VoiceTranscriptionMode? = nil,
+        appleSpeechAvailability: AppleSpeechRecognitionAvailability? = nil
     ) {
         dashboardTicketBridge = bridge
         readAloudGatewayBridge = bridge
         voiceCapabilitySnapshot = snapshot
         self.isVoiceEnabled = isVoiceEnabled
+        if let transcriptionMode { voiceTranscriptionMode = transcriptionMode }
+        if let appleSpeechAvailability { self.appleSpeechAvailability = appleSpeechAvailability }
     }
 
     private func loadVoiceProfilePreferences(profile: String) -> VoiceProfilePreferences {
--- a/Conduit/Voice/HermesVoiceConfigurationService.swift
+++ b/Conduit/Voice/HermesVoiceConfigurationService.swift
@@ -26,6 +26,12 @@ struct VoiceProviderReadiness: Equatable, Identifiable {
     /// The gateway's own picker row label. Provider selection is submitted
     /// under this name so Hermes — not Conduit — owns the row→config mapping.
     let displayName: String
+    /// Structured marker for the managed Nous route, set once at parse time
+    /// from upstream semantics (`managed_nous_feature` for the row's kind, or
+    /// the legacy managed-row label). Selection writes for these rows are
+    /// server-owned: when the toolset provider endpoint is unavailable they
+    /// must fail closed, never fall back to a raw config write of "nous".
+    let isManagedNous: Bool
     let requiredCredentials: [VoiceCredentialStatus]
 }
 
@@ -139,8 +145,10 @@ final class HermesVoiceConfigurationService: ObservableObject {
             sttReadiness: try? stt.get(),
             ttsReadiness: try? tts.get(),
             environment: try? environment.get(),
-            sttEndpointAvailable: (try? stt.get()) != nil,
-            ttsEndpointAvailable: (try? tts.get()) != nil
+            // STT toolset readiness is intentionally NOT passed to the
+            // parser: it cannot influence transcription capability, only the
+            // picker population and diagnostics built from its rows.
+            ttsToolsetConfigAvailable: (try? tts.get()) != nil
         )
         snapshot = parsed
         // Toolset config was added after some public gateways. Its absence is
@@ -160,23 +168,58 @@ final class HermesVoiceConfigurationService: ObservableObject {
     func saveProvider(_ provider: String, kind: VoiceProviderDescriptor.Kind) async -> Bool {
         let providers = kind == .stt ? snapshot.sttProviders : snapshot.ttsProviders
         guard let row = providers.first(where: { $0.descriptor.id == provider })?.readiness else {
-            // Schema-only provider (gateway without toolset readiness): the
-            // raw ID is a plain vendor config value, safe to write directly.
-            let saved = await save(value: provider, for: "\(kind.rawValue).provider")
-            if saved { await reload() }
-            return saved
+            // Schema-only provider (gateway without toolset readiness). The
+            // raw ID is a plain vendor config value, safe to write directly —
+            // except the canonical managed Nous ID, whose meaning is
+            // server-owned and never written raw.
+            guard provider.lowercased() != "nous" else {
+                errorMessage = "Hermes must translate this managed selection; its provider endpoint is unavailable."
+                return false
+            }
+            return await saveVendorSelection(provider, kind: kind)
         }
         if await saveProviderSelection(rowName: row.displayName, kind: kind) { return true }
-        // The toolset endpoint failed. Managed "Nous Subscription" must fall
-        // out loud — writing "nous" raw is only meaningful on gateways whose
-        // provider endpoint translates the row. Vendor rows carry IDs that
-        // ARE the config values upstream persists, so their legacy write
-        // stays safe.
-        guard row.displayName.lowercased() != "nous subscription" else { return false }
+        // The toolset endpoint failed. Managed Nous rows must fail closed:
+        // "nous" is only meaningful on gateways whose provider endpoint
+        // translates the row, so it is never raw-written through the legacy
+        // config fallback. Vendor rows carry IDs that ARE the config values
+        // upstream persists, so their legacy write stays safe.
+        guard !row.isManagedNous else { return false }
         errorMessage = nil
-        let saved = await save(value: provider, for: "\(kind.rawValue).provider")
-        if saved { await reload() }
-        return saved
+        return await saveVendorSelection(provider, kind: kind)
+    }
+
+    /// Legacy provider-selection fallback for gateways whose toolset provider
+    /// endpoint is unavailable. Writes the section atomically the way
+    /// upstream's single-provider selection model expects: the vendor value
+    /// lands in `<section>.provider` and any stale `use_gateway`
+    /// gateway-routing intent is removed — a leftover `use_gateway: true`
+    /// would override the user's fresh BYOK selection on legacy runtimes.
+    /// Like the other config editors, this is a full-document GET→PUT, so a
+    /// concurrent profile-config edit from another writer could be clobbered;
+    /// the single-writer UI flow makes that a deliberate, pre-existing
+    /// trade-off rather than a new one.
+    private func saveVendorSelection(_ vendor: String, kind: VoiceProviderDescriptor.Kind) async -> Bool {
+        let sectionKey = kind.rawValue
+        guard var config = try? await requester.requestJSON(path: profilePath("/api/config"), method: "GET", body: nil) else {
+            errorMessage = "Could not load voice settings to save this change."
+            return false
+        }
+        var section = config[sectionKey] as? [String: Any] ?? [:]
+        section["provider"] = vendor
+        section.removeValue(forKey: "use_gateway")
+        config[sectionKey] = section
+        do {
+            _ = try await requester.requestJSON(
+                path: profilePath("/api/config"), method: "PUT",
+                body: ["config": config]
+            )
+            await reload()
+            return true
+        } catch {
+            errorMessage = "Could not save \(sectionKey).provider: \(error.localizedDescription)"
+            return false
+        }
     }
 
     private func saveProviderSelection(rowName: String, kind: VoiceProviderDescriptor.Kind) async -> Bool {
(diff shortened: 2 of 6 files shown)


kaishi00:
## Summary
Fixes the voice/STT parity regression where Conduit collapsed Hermes' "Nous Subscription" and "OpenAI" picker rows into one identity, disabled the mic via a `needs_auth` first-match, and truncated remote transcriptions at 90s. The change makes provider identities distinct, treats toolset readiness as diagnostic (matching Hermes Desktop's `stt.enabled`-only gate), submits row selections through the gateway's provider endpoint, and ports Desktop's scaled timeout policy.

## Findings

### [minor] Fail-loud guard compares display name instead of the canonical provider id
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:175
**Issue:** After the toolset PUT fails, the "managed Nous must fail loud" decision is `row.displayName.lowercased() != "nous subscription"`. The row already carries the canonical id (`row.id == "nous"`), which `providerID(for:)` derives first from `managed_nous_feature` and only then from the name. If Hermes ever renames the row (e.g. "Nous", "Nous Plan", or any trailing/extra whitespace, since `displayName` is not trimmed like `name` is in `providerID`), the guard evaluates to true and `saveProvider` falls through to writing `stt.provider = "nous"` raw — precisely the older-gateway corruption this branch exists to prevent.
**Suggestion:** Guard on the row's identity instead: `guard row.id == "nous" else { return false }`. This keeps the fail-loud path consistent with the identity rule used during parsing.

### [minor] Real toolset error is silently discarded on the vendor-row fallback
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:176
**Issue:** When `saveProviderSelection` fails for a genuine reason (server `/error` payload or transport failure) and the row is a vendor row, `errorMessage = nil` wipes that diagnosis before the legacy direct write. The direct write masks the endpoint failure from the user even though the endpoint exists and reported a problem.
**Suggestion:** Only clear `errorMessage` if the legacy write fully succeeds, or surface a combined message when the fallback is taken (e.g. "Selected on this gateway, but the provider endpoint reported: …").

### [minor] Success is inferred from absence of an `error` string, ignoring other failure shapes
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:189
**Issue:** `saveProviderSelection` treats any non-throwing, non-`error`-string response as a successful selection. A response like `{"ok": false}` or `{"error": [..]}` (non-string) would be reported as saved. The mocks return `["ok": true]` but the code never consults `ok`.
**Suggestion:** Consider treating a literal `ok == false` (and non-string `error` values, e.g. via `error["message"]`) as failures so the caller doesn't report a saved selection that the gateway rejected.

### [minor] TTS readiness gate may now permanently block voice when the managed route is active on older gateways
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:320 (and `selectedProviderIsReady` at 586)
**Issue:** STT gating dropped readiness, but TTS still requires the *selected* row to be ready. The next step correctly made the managed and direct TTS rows distinct, so on older gateways a stored `tts.provider = "openai"` (gateway-intent managed value) now resolves against the direct "OpenAI TTS" row — which typically reports `needs_keys` — while the managed "nous" row is the ready one. Since `voiceUnavailableReason` (AppState.swift:8262) and read-aloud both gate on `supportsSpeech`, such a profile could lose voice/TTS entirely even though the managed route works. This is the TTS-facing mirror of the fix; worth confirming against upstream's readiness semantics for a stored `openai` value.
**Suggestion:** Verify the upstream TTS readiness behavior for this stored-value case, or extend the same "diagnostic, not gate" treatment to the managed-route case for TTS.

### [nit] Timeout policy includes the data-URL prefix in the payload-length estimate
**File:** Conduit/Voice/HermesVoiceGateway.swift:45
**Issue:** `audio.dataURL.utf8.count` counts the `data:audio/wav;base64,` prefix plus the full MIME string, whereas upstream Desktop's reference policy is quoted as `dataUrl.length`. Since base64 is ASCII the char/utf8 counts agree, and the prefix is only ~19 chars, so this is immaterial at the 180s floor — just noting the comment inside scales from *audio size* while the input is the whole data URL.
**Suggestion:** No change required; the behavior matches upstream within noise. (Verified `Int.max` extremes can't trap: the `Double` product stays well below `Int.max`.)

## Verdict
APPROVE

The core fix is sound: IDs are now distinct, readiness is correctly demoted to diagnostics for STT, selection delegates row→config mapping to the gateway, cancellation/`markAudioEndpointUnavailable` paths are preserved, and the timeout policy is exact to upstream with safe extreme handling. All findings are minor/nits and non-blocking; the display-name guard (first finding) is the one I'd prioritize for robustness. No retain-cycle, force-unwrap, or concurrency issues were introduced — the changed code is consistently `@MainActor` and the new statics are safe.

<a href="https://opencode.ai/s/pYMCDsPJ">New%20session%20-%202026-08-30T19%3A24%3A09.410Z</a>
[opencode session](https://opencode.ai/s/pYMCDsPJ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33330715004)

kaishi00:
## Summary
Fixes the voice/STT parity regression: provider readiness rows now normalize to Hermes' own provider IDs (`nous` vs `openai` kept distinct), STT capability is gated only by `stt.enabled` (matching Hermes Desktop), row-backed selections are submitted to the gateway toolset provider endpoint with a fail-loud managed-Nous path, and the transcription timeout ports Desktop's `min(600s, max(180s, ceil(0.1×len)))` policy.

## Findings

### [minor] Real toolset error is discarded on the vendor-row legacy fallback
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:181-189
**Issue:** When `saveProviderSelection` fails with a genuine diagnosis (transport error or server `/error`), the failure branch for vendor rows does `errorMessage = nil` before calling `saveVendorSelection`. If the legacy config write then succeeds, the user is told nothing about the provider endpoint having reported a problem — a silent masking of a real server-side failure.
**Suggestion:** Only clear `errorMessage` when the fallback write fully succeeds, or emit a combined message (e.g. "Selected on this gateway, but the provider endpoint reported: …") when the fallback is taken.

### [minor] Selection success is inferred solely from the absence of an `error` string
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:232-235
**Issue:** `saveProviderSelection` treats any non-throwing response without a string `error` key as success. A response like `{"ok": false}` or a structured `{"error": {...}}` would be reported as saved and trigger `reload()`, even though the gateway rejected the selection.
**Suggestion:** Treat a literal `ok == false` and non-string error shapes (e.g. `error["message"]`) as failures so the caller doesn't report a saved selection the gateway refused.

### [minor] TTS row split can flip `supportsSpeech` to false on older gateways, blocking voice conversation and read-aloud
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:349-361, 371, 537-557
**Issue:** `readinessRows` now keeps the managed Nous TTS row and the direct "OpenAI TTS" row distinct. On older gateways where the managed route is stored as `tts.provider = "openai"` + `use_gateway: true`, the stored value now resolves against the direct OpenAI TTS row (typically `needs_keys`/not READY) instead of the previously merged, ready managed row. Since `supportsSpeech = ttsToolsetConfigAvailable && selectedProviderIsReady(...)`, and both `voiceUnavailableReason` (AppState.swift:8262) and read-aloud gate on `supportsSpeech`, a working managed route could lose voice and read-aloud entirely on those gateways. The "No TTS behavior change" claim in the PR description isn't quite accurate — capability detection changed via the row split.
**Suggestion:** Verify upstream's TTS readiness resolution for a stored `openai` value on the legacy gateway generation, or apply the same "diagnostic, not gate" demotion to TTS for the selected managed-route case. At minimum, document the resulting capability state in Settings for that scenario.

### [nit] Schema-only "nous" guard relies on the raw ID, which is correct but worth a comment
**File:** Conduit/Voice/HermesVoiceConfigurationService.swift:175
**Issue:** Not a bug — since a `nous` descriptor can only surface from a readiness row or a schema option, and the row-backed path now fails closed via `row.isManagedNous`, the `provider.lowercased() != "nous"` guard on the no-row path is the right defensive behavior (it prevents raw `nous` writes). The CodeRabbit display-name concern is already covered by the row-identity check at line 187.
**Suggestion:** No change required; the earlier concern is addressed.

No retain-cycle, force-unwrap, or strict-concurrency issues introduced: production code is consistently `@MainActor`, the new timeout helpers are `static let`/pure functions with verified overflow safety (`Int.max` case is covered by tests and `Double` math stays below `Int.max`), and `VoiceCapturedAudio.dataURL` is now computed once and reused.

## Verdict
APPROVE

The core fix is sound and well-tested. All findings are minor/non-blocking; the `errorMessage = nil` wipe on the fallback path and the older-gateway TTS `supportsSpeech` flip are the two I'd prioritize for a follow-up.

<a href="https://opencode.ai/s/cKORSuhG">New%20session%20-%202026-08-30T20%3A14%3A14.581Z</a>
[opencode session](https://opencode.ai/s/cKORSuhG)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33333022796)

PR #116, opened 2026-08-30
fix(profiles): make profile discovery monotonic and failure-tolerant

kaishi00:
## Symptom

App Store build (~140) report: a user with two Hermes profiles intermittently sees only one of them. The missing profile cannot be selected; logout/login restores both; after some time one may vanish again; sometimes both spontaneously reappear.

## Root cause

Two defects combined destructively:

1. `AppState.init` never restored the persisted `conduit.knownProfiles.v1` cache into `profiles`, so every cold launch started the visible list empty.
2. `loadProfiles()`'s catch handler replaced `profiles` with `[activeProfile]` whenever it was empty **and persisted that degraded list over the complete cache**.

So one transient `/api/profiles` failure (bridge still loading, dashboard restart, flaky link) collapsed the picker to a single profile — whichever was active — and clobbered the good cache. The inverse direction (losing profile 2 vs losing Default) depends only on which profile was remembered as active. A later successful discovery explains the spontaneous recovery.

## Fix

Profile discovery is now **monotonic and failure-tolerant**:

- **Init hydration** — `AppState.init` seeds the visible list with `orderedProfiles(persisted cache ∪ activeProfile ∪ "default")` before any discovery runs. As a side effect this heals caches a pre-fix build already degraded (`default` and the active profile are re-added on first launch).
- **Failure path** — a failed refresh commits `orderedProfiles(current ∪ activeProfile ∪ "default")` instead of `[activeProfile]`, and persists only that union. It can add, never shrink, and never overwrites a complete cache.
- **Degenerate success** (review finding) — a 200 with empty/missing `profiles` (dashboard mid-restart, partial deploy) routes through the same monotonic merge instead of collapsing to `["default"]`. A genuinely reduced server list (non-empty names) still replaces authoritatively, so server-side deletions still propagate.
- **Stale-response protection** — both commit paths capture `dashboardTicketBridge` before awaiting and discard the outcome if the bridge was swapped mid-request (server change, re-login, disconnect). This reuses the app's existing bridge-identity commit-gate pattern (same as the dashboard catalog loader, `AppState.swift:6018`). A late reply from an old connection can neither overwrite live state nor repopulate a cache the server-identity wipe already cleared.

### Server-identity scoping

Not re-namespaced. `prepareChatResumeForConnection(to:)` already wipes `profiles` and `knownProfilesKey` when the normalized server identity changes, which prevents profile names from server A appearing while connected to server B. Remaining limitation (documented in code): between init hydration and the identity check the previous server's names briefly exist in memory; the picker is unreachable pre-connection and the window is never persisted.

### Non-goals (unchanged by this PR)

Profile switching semantics, transcript reconciliation, presentation cache, YOLO bookkeeping, session hydration (#106), auth flow — all untouched. This is a narrow discovery/persistence fix.

## Tests

New `ConduitTests/ProfileDiscoveryTests.swift` (11 tests) via a `profileDiscoveryLoader` init seam (mirrors `sessionCatalogLoader`) plus a test-only bridge installer to model connection swaps:

- init hydrates the persisted cache; heals a degraded legacy cache
- discovery failure preserves the complete cache (active = profile2 and active = Default)
- no-cache fallback presents `["default", active]` and a later success converges
- successful discovery refreshes the visible list and persisted cache
- transient failure → success converges without relogin
- late success/failure from a replaced bridge is discarded (MainActor gate suspends the loader mid-flight)
- late success/failure after a server change repopulates nothing
- degenerate empty payloads (empty array / missing key) keep known profiles

## Verification

- Focused: `ProfileDiscoveryTests` 11/11 pass (Mac build host, iPhone 17 Pro).
- Full `ConduitTests` suite: **1426/1426, 0 failures** (1415 baseline + 11 new), including all profile-switch, YOLO, and cross-profile cache regression suites.
- `git diff --check` clean; no lint tooling configured in the repo.
- Caveat: the local `ConduitUITests` runner failed to bootstrap ("Timed out waiting for AX loaded notification") — the known simulator AX wedge. It reproduced identically on the pre-fix commit, the post-fix commit, and in isolation, so it is environmental and unrelated to this diff; branch CI will cover the UI target.

## Review

Local multi-model review (three reviewers, parallel): **APPROVE × 3, zero blockers**. Both WARNING-class findings were adopted (degenerate empty-success guard; the reviewer NIT suggesting removal of the dashboard-URL test seed proved incorrect and was reverted with an explanatory comment — the seed is what feeds `initialChatResumeServerIdentity` at init).

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved visible profiles when discovery temporarily fails, returns no profiles, or encounters delayed responses.
  * Restored saved profiles during app launch, including recovery of incomplete legacy profile data.
  * Kept active and default profiles available when discovery results are unavailable.
  * Automatically switched to a fallback profile if the active profile was deleted externally and reconnected when needed.
  * Applied profile removals correctly when confirmed by authoritative discovery results.

* **Tests**
  * Added coverage for profile loading, recovery, refreshes, connection changes, failures, and empty responses.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -6546,15 +6546,25 @@ final class AppState: ObservableObject {
             // old connection must not overwrite the live list, and must not
             // repopulate a persisted cache that
             // prepareChatResumeForConnection(to:) already cleared with the
-            // previous server's profile names.
+            // previous server's profile names. Discarded silently — profile
+            // discovery has no retry loop to feed an error into.
             guard dashboardTicketBridge === bridge else { return }
             let values = response["profiles"] as? [Any] ?? []
             let names = values.compactMap { value -> String? in
                 if let name = value as? String { return name }
                 return (value as? [String: Any])?["name"] as? String
             }
-            profiles = orderedProfiles(names + ["default"])
-            defaults.set(profiles, forKey: knownProfilesKey)
+            // A 200 with no profile names is a degraded payload (dashboard
+            // mid-restart, partial deploy, proxy), not authoritative evidence
+            // that every profile vanished — Hermes always has `default`. The
+            // next line would otherwise collapse the picker and persist that
+            // degraded list over the complete cache, the exact bug this
+            // function exists to prevent, just via a 200 instead of an error.
+            let nextProfiles = names.isEmpty
+                ? orderedProfiles(profiles + [activeProfile, "default"])
+                : orderedProfiles(names + ["default"])
+            profiles = nextProfiles
+            defaults.set(nextProfiles, forKey: knownProfilesKey)
         } catch {
             guard dashboardTicketBridge === bridge else { return }
             // Profile discovery is additive and monotonic. A failed refresh
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -6625,21 +6625,28 @@ final class AppState: ObservableObject {
     /// that reconnect lands.
     private func adoptAuthoritativeFallbackProfile(_ fallback: String) {
         guard fallback != activeProfile else { return }
+        // A profile change replaces the voice gateway; any in-flight read
+        // aloud belongs to the outgoing profile (same as switchProfile).
+        messageReadAloudController.stop()
         flushPendingPresentationCache()
         sessions = []
         cronSessions = []
         archivedSessions = []
+        projects = []
+        supportsProjects = false
+        projectsLoading = false
         slashCommands = Self.builtInSlashCommands
-        // The in-memory transcript, streaming text, and reasoning state all
-        // belong to the deleted profile's session; the flush above already
-        // persisted them under that profile's namespace. Showing them under
-        // the fallback's restored session identity would mislabel them, and
-        // the unguarded presentation-cache writes on user actions would
-        // persist them there. Same reset set as the other hard boundaries
-        // (prepareChatResumeForConnection, sign-out).
         messages = []
         clearStreamingText()
         resetReasoningTurn()
+        // The next profile's approval mode is unknown until its first session
+        // snapshot arrives; don't let the deleted profile's approval floor or
+        // YOLO state leak across the re-home — neutralize to the safe display
+        // until the fallback profile resolves them (same neutralization as
+        // switchProfile).
+        runtime.approvalsMode = nil
+        runtime.yolo = false
+        lastReportedSessionYolo = nil
         // Drop the deleted profile's persisted local bookkeeping so a later
         // re-creation with the same name starts fresh instead of
         // resurrecting obsolete titles and pins.
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -755,6 +755,10 @@ final class AppState: ObservableObject {
     private let sessionTitleRecoveryTracker = SessionTitleRecoveryTracker()
     private let sessionRenameOperationsOverride: SessionRenameOperation.Operations?
     private let sessionCatalogLoaderOverride: ((Bool) async throws -> [SessionSummary])?
+    /// Test seam mirroring `sessionCatalogLoader`: overrides the dashboard
+    /// `/api/profiles` fetch inside `loadProfiles()` so success, failure, and
+    /// late-response ordering can be modeled without a live WebKit bridge.
+    private let profileDiscoveryLoaderOverride: (@MainActor () async throws -> [String: Any])?
     private var reconnectAttempts = 0
     /// Whether the UI scene is active. Backgrounded scene updates must
     /// complete within ~10s of wall clock before the watchdog kills the app
@@ -1006,6 +1010,7 @@ final class AppState: ObservableObject {
         },
         sessionRenameOperations: SessionRenameOperation.Operations? = nil,
         sessionCatalogLoader: ((Bool) async throws -> [SessionSummary])? = nil,
+        profileDiscoveryLoader: (@MainActor () async throws -> [String: Any])? = nil,
         reconnectScheduler: ChatResumeReconnectScheduler? = nil,
         reconnectExecutor: ChatResumeReconnectExecutor? = nil,
         chatResumeLifecycleOperations: ChatResumeLifecycleOperations = .live,
@@ -1025,6 +1030,7 @@ final class AppState: ObservableObject {
         self.clearSessionPresentationCache = clearSessionPresentationCache
         sessionRenameOperationsOverride = sessionRenameOperations
         sessionCatalogLoaderOverride = sessionCatalogLoader
+        profileDiscoveryLoaderOverride = profileDiscoveryLoader
         self.reconnectScheduler = reconnectScheduler ?? scheduleChatResumeReconnectTask
         self.reconnectExecutor = reconnectExecutor
         self.chatResumeLifecycleOperations = chatResumeLifecycleOperations
@@ -1057,6 +1063,16 @@ final class AppState: ObservableObject {
            let stored = try? JSONDecoder().decode([String: [String]].self, from: data) {
             pinnedSessionIDsByProfile = stored
         }
+        // Hydrate the visible profile list from the persisted known-profile
+        // cache before any discovery runs. A cold launch must not present an
+        // effectively empty list — starting from `[]` is what let a single
+        // failed /api/profiles refresh collapse the picker to one entry and
+        // persist that degraded list over the complete cache. The union also
+        // heals a cache a pre-fix build already degraded: `default` and the
+        // remembered active profile are re-added on first launch.
+        profiles = orderedProfiles(
+            (defaults.stringArray(forKey: knownProfilesKey) ?? []) + [activeProfile, "default"]
+        )
         restoreActiveSessionState(for: activeProfile)
         restorePinnedSessions(for: activeProfile)
         if shouldLoadSavedConnection {
@@ -6516,9 +6532,22 @@ final class AppState: ObservableObject {
     }
 
     func loadProfiles() async {
-        guard let dashboardTicketBridge else { return }
+        guard let bridge = dashboardTicketBridge else { return }
         do {
-            let response = try await dashboardTicketBridge.requestJSON(path: "/api/profiles")
+            let response: [String: Any]
+            if let profileDiscoveryLoaderOverride {
+                response = try await profileDiscoveryLoaderOverride()
+            } else {
+                response = try await bridge.requestJSON(path: "/api/profiles")
+            }
+            // Commit gate, same identity pattern as the dashboard catalog
+            // loader: a bridge swapped mid-request (server change, re-login,
+            // disconnect) makes this response foreign. A late reply from an
+            // old connection must not overwrite the live list, and must not
+            // repopulate a persisted cache that
+            // prepareChatResumeForConnection(to:) already cleared with the
+            // previous server's profile names.
+            guard dashboardTicketBridge === bridge else { return }
             let values = response["profiles"] as? [Any] ?? []
             let names = values.compactMap { value -> String? in
                 if let name = value as? String { return name }
@@ -6527,10 +6556,19 @@ final class AppState: ObservableObject {
             profiles = orderedProfiles(names + ["default"])
             defaults.set(profiles, forKey: knownProfilesKey)
         } catch {
-            // Profile discovery is additive. A working chat must not be
-            // replaced by an error just because an older dashboard lacks it.
-            if profiles.isEmpty { profiles = orderedProfiles([activeProfile]) }
-            defaults.set(profiles, forKey: knownProfilesKey)
+            guard dashboardTicketBridge === bridge else { return }
+            // Profile discovery is additive and monotonic. A failed refresh
+            // (bridge still loading, dashboard restart, transient 5xx) must
+            // never shrink the visible list or overwrite a complete persisted
+            // cache with a degraded fallback — that degradation is exactly
+            // how a known profile "disappeared" from the picker until the
+            // next successful discovery or a logout/login cycle. Union with
+            // the current list so the active profile and the `default`
+            // invariant are always represented, then persist the union (a
+            // no-op write when the cache is already complete).
+            let merged = orderedProfiles(profiles + [activeProfile, "default"])
+            profiles = merged
+            defaults.set(merged, forKey: knownProfilesKey)
         }
     }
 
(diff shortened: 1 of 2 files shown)


kaishi00:
## Summary
This PR fixes the intermittent disappearing-profile bug by making `/api/profiles` discovery monotonic and failure-tolerant: `AppState.init` hydrates the persisted cache before discovery, failed/degenerate refreshes merge instead of collapse, and both commit paths gate on dashboard-bridge identity to discard stale responses. 303 lines of focused regression tests cover hydration, healing, failure, degenerate-success, and late-response cases.

## Findings

### [minor] Authoritative success can silently drop the active profile, breaking the invariant the degraded paths guarantee
**File:** Conduit/Services/AppState.swift:6565
**Issue:** The failure and degenerate-empty paths guarantee `activeProfile` and `"default"` are always represented, but the non-empty success path commits `orderedProfiles(names + ["default"])` with no check for `activeProfile`. If the server removes the currently-active profile, `activeProfile` retains the dropped name while `profiles` no longer contains it; `ProfilePickerSheet.visibleProfiles` (ProfilePickerSheet.swift:13-15) then cannot present the active selection. This is pre-existing behavior and server authority is intentional (deletions propagate), but it is the same invariant this PR centralizes and now leaves asymmetric.
**Suggestion:** On authoritative commit, if `activeProfile` ∉ `nextProfiles`, either re-add it for the current session (leave it to a future success to converge) or reset `activeProfile` to "default" so no unrepresentable active state persists. At minimum, document this asymmetry next to the monotonic-invariant comments.

### [minor] Commit gate leaves a window between server-identity wipe and bridge swap where a stale reply can still commit
**File:** Conduit/Services/AppState.swift:6551 (also 6569)
**Issue:** The `dashboardTicketBridge === bridge` guard only fires if the bridge has already been swapped. In the resume flow, `prepareChatResumeForConnection(to:)` wipes `profiles`/cache (line 1678, 1691) before the new dashboard bridge is installed (line 1992/2183). If a suspended `loadProfiles` from the previous connection resumes after the wipe but before the bridge swap, the guard passes and re-persists previous-server profile names into the newly-cleared cache. A subsequent successful discovery converges, so this is non-destructive, but it is a thinner guarantee than the tests claim ("a late reply…can neither overwrite live state nor repopulate a cache the server-identity wipe already cleared").
**Suggestion:** Also compare the captured chat-resume server identity (e.g., `initialChatResumeServerIdentity` vs. current `chatResumeServerIdentityKey`) inside the guard, or introduce a connection-generation counter captured before the await (matching the 3-way `profile/client/bridge` gate used by the catalog loader at line 6033).

### [nit] Same-server re-login with changed Cloudflare access discards a valid response
**File:** Conduit/Services/AppState.swift:6551
**Issue:** `prepareDashboardBridge` rebuilds the bridge when `cloudflareAccess` changes even for the same URL (line 2220-2223). A re-login to the same server changing access would fail the `===` guard and silently drop a perfectly valid response, leaving the list stale until a later discovery. Conservative (no data loss), but slightly over-aggressive.
**Suggestion:** Acceptable as-is; if desired, strengthen the guard to compare normalized server identity plus bridge identity rather than instance identity alone.

### [nit] Test-only installer ships in the production target
**File:** Conduit/Services/AppState.swift:8603
**Issue:** `installDashboardTicketBridgeForTesting` is a public symbol in the app target solely for tests. It hard-codes a non-production value into `dashboardTicketBridge` with no `#if DEBUG` guard.
**Suggestion:** Wrap in `#if DEBUG` (tests build with `@testable` can still call it) or move the seam behind a protocol.

### [nit] Degenerate-empty test second half relies on state persisted by the first half
**File:** ConduitTests/ProfileDiscoveryTests.swift:258-265
**Issue:** `missingKeyState` is created without explicitly seeding its known-profile cache; it correctly passes only because the preceding block persisted `["default", "profile2"]` and because `activeProfileKey` is seeded. If the first block is edited or removed, the missing-key assertion silently depends on that coupling.
**Suggestion:** Call `seedKnownProfiles(["default", "profile2"], activeProfile: "profile2")` immediately before creating `missingKeyState` to make the setup self-contained.

## Verdict
APPROVE — the fix is correct in its core behavior: no force-unwraps, no retain cycles introduced (the loader override closure is set externally and captures nothing from `self`), the class is `@MainActor`-isolated with no data-race exposure, and the commit path has no suspension point after the await so commits are atomic on the actor. Findings above are minors/nits and non-blocking.

<a href="https://opencode.ai/s/CC2tUNLG">New%20session%20-%202026-08-30T23%3A13%3A02.349Z</a>
[opencode session](https://opencode.ai/s/CC2tUNLG)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33341166604)

kaishi00:
## Summary
This PR fixes the intermittent disappearing-profile bug by making `/api/profiles` discovery monotonic and failure-tolerant: `AppState.init` hydrates the visible list from the persisted cache, failed/degenerate-empty refreshes union instead of collapse, both commit paths gate on dashboard-bridge identity, and a large new test file covers hydration, healing, failure, late-response, and degenerate cases.

## Findings

### [major] Cross-server authoritative discovery destructively prunes a valid profile's global bookkeeping
**File:** Conduit/Services/AppState.swift:6574-6589 and 6626-6654
**Issue:** The new re-home path fires whenever a *non-empty* authoritative response lacks `activeProfile`. `activeProfile` is global and is never reset on server change/logout (`prepareChatResumeForConnection` clears `profiles`/`knownProfilesKey` but not `activeProfile`), and `connect(with:profile:)` is always invoked with `profile: activeProfile` (lines 2184, 2204). So connecting to a different Hermes server whose profile list doesn't contain the current active profile — a routine multi-server scenario — triggers `adoptAuthoritativeFallbackProfile`, which prunes `activeSessionTitlesByProfile`/`pinnedSessionIDsByProfile` (lines 6646-6649). Those keys are **global**, not server-scoped, so a profile that still exists on server A silently loses its persisted titles/pins when the user merely connects to server B. It also nukes the in-memory chat state and force-flips `activeProfile` to "default" with no user action. Pre-PR, this path only replaced `profiles`/cache and left the profile's bookkeeping intact. The new tests only cover the same-server deletion case.
**Suggestion:** Only re-home/prune when discovery ran against the *same* server identity as before it started (capture the normalized `chatResumeServerIdentityKey` at `loadProfiles()` entry and compare at commit, in addition to the bridge `===` gate). If the identity changed, skip the prune and just leave the freshly-wiped cache alone. At minimum, don't prune bookkeeping for a profile that belongs to a previously-connected server.

### [minor] Older authoritative response can overwrite a newer one on the same bridge
**File:** Conduit/Services/AppState.swift:6534-6567
**Issue:** `loadProfiles()` is re-entrant (called from `connect` at line 2016 and the resume pipeline at 4017). Two overlapping refreshes on the *same* bridge both pass the `dashboardTicketBridge === bridge` gate; if the older response commits last it reverts the list that the newer response just established (the non-empty branch is authoritative and replaces wholesale). Unlike the catalog loader's generation-based commit (line 6050), there is no ordering guard here.
**Suggestion:** Serialize refreshes (e.g., skip if a discovery is already in flight) or add a monotonically increasing request generation captured before the await and compared after it, so an older reply is discarded.

### [minor] Commit gate ignores the server-identity wipe window
**File:** Conduit/Services/AppState.swift:6551 (and 6592)
**Issue:** The guard only compares bridge identity. In the resume flow, `prepareChatResumeForConnection(to:)` wipes `profiles`/`knownProfilesKey` (1678, 1691) *before* the new dashboard bridge is installed (2220-2223). A suspended `loadProfiles` from the old connection that resumes in that window passes the `===` guard and re-persists `[activeProfile, "default"]` (or prior names) into the cache the wipe just cleared — the exact stale-repopulation the PR claims to prevent. Converges on next success, but a transient failure path could keep the foreign names in the cache indefinitely.
**Suggestion:** Fold the server-identity comparison into the commit gate (capture identity at entry, compare at commit), which also addresses the major finding above.

### [minor] Test-only bridge installer ships in the production target
**File:** Conduit/Services/AppState.swift:8677
**Issue:** `installDashboardTicketBridgeForTesting` is a public function in the app target with no `#if DEBUG` guard, inconsistent with the existing `#if DEBUG`-wrapped `presentationCacheFlushOperationForTesting` (line 931-939). It mutates a `@Published` connection-swap seam in release builds.
**Suggestion:** Wrap in `#if DEBUG` (the `@testable` test target still sees it).

### [nit] Test-fixture implicitly unwrapped optionals remain at HEAD
**File:** ConduitTests/ProfileDiscoveryTests.swift:18-19
**Issue:** `suiteName`/`defaults` are still `String!`/`UserDefaults!` at HEAD despite the review comment marking them addressed; a premature helper/teardown access would trap.
**Suggestion:** Initialize to nonoptional values in `setUp()` or use optional access with explicit guards.

## Verdict
REQUEST_CHANGES — the core monotonic/commit-gate logic is sound, MainActor-isolated, and free of force-unwraps and retain cycles, with solid test coverage. But the new authoritative re-home path can destructively prune globally-persisted profile bookkeeping and reset the active profile on a cross-server connection where the current profile is merely absent on the other server — a previously-preserved scenario now untested that warrants a server-identity check in the commit path before merging.

<a href="https://opencode.ai/s/ME4ohSzd">New%20session%20-%202026-08-30T23%3A47%3A08.787Z</a>
[opencode session](https://opencode.ai/s/ME4ohSzd)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33342670174)

kaishi00:
## Summary
This PR fixes the intermittent disappearing-profile bug by making `/api/profiles` discovery monotonic and failure-tolerant: `AppState.init` hydrates the persisted cache, failed/degenerate-empty refreshes union instead of collapse, both commit paths gate on dashboard-bridge identity, and an authoritative success that drops the active profile re-homes to a fallback. A new 423-line test file covers hydration, healing, failure, late-response, and degenerate cases.

## Findings

### [major] Authoritative re-home can destructively prune a still-valid profile's persisted bookkeeping on a same-server partial response
**File:** Conduit/Services/AppState.swift:6574-6589, 6626-6660
**Issue:** The guard only rejects a response whose bridge was swapped. A non-empty-but-truncated 200 (the exact "degraded payload" class the empty-names union at 6563-6565 defends against) is treated as fully authoritative purely because `names` is non-empty. If it happens to omit only the active profile (dashboard mid-restart, partial deploy), the new `adoptAuthoritativeFallbackProfile` path permanently removes that profile from `activeSessionTitlesByProfile`/`pinnedSessionIDsByProfile` via `removeValue` + `persist*` (6653-6656), force-flips `activeProfile` to "default" (6657-6660), tears down the in-memory transcript/catalogs, and fires a reconnect (6587-6589) — the same transient-signal-destroys-valid-state failure mode this PR exists to eliminate, reintroduced via a different boolean. No test covers a same-server non-empty payload missing only the active profile (the tests only cover full deletion, empty payloads, and cross-server cases where `prepareChatResumeForConnection` already wiped the dicts, making the prune a no-op). Note the cross-server framing in prior reviews overstates damage: on an identity change `prepareChatResumeForConnection` already clears the dicts (1685-1689), so the prune is only destructive in the same-server/partial case.
**Suggestion:** Capture the normalized `chatResumeServerIdentityKey` at `loadProfiles()` entry and include it (plus a request-generation token) in the re-home gate, so the prune/reset only fires when the response provably reflects the same server state that was in effect before the request. At minimum, add a same-server partial-payload test documenting the intended behavior, or shrink the destructive half (skip the bookkeeping prune on a same-server transient, defer it to a subsequent confirming observation).

### [minor] No ordering guard for overlapping refreshes on the same bridge
**File:** Conduit/Services/AppState.swift:6534-6605
**Issue:** `loadProfiles()` is re-entrant (called from `connect` at 2016 and the resume pipeline at 4052/4054). Two overlapping refreshes on the same bridge both pass the `===` gate; the older authoritative reply committing last can revert the list/cache the newer one just wrote. The catalog loader avoids this with a generation-based commit (6050); this path has no equivalent.
**Suggestion:** Serialize refreshes (early-return if a discovery is already in flight) or add a monotonically increasing request generation captured before the await and compared after it.

### [minor] Test-only bridge installer ships unguarded in the production target
**File:** Conduit/Services/AppState.swift:8678-8680
**Issue:** `installDashboardTicketBridgeForTesting` is a public symbol that mutates the `@Published` dashboard-bridge seam with no `#if DEBUG` guard, inconsistent with the existing `#if DEBUG`-wrapped `presentationCacheFlushOperationForTesting` (931-939).
**Suggestion:** Wrap in `#if DEBUG`; the `@testable` test target still sees it.

### [nit] Commit gate is weaker than the tests claim (server-identity wipe window)
**File:** Conduit/Services/AppState.swift:6551, 6592
**Issue:** The gate compares only bridge identity. On the MainActor the bridge swap is synchronous before the first await in `connect`, so the window between the `prepareChatResumeForConnection` wipe and the swap is practically unreachable — but the tests (197-241) assert a guarantee the guard alone does not provide. If a future change introduces an await in that span, a suspended late reply would re-persist previous-server names.
**Suggestion:** Fold the captured server identity into the commit gate (also closes the major finding above).

### [nit] Fire-and-forget reconnect Task races the in-flight connect
**File:** Conduit/Services/AppState.swift:6587-6589
**Issue:** `loadProfiles()` is itself awaited inside `connect` (2016) after `isConnected = true`; scheduling `Task { await reconnect() }` here spawns a nested connect concurrently with the connect that's still mid-sequence. The automatic-work-token fences likely make it safe (reconnect is guarded by `isSceneActive` and continuation identity), but this interleaving is untested.
**Suggestion:** Acceptable; consider documenting, or await the reconnect instead of fire-and-forget now that the fences are in place.

### [nit] Test fixture still uses implicitly unwrapped optionals
**File:** ConduitTests/ProfileDiscoveryTests.swift:18-19
**Issue:** `suiteName`/`defaults` remain `String!`/`UserDefaults!` at HEAD despite the prior comment being marked addressed; they trap on any helper/teardown access before `setUp`. Low risk because XCTest guarantees `setUp` runs first, but the pattern was explicitly flagged.
**Suggestion:** Initialize nonoptional values inline in `setUp` or make them plain optionals with guards. (The "second half depends on first half" nit from the first review is correctly fixed at HEAD — `seedKnownProfiles` is explicitly re-invoked at line 260.)

## Verdict
REQUEST_CHANGES — the core monotonic/commit-gate logic is sound: no force-unwraps or retain cycles introduced, `AppState` stays `@MainActor`-isolated with atomic sync commits after the await, and the new tests are thorough on the documented paths. However, the new authoritative re-home path reintroduces the PR's core failure mode through an unguarded route — a same-server non-empty partial response can permanently prune a still-valid profile's persisted titles/pins and silently reset the user's active profile — and no test exercises that case. Requiring a server-identity/request-generation gate on the re-home (which also closes the wipe-window nit) before merging.

Note: I did not run builds or tests; this review is based on static analysis of the diff.

<a href="https://opencode.ai/s/j2Fy2XHh">New%20session%20-%202026-08-31T00%3A37%3A13.136Z</a>
[opencode session](https://opencode.ai/s/j2Fy2XHh)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33345058608)

PR #118, opened 2026-08-31
fix(chat): hydrate persisted transcripts as bounded tail-anchored pages

kaishi00:
## Summary

Follow-up to PR #110 / issue #106: opening a conversation now has **bounded transport and normalization cost** regardless of session size (200 / 2,000 / 20,000 / 100,000 persisted rows). PR #110 removed giant transcripts from `session.resume` via `omit_messages`, but Conduit still assumed `GET /api/sessions/{id}/messages` returns the entire transcript in one response — no longer true for current Hermes, which serves it as a paged history API. Conduit now matches Hermes Desktop's division of responsibility: compact `session.resume(omit_messages: true)` for runtime/inflight state, bounded tail pages of `/messages` for persisted display history.

## The pagination model

- **Request contract (explicit, never the implicit default):** `?limit=120&offset=0&order=latest&include_compacted=true` — `PersistedTranscriptPagination.pageSize = 120`, matching Desktop's `LATEST_SESSION_MESSAGES_LIMIT`. Older pages continue the same contract at the next offset.
- **`PersistedTranscriptWindowState`** (per active session): requested/resolved session IDs, profile, pageSize, nextOffset (rows of the newest tail covered locally), canLoadEarlier, isLoadingEarlier, hasBackfilledPrefix. Identity-safe across session switch, profile switch, disconnect/reconnect, runtime↔stored ID remapping, and stale async backfills (multi-field discard guard + alias-aware ownership gate).
- **Backfill:** "Load earlier messages" prepends older pages. Offset advances by fetched row count (self-corrects `order=latest` drift); overlap is deduplicated by durable row identity (`ChatMessage.id` = SQLite row id, never visible text), and only the strictly-older region of a page is prepended. A short/empty page or a response that lost the tail contract retires the affordance; failures never loop — only a structurally gone source retires it, transient trouble stays tap-retryable. Every page validates its returned `session_id` against the requested ID, the window's resolved stored ID, and known aliases before normalizing — foreign responses never prepend nor advance coverage.
- **Repeated-reconnect prefix preservation:** a reconnect re-reads only the newest page; the refreshed tail grafts onto the backfilled prefix via the explicit `hasBackfilledPrefix` flag (alias-aware gate) — set when a backfill prepends, carried by every successful graft, cleared when a refreshed tail has no safe anchor. The prefix therefore survives reconnect #2, #3, … even though network coverage resets to the fresh tail; the next backfills retrace the prefix in deduped page-sized increments (harmless duplicates, never gaps).

## Current Hermes vs legacy Hermes (shape detection, no version sniffing)

- **pagination echo with `order:"latest"`** → current tail-anchored contract (offset measured back from the newest row; pages arrive chronological — verified against upstream Hermes and Desktop's client).
- **pagination echo without `order`** → an older dashboard generation whose offsets page from the *oldest* end (deployed installs of that era silently ignore `order`/`include_compacted`). Conduit re-reads the transcript one-shot at most once and treats it as legacy — never walks forward from row zero. The same one-shot re-read covers a paginated echo whose rows lack durable persisted IDs (positional fallback IDs can never enter dedup).
- **no pagination object** → legacy one-shot full transcript; hydrated exactly as before, no window, no affordance.

## Fallback policy (final contract)

**Structural endpoint absence — exactly one legacy full-transcript resume fallback:**

```text
404 / 410 / 501
→ the endpoint cannot serve any page; exactly one legacy resume, never a cascade
```

**Transient / bounded-history failures — DO NOT fall back to the full-transcript WebSocket resume:**

```text
status 0 · notReady / request timeout · 408 · 429 · ordinary 5xx · oversized bounded response
→ surface a retryable restore error; retry the BOUNDED request through reconnect recovery only
```

This is the giant-session safety property: current Hermes can serve the bounded `include_compacted=true` page slowly for heavily compacted sessions until upstream #97440 bounds that work server-side, so a slow bounded read fails boundedly and never silently becomes "load the entire transcript over the WebSocket". Authentication and unclassified failures also surface.

## Oversized responses: generic vs transcript-specific

The bridge-level `oversizedResponse` error carries **neutral copy** ("This response is too large to load safely.") — it can arise from any oversized dashboard operation, not only transcripts. The transcript layer maps a **legacy one-shot** oversize (an old backend attempting the entire conversation) to the compatibility message ("This conversation is too large to load safely with this Hermes version. Update Hermes to enable paginated conversation history."), while a **bounded current-Hermes page** oversize keeps the neutral copy — it never claims the backend lacks pagination. No global response-size guard was weakened or raised.

## Viewport stability

Prepends re-pin the currently visible row via the pure `ChatViewportController`: the anchor is captured at tap time, consumed only by the prepend's front-row replacement (streaming growth, turn completion, tail grafts, and prefix-dropping rewrites never consume it or re-pin early), one unanimated scroll + delayed generation-stamped retry, ownership re-validated against rendered *and* active session at landing, and silently discharged when the captured row no longer exists. Following-latest holds via the existing bottom-follow system; discharge is session-scoped. The backfill task is view-lifetime-aware (cancelled on disappear, cancellation-checked before viewport mutation). Saved viewport anchors older than the loaded tail fall back to latest through the existing `ChatResumeViewportResolver` — opening never becomes an N-page anchor chase.

## Tool calls across a page boundary

A single assistant turn can carry any number of tool calls. When a call row ends an older page and its result rows begin the newer page, `PersistedTranscriptWindow.reconcilingToolCallsAcrossBoundary` folds the trailing unmatched call cards together with the matching held standalone result cards **by durable `ToolActivity.id` (Hermes `tool_call_id`) — unbounded, with no pair ceiling; identity only, never name, text, or position**. One logical tool run renders as one completed card with the right input/output; no orphan call cards, no duplicate standalone results. Regression coverage pins a 12-call run split across the boundary reconstructing fully.

## Replaced PR #110 assumptions

The old `CompactResumeTranscriptTests` assumption that a giant REST response carries the whole transcript is now the explicitly-pinned *legacy/one-shot compatibility* path. Regression coverage (29 new tests): bounded explicit request contract, tail-only hydration of a 2,000-row session, drift-overlap backfill dedup, terminal short page (repeated taps issue no requests), stale backfill discard on session switch, foreign-response ownership rejection, compacted display-row projection after prepend (hidden carriers, display_content, model-switch, auto-continue), legacy one-shot acceptance, oldest-anchored-echo and durable-ID-less fallbacks, oversized-history no-WS-retry (neutral + compatibility copies), live paginated resume inflight behavior, reconnect graft across repeated reconciles (same-id and runtime-alias), transient-failure retryability, duplicate-only coverage advance, malformed page and id-less page retirement, structural/transient classification matrices, and the viewport-controller anchor state machine (arming, landing, discharge, ownership races, mid-flight suffix changes, deleted anchors).

## Not changed

Display projection/`MessageNormalizer` architecture, live/inflight reconciliation (running, inflight continuation, persisted prefix vs streaming suffix, pending decisions, YOLO), compact WS resume behavior, session catalogs, profile discovery, voice, WebSocket/REST size limits, chat chrome, Markdown pipeline, Hermes backend code, page size, and the manual "Load earlier messages" UX.

## Verification (at head `01b24af`)

- Focused: `CompactResumeTranscriptTests` + `ChatViewportControllerTests` — **125/125** on the Mac build host (iPhone 17 Pro), including the >8-call (12-call) page-boundary regression: all results reconstruct, zero orphan cards.
- Full unit suite: **1458/1458, 0 failures** (PR #116 baseline 1429 + 29 new) at the final commit.
- `git diff --check` clean; no lint tooling exists in the repo.
- Multi-model review gate: base implementation 3 reviewer rounds (final APPROVE); the hardening pass (repeated-reconcile prefix, structural-only fallback, ownership validation, tool-boundary reconstruction, oversized split) was delta-reviewed — its one BLOCKER (all-held-folded duplicate) fixed; the final cap-lift delta was reviewed 3× APPROVE.
- Known accepted trade-offs (documented in code): after a reconnect the window resets coverage, so backfills retrace the backfilled prefix in deduped page-sized increments before reaching fresh rows; a cold bridge or slow bounded read now fails the session open with a visible error instead of degrading to legacy content.

Out of scope: upstream Hermes tagging/writing fixes; Conduit never loads the complete transcript automatically at open and does not chase ancient scroll anchors through REST pages.

## Summary by CodeRabbit

* **New Features**
  * Added paginated transcript history with a “Load earlier messages” control.
  * Older messages load in bounded batches while preserving scroll position.
  * Tool calls and results remain correctly connected across loaded pages.
* **Bug Fixes**
  * Prevented duplicate or unrelated messages during history loading.
  * Preserved previously loaded messages when transcripts refresh.
  * Improved handling of oversized, unavailable, or temporarily inaccessible history.
  * Prevented stale or cancelled loads from affecting the active conversation.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -31,7 +31,17 @@ struct ChatResumeLifecycleOperations {
     var loadCatalog: (@MainActor (HermesClient, Bool) async throws -> [SessionSummary])?
     var mintTicket: (@MainActor (String) async throws -> String)?
     var openSession: (@MainActor (HermesClient, String, Bool) async throws -> SessionResumeResult)?
+    /// Seam for the initial persisted-history fetch. Receives BOTH the
+    /// bounded tail-page request and the legacy one-shot re-read (an echo
+    /// without the `order=latest` contract) — the query argument is built by
+    /// the production path builder and is not threaded through the seam, so
+    /// callables distinguish the two requests by call order (bounded first,
+    /// one-shot re-read second, at most once).
     var persistedTranscript: (@MainActor (String, String) async -> PersistedTranscriptFetchOutcome)?
+    /// Seam for older-page backfill requests only (`offset` = the window's
+    /// next offset). Distinct from `persistedTranscript` so backfill tests
+    /// never collide with initial-hydration fixtures.
+    var loadEarlierTranscriptPage: (@MainActor (String, String, Int) async -> PersistedTranscriptFetchOutcome)?
     var branchSession: (@MainActor (
         HermesClient,
         String,
@@ -69,6 +79,7 @@ struct ChatResumeLifecycleOperations {
         mintTicket: (@MainActor (String) async throws -> String)? = nil,
         openSession: (@MainActor (HermesClient, String, Bool) async throws -> SessionResumeResult)? = nil,
         persistedTranscript: (@MainActor (String, String) async -> PersistedTranscriptFetchOutcome)? = nil,
+        loadEarlierTranscriptPage: (@MainActor (String, String, Int) async -> PersistedTranscriptFetchOutcome)? = nil,
         branchSession: (@MainActor (
             HermesClient,
             String,
@@ -105,6 +116,7 @@ struct ChatResumeLifecycleOperations {
         self.mintTicket = mintTicket
         self.openSession = openSession
         self.persistedTranscript = persistedTranscript
+        self.loadEarlierTranscriptPage = loadEarlierTranscriptPage
         self.branchSession = branchSession
         self.setSessionTitle = setSessionTitle
         self.refreshContext = refreshContext
@@ -150,6 +162,10 @@ enum PersistedTranscriptOutcome {
 struct PersistedSessionTranscript {
     let resolvedSessionId: String?
     let messages: [ChatMessage]
+    /// Pagination echo of the response this transcript was parsed from.
+    /// Nil (or an echo without the `order=latest` tail contract) means the
+    /// backend served a one-shot full transcript.
+    var page: PersistedTranscriptPagination.PageInfo?
 }
 
 struct ComposerSubmissionContext: Equatable {
@@ -419,6 +435,11 @@ final class AppState: ObservableObject {
         }
     }
     @Published private(set) var activeSessionTitle = "New conversation"
+    /// Persisted-history pagination window of the active conversation.
+    /// Drives the "Load earlier messages" affordance; nil means the current
+    /// transcript is either a legacy one-shot hydration or not persisted-
+    /// history-backed at all, and no older pages exist to fetch.
+    @Published private(set) var persistedTranscriptWindow: PersistedTranscriptWindowState?
     @Published private(set) var isChatRefreshing = false
     @Published private(set) var chatResumeBehavior: ChatResumeBehavior = .continueWhereLeftOff
     @Published private(set) var chatReturnSurface: ChatReturnSurface = .conversation
@@ -1680,6 +1701,7 @@ final class AppState: ObservableObject {
         activeSessionId = nil
         activeSessionTitle = "New conversation"
         messages = []
+        persistedTranscriptWindow = nil
         clearStreamingText()
         resetReasoningTurn()
         activeSessionTitlesByProfile = [:]
@@ -2125,6 +2147,7 @@ final class AppState: ObservableObject {
         sessionCatalogCache.removeAll()
         pinnedSessionIDs = []
         messages = []
+        persistedTranscriptWindow = nil
         setActiveSessionState(id: nil, title: "New conversation")
         clearStreamingText()
         resetReasoningTurn()
@@ -2776,6 +2799,41 @@ final class AppState: ObservableObject {
                     resumedSessionId: result.sessionId
                 )
             } ?? false
+
+            // Persisted-history window bookkeeping. The window describes the
+            // REST-fetched display history only — runtime state above is
+            // untouched. A pagination echo carrying the `order=latest` tail
+            // contract opens the bounded window; a legacy one-shot hydration
+            // (or a hydration that failed the session-identity gate) leaves
+            // no window, because no older page exists to fetch.
+            //
+            // The refresh resets coverage to the fetched page even when the
+            // graft below kept older rows. The next backfills then retrace
+            // the backfilled prefix in page-sized increments, each deduped
+            // against held rows, until coverage catches up — harmless
+            // duplicate requests that self-correct, never gaps. Under-
+            // counting is the safe direction: trusting the prior coverage
+            // after a server-side rewrite could silently skip rows. A
+            // conversation that was already fully backfilled re-lights the
+            // affordance once and re-retires on its first terminal page.
+            //
+            // Capture the pre-reconcile transcript and window for the graft
+            // decision below before either is replaced.
+            let previousTranscriptMessages = messages
+            let priorWindow = persistedTranscriptWindow
+            if let transcript, transcriptMatches, let page = transcript.page,
+               page.honorsTailContract {
+                persistedTranscriptWindow = PersistedTranscriptWindowState(
+                    requestedSessionID: sessionId,
+                    profile: profile,
+                    pageSize: PersistedTranscriptPagination.pageSize,
+                    resolvedSessionID: transcript.resolvedSessionId ?? result.sessionId,
+                    nextOffset: page.rawReturned,
+                    canLoadEarlier: page.mayHaveOlderRows(fetchedRowCount: page.rawReturned)
+                )
+            } else {
+                persistedTranscriptWindow = nil
+            }
             if let transcript, transcriptMatches, result.snapshot.hasLiveProjection, !transcript.messages.isEmpty {
                 // Desktop keeps its live projection during an active turn. Seed
                 // the same durable presentation details first so the completed
@@ -2798,9 +2856,21 @@ final class AppState: ObservableObject {
                 && (!result.snapshot.hasLiveProjection || !resumeCarriedTranscript)
             let presentationResult: SessionResumeResult
             if let transcript, shouldUsePersistedTranscript {
+                // A re-reconciliation replaces only the newest page. Keep
+                // everything "Load earlier messages" already backfilled by
+                // re-anchoring the refreshed tail onto the older prefix;
+                // without this, any reconnect during a browsed long session
+                // would silently truncate the visible history back to one
+                // page.
+                let graftedTail = PersistedTranscriptWindow.grafting(
+                    refreshedTail: transcript.messages,
+                    ontoBackfilled: previousTranscriptMessages,
+                    window: priorWindowOwnsThisConversation(priorWindow, requestedSessionId: sessionId, profile: profile)
+                        ? priorWindow : nil
+                )
                 presentationResult = SessionResumeResult(
                     sessionId: result.sessionId,
-                    messages: transcript.messages,
+                    messages: graftedTail,
                     snapshot: result.snapshot
                 )
             } else {
@@ -2955,7 +3025,15 @@ final class AppState: ObservableObject {
                 return false
             }
             turnState = .reconnecting
-            errorMessage = "Failed to restore this conversation: \(error.localizedDescription)"
+            if case DashboardTicketBridgeError.oversizedResponse = error {
+                // A known-oversized one-shot history response from an old
+                // backend: retrying the same giant transcript over the
+                // bounded WebSocket is not a recovery. Surface the
+                // controlled compatibility error instead.
+                errorMessage = error.localizedDescription
+            } else {
+                errorMessage = "Failed to restore this conversation: \(error.localizedDescription)"
+            }
             settleReconciliation(token, automaticSyncOperationID: automaticSyncOperationID)
             chatResumeCoordinator.abandonPendingAutomaticSync()
             return false
@@ -3045,6 +3123,7 @@ final class AppState: ObservableObject {
             markChatViewportReplacement()
             setActiveSessionState(id: runtimeSessionID, title: "New conversation")
             messages = []
+            persistedTranscriptWindow = nil
             noteChatViewportTranscriptReplacement()
             clearStreamingText()
             activeAssistantMessageId = nil
@@ -4525,6 +4604,7 @@ final class AppState: ObservableObject {
         markChatViewportReplacement()
         setActiveSessionState(id: nil, title: "New conversation")
         messages = []
+        persistedTranscriptWindow = nil
         clearStreamingText()
         activeAssistantMessageId = nil
         resetReasoningTurn()
@@ -4603,6 +4683,7 @@ final class AppState: ObservableObject {
         let acceptedSessionIDs = knownSessionIDs(for: sessionId)
         setActiveSessionState(id: sessionId)
         messages = []
+        persistedTranscriptWindow = nil
         clearStreamingText()
         activeAssistantMessageId = nil
         resetReasoningTurn()
@@ -6109,30 +6190,25 @@ final class AppState: ObservableObject {
     /// reconciliation distinguish a missing history source — which triggers
     /// the legacy full-transcript resume — from an unrelated failure, which
     /// must surface instead of silently degrading.
+    ///
+    /// The request is explicitly bounded (tail-anchored page of
+    /// `PersistedTranscriptPagination.pageSize` rows) so the transport and
+    /// normalization cost of opening a conversation is bounded no matter how
+    /// large the session grew. A response that carries the `order=latest`
+    /// pagination echo answers under the paginated tail contract; one
+    /// without it is a legacy one-shot full transcript and hydrates as
+    /// before.
     private func persistedTranscriptOutcome(
         sessionId: String,
         profile: String,
         using bridge: DashboardTicketBridge?
     ) async -> PersistedTranscriptOutcome {
-        let fetchOutcome: PersistedTranscriptFetchOutcome
-        if let persistedTranscript = chatResumeLifecycleOperations.persistedTranscript {
-            fetchOutcome = await persistedTranscript(sessionId, profile)
-        } else {
-            guard let bridge else { return .unavailable }
-            do {
-                fetchOutcome = .payload(try await bridge.requestJSON(
-                    path: sessionMessagesPath(sessionId: sessionId, profile: profile)
-                ))
-            } catch {
-                // requestJSON's readiness poll is the bounded wait for a cold
-                // or reloading bridge; whatever it raises is classified with
-                // the seam-sourced failures below. Note the bridge caps REST
-                // response size (DataURLLimits.maxJSONResponseBytes), so an
-                // oversized transcript also lands here rather than silently
-                // degrading the resume.
-                fetchOutcome = .failed(error)
-            }
-        }
+        let fetchOutcome = await fetchPersistedHistoryPayload(
+            sessionId: sessionId,
+            profile: profile,
+            query: PersistedTranscriptPagination.tailQuery(offset: 0),
+            using: bridge
+        )
 
         switch fetchOutcome {
         case .unavailable:
@@ -6142,8 +6218,9 @@ final class AppState: ObservableObject {
             // structural endpoint gaps, a bridge that never became ready
             // inside its bounded poll, and transient transport trouble
             // (429, 5xx, status-0 WebKit/network failures, request timeouts)
-            // degrade to the single legacy resume; authentication and every
-            // other failure must surface instead of silently degrading.
+            // degrade to the single legacy resume; authentication, a known-
+            // oversized one-shot history response, and every other failure
+            // must surface instead of silently degrading.
             if Self.historySourceIsUnavailable(error) {
                 sessionCatalogLog.debug("Persisted transcript unavailable for \(sessionId, privacy: .public): \(error.localizedDescription, privacy: .public)")
                 return .unavailable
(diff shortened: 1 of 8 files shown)


kaishi00:
## Summary
Implements bounded, tail-anchored pagination for persisted-transcript hydration (`?limit=120&offset=…&order=latest&include_compacted=true`), with identity-safe per-session window state, a "Load earlier messages" backfill affordance, reconnect tail-grafting, legacy/one-shot shape detection, a typed oversized-response error, and viewport re-pinning across prepends.

## Findings

### [minor] Unstructured backfill `Task` outlives the view and mutates `@State` after disappear
**File:** Conduit/Views/ChatView.swift:135-146
**Issue:** The button launches an unstructured `Task { @MainActor in … }` that is not tied to the view lifecycle. If the user leaves `ChatView` while a backfill is in flight, the task still runs `appState.loadEarlierMessages()` and mutates `viewport` (`@State`). `onDisappear`/`viewDisappeared()` has already cleared `pendingPrependAnchor`, so the discharge is a no-op and the prepend is applied with no re-pin — older rows silently appear when the user returns, and the fetch/transport is burned for a conversation the user left. The AppState ownership guard checks session/profile currency but not view lifecycle, so this path is reachable.
**Suggestion:** Store the task (e.g., `@State private var backfillTask: Task<Bool, Never>?`), cancel it in `onDisappear`, and guard on `Task.isCancelled` inside before mutating viewport state.

### [minor] Dedup in `prepending` silently breaks if a page row lacks a payload id
**File:** Conduit/Services/PersistedTranscriptWindow.swift:126-146 (with Conduit/Services/HermesClient.swift:1522-1526)
**Issue:** `prepending` keys dedup on `ChatMessage.id`. When a payload row has no `id`/`message_id`, `MessageNormalizer` falls back to `String(index)` — the page-local enumerated index. A boundary row shared between two consecutive pages then normalizes to different ids (e.g. `"119"` vs `"0"`), so the overlap boundary is never found and either the whole page is prepended (duplicate rows) or `firstIndex` collides on an unrelated row and the merge silently returns nil (affordance spins fetching duplicates forever while advancing offset).
**Suggestion:** This is documented as unreachable for real Hermes rows (they carry SQLite ids), but the failure mode is silent. Consider asserting/ignoring the positional fallback for backfilled pages (drop rows without a durable id from dedup consideration), so a future backend regression fails loudly instead of corrupting the transcript.

### [minor] Reconnect graft silently dropped when the session catalog is stale
**File:** Conduit/Services/AppState.swift:6378-6390
**Issue:** `priorWindowOwnsThisConversation` relies on `knownSessionIDs(for:)` for the runtime↔stored alias match. Right after a reconnect, before the catalog refreshes, `knownSessionIDs` may be empty, so the alias branch returns false. For a window opened under a runtime alias, `grafting` is then disabled and the reconnect replaces the entire transcript with the single newest page — truncating all backfilled history, the exact outcome the graft exists to prevent. The non-alias spelling is safe (direct id equality).
**Suggestion:** Fall back to the direct-continuity check when catalog knowledge is unavailable (e.g., compare against `previousTranscriptMessages.first?.id` anchor presence), or accept transient truncation but log it distinctly so the trade-off is observable.

### [minor] Misleading compatibility error when a paginated tail page itself is oversized
**File:** Conduit/Services/DashboardTicketBridge.swift:823-827 and Conduit/Services/AppState.swift:3028-3033
**Issue:** `oversizedResponse` originates from `response_too_large` on *any* `/messages` fetch. When a current (paginated) backend returns a page that exceeds `maxJSONResponseBytes` (e.g. `include_compacted=true` with multi-MB carrier rows transported before being dropped), the user is told "Update Hermes to enable paginated conversation history" — wrong advice for an up-to-date Hermes, and there is no retry path a smaller page could satisfy.
**Suggestion:** Distinguish tail-page oversize (retryable with a smaller `limit`, or surface a separate message) from the legacy one-shot oversize that motivated the typed condition.

### [nit] Pedantic: lingança — anchor can linger armed when the prepend landing deletes the front row
**File:** Conduit/Services/ChatViewportController.swift:427-456
**Issue:** The anchor-consumption guard intentionally skips when the previous front row is not retained (compaction rewrite "must not consume the anchor — the in-flight backfill's own landing still owns it"). But that landing is this same change; if it dropped the front row, `didPrepend == true` so the view never calls `prependAnchorDischarged`, leaving `pendingPrependAnchor` armed until a later session switch/drag/disappear — an unexpected, unrelated front-replacement change could then consume it and emit a stale scroll.
**Suggestion:** On the true prepend landing, discharge the anchor even when the retention guard prevents the scroll (consume-and-suppress), so it cannot later re-pin on an unrelated change.

### [nit] Positional fallback in `MessageNormalizer` weakens the "only strictly-older region" guarantee for non-Hermes payloads
**File:** Conduit/Services/HermesClient.swift:1526
**Issue:** Rows normalized without a payload id get `String(index)` identities that are page-local, defeating the `prepending` boundary math for any backend not supplying durable ids. Same root cause as the minor above; noted separately only because the docs in `prepending` assert this is unreachable.

## Verdict
COMMENT

The core pagination/window logic is well-reasoned, heavily guarded, and thoroughly tested (23 new tests, all green, per the PR description); I found no critical or major defect. The findings above are edge-case correctness/robustness concerns (view-lifetime task, positional-id dedup fragility, catalog-stale graft disable, oversize-error wording) that are worth a follow-up but do not block the change.

<a href="https://opencode.ai/s/TmrqXaMc">New%20session%20-%202026-08-31T09%3A18%3A45.691Z</a>
[opencode session](https://opencode.ai/s/TmrqXaMc)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33376933844)

kaishi00:
## Review-hardening pass (2eb30c4)

Targeted correctness pass addressing the remaining review findings — the bounded pagination architecture is unchanged.

**1. Backfilled history survives repeated reconciles.** An explicit `PersistedTranscriptWindowState.hasBackfilledPrefix` now carries the prefix fact: set when a backfill prepends, kept by a reconcile that grafts onto the prefix, cleared when a refreshed tail has no safe anchor. The graft gates on the flag instead of `nextOffset > pageSize`, so reconcile #2, #3, … keep the prefix even though coverage resets to the fresh tail. Regression: `testReconcileRefreshGraftsBackfilledPrefix` drives backfill → reconcile → reconcile (asserting the prefix and flag after each) plus the runtime/stored alias variant (`testGraftSurvivesRuntimeAliasReconcile`, also two reconciles).

**2. A slow bounded history request never becomes a giant WebSocket retry.** `historySourceIsUnavailable` is now structural-only (404/410/501 → exactly one legacy resume, unchanged). Timeouts, status-0, notReady (deadline or cold bridge), 408/429, other 5xx, oversized, and auth failures surface as a retryable restore error and retry the bounded request through reconnect recovery — never `openSession(compact: false)`. Upstream note: current Hermes can serve `include_compacted=true` slowly for heavily compacted sessions until upstream #97440 bounds that work; Conduit now fails/retries boundedly by design. Regressions: `testTransientHistoryFailuresSurfaceWithoutLegacyResume` (408/429/500/503/status-0/notReady → compact 1, legacy 0, error surfaced), `testColdDashboardBridgeBeginsCompactAndNeverEscalatesToLegacy`, `testStructuralHistoryEndpointGapsFallBackExactlyOnce` (unchanged), classification matrix.

**3. Backfill response ownership.** Every older page validates its returned `session_id`/`sessionId` against the requested ID, the window's resolved stored ID, and known aliases before normalizing — foreign responses never prepend nor advance coverage. Regression: `testForeignBackfillResponseRejected`. The stale session-switch discard test is retained (separate protection).

**4. Tool-call/page-boundary reconstruction.** Yes, production changes were required: `PersistedTranscriptWindow.reconcilingToolCallsAcrossBoundary` folds a trailing unmatched call card into its held standalone result card by durable `ToolActivity.id` (≤8 pairs) before prepending. Regression: `testToolCallSplitAcrossPageBoundaryReconstructsOneRun` (one card, correct input/output/completion, no orphan, no duplicate). Additionally, pages lacking durable persisted row IDs (including empty/zero IDs) can never enter dedup: initial hydration falls back to the one-shot read, backfill retires — `testBackfillPageWithoutDurableIdsRetiresAffordance`.

**5. Generic vs transcript-specific oversized.** The bridge error is now neutral ("This response is too large to load safely.") for any oversized dashboard operation; a legacy one-shot oversize maps to `LegacyTranscriptOversizedError` (compatibility copy) in the transcript layer, while a bounded-page oversize keeps the neutral copy — it never claims the backend lacks pagination. Tests: `testGenericOversizedBridgeErrorUsesNeutralCopy`, `testOversizedBoundedHistorySurfacesNeutralError`, `testOversizedLegacyOneShotSurfacesCompatibilityCopy`.

**Additional hardening:** the backfill task is view-lifetime-aware (stored, cancelled on disappear, cancellation-checked before viewport mutation); a prepend anchor whose row was deleted mid-flight discharges silently instead of scrolling to a dead ID.

**Verification:** focused suites 124/124; full unit suite 1457/1457 (0 failures) at 2eb30c4; `git diff --check` clean. Review gate: two-reviewer delta review; the one BLOCKER found (all-held-folded corner resurrecting a duplicate via the unfolded-page fallback) is fixed in this commit.

**Accepted trade-offs (documented):** a cold bridge or slow bounded read now fails the session open with a visible error instead of degrading to legacy content — the bounded retry rides reconnect recovery; the all-held-folded defensive branch is unreachable through public seams (a backfillable window requires ≥120 held rows, folds cap at 8) and is fixed defensively without a seam-level test.

kaishi00:
## Summary
Adds bounded, tail-anchored pagination for persisted-transcript hydration (a 120-row tail page via `?limit=120&offset=…&order=latest&include_compacted=true`), a single-flight "Load earlier messages" backfill with durable-ID dedup, reconnect tail-grafting that preserves the backfilled prefix, legacy/one-shot shape detection, a typed oversized-response error, and viewport re-pinning across prepends.

## Findings

### [minor] Backfill fold can silently skip the viewport re-pin and leave the anchor armed
**File:** Conduit/Services/AppState.swift:6509-6535, Conduit/Services/ChatViewportController.swift:427-436
**Issue:** When `reconcilingToolCallsAcrossBoundary` folds a trailing call card into a held standalone result card, the helper returns `foldedFromHeld` and `loadEarlierMessages` drops that many leading held rows (`messages.dropFirst(foldedFromHeld)`). The anchored row is captured from `viewport.stableTopMessageID` at tap time — and at the very top of the transcript (where the "Load earlier" button lives) that row is `messages[0]`, exactly the row the fold deletes when a page boundary splits a tool pair (bounded at ≤8 pairs). The resulting prepend therefore changes the front row while the old front row is **not retained**, so `transcriptChanged`'s retention guard (`frontRowIDBefore == nil || contains(frontRowIDBefore)`) fails, the anchor is neither consumed nor re-issued, **and** because `didPrepend == true` the view task never calls `prependAnchorDischarged`. The user's visible row is folded away without any re-pin, and `pendingPrependAnchor` lingers until a future front-row-replacing change consumes it (issuing a delayed scroll to a now-dead id, a silent no-op). This is reachable for tool-heavy transcripts across boundary backfills — not just the theoretical all-held-folded corner.
**Suggestion:** In the fold path, discharge the anchor when the previously-front row is among the folded/dropped held rows (consume-and-suppress: clear `pendingPrependAnchor` without scrolling), mirroring the "deleted mid-flight discharges silently" rule already used on the consume side.

### [minor] "Load earlier messages" scrolls off-screen after every prepend
**File:** Conduit/Views/ChatView.swift:839, Conduit/Services/ChatViewportController.swift:384-392
**Issue:** The prepend re-pin scrolls the anchored front message to the viewport top with `anchor: .top`. The backfill affordance sits *above* the messages in the same LazyVStack (ChatView.swift:245-247), so each prepend pushes it off the visible area by its own height; the user must scroll up ~44pt between every backfill to tap it again, making multi-page backfilling tedious.
**Suggestion:** Either scroll with a leading-inset anchor so the affordance stays visible above the re-pinned row, or keep the button pinned as a fixed `.overlay` header (like the existing bottom-follow button) so it remains reachable across prepends.

### [minor] Alias-aware graft/ownership gates depend on catalog freshness right after a reconnect
**File:** Conduit/Services/AppState.swift:6384-6396, 6599-6607, 4589-4596
**Issue:** `priorWindowOwnsThisConversation` and `persistedTranscriptWindowOwnershipIsCurrent` fall back to `knownSessionIDs(for:)`, which returns just `{sessionID}` when the session is absent from `sessions + cronSessions`. A window opened under a runtime alias loses the alias branch until the catalog refreshes -- so on the very reconcile the graft exists to protect, the gate can fail and the reconnect truncates the backfilled prefix back to one page. `hasBackfilledPrefix` fixed the repeated-reconcile case but not the stale-catalog gate itself.
**Suggestion:** When the catalog returns no alias knowledge, fall back to a structural check (e.g., grafting can additionally anchor on the prior window's resolved/reported `session_id` from the response) or log the disabled-graft distinctly so the trade-off is observable.

### [minor] `transcriptBackfillAvailable` gates on profile but not session identity
**File:** Conduit/Views/ChatView.swift:94-102
**Issue:** The affordance shows when `window.profile == activeProfile` and `!messages.isEmpty`, but the window's *session* is not compared to the active session. During a session switch the old window can briefly remain while the new transcript renders, showing a "Load earlier messages" button that belongs to the previous conversation (ownership is re-validated only when tapped, inside AppState).
**Suggestion:** Include the session check (e.g., `activeSessionId` equivalence or a direct `persistedTranscriptWindowOwnershipIsCurrent`-style check) in `transcriptBackfillAvailable` so the button never appears for a stale window.

### [nit] Delayed prepend-anchor retry never re-validates target existence
**File:** Conduit/Services/ChatViewportController.swift:1029-1034
**Issue:** `isCommandCurrent` for `.prependAnchor` checks session/mode/generation but not that the anchored row still exists in `targets`; the initial consume checks existence at that moment, but a rewrite within the 150 ms retry window leaves a silent no-op scroll. Harmless, but the retry then doesn't re-pin at all.
**Suggestion:** Re-check `targetCache.targets.contains(id)` inside `isCommandCurrent`'s `.prependAnchor` case so a dead anchor retry is suppressed consistently.

### [nit] Anchor armed under a nil session key is discharged across sessions
**File:** Conduit/Services/ChatViewportController.swift:377-382
**Issue:** `prependAnchorDischarged(matching:)` clears any anchor whose `sessionKey == nil` regardless of the passed session. If a backfill was armed before the view had any session key, a *different* session's discharge would clear it. Requires the backfill to complete before the view learns a session key; strictly theoretical given the button requires a hydrated transcript, but the guard is asymmetric with the non-nil case.
**Suggestion:** Require at least one side to be non-nil, or match nil-with-nil only.

## Verdict
COMMENT

The architecture is sound and unusually well-guarded (identity-safe window state, durable-ID dedup, structural-only retirement, viewport re-pin state machine) and the claimed regression coverage is extensive. All findings above are edge-case correctness/UX concerns in the backfill/graft interaction; none block the change, but the fold-vs-viewport conflict (first finding) is worth a follow-up since it is reachable in tool-heavy sessions.

<a href="https://opencode.ai/s/vjBwo5tx">New%20session%20-%202026-08-31T11%3A16%3A35.055Z</a>
[opencode session](https://opencode.ai/s/vjBwo5tx)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33386120722)

kaishi00:
## Summary
Adds bounded, tail-anchored pagination for persisted-transcript hydration (`?limit=120&offset=…&order=latest&include_compacted=true`), with identity-safe per-session window state, a single-flight "Load earlier messages" backfill with durable-ID dedup, reconnect tail-grafting that preserves the backfilled prefix across repeated reconciles, legacy/one-shot shape detection, a typed oversized-response error, and viewport re-pinning across prepends (23 new tests, all green per the PR).

## Findings

### [minor] Cross-page tool-fold can consume the anchored row without re-pinning, leaving the prepend anchor armed
**File:** Conduit/Services/ChatViewportController.swift:422-439 (with Conduit/Services/AppState.swift:6509-6535)
**Issue:** When a page boundary splits a tool call/result pair, `reconcilingToolCallsAcrossBoundary` folds the leading held rows (`messages.dropFirst(foldedFromHeld)`) into the older page. The anchored row is `stableTopMessageID` captured at tap time — because the button renders above the first message row, that anchor is normally `messages[0]`, exactly the row the fold deletes when the boundary splits a tool pair. The consuming guard in `transcriptChanged` requires the previous front row to be retained (`frontRowIDBefore == nil || targets.contains(frontRowIDBefore)`); when the fold dropped it, the guard fails, so the anchor is neither consumed nor re-issued, and since `didPrepend == true` the view's discharge is never called. The re-pin is silently skipped and `pendingPrependAnchor` stays armed against a now-dead ID until some later front-replacing change consumes it (clearing it without scrolling). I verified the "unfolded-page fallback resurrecting a duplicate" variant flagged earlier is genuinely unreachable (folded vs. unfolded pages carry identical row IDs), but the missed re-pin/lingering anchor is not.
**Suggestion:** When the prepend's own landing drops the previously-front row (the fold case), treat it as consume-and-suppress — clear `pendingPrependAnchor` without scrolling — mirroring the deleted-mid-flight rule already applied on the inner guard.

### [minor] Backfill affordance gates on profile but not session identity
**File:** Conduit/Views/ChatView.swift:94-101
**Issue:** `transcriptBackfillAvailable` shows when the window's profile matches `activeProfile` and `messages` is non-empty, but never compares the window's `requestedSessionID` to the active session. During a session switch the previous window can briefly remain (its reset in AppState happens on the switch path), presenting a "Load earlier messages" button belonging to the prior conversation. Tapping is safe — `loadEarlierMessages` re-validates ownership and no-ops — but the button and its loading row can mislead during the race, and double-tap protection reads `isLoadingEarlier` from the stale window.
**Suggestion:** Include a session-equivalence check (e.g., `activeSessionId` matching, or reuse of the AppState ownership gate) so the affordance never renders for a stale window.

### [minor] Prepending re-pins to `.top`, pushing the "Load earlier" button out of the viewport after every page
**File:** Conduit/Views/ChatView.swift:841, 245-247 (with ChatViewportController.swift:378-380)
**Issue:** The prepend re-pin scrolls the anchored row to `anchor: .top`, but the affordance sits above the messages in the same `LazyVStack`. Each prepend therefore drives the button ~44pt off-screen, and multi-page backfilling requires the user to scroll back up between every tap.
**Suggestion:** Scroll with a leading-inset anchor so the control stays visible above the re-pinned row, or render the control as a fixed `.overlay` header (like the existing bottom-follow button) so it remains reachable across prepends.

### [nit] Delayed `.prependAnchor` retry never re-validates that the anchored row still exists
**File:** Conduit/Services/ChatViewportController.swift:1029-1034
**Issue:** `isCommandCurrent` for `.prependAnchor` checks session, mode, and generation but not whether the anchor ID is still present in the target cache. The initial consume validates existence at that moment, but a rewrite within the 150 ms retry window yields a scroll command for a dead ID — a silent no-op that leaves the prepend unanchored.
**Suggestion:** Re-check `targetCache.targets.contains(id)` inside the `.prependAnchor` case so a dead-anchor retry is suppressed consistently with the initial-consume check.

### [nit] `prependAnchorDischarged` clears an anchor armed under a nil session key across sessions
**File:** Conduit/Services/ChatViewportController.swift:377-383
**Issue:** `if armed.sessionKey == nil || identity.areEquivalent(armed.sessionKey, sessionKey)` clears any anchor whose `sessionKey` is nil regardless of the session passed, while the non-nil case is strictly matched. Asymmetric; requires a backfill armed before the view learned a session key (theoretically impossible given the button needs a hydrated transcript), but the guard costs nothing to tighten.
**Suggestion:** Match nil-with-nil only (require at least one side non-nil), mirroring the conservative non-nil path.

### [nit] Docstring coverage below the configured threshold
**File:** ConduitTests/CompactResumeTranscriptTests.swift and others (pre-merge check)
**Issue:** CodeRabbit's pre-merge check reports 73.91% docstring coverage (< 80%) across the 69 touched functions. This is a repo-tooling gate, not a functional defect, but the new `loadEarlierMessages`/window logic and the backfill controls are exactly the kind of stateful code a docstring gap obscures.
**Suggestion:** Add doc comments to the uncovered touched functions to satisfy the 80% gate.

## Verdict
COMMENT

The architecture is well-reasoned and unusually thoroughly guarded (identity-safe window state, durable-ID dedup with a loud rejection path, structural-only retirement, the graft-preserved-prefix flag, consume-and-suppress for dead anchors, view-lifetime-cancelled backfill task). I found no critical or major defect and no new retain cycles, force unwraps, or concurrency hazards; the false-claim edge cases from earlier rounds (unfolded-page duplicate resurrection, oversized copy leaking to non-transcript operations, overlap-gap correctness) have been correctly resolved. The findings above are edge-case UX/robustness concerns in the backfill/graft interaction — the fold-vs-anchor conflict (first finding) is the most substantive and is worth a follow-up since it's reachable in tool-heavy conversations.

<a href="https://opencode.ai/s/e2dxdQsa">New%20session%20-%202026-08-31T14%3A49%3A13.947Z</a>
[opencode session](https://opencode.ai/s/e2dxdQsa)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33404749135)

PR #25, opened 2026-08-09
Fix Cloudflare Access token injection origin leakage

kaishi00:
## Summary

- serialize Cloudflare Access client ID, secret, and expected dashboard URL as JSON string literals;
- only attach service-token headers to requests whose document and resolved request origins match the configured dashboard origin;
- restrict Cloudflare user scripts to WebKit main frames in both authentication WebViews.

## Security impact

Previously, the injected fetch/XHR hooks could expose service-token headers to subframes and cross-origin requests, and manual escaping could produce unsafe JavaScript for values containing backslashes or control characters. This PR fails closed for invalid dashboard URLs and preserves native initial requests and same-origin dashboard API/ticket requests.

This addresses the Cloudflare token-injection finding in #20. The other findings in that issue are intentionally out of scope for this PR.

## Verification

- Focused Cloudflare suite: 16/16 passed
- Full simulator suite: 353/353 passed
- `git diff --check`: passed

Related to #20

## Summary by CodeRabbit

* **Security Improvements**
  * Cloudflare Access credentials are sent only with valid, same-origin dashboard requests.
  * Credential injection is restricted to the dashboard’s main frame.
  * Invalid URLs, configuration, or request inputs fail safely without attaching credentials.
  * Supports fetch requests, request objects, and XMLHttpRequests while preventing forged cross-origin eligibility.
  * Credentials and URLs are safely serialized, including special characters.

* **Documentation**
  * Added design and implementation documentation covering credential origin protection.

PR #119, opened 2026-08-31
fix(auth): restore Cloudflare Access sign-in and keyboard-safe login form

### LINKED ISSUE #117: BUG: Connection to Agent behind Cloudflare not Possible
Hello

I have my Hermes Agent running on my VPS and it is reachable through Cloudflare. I also have an CF authentication setup to protect access to my agent.
After installing Hermes Conduit to my iPhone I entered my https URL and Dashboard User / PW into your App.
Next a popup comes up from Cloudflare asking me to logon using a CF User first. However, the popup does not accept my user and pass and throws an error instead: there was a problem with verification.
So I cannot login and cannot use the app. I also created a cloudflare service token and put it into the app, which is a real pain because the keyboard covers the input fields! Then the CF Popup came up again and failed again with the verification.

When I use my browser in IOS I can pass the CF Logon without any Problem and access my Hermes Agent using my browser. Only with your App-embedded CF Logon popup it fails.

Can you please help me to fix this probem or provide a fixed app version? I wanted to use the app because it claims to support Cloudflare but it does not work for me.

Thank you so much.

kaishi00:
Closes #117.

## Root cause — Conduit's navigation policy blocked documents Cloudflare's Turnstile requires

Regression archaeology (diffing every commit that touched `LoginView.swift`/`AuthWebView`, plus #25, #38, and the native-auth rework) shows the WebView shell is essentially unchanged since the initial public release — this is not a code regression but a latent incompatibility with how Cloudflare's interactive verification operates inside WebViews.

`AuthWebView.Coordinator.decidePolicyFor` (and `DashboardTicketBridge`'s, identically) applied the HTTPS-only transport guard **before** the subframe allowance:

```
guard ConnectionURLPolicy.isAllowedTransport(url) else { cancel }   // https/http + host only
if !targetFrame.isMainFrame { allow }                                // never reached for about:
```

Cloudflare's [Turnstile mobile implementation requirements](https://developers.cloudflare.com/turnstile/get-started/mobile-implementation/) state that an embedding WebView must **"Allow connections to `about:blank` and `about:srcdoc`"** (alongside JavaScript, storage, a consistent User Agent, and reachability of `challenges.cloudflare.com`). Conduit cancelled every `about:` subframe navigation — including the two Turnstile names — so verification could not complete inside the app while the same login succeeded in Safari. This produces exactly the reported failure: the Cloudflare login page renders, then "there was a problem with verification".

### Runtime evidence (WKWebView, iPhone 17 Pro simulator, iOS 26.5)

A probe page embedding `<iframe src="about:blank">` and `<iframe srcdoc="…">` was loaded under two navigation delegates:

- **PR's previous policy (subframe `about:` cancelled):** WebKit consulted the delegate for **both** `about:blank` and `about:srcdoc` subframe navigations, both were cancelled, and the `srcdoc` iframe's document was never instantiated (`contentDocument.body.innerHTML === ""`).
- **Fixed policy (subframe allowance):** both navigations allowed; the `srcdoc` document instantiates (`<p>…</p>`).

So the delegate is demonstrably in the path for exactly the documents Cloudflare requires, and cancellation demonstrably suppresses them. (The originally suspected mechanism — Cloudflare rejecting the default WKWebView User Agent — was **withdrawn**: the same Cloudflare page requires only a *consistent* User Agent and explicitly recommends maintaining the default WebView UA. The `WebViewUserAgent` spoof from the previous iteration of this PR has been removed, and with it the tests that asserted the unsupported causal claim.)

### 1. Narrow subframe allowance (`about:blank` / `about:srcdoc`)

- `ConnectionURLPolicy.isTurnstileRequiredSubframeDocument` recognizes exactly `about:blank` and `about:srcdoc`; `isAllowedWebViewSubframeTransport` = ordinary HTTP(S) transport **or** those documents.
- Both `decidePolicyFor` implementations now classify **subframes first**: HTTP(S) subframes (identity providers, `challenges.cloudflare.com`) and the two Turnstile documents are allowed; every other scheme (`about:` variants, custom schemes, `data:`) is denied.
- The allowance is **subframe-only**. `about:` documents are opaque and cannot navigate the top level, so the main-frame rules are untouched: unpinned sign-in may still traverse the legitimate Cloudflare/IdP redirect chain; once the dashboard origin pins, top-level navigation still cannot escape it (`about:` main-frame navigations remain cancelled — pinned or not). The User Agent is left exactly as WebKit ships it.
- The auth WebView's policy is extracted into `AuthWebViewNavigationPolicy.decide(url:isMainFrame:hasPinnedDashboardOrigin:expectedDashboardURL:)` so the entire boundary is unit-testable without hosting a WKWebView.

### 2. Service-token mode (unchanged, re-verified)

- `CF-Access-Client-Id/Secret` remain attached to every dashboard-origin native request (providers → password-login → ws-ticket → WebSocket upgrade) and to the initial WebView `/login` request; the injected fetch/XHR script stays scoped to same-origin dashboard requests only.
- Rejected tokens remain a **typed diagnostic**: provider discovery still redirected to a `*.cloudflareaccess.com` login origin with a token configured throws `AuthClientError.cloudflareServiceTokenRejected` ("…verify the Client ID / Secret and that the token is allowed by a Service Auth policy for this Access application, or turn off \"Use Cloudflare Access service token\" to sign in interactively."). Token values are never logged, never in errors, never replayed cross-origin (regression-tested: 0 requests reach the tenant host).
- No new auth mode; normal non-Cloudflare password login is untouched.

### 3. Keyboard-safe login form (unchanged from review)

Responsive `ScrollView`/`ScrollViewReader` layout, focus-driven scrolling of the Cloudflare fields, keyboard **Done** control, return-key focus chain (password return hands focus to the Cloudflare Client ID instead of triggering Connect), typed token-rejection diagnostic, Keychain storage with origin binding. Two review findings addressed: the stale backdrop tap-to-dismiss gesture (dead behind the full-screen scroll view) is removed with an accurate comment — dismissal is owned by interactive scroll-dismiss, the Done control, and field-to-field focus movement — and Connect-button enablement now uses trimmed emptiness, consistent with `connect()`'s guard.

## Tests

- `AuthWebViewNavigationPolicyTests` (new) — the full boundary matrix: Turnstile subframe documents allowed; HTTPS subframes (identity providers, `challenges.cloudflare.com`) allowed; `about:` variants / custom schemes / `data:` / nil denied; unpinned main-frame auth chain allowed while insecure/malformed/about: transports stay blocked; pinned main-frame stays dashboard-origin-bound (including suffix-spoof hosts and `about:` escapes) and fails closed on an unknown expected origin.
- `TurnstileSubframeBoundaryTests` (new) — live-WKWebView regression: the shipped subframe policy lets the `about:blank` / `about:srcdoc` documents instantiate in a real engine (the srcdoc document is asserted present), and custom-scheme subframes stay denied.
- `NativeAuthClientTests` — token headers present on every dashboard-origin request and absent cross-origin (redirect cancelled, 0 requests to the tenant host); provider-discovery matrix (no token → WebView fallback, rejected token → typed error, non-CF redirect → fallback, normal/3xx/5xx preserved); full-sequence token propagation with a cookie-gated ws-ticket; redirect-classifier adversarial cases.
- `LoginFieldNavigationTests` — the focus-chain contract under both token-entry states.
- `LoginKeyboardUITests` — real-gesture keyboard coverage on an iPhone simulator.
- UA-composition tests from the previous iteration were **removed** with the withdrawn mechanism.

## Verification

- Focused suites (AuthWebViewNavigationPolicy, TurnstileSubframeBoundary, CloudflareAccess, NativeAuthClient, NativeAuthClientIntegration, LoginFieldNavigation): all green.
- Full `ConduitTests`: green on the Mac build host (iPhone 17 Pro simulator) at the pushed commit.
- `git diff --check` clean.
- Multi-model review cycle run on the earlier iteration; this rework supersedes its UA-related items with the documented evidence above.
- Known local limitation: the UI-test runner on the Mac build host wedges at bootstrap ("Timed out waiting for AX loaded notification") — a recurring local-environment failure; CI's UI lane is the execution gate for `LoginKeyboardUITests`.
- A real-tenant interactive Cloudflare Access login was not available for manual verification; the mechanism is established by Cloudflare's documented requirements plus the WKWebView probe above.

## Security hardening: service tokens are HTTPS/WSS-only

`CloudflareAccessCredentials.applying(to:)` now fails closed for any non-HTTPS/WSS URL, and that single guard covers every application path (native provider discovery, password login, ticket minting, the WebSocket upgrade, the auth WebView's initial `/login` request, and the ticket bridge's `/api/status` load). The document-start fetch/XHR injection shares the invariant: a plain-HTTP configured origin gets no script at all, so the secret is never embedded in page JavaScript either. Plain-HTTP/wss-less transports silently omit the credentials — trusted-LAN Hermes connectivity is unaffected, and a configured-but-unusable token over HTTP never blocks a local login. Regression tests: cleartext HTTP and `ws://` requests carry no CF headers while `https`/`wss` carry both (both header names asserted); the WebView initial request over HTTP omits them; an HTTP LAN dashboard completes the full native sequence with a configured token and zero CF headers on any request; the injection script fails closed for HTTP origins; existing cross-origin/IdP/Cloudflare-login-origin leak tests remain green.

## Out of scope

No new auth mode, no changes to Hermes server auth, no weakening of origin scoping or of the main-frame dashboard-origin boundary, no unrelated login-screen redesign.

## Summary by CodeRabbit

- **Security**
  - Cloudflare Access credentials are now sent only over secure HTTPS and WSS connections.
  - Credentials are no longer embedded in cleartext WebView requests or scripts.

- **Bug Fixes**
  - Added a clear error when a Cloudflare Access service token is rejected.
  - Improved WebView sign-in compatibility with Turnstile and identity-provider pages.
  - Login validation now ignores whitespace-only entries.

- **Usability**
  - Login fields and the Connect button remain accessible while the keyboard is open.

### DIFF
--- a/Conduit/Services/CloudflareAccess.swift
+++ b/Conduit/Services/CloudflareAccess.swift
@@ -1,30 +1,4 @@
 import Foundation
-import WebKit
-
-/// Cloudflare's Access login page runs a Turnstile-backed verification that
-/// scores the bare WKWebView user agent as automation, so the interactive
-/// sign-in fails inside the app while the same login succeeds in Safari on
-/// the same device (issue #117). Appending the tokens Safari carries but a
-/// default WKWebView omits ("Version/… Safari/…") before the first load
-/// lets the legitimate Cloudflare/IdP flow verify. "Safari/604.1" is Apple's
-/// frozen WebKit build marker (an identity token, not an OS version — it has
-/// been constant across every iOS Safari release), and "Version/…" tracks
-/// the current OS; the device part of the UA keeps coming from WebKit
-/// itself. No credentials or navigation policy are affected.
-enum WebViewUserAgent {
-    static func safariCompatibilitySuffix() -> String {
-        let version = ProcessInfo.processInfo.operatingSystemVersion
-        // Safari's Version token is major.minor only — the OS patch level
-        // never appears in a real iOS UA, so omit it to stay byte-faithful.
-        return "Version/\(version.majorVersion).\(version.minorVersion) Safari/604.1"
-    }
-
-    /// Must run before the web view loads anything: WebKit ignores user-agent
-    /// changes once a page has started loading.
-    static func apply(to configuration: WKWebViewConfiguration) {
-        configuration.applicationNameForUserAgent = safariCompatibilitySuffix()
-    }
-}
 
 struct CloudflareAccessKeychainRecord: Codable, Equatable {
     let clientID: String
--- a/Conduit/Services/CloudflareAccess.swift
+++ b/Conduit/Services/CloudflareAccess.swift
@@ -14,11 +14,9 @@ import WebKit
 enum WebViewUserAgent {
     static func safariCompatibilitySuffix() -> String {
         let version = ProcessInfo.processInfo.operatingSystemVersion
-        var versionText = "\(version.majorVersion).\(version.minorVersion)"
-        if version.patchVersion > 0 {
-            versionText += ".\(version.patchVersion)"
-        }
-        return "Version/\(versionText) Safari/604.1"
+        // Safari's Version token is major.minor only — the OS patch level
+        // never appears in a real iOS UA, so omit it to stay byte-faithful.
+        return "Version/\(version.majorVersion).\(version.minorVersion) Safari/604.1"
     }
 
     /// Must run before the web view loads anything: WebKit ignores user-agent
--- a/Conduit/Services/CloudflareAccess.swift
+++ b/Conduit/Services/CloudflareAccess.swift
@@ -1,4 +1,32 @@
 import Foundation
+import WebKit
+
+/// Cloudflare's Access login page runs a Turnstile-backed verification that
+/// scores the bare WKWebView user agent as automation, so the interactive
+/// sign-in fails inside the app while the same login succeeds in Safari on
+/// the same device (issue #117). Appending the tokens Safari carries but a
+/// default WKWebView omits ("Version/… Safari/…") before the first load
+/// lets the legitimate Cloudflare/IdP flow verify. "Safari/604.1" is Apple's
+/// frozen WebKit build marker (an identity token, not an OS version — it has
+/// been constant across every iOS Safari release), and "Version/…" tracks
+/// the current OS; the device part of the UA keeps coming from WebKit
+/// itself. No credentials or navigation policy are affected.
+enum WebViewUserAgent {
+    static func safariCompatibilitySuffix() -> String {
+        let version = ProcessInfo.processInfo.operatingSystemVersion
+        var versionText = "\(version.majorVersion).\(version.minorVersion)"
+        if version.patchVersion > 0 {
+            versionText += ".\(version.patchVersion)"
+        }
+        return "Version/\(versionText) Safari/604.1"
+    }
+
+    /// Must run before the web view loads anything: WebKit ignores user-agent
+    /// changes once a page has started loading.
+    static func apply(to configuration: WKWebViewConfiguration) {
+        configuration.applicationNameForUserAgent = safariCompatibilitySuffix()
+    }
+}
 
 struct CloudflareAccessKeychainRecord: Codable, Equatable {
     let clientID: String
--- a/Conduit/Services/CloudflareAccess.swift
+++ b/Conduit/Services/CloudflareAccess.swift
@@ -40,6 +40,14 @@ struct CloudflareAccessCredentials: Equatable, CustomStringConvertible {
 
     func applying(to request: URLRequest) -> URLRequest {
         guard isConfigured else { return request }
+        // Service tokens are HTTPS/WSS-only. Cleartext transports — plain
+        // HTTP dashboards and ws:// upgrades on trusted LANs — must never
+        // carry the long-lived Cloudflare credentials, even when a token is
+        // configured for the session; local Hermes connectivity works
+        // without them, so they are silently omitted rather than blocking
+        // the login.
+        let scheme = request.url?.scheme?.lowercased()
+        guard scheme == "https" || scheme == "wss" else { return request }
         var request = request
         request.setValue(clientID, forHTTPHeaderField: "CF-Access-Client-Id")
         request.setValue(clientSecret, forHTTPHeaderField: "CF-Access-Client-Secret")
@@ -58,10 +66,14 @@ struct CloudflareAccessCredentials: Equatable, CustomStringConvertible {
     /// `fetch()` and `XMLHttpRequest` calls inside the WKWebView include the
     /// Cloudflare Access service-token headers. The script is intentionally
     /// scoped to the configured dashboard origin; native requests still apply
-    /// the headers directly to the initial dashboard request.
+    /// the headers directly to the initial dashboard request. Shares the
+    /// HTTPS-only invariant of `applying(to:)`: plain-HTTP dashboard origins
+    /// get no script at all, so the credentials are never embedded in page
+    /// JavaScript either.
     func fetchInjectionUserScript(expectedBaseURL: String) -> String {
         guard isConfigured,
               let normalizedBaseURL = try? ConnectionURLPolicy.normalizedBaseURL(expectedBaseURL),
+              normalizedBaseURL.lowercased().hasPrefix("https://"),
               let baseURLLiteral = javaScriptStringLiteral(normalizedBaseURL),
               let idLiteral = javaScriptStringLiteral(clientID),
               let secretLiteral = javaScriptStringLiteral(clientSecret) else { return "" }
--- a/Conduit/Services/ConnectionURLPolicy.swift
+++ b/Conduit/Services/ConnectionURLPolicy.swift
@@ -26,6 +26,34 @@ enum ConnectionURLPolicy {
         return scheme == "https" || (scheme == "http" && isInsecureTransportAllowed(host))
     }
 
+    /// The `about:` documents Cloudflare's Turnstile WebView requirements
+    /// call out ("Allow connections to `about:blank` and `about:srcdoc`").
+    /// Allowed only in SUBframes. Those documents inherit the parent page's
+    /// origin, so script inside them could attempt a top-level navigation —
+    /// but any such navigation is evaluated as a MAIN frame against the
+    /// strict transport/dashboard-origin rules, so this allowance never
+    /// widens the top-level boundary.
+    static func isTurnstileRequiredSubframeDocument(_ url: URL?) -> Bool {
+        // Classify from the absolute string: WebKit's own URL objects for
+        // these subframe navigations do not necessarily expose the document
+        // tail through `path`/`host` the way a string-constructed URL does
+        // (observed directly in the boundary tests).
+        guard var raw = url?.absoluteString.lowercased() else { return false }
+        guard raw.hasPrefix("about:") else { return false }
+        raw.removeFirst("about:".count)
+        // Strip any trailing slashes WebKit may append to about: URLs.
+        while raw.hasSuffix("/") { raw.removeLast() }
+        return raw == "blank" || raw == "srcdoc"
+    }
+
+    /// Transport policy for WebView SUBFRAME navigations: ordinary HTTP(S)
+    /// (identity providers, challenges.cloudflare.com) plus the about:
+    /// documents Turnstile requires. Everything else (custom schemes,
+    /// other about: variants, data:) is denied.
+    static func isAllowedWebViewSubframeTransport(_ url: URL?) -> Bool {
+        isAllowedTransport(url) || isTurnstileRequiredSubframeDocument(url)
+    }
+
     static func normalizedBaseURL(_ value: String) throws -> String {
         let trimmed = value.trimmingCharacters(in: .whitespacesAndNewlines)
         guard !trimmed.isEmpty,
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -274,7 +274,6 @@ final class DashboardTicketBridge: NSObject {
         self.readinessPollInterval = readinessPollInterval
         self.requestDeadlineGraceMilliseconds = max(0, requestDeadlineGraceMilliseconds)
         let configuration = WKWebViewConfiguration()
-        WebViewUserAgent.apply(to: configuration)
         configuration.websiteDataStore = .default()
         if let script = cloudflareAccess?.fetchInjectionUserScript(expectedBaseURL: normalizedBaseURL), !script.isEmpty {
             configuration.userContentController.addUserScript(
@@ -682,12 +681,22 @@ extension DashboardTicketBridge: WKNavigationDelegate {
         decidePolicyFor navigationAction: WKNavigationAction,
         decisionHandler: @escaping (WKNavigationActionPolicy) -> Void
     ) {
-        guard ConnectionURLPolicy.isAllowedTransport(navigationAction.request.url) else {
-            decisionHandler(.cancel)
+        // Subframe check must come BEFORE the main-frame transport guard:
+        // Turnstile's WebView requirements call for about:blank and
+        // about:srcdoc subframe documents, which plain HTTP(S) transport
+        // rules would reject. The allowance is confined to subframes —
+        // a top-level navigation attempt from such a document re-enters
+        // this delegate as a MAIN frame and still must match the dashboard
+        // origin below.
+        if let targetFrame = navigationAction.targetFrame, !targetFrame.isMainFrame {
+            decisionHandler(
+                ConnectionURLPolicy.isAllowedWebViewSubframeTransport(navigationAction.request.url)
+                    ? .allow : .cancel
+            )
             return
         }
-        if let targetFrame = navigationAction.targetFrame, !targetFrame.isMainFrame {
-            decisionHandler(.allow)
+        guard ConnectionURLPolicy.isAllowedTransport(navigationAction.request.url) else {
+            decisionHandler(.cancel)
             return
         }
         guard let expectedURL = URL(string: baseURL),
--- a/Conduit/Services/DashboardTicketBridge.swift
+++ b/Conduit/Services/DashboardTicketBridge.swift
@@ -274,6 +274,7 @@ final class DashboardTicketBridge: NSObject {
         self.readinessPollInterval = readinessPollInterval
         self.requestDeadlineGraceMilliseconds = max(0, requestDeadlineGraceMilliseconds)
         let configuration = WKWebViewConfiguration()
+        WebViewUserAgent.apply(to: configuration)
         configuration.websiteDataStore = .default()
         if let script = cloudflareAccess?.fetchInjectionUserScript(expectedBaseURL: normalizedBaseURL), !script.isEmpty {
             configuration.userContentController.addUserScript(
--- a/Conduit/Services/NativeAuthClient.swift
+++ b/Conduit/Services/NativeAuthClient.swift
@@ -107,12 +107,14 @@ struct NativeAuthClient {
         case 301, 302, 303, 307, 308:
             // The SecureRedirectDelegate cancels cross-origin redirects, so a
             // 3xx final response here is the edge bouncing us to its sign-in
-            // page. Without a token that is the expected interactive-auth
-            // signal (empty list → WebView fallback). With a configured
-            // service token it means Cloudflare rejected that token — say so
-            // instead of presenting the same login page that should have
-            // been bypassed.
-            if cloudflareAccess != nil,
+            // page. Without a configured token that is the expected
+            // interactive-auth signal (empty list → WebView fallback). With
+            // an actually configured service token it means Cloudflare
+            // rejected that token — say so instead of presenting the same
+            // login page that should have been bypassed. Match
+            // `applying(to:)`'s configuration state: a non-nil but empty
+            // credentials value sends no headers and must fall back too.
+            if cloudflareAccess?.isConfigured == true,
                Self.redirectsToCloudflareAccessLogin(http) {
                 throw AuthClientError.cloudflareServiceTokenRejected
             }
--- a/Conduit/Services/NativeAuthClient.swift
+++ b/Conduit/Services/NativeAuthClient.swift
@@ -21,6 +21,12 @@ enum AuthClientError: LocalizedError {
     case loginFailed(String)
     case ticketFailed(String)
     case providerDiscoveryFailed(String)
+    /// A configured Cloudflare Access service token did not satisfy the
+    /// edge: provider discovery was still answered with a redirect to the
+    /// Cloudflare Access login page. Distinct from `.loginFailed`/`.ticketFailed`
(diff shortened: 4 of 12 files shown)


kaishi00:
## Summary
Fixes Cloudflare Access interactive sign-in (issue #117) by appending Safari-style UA tokens to the WKWebView before first load, adds a typed `cloudflareServiceTokenRejected` diagnostic when a configured service token still bounces provider discovery to a `*.cloudflareaccess.com` login origin, and reworks the login form into a keyboard-aware, focus-chain-navigable ScrollView with unit + UI tests.

## Findings

### [severity: minor] Backdrop tap-to-dismiss is now unreachable (keyboard dismissal regression)
**File:** Conduit/Views/LoginView.swift:36-44, 52-78
**Issue:** The backdrop's `onTapGesture { focusedField = nil }` sits on `ConduitBackdrop` in the `ZStack`, but it is now fully covered by the new `GeometryReader`→`ScrollView` layer that fills the entire screen. A `ScrollView` hit-tests across its whole frame, so taps on the previously-empty backdrop area land on the scroll view and never reach the gesture. The documented behavior ("Tapping the backdrop … clears focus so the keyboard dismisses") no longer works; dismissal now relies solely on the Done toolbar and interactive scroll.
**Suggestion:** Move the tap-to-dismiss onto the ScrollView's empty content (e.g. a transparent hit-testable overlay inside the ScrollView, or `scrollDismissesKeyboard(.immediately)` for an extra dismissal path), or update the comment to reflect that backdrop taps are no longer a dismissal path.

### [severity: minor] Whitespace-only credentials pass the button check but fail `connect()` guard
**File:** Conduit/Views/LoginView.swift:269 vs 304-309
**Issue:** Connect's `.disabled(... username.isEmpty || password.isEmpty ...)` uses raw `.isEmpty`, while `connect()` rejects *trimmed*-empty values. A username/password of only spaces leaves the button enabled and causes a remote 401-style error flow instead of the intended local "Enter your dashboard username and password." message.
**Suggestion:** Apply the same trimmed check (or a `nil`-coalesced trimmed value) to the `.disabled` condition so button state and validation agree.

### [severity: nit] Programmatic focus-scroll may fight the keyboard's own auto-scroll
**File:** Conduit/Views/LoginView.swift:59-64
**Issue:** Every focus change triggers `withAnimation(easeOut 0.25) { proxy.scrollTo(field, anchor: .center) }`. iOS also auto-scrolls a first responder into view in a ScrollView, so the two can animate concurrently and the final position can settle differently than intended (about the keyboard height) on slower devices. The reveal-triggered scroll (`:65-76`) additionally mutates `revealCloudflareSection` inside `onChange` and defers via `DispatchQueue.main.async`, which can scroll to a no-longer-present id if the toggle is flipped again before the block runs (harmless no-op, but fragile).
**Suggestion:** Consider scrolling only when the focused field is not already within the safe-area-visible bounds (e.g. using the ScrollView proxy's frame computation), and invalidating the deferred scroll work if the section is toggled off before it fires.

### [severity: nit] Token-rejection classifier can hard-block previously-working interactive paths
**File:** Conduit/Services/NativeAuthClient.swift:115-118, 279-284
**Issue:** With any non-nil token, a 3xx to `*.cloudflareaccess.com` now **throws** and fully blocks the WebView fallback (the user cannot reach interactive sign-in until they toggle off "Use Cloudflare Access service token"). If an Access policy is path-scoped (token valid for `/login` but not `/api/auth/providers`), this is arguably still correct — the WebView would carry the same headers and bounce identically — so the diagnostic text is the better outcome. Noting it only as a UX consequence worth a sanity check: the classifier is host-exact and spoof-resistant, and relative `Location` values correctly resolve to the dashboard origin, so no correctness or security issue was found.
**Suggestion:** None required if the intended UX is "diagnostic over silent detour"; optional hardening would be to only classify when the redirect chains from a *providers* request (already the only call site here).

## Verdict
APPROVE

No retain cycles, force unwraps, or concurrency violations were introduced (the new `connect()` `@MainActor` annotation and `[weak self]`/`[weak webView]` captures are correct; the classifier is dismissed as safe against spoofing). The two minor findings are UX regressions/inconsistencies, not blockers.

<a href="https://opencode.ai/s/jMBhOtFB">New%20session%20-%202026-08-31T17%3A03%3A43.652Z</a>
[opencode session](https://opencode.ai/s/jMBhOtFB)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33417392892)

kaishi00:
## Summary
Fixes Cloudflare Access interactive sign-in (#117) by allowing the `about:blank`/`about:srcdoc` subframe documents Turnstile requires (classified subframe-first in both the auth WebView and the ticket bridge), extracts the navigation boundary into a unit-testable `AuthWebViewNavigationPolicy`, and reworks the login form into a keyboard-aware ScrollView with a return-key focus chain plus added unit/live-WKWebView/UI tests.

## Findings

### [severity: minor] Service-token credentials are applied to non-HTTPS (cleartext) requests
**File:** Conduit/Services/NativeAuthClient.swift:218 (also Conduit/Views/LoginView.swift:497, Conduit/Services/DashboardTicketBridge.swift:657, and the fetch-injection origin guard in Conduit/Services/CloudflareAccess.swift:84)
**Issue:** `cloudflareAccess?.applying(to:)` prefixes `CF-Access-Client-Id` / `CF-Access-Client-Secret` onto every native request, the WebView's initial `/login` request, the bridge's initial `/api/status` request, and same-origin `fetch`/XHR — with no scheme check. `ConnectionURLPolicy.isAllowedTransport` deliberately permits plain HTTP for loopback/private-LAN/Tailscale hosts, so a token configured against an HTTP dashboard transmits this long-lived access credential as cleartext HTTP headers, sniffable by any co-tenant on that local network. For example, a user reaching a dashboard at `http://192.168.1.x:9119` with a token would leak the Cloudflare Access secret for nothing in return.
**Suggestion:** Gate token application on `https` (e.g., in `applying(to:)`, return the request unchanged unless `url.scheme == "https"`, or have each construction site require HTTPS before passing `cloudflareAccess`), and add the same scheme check to `shouldAttach` in the injection script so the boundaries stay consistent. Preserves behavior for the HTTPS cases the feature targets.

### [severity: minor] Token-rejection classifier checks `cloudflareAccess != nil` instead of `.isConfigured`
**File:** Conduit/Services/NativeAuthClient.swift:115
**Issue:** `CloudflareAccessCredentials` can be non-`nil` yet unconfigured (empty client ID/secret — reachable via the public memberwise init and in tests). Such an instance adds no headers (`applying(to:)` no-ops), but `authProviders()` would still throw `cloudflareServiceTokenRejected` on a `*.cloudflareaccess.com` redirect, suppressing the interactive WebView fallback that the surrounding comment promises for the no-token case. Not reachable through LoginView (its factory returns `nil` when unconfigured), but the API permits it.
**Suggestion:** Use `cloudflareAccess?.isConfigured == true` before classifying.

### [severity: nit] Whitespace-only dashboard URL returns silently via the keyboard submit path
**File:** Conduit/Views/LoginView.swift:301
**Issue:** `connect()` still `guard !cleaned.isEmpty else { return }`. The Connect button now uses `connectInputsArePresent` (trimmed), but the return-key chain can still reach `connect()` with a whitespace-only server URL (e.g., password return with service token off), producing no feedback — the same inconsistency CodeRabbit flagged.
**Suggestion:** Set the `invalidURL` error and focus `.server` instead of returning silently, matching the button's trimmed check.

### [severity: nit] Deferred focus-scroll can fire after its target unmounts
**File:** Conduit/Views/LoginView.swift:59-70
**Issue:** `revealCloudflareSection` is a one-shot flag (cleared inside `onChange`) with the actual scroll deferred via `DispatchQueue.main.async`. If the user flips the Cloudflare toggle off again before the block runs, the deferred `scrollTo(.cloudflareClientID)` targets a field that no longer exists (silent no-op, harmless) and the two-phase write makes the state read oddly. Additionally, every focus change triggers `scrollTo(anchor: .center)`, which can race iOS's own auto-scroll of a first responder on slower devices.
**Suggestion:** Guard the deferred block with a `guard revealCloudflareSection`-style re-check (or only scroll when the field's current frame escapes the visible-bounds), so a torn-down section can't be scrolled to.

No retain cycles, force unwraps, or concurrency violations were introduced: `Coordinator.parent` holds the value-type `AuthWebView`, `authenticatedWebView` is held weakly, the bridge's script handler proxies weakly, `connect()` is `@MainActor`, and the classifier is a pure static closed over a host-exact match. The `about:` allowance is correctly subframe-only, pinned main frames stay origin-bound (including `about:` escape attempts), and the live-WKWebView `TurnstileSubframeBoundaryTests` substantiate the root-cause claim — the fix is minimal and well-tested.

## Verdict
APPROVE

The findings are non-blocking hardening/consistency items; the primary security one (HTTPS gating of token headers) is defensive but worth folding into a follow-up.

<a href="https://opencode.ai/s/rSDVuLaL">New%20session%20-%202026-08-31T17%3A59%3A24.466Z</a>
[opencode session](https://opencode.ai/s/rSDVuLaL)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33422432175)

kaishi00:
## Summary
Restores Cloudflare Access interactive sign-in (#117) by allowing the `about:blank`/`about:srcdoc` subframe documents Turnstile requires (subframe-classified before the transport guard in both the auth WebView and the ticket bridge), extracts the boundary into a unit-testable `AuthWebViewNavigationPolicy`, gates service-token transmission to HTTPS/WSS-only, and reworks the login form into a keyboard-aware, focus-chain-navigable ScrollView with unit/live-WKWebView/UI coverage.

## Findings

### [severity: minor] Whitespace-only dashboard URL still returns silently via the keyboard submit path
**File:** Conduit/Views/LoginView.swift:300-301
**Issue:** `connect()` still guards `let cleaned = serverUrl.trimmingCharacters(...); guard !cleaned.isEmpty else { return }` with no feedback. The Connect button is now correctly disabled for whitespace-only input (`connectInputsArePresent` at :280-283), but the return-key chain lands on Connect from the password/Client Secret fields and reaches `connect()` regardless — a whitespace-only server URL produces a silent no-op instead of the `invalidURL` message the form otherwise shows.
**Suggestion:** Mirror the button: `guard !cleaned.isEmpty else { error = ConnectionURLPolicyError.invalidURL.localizedDescription; focusedField = .server; return }`.

### [severity: minor] Token-rejection classifier checks `!= nil` instead of `isConfigured`
**File:** Conduit/Services/NativeAuthClient.swift:115
**Issue:** `if cloudflareAccess != nil, Self.redirectsToCloudflareAccessLogin(http)` throws `cloudflareServiceTokenRejected` for any non-nil instance, including an unconfigured (empty ID/secret) one constructed via the public memberwise init — which adds no headers (`applying(to:)` no-ops) but would still suppress the interactive WebView fallback the surrounding comment promises for the no-token case. Not reachable through `LoginView` (its `configuredCloudflareAccess` uses `from()`, which returns `nil` when unconfigured), so low impact, but the API permits it.
**Suggestion:** Use `cloudflareAccess?.isConfigured == true` before classifying, matching `applying(to:)`'s own check.

### [severity: nit] Deferred focus-scroll can fire against a torn-down Cloudflare section
**File:** Conduit/Views/LoginView.swift:59-70
**Issue:** `revealCloudflareSection` is a one-shot flag cleared inside `onChange`, with the actual scroll deferred via `DispatchQueue.main.async`. If the user flips the Cloudflare toggle off before the block runs, the deferred `scrollTo(.cloudflareClientID)` targets a field that was just unmounted (silent no-op, harmless) and the two-phase state write reads oddly. Every focus change also triggers `scrollTo(anchor: .center)` which can race iOS's own first-responder auto-scroll on slower devices.
**Suggestion:** Re-check the section state inside the deferred block (or only scroll when the focused field's frame escapes the visible bounds), and optionally invalidate the deferred work when the section is toggled off.

### [severity: nit] `insecureTransport` message shown for any URL normalization failure
**File:** Conduit/Views/LoginView.swift:312-314
**Issue:** `guard let normalized = try? ConnectionURLPolicy.normalizedBaseURL(cleaned)` collapses `.invalidURL` and `.insecureTransport` into a single catch that always displays "Remote dashboards must use HTTPS…". A garbage string therefore gets the wrong diagnostic. Pre-existing behavior (the lines are unchanged), and only touchable while editing this function.
**Suggestion:** `do/catch` on the typed errors and surface `.invalidURL`/`.insecureTransport` separately.

## Verdict
APPROVE

The primary security concern from earlier review rounds — service tokens over cleartext HTTP/WS — is properly resolved (`applying(to:)` now gates on `https`/`wss` and the fetch-injection fails closed on HTTP origins, with matching regression tests). No retain cycles, force unwraps, or concurrency violations were introduced: the Coordinator's `authenticatedWebView` is weak, the bridge's script handler is a `@MainActor` weak proxy, `connect()` is `@MainActor`, and the extracted `about:` allowance is demonstrably subframe-only with main-frame origin pinning intact. The two minor items are consistency/UX gaps that are unreachable or nearly unreachable through the UI and can be folded into a follow-up.

<a href="https://opencode.ai/s/vzWlc6qE">New%20session%20-%202026-08-31T19%3A57%3A50.228Z</a>
[opencode session](https://opencode.ai/s/vzWlc6qE)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33433303436)

kaishi00:
## Summary

Restores Cloudflare Access interactive sign-in (issue #117) by allowing Turnstile's `about:blank`/`about:srcdoc` subframe documents (classified subframe-first in both the auth WebView and the dashboard ticket bridge, extracted into a unit-testable `AuthWebViewNavigationPolicy`), gates service-token transmission to HTTPS/WSS-only, adds a typed `cloudflareServiceTokenRejected` diagnostic, and reworks the login form into a keyboard-aware, focus-chain-navigable scroll view — with strong unit, live-WKWebView, and UI coverage.

## Findings

### [severity: minor] WebSocket redirects bypass the HTTPS/WSS-only credential gate
**File:** Conduit/Services/HermesClient.swift:360-394
**Issue:** `CloudflareAccessCredentials.applying(to:)` now adds `CF-Access-Client-Id/Secret` only when the *initial* URL scheme is `https`/`wss`, and the debug claim is that the secret "never" rides cleartext. But `URLSessionWebSocketTransport.makeSocket` uses a `WebSocketOpenDelegate` that implements only `didOpenWithProtocol`/`didCloseWith` — not `willPerformHTTPRedirection`. URLSession follows HTTP(S) redirects for `URLSessionWebSocketTask` by default and forwards custom headers (CF-Access headers are not the stripped `Authorization` header) to the redirect target. A dashboard answering the `wss://` upgrade with, e.g., a cross-origin or `ws://` downgrade redirect therefore forwards the long-lived secret (and the one-time ticket in the query string) out of the intended origin — exactly the category of leak this PR's hardening is meant to prevent. Pre-existing (headers used to attach to all schemes), so not a regression, but it directly contradicts the stated invariant.
**Suggestion:** Implement `urlSession(_:task:willPerformHTTPRedirection:newRequest:completionHandler:)` on the WebSocket delegate: reject scheme downgrades (`wss→ws`) and cross-origin redirects via `completionHandler(nil)`, and only re-apply CF headers to same-origin `wss` requests. Add a regression test asserting neither header reaches a downgraded/cross-origin target.

### [severity: minor] Whitespace-only URL path and error messaging consistency
**File:** Conduit/Views/LoginView.swift:326-328
**Issue:** The wire-up of the whitespace-only fix is good (`connectInputsArePresent` uses trimmed emptiness and the return-key path surfaces `invalidURL`). But the normalization failure branch still collapses `.invalidURL` and `.insecureTransport` with `try?` into a single message: a garbage string ("garbage", ":") shows "Remote dashboards must use HTTPS…" even though the real error is "Enter a valid dashboard URL." Pre-existing, but this file is the natural place to correct it while the validation contract is already under test.
**Suggestion:** `do/catch` the typed error from `normalizedBaseURL` and surface `.invalidURL`/`.insecureTransport` separately (matches `ConnectionURLPolicyError.localizedDescription`, which the new empty-URL branch already uses).

### [severity: nit] Deferred focus-scroll can target unmounted field IDs
**File:** Conduit/Views/LoginView.swift:59-70
**Issue:** `revealCloudflareSection` is a one-shot flag cleared inside `onChange` with the actual scroll deferred via `DispatchQueue.main.async`. If the toggle is flipped off again before the block runs, `scrollTo(.cloudflareClientID)` targets an id that was just unmounted (harmless silent no-op, but fragile two-phase state). Separately, every focus change triggers `scrollTo(anchor: .center)`, which can race iOS's own first-responder auto-scroll and settle the layout differently on slower devices.
**Suggestion:** Re-check the section state inside the deferred block before scrolling (or only scroll when the field's frame escapes the visible bounds), and consider invalidating the deferred work when the section is toggled off.

### [severity: nit] UI test relies on implicit keyboard state and explicit scrolling
**File:** ConduitUITests/LoginKeyboardUITests.swift:50-57
**Issue:** The test asserts hittability via busy-poll `Thread.sleep(forTimeInterval: 0.1)` and swallows it with an `app.swipeUp()` fallback. It also assumes `typeText("\n")` triggers the return-key chain, which can silently behave differently if a hardware keyboard is connected to the simulator. These are acceptable but flaky-prone patterns.
**Suggestion:** Consider `wait(for:timeout:)` helpers over raw sleep-polls, and drive with `accessibilityActivate()`/keyboard-type where supported; fine to leave as-is given the documented CI-gating.

### [severity: nit] Accessibility: no explicit labels beyond placeholders
**File:** Conduit/Views/LoginView.swift:157-231
**Issue:** TextFields rely on placeholder text as their accessibility labels; the keyboard `Done` control and Toggles use visible text, so nothing is missing functionally. Low priority, but VoiceOver users get no explicit `accessibilityLabel`/`accessibilityHint` on the server/cloudflare fields (hints like "Enter dashboard URL" help). Not a blocker.
**Suggestion:** Optional — add `.accessibilityHint` to the two Cloudflare fields describing the token purpose.

No retain cycles, force unwraps, or concurrency violations were introduced. `connect()` is correctly `@MainActor`, the Coordinator holds the value-type `AuthWebView` with a weak `authenticatedWebView` and a properly-removed script-message handler (`dismantleUIView`), and the `about:` allowance is demonstrably subframe-only (`AuthWebViewNavigationPolicyTests` pins that pinned/unpinned `about:` main-frame navigations stay cancelled, and the live `TurnstileSubframeBoundaryTests` substantiate the root-cause claim). The token-rejection classifier is host-exact, the HTTPS-only gate is consistently applied across native/WebView/injection paths, and the unreachable backdrop-tap and button-consistency issues from earlier reviews are resolved.

## Verdict

APPROVE — the WebSocket redirect finding is worth a follow-up commit (it sits squarely in the PR's own security-hardening surface), but it is pre-existing and narrows rather than regresses; the remaining items are non-blocking consistency/UX nits.

<a href="https://opencode.ai/s/5hC9Go9V">New%20session%20-%202026-08-31T21%3A56%3A34.422Z</a>
[opencode session](https://opencode.ai/s/5hC9Go9V)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33443747180)

PR #120, opened 2026-09-01
fix(chat): make transcript backfill ownership self-contained, not catalog-dependent

kaishi00:
Fixes the production regression where the **"Load earlier messages"** button renders but tapping it does absolutely nothing — no spinner, no error, no older rows, no network request. The initial bounded tail hydration itself was correct.

## Root cause (the actual early-return)

A session carries three identities: the catalog/requested ID, the persisted stored ID, and the runtime ID that `session.resume` returns. After resume, `applyChatResume(_:)` re-homes `activeSessionId` to the **runtime** ID, but the session catalog does not learn the `stored ↔ runtime` alias (for catalog-opened sessions it often never does — only the *new-session* path records `alternateIds`).

`loadEarlierMessages()` began with an ownership gate that recognized only the window's **requested** ID:

```swift
persistedTranscriptWindowOwnershipIsCurrent(
    requestedSessionId: window.requestedSessionID,
    profile: window.profile
)
```

That helper depended entirely on live catalog aliases (`knownSessionIDs(for:)`) to equate `stored-a ↔ runtime-a`. With an alias-stale catalog it returned `false` **before** `isLoadingEarlier = true`, before any REST request, before any user-visible failure — the exact observed no-op. Every PR #118 backfill test preloaded `session("stored-a", alternateIDs: ["runtime-a"])` into the catalog, which recreated the synthetic condition that hid the bug.

## What changed

**1. Window models the conversation identity explicitly** — `PersistedTranscriptWindowState` now captures all three identities **from the same accepted hydration/resume transaction**:

- `requestedSessionID` — ID used to open/resume
- `resolvedSessionID` — `session_id` echoed by `/messages` (now strictly that; the old `?? result.sessionId` coalescing that conflated it with the runtime ID is gone)
- `runtimeSessionID` — `session.resume` `result.sessionId` (new)

Exposed as `trustedSessionIDs`. These are trusted together because they arrived in one accepted transaction after the existing transcript identity validation — never reconstructed from a later catalog refresh.

**2. One centralized, self-contained ownership rule** — `PersistedTranscriptWindow.ownershipHolds(window:conversationSessionIds:profile:windowAliasIds:conversationAliasIds:)`:

- profile mismatch always rejects;
- primary truth: explicit ID overlap between the window's transaction-captured set and the checked conversation's IDs;
- catalog aliases remain a compatibility source (either direction), but catalog freshness is **never required**.

The same rule now serves `loadEarlierMessages()`'s entry guard, the stale-response re-validation, the reconnect graft gate (`priorWindowOwnsThisConversation`, now comparing against the new transaction's requested + resolved + runtime identities), and the response-echo acceptance check — one identity semantics, no divergent copies.

**3. Older pages fetch by the durable stored ID** — requests go to `window.resolvedSessionID ?? window.requestedSessionID`, mirroring Hermes Desktop's `BackfillRequest.storedSessionId → getOlderSessionMessages(storedSessionId, ...)`. The runtime WebSocket ID is never used for persisted-history reads; `profile` routing is unchanged.

**4. UI availability = action ownership** — new AppState-owned `canLoadEarlierMessagesForActiveConversation` (window state + non-empty transcript + the exact ownership predicate the action enforces). `ChatView.transcriptBackfillAvailable` now just reads it, so the control can no longer render for a window the action would silently reject. A genuinely stale window from another conversation no longer shows an actionable button.

**5. Observability** — an unexpected ownership rejection of an otherwise-actionable window logs a diagnostic `debug` line with requested/resolved/runtime/active IDs and both profiles (non-sensitive; `privacy: .public`). Stale session-switch races stay silent — no user-facing error.

**6. Review-hardening extras** — the single-flight `isLoadingEarlier` flag is released when ownership is lost mid-flight while the armed window itself survived (previously it could strand the affordance forever — pre-existing, fixed here); ID normalization and alias expansion are shared helpers; response-echo acceptance routes through the same centralized rule.

## Reconnect/graft behavior

The graft gate now compares the previous window's transaction-derived identity set against the new reconciliation's full identity set, so a reconnect **keeps already-loaded older history even when the catalog is temporarily stale** about the alias. The existing safety rule is preserved: with no safe identity/row-ID anchor proving continuity, the refreshed tail stays authoritative, and grafting still requires the refreshed tail's first row to actually exist inside the previous transcript — never grafted across unrelated sessions.

## Regression coverage (10 new tests, `CompactResumeTranscriptTests`)

- **The production no-op repro**: catalog contains `stored-a` only, resume returns `runtime-a`, **no alias preloaded** → active is `runtime-a`, window owned, `canLoadEarlierMessagesForActiveConversation == true`, `loadEarlierMessages()` does **not** early-return, loader invoked exactly once with `("stored-a", "default", 120)`, rows prepend, `nextOffset` advances. Fails on the pre-fix implementation (zero calls, silent `false`).
- Requested runtime alias resolving to a different stored ID → backfill fetches the **stored** ID and stays owned (fails on old code: it requested the runtime alias).
- Session-switch discard with an alias-less catalog (B untouched, no coverage leak).
- Truly foreign active session → affordance hidden, action rejects, zero network requests.
- Catalog learning the alias later → window unchanged, ownership intact.
- Reconnect/graft with a stale catalog end to end: backfill → reconnect (catalog still alias-less) → prefix survives → ownership holds → continued backfill still routes by the stored ID. Fails on old code (backfill no-op; prefix would truncate).
- Mid-flight ownership loss releases the loading flag and re-arms.
- No-echo hydration falls back to the requested ID for older pages.
- Response echoing the transaction-vouched runtime ID is accepted (not foreign).
- The pure identity rule (explicit overlap primary, catalog aliases assist, profile mismatch always rejects).

## Architecture unchanged

All PR #118 invariants preserved: initial page `limit=120&offset=0&order=latest&include_compacted=true`; manual, bounded, tail-anchored, raw-row-offset older pages; `hasBackfilledPrefix`; overlap dedupe by durable persisted row ID; cross-page tool-call reconstruction; compact `session.resume(omit_messages=true)`; structural-only legacy fallback; no transient-failure → giant WebSocket fallback; viewport prepend anchoring; current response-size/WebSocket limits. No automatic scrolling, no full-session auto-load, no Markdown/rendering changes, no Hermes backend changes.

## Verification

- Focused: `CompactResumeTranscriptTests` 44/44 (34 existing + 10 new) + `ChatViewportControllerTests` 91/91 = 135/135 on iPhone 17 Pro (per-class executed counts audited from the xcresult).
- Full unit suite on the build Mac: 1485 passed, 0 assertion failures; the runner then aborted with the **known end-of-day host audio degradation** (`SIGABRT` in `AURemoteIO::_ReportRPCTimeout`, untouched suites — same signature hit on 2026-08-31 during #117 work). A focused re-run of the interrupted classes passed another 145 with 0 failures before the same host abort. CI's fresh runners gate the complete matrix.
- `git diff --check` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript history loading when conversations use different runtime and stored session IDs.
  * Prevented older-message requests for unrelated or outdated conversations.
  * Ensured the history-loading state resets correctly after ownership changes.
  * Improved handling of missing or alternate session identifiers in backfill responses.

* **Improvements**
  * More accurately determines when earlier transcript messages are available.
  * Added comprehensive coverage for conversation ownership and transcript backfill behavior.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2830,6 +2830,8 @@ final class AppState: ObservableObject {
             let priorWindowForGraft = priorWindowOwnsThisConversation(
                 priorWindow,
                 requestedSessionId: sessionId,
+                resolvedSessionId: transcript?.resolvedSessionId,
+                runtimeSessionId: result.sessionId,
                 profile: profile
             ) ? priorWindow : nil
             let presentationResult: SessionResumeResult
@@ -2879,11 +2881,16 @@ final class AppState: ObservableObject {
             // affordance once and re-retires on its first terminal page.
             if let transcript, transcriptMatches, let page = transcript.page,
                page.honorsTailContract {
+                // The window records EVERY identity the accepted
+                // transaction produced — requested, resolved stored, and
+                // runtime — so ownership stays self-contained even before
+                // the session catalog learns the runtime↔stored alias.
                 persistedTranscriptWindow = PersistedTranscriptWindowState(
                     requestedSessionID: sessionId,
                     profile: profile,
                     pageSize: PersistedTranscriptPagination.pageSize,
-                    resolvedSessionID: transcript.resolvedSessionId ?? result.sessionId,
+                    resolvedSessionID: transcript.resolvedSessionId,
+                    runtimeSessionID: result.sessionId,
                     nextOffset: page.rawReturned,
                     canLoadEarlier: page.mayHaveOlderRows(fetchedRowCount: page.rawReturned),
                     hasBackfilledPrefix: graftedBackfilledPrefix
@@ -6381,18 +6388,30 @@ final class AppState: ObservableObject {
         )
     }
 
+    /// Whether a persisted-history window from the previous reconciliation
+    /// belongs to the conversation this reconcile just accepted. Identity is
+    /// compared through the ONE centralized rule with the new transaction's
+    /// full identity set (requested, resolved stored, runtime), so a
+    /// reconnect keeps the backfilled prefix even when the catalog is
+    /// temporarily stale about the stored↔runtime alias.
     private func priorWindowOwnsThisConversation(
         _ window: PersistedTranscriptWindowState?,
         requestedSessionId: String,
+        resolvedSessionId: String?,
+        runtimeSessionId: String?,
         profile: String
     ) -> Bool {
-        guard let window, window.profile == profile else { return false }
-        if window.requestedSessionID == requestedSessionId { return true }
-        // Alias-aware, exactly like the backfill ownership gate: the window
-        // may have been opened under a runtime ID while this reconcile runs
-        // under the catalog's stored ID (or vice versa).
-        return knownSessionIDs(for: requestedSessionId).contains(window.requestedSessionID)
-            || knownSessionIDs(for: window.requestedSessionID).contains(requestedSessionId)
+        guard let window else { return false }
+        let conversationIds = PersistedTranscriptWindow.normalizingSessionIDs([
+            requestedSessionId, resolvedSessionId, runtimeSessionId
+        ])
+        return PersistedTranscriptWindow.ownershipHolds(
+            window: window,
+            conversationSessionIds: conversationIds,
+            profile: profile,
+            windowAliasIds: catalogAliasIDs(for: window.trustedSessionIDs),
+            conversationAliasIds: catalogAliasIDs(for: conversationIds)
+        )
     }
 
     /// Fetches the next older persisted-history page for the active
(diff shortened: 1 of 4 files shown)


kaishi00:
## Summary
Fixes the "Load earlier messages" production no-op by making backfill ownership derive from the window's own transaction-captured identity set (requested/resolved-stored/runtime) via one centralized rule instead of catalog-alias freshness, routes older pages to the durable stored ID, and mirrors the same predicate into the ChatView affordance. 724 additions, 64 deletions across 4 files with 10 new tests.

## Findings

### [minor] Trusted pairing is unvalidated when `/messages` echoes no `session_id`
**File:** Conduit/Services/PersistedTranscriptWindow.swift:119-129 (with Conduit/Services/AppState.swift:6696-6698)
**Issue:** `trustedSessionIDs` treats requested + runtime as one conversation, but the transaction that produced them is cross-validated by `transcriptMatchesSession` only through the echoed `session_id`. When `resolvedSessionID` is nil, `transcriptMatchesSession` returns `true` with no cross-check at all (AppState.swift:6696), so the requested↔runtime pairing rests entirely on `session.resume` fidelity. If a resume ever returned an unrelated runtime ID, the new rule would consider the window owned by that runtime conversation and `loadEarlierMessages` would proceed (wire = `requestedSessionID` fallback, AppState.swift:6456) and prepend rows. Pre-fix, an alias-less catalog produced a safe no-op. This is a deliberate enlargement of trust to fix the fresh-runtime/no-echo path, and it only extends rows that initial hydration already displayed, so it's not cross-conversation contamination — but it deserves documenting and, if feasible, a cheap sanity check (e.g. require an echo or catalog proof when the runtime ID differs from requested).
**Suggestion:** Document the nil-echo fallback explicitly at window creation and consider validating requested↔runtime when no echo exists, accepting that the no-echo legacy path inherently cannot prove the pairing.

### [minor] Reachability of the ownership-reject diagnostic contradicts its comment
**File:** Conduit/Services/AppState.swift:6435-6451
**Issue:** The comment claims the branch is "normally unreachable from a tap" while simultaneously acknowledging it "also fires on the legitimate render-to-tap race during a session switch." A disconnect that clears `activeSessionId` while the window survives lands here too, so rapid session switching can spam the `debug` log (session/profile IDs printed per event). Not a correctness problem, but the comment is self-contradictory about how often it fires.
**Suggestion:** Tighten the comment to say the privileged-race case (render → tap) is rare while the disconnect/mid-flight case is expected, or rate-limit the diagnostic.

### [nit] `knownSessionIDs` linear catalog scan runs on every affordance recompute
**File:** Conduit/Services/AppState.swift:6646-6650 (called from AppState.swift:6678-6686)
**Issue:** `canLoadEarlierMessagesForActiveConversation` executes `catalogAliasIDs` → `knownSessionIDs` (a linear scan of `sessions + cronSessions`) each time ChatView's body re-evaluates. This matches pre-existing patterns (`sessionMatchesActiveSession`), so it's not a regression, but per-frame cost scales with catalog size.
**Suggestion:** Optional — no change needed; flagging only for awareness.

### [nit] Redundant alias-assisted overlap cases in `ownershipHolds`
**File:** Conduit/Services/PersistedTranscriptWindow.swift:167-172
**Issue:** Because `knownSessionIDs(for:)` always returns the input ID itself, `windowAliasIds ⊇ windowIDs` and `conversationAliasIds ⊇ conversationSessionIds`; thus the two `isDisjoint` alias checks (lines 170-171) partially re-cover the primary overlap. Harmless, but the rule reads as larger than the effective semantics.
**Suggestion:** Keep as-is for clarity, or add a short comment noting the container relationship so readers don't expect four independent classes of evidence.

No issues found in the focus areas covered by the review: the new code adds no `!` force unwraps, no new closures capture `self` outside `@MainActor` (all state mutation stays on the main actor), `applyChatResume` re-homing, window rebuild, and mid-flight flag release are consistent with the transaction semantics, and the affordance retains its accessibility label/hint. The single-flight `isLoadingEarlier` release on mid-flight ownership loss (AppState.swift:6483-6491) is a genuine bug fix over the pre-existing strand.

## Verdict
APPROVE

<a href="https://opencode.ai/s/gwekfMKj">New%20session%20-%202026-09-01T03%3A45%3A54.453Z</a>
[opencode session](https://opencode.ai/s/gwekfMKj)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33467313099)

PR #121, opened 2026-09-01
fix(chat): end long-transcript reasoning churn and composer cursor corruption

kaishi00:
## Problem

A deep-agent-session user report: with >200K-token transcripts the UI grows sluggish, CPU climbs, the device heats and drains battery, and — while typing on a hardware keyboard — UI refreshes move the cursor and reorder input. This is a long-transcript scaling + editor-isolation bug, not a generic battery issue.

## Root causes (verified against current main)

**Composer cursor corruption.** `ComposerPasteTextView.Coordinator.apply` treated `textView.text != binding` as an intentional programmatic replacement. UIKit delivers `textViewDidChange` after the raw text-storage edit, so an unrelated SwiftUI invalidation landing inside that window saw a stale binding, "reverted" the user's in-flight keystroke via `textView.text = …`, and re-clamped the selection. The window opens whenever `ComposerBar` re-renders — i.e. on every AppState publication — and IME marked text was destroyed by the same path.

**Long-transcript CPU.** Every ~50 ms reasoning publish did an O(n) `messages.firstIndex` scan + O(n) CoW `messages` write + `chatTranscriptRevision` bump → `viewport.transcriptChanged` → scroll-target-cache O(n) prefix walk + O(prefix) fingerprint-`Set` build + suffix rebuild → `ChatView` body re-eval with an O(n) `Array(enumerated())` ForEach copy. At ~20 Hz over hundreds of rows, that is the sustained CPU/thermal load.

### Live reasoning projection (no per-tick transcript mutation)
- New `@Published liveReasoningSegment: LiveReasoningSegment?` (fixed id/timestamp, content advancing at the existing 50 ms cadence). ChatView renders it as a normal `ThinkingCard` between the transcript and the streaming bubble — same presentation, chronology preserved.
- The settled `.reasoning` row commits into `messages` **exactly once per segment boundary** (`settleReasoningSegmentIntoTranscript`): tool start/complete, completion (incl. the drain-race window and the completion-carried trace via `finalizeReasoning`), error, interruption, new turn — before the boundary's own row lands, so ordering, dedup, resume, reconciliation, and session-switch semantics are unchanged.
- **Mid-turn rows settle first**: steer/redirect corrections, slash output, review summaries, clarify and approval cards, and the send-path outgoing bubble all commit the live segment before appending, preserving pre-projection chronology (resumed reasoning mounts a fresh segment, matching the tool-boundary precedent).

### Composer programmatic revision
- All intentional writes (draft restore, prefill, slash insertion, clear-after-send, failed-send restore, session handoff) route through `ComposerBar.replaceComposerText(_:)`, advancing `programmaticRevision`.
- The bridge applies a replacement **only on a revision advance, exactly once**; binding values that merely echo what UIKit last reported are never written back; a replacement arriving during IME composition is **deferred** until it ends (and stays the newest instruction); identity rotations reset bookkeeping and adopt the binding as ground truth.
- `updateUIView` no longer force-layouts on every unrelated publication, and editor identity (`.id(editorIdentity)`) is unchanged — it only rotates on handoff/submit, as before.

### Transcript-sized body work
- `ForEach` iterates `chatMessageScrollTargets` directly; `ChatMessageScrollTarget` gains an `order: Int` maintained by the cache (suffix rebuilds use `baseOrder:`, rendering-only replacements preserve it). Row identity (`message.id`) is unchanged.

## Before / after (800-row hosted fixture, 12 live reasoning deltas, DEBUG-only deterministic counters)

| Counter | before | after |
|---|---|---|
| reasoning transcript mutations | 8 | **0** |
| `transcriptChanged` calls | 9 | **0** |
| cache prefix comparisons | 7,200 | **0** |
| prefix fingerprint-Set builds | 9 | **0** |
| settled Markdown body evals | 1 | **0** |
| stale-echo seam probe | keystroke reverted, cursor moved, 1 assignment | **0 assignments, text + cursor intact** |

## Deliberate test-contract change

`AppStateReasoningStreamTests`: mid-stream reasoning assertions now read `liveReasoningSegment` — a 75-delta burst produces **0** `$messages` publications (previously 1), and the scheduled publish produces 0. Boundary commit, ordering, dedup, turn-flag, drain-race, and session-switch semantics keep their original assertions (plus a new mid-turn ordering test).

## Review gate

Two rounds of the three-reviewer cycle. Round 1: APPROVE / REQUEST CHANGES / APPROVE, no blockers — the one substantive finding (mid-turn appends committing below the live card) is fixed in this PR with regression tests. Round 2: APPROVE ×3; residual suggestions adopted (IME-deferral precedence over revision adoption, docs).

## Verification

- Focused suites on the final commit: **52/52** and **221/221** (composer bridge/draft/return-key, reasoning stream, scroll state + incremental cache, transcript-performance fixtures, settled-message isolation, viewport controller).
- Full unit suite: 1,388 + 88 + 51 passed, 0 assertion failures. 11 tests in four untouched suites (`AppStateChatResumeTests` ×4, `ChatReturnSurfaceTests` ×6, `ChatTypographyTests` ×1) could not complete on the local Mac build host — an `AudioToolboxCore/AURemoteIO` SIGABRT aborts the runner mid-suite; a control run of the same suites on **reverted base code crashed identically**, so this is host state, not the diff. CI's fresh runners are the gate for those.
- Release configuration builds clean; `git diff --check` clean.

## Out of scope

No AppState architecture rewrite, no changes to settled-Markdown Equatable gates, viewport ownership/restoration, Markdown rendering, streaming cadence, Kanban/voice/auth, or the 50 ms reasoning cadence (the per-tick mutation was the problem, not the cadence). Physical-device profiling (CPU/thermal under hardware-keyboard typing on a large transcript) remains worth a manual pass; not blocking per the deterministic source-level regressions above.

## Summary by CodeRabbit

- **Improvements**
  - Streaming reasoning now appears smoothly in a live view and is committed cleanly at completion or turn boundaries.
  - Long conversations should scroll and render more efficiently.
  - Composer text, cursor position, and selections remain stable during updates, including pasted and programmatically inserted text.

- **Accessibility**
  - Added reliable accessibility identifiers to login controls, improving keyboard navigation and automated interaction.

- **Reliability**
  - Improved handling of interrupted sessions, session changes, and transcript updates.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,8 +102,8 @@ jobs:
           echo "ui-estimates=$(jq -r '.ui_lane.class_estimates // ""' ci-artifacts/plan.json)" >> "$GITHUB_OUTPUT"
           echo "ui-target=$(jq -r '.ui_lane.target // ""' ci-artifacts/plan.json)" >> "$GITHUB_OUTPUT"
           echo "ui-predicted=$(jq -r '.ui_lane.predicted_s // 0' ci-artifacts/plan.json)" >> "$GITHUB_OUTPUT"
-          echo "ui-timeout=$(jq -r '.ui_lane.timeout_s // 600' ci-artifacts/plan.json)" >> "$GITHUB_OUTPUT"
-          echo "ui-job-timeout=$(jq -r '.ui_lane.job_timeout_min // 25' ci-artifacts/plan.json)" >> "$GITHUB_OUTPUT"
+          echo "ui-timeout=$(jq -r '.ui_lane.timeout_s // 900' ci-artifacts/plan.json)" >> "$GITHUB_OUTPUT"
+          echo "ui-job-timeout=$(jq -r '.ui_lane.job_timeout_min // 65' ci-artifacts/plan.json)" >> "$GITHUB_OUTPUT"
 
       - name: Plan summary
         run: cat ci-artifacts/plan-summary.md >> "$GITHUB_STEP_SUMMARY"
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -459,6 +459,14 @@ final class AppState: ObservableObject {
     @Published private(set) var busyInputMode: BusyInputMode = .steer
     @Published private(set) var displayPreferences = ProfileDisplayPreferences()
     @Published var streamingText = ""
+    /// Live, frequently-changing reasoning projection. Streaming reasoning
+    /// renders from here at display cadence WITHOUT mutating the settled
+    /// `messages` array — per-publish transcript mutation (O(message count)
+    /// index scan + copy-on-write copy + revision bump + scroll-target cache
+    /// walk) is what made deep agent sessions burn CPU and battery. The card
+    /// commits into `messages` exactly once per segment boundary via
+    /// `settleReasoningSegmentIntoTranscript()`.
+    @Published private(set) var liveReasoningSegment: LiveReasoningSegment?
     /// An explicit user-send request lets ChatView scroll after SwiftUI has
     /// inserted the outgoing bubble, even if the user previously browsed up.
     @Published private(set) var chatScrollRequest = 0
@@ -661,6 +669,15 @@ final class AppState: ObservableObject {
         let reasoning: String?
     }
 
+    /// One live reasoning segment as projected to the UI: the identity and
+    /// mount timestamp are fixed for the segment's lifetime, only `content`
+    /// advances at display cadence. Rendering treats it exactly like a
+    /// settled `.reasoning` row (same ThinkingCard presentation).
+    struct LiveReasoningSegment: Equatable {
+        let id: String
+        let timestamp: String
+        var content: String
+    }
     private var reconciliationToken = UUID()
     private var reconciliation: Reconciliation?
     private var activeClientEpoch = UUID()
@@ -8188,9 +8205,10 @@ final class AppState: ObservableObject {
         switch event {
         case .messageStart:
             finalizePendingStreamingCompletion()
-            // A new turn ends any live reasoning card; flush first so the
-            // previous segment keeps its exact buffered text.
-            flushReasoningPublish()
+            // A new turn ends any live reasoning card; settle first so the
+            // previous segment keeps its exact buffered text in the
+            // transcript.
+            settleReasoningSegmentIntoTranscript()
             resetReasoningTurn()
             setRunning(true)
             notifyVoiceAssistant(.started(sessionID: streamSessionId))
@@ -8218,7 +8236,7 @@ final class AppState: ObservableObject {
             notifyVoiceAssistant(.completed(sessionID: streamSessionId, content: content))
 
         case .messageError(_, let message):
-            flushReasoningPublish()
+            settleReasoningSegmentIntoTranscript()
             resetReasoningTurn()
             clearStreamingText()
             errorMessage = message
@@ -8226,7 +8244,7 @@ final class AppState: ObservableObject {
             notifyVoiceAssistant(.failed(sessionID: streamSessionId, message: message))
 
         case .messageInterrupted:
-            flushReasoningPublish()
+            settleReasoningSegmentIntoTranscript()
             resetReasoningTurn()
             clearStreamingText()
             setRunning(false)
@@ -8260,11 +8278,12 @@ final class AppState: ObservableObject {
 
         case .toolStart(_, let name, let input):
             if name.lowercased() == "clarify" { break }
-            // The tool card must land after a complete reasoning card; flush
-            // the coalesced buffer before the boundary reorders the transcript.
-            // A tool ends the reasoning SEGMENT only — the turn flag survives
-            // so completion-carried reasoning cannot duplicate this segment.
-            flushReasoningPublish()
+            // The tool card must land after a complete reasoning card; commit
+            // the coalesced segment before the boundary reorders the
+            // transcript. A tool ends the reasoning SEGMENT only — the turn
+            // flag survives so completion-carried reasoning cannot duplicate
+            // this segment.
+            settleReasoningSegmentIntoTranscript()
             resetReasoningSegment()
             flushStreamingPartial()
             messages.append(ChatMessage(
@@ -8277,7 +8296,7 @@ final class AppState: ObservableObject {
 
         case .toolComplete(_, let name, let output):
             if name.lowercased() == "clarify" { break }
-            flushReasoningPublish()
+            settleReasoningSegmentIntoTranscript()
             resetReasoningSegment()
             // Update the matching running tool card in place instead of
             // appending a duplicate. This keeps input + output together in
@@ -8430,9 +8449,11 @@ final class AppState: ObservableObject {
     /// A reasoning delta belongs exactly where Hermes emitted it. Gateways can
     /// send either deltas or repeated cumulative snapshots, so merge both into
     /// one live card rather than creating duplicate thinking boxes. The first
-    /// delta of a segment mounts the card immediately so the thinking box
-    /// appears promptly; every later delta coalesces through
-    /// `reasoningBuffer` and republishes at display cadence.
+    /// delta of a segment mounts the live projection immediately so the
+    /// thinking box appears promptly; every later delta coalesces through
+    /// `reasoningBuffer` and republishes the PROJECTION at display cadence.
+    /// The settled transcript is touched only once per segment, at the
+    /// boundary commit.
     private func appendReasoning(_ text: String) {
         guard !text.isEmpty else { return }
         reasoningBuffer = mergedReasoning(
@@ -8441,14 +8462,13 @@ final class AppState: ObservableObject {
         )
         guard activeReasoningMessageId != nil else {
             let id = "reasoning-\(Date().timeIntervalSince1970)"
-            messages.append(ChatMessage(
-                id: id,
-                role: .reasoning,
-                content: reasoningBuffer,
-                timestamp: Self.localTimestamp(),
-                author: activeProfile
-            ))
             activeReasoningMessageId = id
+            TranscriptPerf.note(.reasoningProjectionPublish)
+            liveReasoningSegment = LiveReasoningSegment(
+                id: id,
+                timestamp: Self.localTimestamp(),
+                content: reasoningBuffer
+            )
             return
         }
         scheduleReasoningPublish()
@@ -8461,12 +8481,11 @@ final class AppState: ObservableObject {
 
         reasoningPublishTask = Task { @MainActor [weak self] in
             do {
-                // Reasoning updates mutate the published transcript, which is
-                // heavier than the streaming-text projection: an expanded
-                // ThinkingCard restyles its attributed text and remeasures a
-                // growing height on every commit. ~20 fps keeps the stream
-                // readable while leaving the main actor free between layout
-                // passes.
+                // Reasoning updates republish the live projection at this
+                // cadence; an expanded ThinkingCard restyles its attributed
+                // text and remeasures a growing height on every commit, so
+                // ~20 fps keeps the stream readable while leaving the main
+                // actor free between layout passes.
                 try await Task.sleep(for: .milliseconds(50))
             } catch {
                 return
@@ -8484,19 +8503,60 @@ final class AppState: ObservableObject {
     /// nothing can mutate a finalized or replaced transcript.
     private func publishReasoningBuffer(liveCardID: String?) {
         guard let liveCardID, liveCardID == activeReasoningMessageId,
-              let index = messages.firstIndex(where: { $0.id == liveCardID }),
-              messages[index].content != reasoningBuffer else { return }
-        messages[index].content = reasoningBuffer
+              var segment = liveReasoningSegment, segment.id == liveCardID,
+              segment.content != reasoningBuffer else { return }
+        TranscriptPerf.note(.reasoningProjectionPublish)
+        segment.content = reasoningBuffer
+        liveReasoningSegment = segment
     }
 
-    /// Publish any coalesced reasoning immediately so the live card is exact
-    /// before a semantic boundary (completion, tool, error, interruption,
-    /// next turn) reads or reorders the transcript.
+    /// Publish any coalesced reasoning into the live projection immediately.
+    /// Non-boundary callers (the sidebar closing) keep the segment streaming;
+    /// boundary callers need `settleReasoningSegmentIntoTranscript()`.
     private func flushReasoningPublish() {
         reasoningPublishTask?.cancel()
         reasoningPublishTask = nil
         hasScheduledReasoningPublish = false
         publishReasoningBuffer(liveCardID: activeReasoningMessageId)
+    }
+
+    /// Commit the live reasoning segment into the settled transcript EXACTLY
+    /// ONCE at a semantic boundary: tool card, completion, error,
+    /// interruption, or a new turn. `explicitContent` supersedes the buffered
+    /// text (the completion-carried trace replaces what streamed, exactly as
+    /// it replaced the in-transcript card content before the projection
+    /// existed). Paired with the boundary's `resetReasoning*` call, this
+    /// preserves the pre-projection transcript shape: the card is in
+    /// `messages`, complete, and no stale publish can touch it afterwards.
+    private func settleReasoningSegmentIntoTranscript(explicitContent: String? = nil) {
+        reasoningPublishTask?.cancel()
+        reasoningPublishTask = nil
+        hasScheduledReasoningPublish = false
+
+        func commit(id: String, timestamp: String, content: String) {
+            guard !content.isEmpty else { return }
+            TranscriptPerf.note(.reasoningTranscriptMutation)
+            messages.append(ChatMessage(
+                id: id,
+                role: .reasoning,
+                content: content,
+                timestamp: timestamp,
+                author: activeProfile
+            ))
+            reasoningBuffer = ""
+            activeReasoningMessageId = nil
+            liveReasoningSegment = nil
+        }
+
+        if let segment = liveReasoningSegment, activeReasoningMessageId == segment.id {
+            commit(id: segment.id, timestamp: segment.timestamp, content: explicitContent ?? reasoningBuffer)
+        } else if let text = explicitContent {
+            commit(
+                id: "reasoning-\(Date().timeIntervalSince1970)",
+                timestamp: Self.localTimestamp(),
+                content: text
+            )
+        }
     }
 
     /// End the active reasoning SEGMENT without publishing. Used at tool
@@ -8509,6 +8569,7 @@ final class AppState: ObservableObject {
         hasScheduledReasoningPublish = false
         reasoningBuffer = ""
         activeReasoningMessageId = nil
+        liveReasoningSegment = nil
     }
 
     /// Restore the ENTIRE per-turn reasoning state machine to its initial
@@ -8533,19 +8594,13 @@ final class AppState: ObservableObject {
     /// Completion sometimes carries the full reasoning trace as well as the
     /// deltas. Prefer that complete value without duplicating an already
     /// streamed card; gateways that only provide completion still get a card
-    /// immediately before their final answer.
+    /// immediately before their final answer. Completion is a boundary: the
+    /// trace commits straight into the settled transcript rather than the
+    /// live projection — a projection left behind would be discarded by the
+    /// turn reset that immediately follows.
     private func finalizeReasoning(_ text: String) {
         guard !text.isEmpty else { return }
-        if let id = activeReasoningMessageId,
-           let index = messages.firstIndex(where: { $0.id == id }) {
-            if text.hasPrefix(messages[index].content) {
-                messages[index].content = text
-            } else if messages[index].content != text {
-                messages[index].content = text
-            }
-            return
-        }
-        appendReasoning(text)
+        settleReasoningSegmentIntoTranscript(explicitContent: text)
     }
 
     func requestChatScrollToLatest() {
@@ -9243,7 +9298,7 @@ final class AppState: ObservableObject {
     ) {
         // messageComplete is a semantic boundary: the thinking card must show
         // its full buffered reasoning immediately, not one cadence later.
-        flushReasoningPublish()
+        settleReasoningSegmentIntoTranscript()
         streamingCompletionTask?.cancel()
         streamingPublishTask?.cancel()
         streamingPublishTask = nil
@@ -9321,7 +9376,7 @@ final class AppState: ObservableObject {
     ) {
         // Reasoning that raced the drain window must not be discarded when
         // the pending completion finalizes.
-        flushReasoningPublish()
+        settleReasoningSegmentIntoTranscript()
         removeAllPartials()
         // Some gateways repeat the full trace in completion after already
         // emitting reasoning events. Use it only when streaming supplied no
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -5231,6 +5231,9 @@ final class AppState: ObservableObject {
             author: nil,
             attachments: attachments.isEmpty ? nil : attachments
         )
+        // Mid-turn ordering rule: the outgoing bubble must not land below a
+        // still-live reasoning card's eventual commit.
+        settleReasoningSegmentIntoTranscript()
         messages.append(userMessage)
         requestChatScrollToLatest()
         cacheMessagePresentation(for: [sessionId])
@@ -5787,6 +5790,9 @@ final class AppState: ObservableObject {
         if let context {
             guard isCurrentComposerSubmission(context) else { return }
         }
+        // Mid-turn ordering rule: slash output must not land below the live
+        // reasoning card's eventual commit.
+        settleReasoningSegmentIntoTranscript()
         messages.append(ChatMessage(
             id: "slash-\(Date().timeIntervalSince1970)",
             role: .system,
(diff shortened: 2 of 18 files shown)


kaishi00:
## Summary
This PR fixes two long-transcript UX bugs: (1) live reasoning no longer mutates the settled `messages` array per-publish (a `@Published liveReasoningSegment` projection renders instead, committing into the transcript exactly once per boundary), and (2) the UIKit composer bridge no longer rewrites editor text/selection on unrelated SwiftUI invalidations, guarded by a `programmaticRevision` generation token with IME-deferral. It also removes the O(n) `Array(enumerated())` per-body copy and adds a 597-line scaling fixture suite.

## Findings

### [minor] Live→settled transition resets the ThinkingCard's expanded state
**File:** Conduit/Views/ChatView.swift:282, Conduit/Views/ChatView.swift:1659
**Issue:** The live card is a conditional `if let` at line 282 with `.id(segment.id)`, while the settled `.reasoning` row lives inside the `ForEach` (line 257→MessageBubble→ThinkingCard). `SettledThinkingCardContent.expanded` is `@State` (line 1667). When a boundary commits, the live view is destroyed and a structurally distinct row is created, so the `@State` resets to `false` — a thinking box the user opened mid-stream collapses at every tool/completion boundary (the pre-projection card kept its expansion state because the row identity was continuous).
**Suggestion:** Preserve expansion across the transition by keying expansion on `segment.id` in a shared store (e.g., `Set<String>` of expanded reasoning ids in the view model), or render both states from one identity-stable container.

### [minor] In-flight reasoning is now lost on non-boundary termination / resume
**File:** Conduit/Services/AppState.swift:8551
**Issue:** Pre-fix, the reasoning card lived in `messages` from the first delta, so a background-termination snapshot or the `applyRuntime` inflight-text recovery (AppState.swift:3368) could preserve it. Now the segment exists only in `liveReasoningSegment` until a boundary, and the seven `resetReasoningTurn()` paths (e.g., disconnect, profile switch) drop it without commit. A session killed mid-reasoning restores with the partial thought missing.
**Suggestion:** Acceptable if intentional, but document it; or reconcile the live segment into the persisted snapshot's inflight data rather than `messages`.

### [minor] New fixture suite relies on wall-clock pacing
**File:** ConduitTests/LongContextScalingFixtureTests.swift:105
**Issue:** `feedReasoningDeltas` sleeps 55 ms per delta to cross the 50 ms publish cadence and `settleQuiet` drains with `RunLoop` steps. On a loaded CI runner a scheduled 50 ms `Task.sleep` may fire late, and the whole assertion window depends on live re-rendering reaching silence. Guards (`reasoningProjectionPublishes >= 5`, quiet-gated adjustments, 15 s cap) mitigate, but the "0 mutations/0 prefix walks" assertions could still be vacuous if publishes were starved.
**Suggestion:** Add a held-count sanity assertion (as done for publishes ≥5) before each zero-assertion, and consider a deterministic flush hook instead of sleep.

### [nit] Possible reasoning id collisions for segments mounted in the same millisecond
**File:** Conduit/Services/AppState.swift:8484
**Issue:** Segments use `"reasoning-\(Date().timeIntervalSince1970)"`. Two segments beginning within the same ms (delta → toolStart → delta burst) can share an id, placing a live card and a settled row with the same `.id` in one LazyVStack. Pre-existing (the old append used the same formula), so not introduced, but the new dual-branch rendering makes a transient duplicate more observable.
**Suggestion:** Include a per-segment counter/UUID suffix in the segment id.

### [nit] `settleReasoningSegmentIntoTranscript` `commit` guard can strand state
**File:** Conduit/Services/AppState.swift:8556
**Issue:** `guard !content.isEmpty else { return }` in `commit` returns without clearing `activeReasoningMessageId`/`liveReasoningSegment`. Unreachable today (`appendReasoning` guards empty input and always mounts non-empty content), but if a future caller ever passes an empty `explicitContent` with a nil segment, the state machine stays "live" and subsequent deltas keep deferring.
**Suggestion:** Move the clear-outside-guard or assert non-empty; prefer clearing even when refusing to commit.

### [nit] Intentional replacement equal to the echoed binding can be silently skipped
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:243
**Issue:** When an intentional revision advances but the editor has advanced past the binding (delegate not yet flushed), `text == lastTextReportedByUIKit` still applies (adopt path) and the editor keeps the user's newer text, skipping the intentional replacement silently and marking the revision applied so no retry occurs. Safe (never clobbers live input) but the deliberate change is dropped in that race.
**Suggestion:** Fine as a safety posture; consider recording this "superseded by live input" case for diagnostics, or retaining the pending replacement until the editor actually matches.

### [nit] Live reasoning card is outside the scroll-target cache
**File:** Conduit/Views/ChatView.swift:282
**Issue:** The projection card produces no `ChatMessageScrollTarget` and no `ChatRenderedScrollTargets.row` frame. Bottom-follow still works via the bottom-anchor drift mechanism, but semantic id-based scroll anchoring and row-order bookkeeping can't address the live card while it streams, and `order` continuity for the card itself is untracked until settle.
**Suggestion:** Optionally emit a synthetic scroll target (or a frame preference) for the active live segment so restore/anchor paths can locate it.

## Verdict
APPROVE — the core fixes are well-designed and thoroughly tested (revision-gated composer writes, exactly-once settle, all mid-turn append paths covered, `@MainActor` isolation and weak captures preserved). The findings are minor UX/robustness regressions, none blocking.

<a href="https://opencode.ai/s/7QoZSNuP">New%20session%20-%202026-09-01T19%3A40%3A24.372Z</a>
[opencode session](https://opencode.ai/s/7QoZSNuP)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33550755184)

kaishi00:
Follow-up hardening pushed as c3efbb0: the composer bridge's equal-text revision-adoption shortcut now verifies against the **live** `textView.text`, not only the last delegate report (which can lag in-flight user input), so a genuine intentional revision can never be acknowledged without being applied. Regression: `testIntentionalRevisionAppliesWhenLiveEditorAdvancedPastReportedText`. The revision-gated design is unchanged; Mac verification 27/27 focused.

kaishi00:
## Summary
This PR fixes two long-transcript chat bugs: (1) streaming reasoning no longer mutates the settled `messages` array per publish — a new `@Published liveReasoningSegment` projection renders in place of the per-tick transcript write and commits into `messages` exactly once per semantic boundary — and (2) the UIKit composer bridge no longer rewrites editor text/selection on unrelated SwiftUI invalidations, gated by a `programmaticRevision` generation token with IME-deferral and identity-rotation bookkeeping. It also removes the O(n) `Array(enumerated())` copy in ChatView's `ForEach` and adds a large deterministic instrumented fixture suite.

## Findings

### [minor] In-flight reasoning is silently dropped at non-boundary teardown
**File:** Conduit/Services/AppState.swift:8608 (`resetReasoningTurn`), called from :2170 (disconnect), :3368, :3171, :4651, :4730, :1723
**Issue:** Pre-fix, the reasoning card lived in `messages` from the first delta, so a background termination / disconnect / snapshot surface preserved the partial thought. Now the segment exists only in the transient `liveReasoningSegment` until a boundary, and `resetReasoningTurn()`/`resetReasoningSegment()` discard it without committing. A user killed mid-reasoning on a resumed connection loses the in-flight trace that used to be persisted.
**Suggestion:** Accept as intentional (reasoning is ephemeral chain-of-thought), but it should be documented; or settle the live segment into the persisted snapshot path rather than `messages` on teardown.

### [minor] Thinking-card expanded state collapses at every live→settled boundary
**File:** Conduit/Views/ChatView.swift:282 and :1659 (`SettledThinkingCardContent.expanded` @State at :1667)
**Issue:** During streaming the card renders from a directly-placed `ThinkingCard` (`if let segment` at :282) with `.id(segment.id)`. At settle the segment clears and the same row reappears as a `MessageBubble` inside the `ForEach` at `target.id`. The `.id` matches, but the structural position changes, so SwiftUI destroys/recreates the view and the `expanded` @State resets to `false`. A thinking box the user opened mid-stream collapses at every tool/completion/interrupt boundary. The settle-function comment at AppState.swift:8558 implies this transition is seamless; the comment should match reality.
**Suggestion:** Key expansion on `segment.id` in a shared store (e.g. a `Set<String>` of expanded reasoning ids in the view model) so it survives the structural move, or render live and settled content from one identity-stable container.

### [minor] Live reasoning card has no scroll-target frame/order while streaming
**File:** Conduit/Views/ChatView.swift:282
**Issue:** The projection card produces no `ChatRenderedScrollTargets.row` and no order entry. Bottom-follow still works via the bottom-anchor drift mechanism, but `updateStableTopMessage` (ChatViewportController.swift:1117) and semantic id-based anchoring can't see the card while it streams, and the settled row pops into frame bookkeeping at its commit order — a boundary mid-scroll can shift the detected top row.
**Suggestion:** Optionally emit a synthetic scroll target/frame preference for the active live segment so stable-top and restoration paths can locate it.

### [nit] Reasoning segment id collisions within the same millisecond are now more observable
**File:** Conduit/Services/AppState.swift:8484 and :8582
**Issue:** Both the projection mount (`appendReasoning`) and the explicit-content settle commit use `"reasoning-\(Date().timeIntervalSince1970)"`. Two segments beginning within the same ms (delta → toolStart → delta burst, or a settle-generated id) can collide, and the new dual-branch (live + settled in one LazyVStack) makes a transient duplicate `.id` more likely. Pre-existing formula, but the split rendering raises the exposure.
**Suggestion:** Append a per-segment counter or UUID suffix.

### [nit] `settleReasoningSegmentIntoTranscript` can strand live state on empty content
**File:** Conduit/Services/AppState.swift:8556-8576
**Issue:** The nested `commit` does `guard !content.isEmpty else { return }` before clearing `liveReasoningSegment`/`activeReasoningMessageId`/`reasoningBuffer`. Unreachable today (`appendReasoning` guards empty input), but if a future caller passes empty `explicitContent` with a nil segment or an empty buffer, the state machine stays "live" and subsequent deltas keep deferring.
**Suggestion:** Clear the pending state outside the guard (or `assert` non-empty), so refusing to commit never leaves a stale live segment.

### [nit] Scaling fixture still leans on wall-clock pacing
**File:** ConduitTests/LongContextScalingFixtureTests.swift:105, :112
**Issue:** `feedReasoningDeltas` sleeps 55 ms per delta to cross the 50 ms cadence and `settleQuiet` drains via `RunLoop` steps. The `reasoningProjectionPublishes >= 5` sanity guard protects the zero-mutation assertions, and both failures bail out loudly, so the risk of vacuous green is bounded — but a starved CI runner could still make the window degenerate.
**Suggestion:** Where feasible prefer a deterministic flush seam (as the sidebar toggle provides in `AppStateReasoningStreamTests`) over sleeps.

## Verdict
APPROVE

The core architectural fixes are sound and impressively tested: revision-gated exactly-once composer writes with IME-deferral, the `@MainActor`/weak-capture discipline, and the projection/settle arbitration all hold up under scrutiny. All findings are minor UX/robustness regressions; none block merge. The two most user-visible (dropped partial reasoning at teardown, expanded-state collapse across boundaries) would benefit from a follow-up or explicit documentation.

<a href="https://opencode.ai/s/gkazwsWS">New%20session%20-%202026-09-01T20%3A09%3A16.576Z</a>
[opencode session](https://opencode.ai/s/gkazwsWS)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33553514553)

kaishi00:
## Summary
This PR eliminates long-transcript CPU/thermal churn by decoupling streaming reasoning from the settled `messages` array (a new `@Published liveReasoningSegment` projection commits to the transcript exactly once per semantic boundary), and fixes composer cursor corruption by gating all programmatic text writes behind a `programmaticRevision` token in the UIKit bridge (with IME deferral and editor-identity rotation bookkeeping). It also removes the O(n) `Array(enumerated())` per-body copy by iterating cached scroll targets with a cache-maintained `order`, and adds a large instrumented scaling fixture suite plus CI watchdog/UI-test stabilization.

## Findings

### [minor] Deferred IME replacement can be silently dropped by a same-revision echo during composition
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:214
**Issue:** The deferral branch unconditionally reassigns `pendingProgrammatic = (text, programmaticRevision)`. During active IME composition, `textViewDidChange` fires on each composition keystroke, writes `parent.text = textView.text`, and the resulting ComposerBar re-render routes back into `apply` with the **same** revision and the echoed user text. That echo clobbers the pending intentional replacement's text. When composition ends, `flushPendingProgrammaticIfCompositionEnded` finds `textView.text == pending.text` and simply "adopts" the revision — the deliberate replacement (prefill, session handoff, slash insertion) never lands, yet the revision is consumed so there is no retry. This contradicts the documented guarantee ("stays the newest instruction") and is not covered by the IME test in `LongContextScalingFixtureTests.swift:422`, which exercises only the no-echo path. It errs on the safe side for user input but silently drops the programmatic intent.
**Suggestion:** Only overwrite `pendingProgrammatic` when the incoming revision is strictly newer than `pending.revision`; a same-revision echo should leave the pending instruction untouched and skip the deferral write. Add a regression test with an echo `apply` interleaved during marked text.

### [minor] In-flight reasoning is dropped at non-boundary teardown instead of persisting
**File:** Conduit/Services/AppState.swift:8608 (`resetReasoningTurn`), callers at :3368, :2170, :1723, :4651, :4730
**Issue:** Pre-fix the partial reasoning row lived in `messages` from the first delta and was carried by the persisted snapshot/presentation cache. Now the segment exists only in the transient `liveReasoningSegment` until a boundary, and the teardown/reset paths discard it without committing. A session disconnected or background-suspended mid-reasoning restores without the partial chain-of-thought that used to survive.
**Suggestion:** Either settle the live segment into the persisted snapshot path (not `messages`) on teardown, or explicitly document that streaming reasoning is intentionally ephemeral and does not survive termination. Acceptable as-is if intentional, but the behavior change should be recorded.

### [minor] Thinking-card expanded state likely resets at every live→settled boundary
**File:** Conduit/Views/ChatView.swift:282 (live card) and :1659/:1667 (`SettledThinkingCardContent.expanded` `@State`)
**Issue:** During streaming the card is a direct `if let` view with `.id(segment.id)`; at settle the same id reappears as a `MessageBubble` row inside the `ForEach`. The `.id` matches, but the node moves between structural containers, so SwiftUI may destroy the `@State` — an opened thinking box collapses at each tool/completion/interrupt boundary. Confidence is moderate (`@State` can survive `.id`-keyed moves), but the transition is not identity-continuous.
**Suggestion:** Hoist expansion to a shared store keyed by segment id (e.g., a `Set<String>` of expanded reasoning ids in the view model), or render live and settled content through one identity-stable container so the transition is seamless in both layout and state.

### [nit] Live reasoning card has no scroll-target order/frame while streaming
**File:** Conduit/Views/ChatView.swift:282
**Issue:** The projection card emits no `ChatRenderedScrollTargets.row` and occupies no cache `order`. Bottom-follow still works via bottom-anchor drift, but stable-top detection and semantic id-based anchoring cannot see the card while it streams, and a boundary in the middle of a scroll can shift which row is detected as the top.
**Suggestion:** Optionally emit a synthetic scroll target/frame preference for the active live segment so restoration/anchor paths can locate it.

### [nit] `lastTextReportedByUIKit` is now write-only dead state
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:127
**Issue:** Commit c3efbb0 changed the adopt-equality check from `text == lastTextReportedByUIKit` to the live `text == textView.text`, so `lastTextReportedByUIKit` is assigned (5 sites) but never read. It is residual state from the fix's earlier iteration.
**Suggestion:** Remove the property and its assignments, or keep it only if intended for diagnostics.

### [nit] Reasoning segment id can collide within the same millisecond
**File:** Conduit/Services/AppState.swift:8484 and :8582
**Issue:** Both the projection mount and the explicit-content commit use `"reasoning-\(Date().timeIntervalSince1970)"`. Two segments beginning within the same ms (delta → toolStart → delta burst) can share an id, and the new dual-branch (live + settled rows in one LazyVStack) makes a transient duplicate `.id` more observable. Pre-existing formula, but exposure is higher now.
**Suggestion:** Append a UUID or per-segment counter suffix to the segment id.

### [nit] `settleReasoningSegmentIntoTranscript` `commit` guard can strand live state
**File:** Conduit/Services/AppState.swift:8556
**Issue:** `commit` does `guard !content.isEmpty else { return }` before clearing `reasoningBuffer`/`activeReasoningMessageId`/`liveReasoningSegment`. Unreachable today (`appendReasoning` guards empty input), but a future empty `explicitContent` with no live segment would leave state machine "live" while the private `appendReasoning` continues deferring new deltas forever.
**Suggestion:** Clear the pending state outside the guard (or `assert` non-empty), so refusing to commit never leaves a stale live segment.

### [nit] Scaling fixture still leans on wall-clock pacing
**File:** ConduitTests/LongContextScalingFixtureTests.swift:105
**Issue:** `feedReasoningDeltas` sleeps 55 ms per delta to cross the 50 ms cadence and `settleQuiet` drains via `RunLoop` steps. The `reasoningProjectionPublishes >= 5` sanity guard bounds the vacuous-green risk for the zero-mutation assertions (both failure paths bail loudly), but a starved CI runner could still make the window degenerate.
**Suggestion:** Add a held-publish sanity assertion before each zero-assertion where the fixture relies on live rendering, and prefer a deterministic flush seam (e.g., the sidebar toggle already used in `AppStateReasoningStreamTests`) where feasible.

## Verdict
APPROVE

The core architecture is sound and unusually well-tested: revision-gated exactly-once composer writes with IME deferral, `@MainActor`/weak-self task discipline, the projection/settle arbitration preserving chronology and dedup, and the `order`-maintained `ForEach` all hold up. No retain cycles, force unwraps, or new data races were introduced, and instrumentation is Debug-only. My findings are non-blocking; the first (same-revision echo clobbering a deferred intentional replacement) is the only one that contradicts a documented guarantee of the fix and would benefit from a one-line guard plus a regression test.

<a href="https://opencode.ai/s/KGvO0hZb">New%20session%20-%202026-09-01T21%3A25%3A44.544Z</a>
[opencode session](https://opencode.ai/s/KGvO0hZb)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33560898632)

kaishi00:
## Summary
This PR fixes two long-transcript chat bugs: (1) streaming reasoning now renders through a `@Published liveReasoningSegment` projection and commits into the settled `messages` array exactly once per semantic boundary (eliminating per-tick O(n) transcript churn), and (2) the UIKit composer bridge gates all programmatic text writes behind a `programmaticRevision` token with IME-deferral so unrelated SwiftUI invalidations can never rewrite editor text or move the cursor. It also removes the O(n) `Array(enumerated())` ForEach copy and adds a large deterministic scaling fixture suite plus CI watchdog/UI-test hardening.

## Findings

### [minor] Deferred IME replacement can be dropped by a same-revision echo during composition
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:214
**Issue:** The deferral branch reassigns `pendingProgrammatic = (text, programmaticRevision)` unconditionally. While marked text is active, each composition keystroke fires `textViewDidChange` → `parent.text = textView.text` → a ComposerBar re-render routes back into `apply` with the **same** revision (ComposerBar never advanced it) and the echoed user text. Because the deferred revision never became `appliedProgrammaticRevision`, the `guard programmaticRevision != appliedProgrammaticRevision` gate passes, and the echo overwrites `pendingProgrammatic.text` with the user's own composition text. When composition ends, `flushPendingProgrammaticIfCompositionEnded` finds `textView.text == pending.text` and merely adopts the revision — the deliberate replacement (prefill, slash insertion, session handoff) never lands, yet the revision is consumed so there is no retry. This contradicts the documented "stays the newest instruction" guarantee and is not covered by `testProgrammaticReplacementDefersDuringMarkedText` (no-echo path only).
**Suggestion:** Overwrite `pendingProgrammatic` only when `pendingProgrammatic == nil || programmaticRevision > pendingProgrammatic!.revision`; a same-revision echo should leave the pending instruction intact. Add a regression test with an echo `apply` interleaved during marked text.

### [minor] In-flight reasoning is dropped on non-boundary teardown instead of persisting
**File:** Conduit/Services/AppState.swift:8593 (`resetReasoningSegment`), callers at :1720, :2170, :3168, :3368, :4651, :4730, :7154
**Issue:** Pre-fix, the partial reasoning row lived in `messages` from the first delta and survived disconnect/reconnect snapshot recovery (e.g. the `applyRuntime` inflight path at :3368). Now the segment exists only in the transient `liveReasoningSegment`, and the seven `resetReasoningTurn()` teardown/session-switch paths discard it without committing. A session killed or background-suspended mid-reasoning restores without the partial chain-of-thought that previously survived.
**Suggestion:** Either settle the live segment into the persisted snapshot path (not `messages`) on teardown, or explicitly document that streaming reasoning is intentionally ephemeral and does not survive termination. Acceptable as-is if deliberate, but the behavior change should be recorded.

### [minor] Thinking-card expanded state collapses at every live→settled boundary
**File:** Conduit/Views/ChatView.swift:282 (live card), :257 (ForEach row), :1659/:1667 (`SettledThinkingCardContent.expanded` @State)
**Issue:** During streaming the card is a direct `if let` view with `.id(segment.id)`; at settle the same id appears as a `MessageBubble` row inside the `ForEach`. The id value matches but the node moves between structurally distinct containers, so SwiftUI destroys/recreates the view and the `@State expanded` resets to `false` — an opened thinking box collapses at every tool/complete/interrupt boundary. The in-code comment at AppState.swift:8558 ("seamless live→settled transition") overstates what happens.
**Suggestion:** Hoist expansion to a shared store keyed by segment id (e.g. a `Set<String>` of expanded reasoning ids in the view model), or render live and settled content through one identity-stable container.

### [nit] Reasoning segment ids can collide within the same millisecond
**File:** Conduit/Services/AppState.swift:8484 and :8582
**Issue:** Both the projection mount and the explicit-content settle commit use `"reasoning-\(Date().timeIntervalSince1970)"`. Two segments beginning within the same ms (delta → toolStart → delta burst, or a settle-generated id after a fresh delta) can share an id; the new dual-branch rendering (live card + settled row in one LazyVStack) makes a transient duplicate `.id` more observable. Pre-existing formula, so not introduced, but exposure is higher now.
**Suggestion:** Append a per-segment counter or UUID suffix to the segment id.

### [nit] `settleReasoningSegmentIntoTranscript` commit guard can strand live state
**File:** Conduit/Services/AppState.swift:8556
**Issue:** The nested `commit` does `guard !content.isEmpty else { return }` before clearing `reasoningBuffer`/`activeReasoningMessageId`/`liveReasoningSegment`. Unreachable today (`appendReasoning` and `finalizeReasoning` guard empty input), but a future empty `explicitContent` with a live segment would return without clearing, leaving the state machine "live" so subsequent deltas keep deferring forever.
**Suggestion:** Clear the pending state outside the guard (or `assert` non-empty) so refusing to commit never leaves a stale live segment.

### [nit] `lastTextReportedByUIKit` is now write-only dead state
**File:** Conduit/Views/Components/ComposerPasteTextView.swift:127
**Issue:** Commit c3efbb0 changed the adopt-equality shortcut from `text == lastTextReportedByUIKit` to the live `text == textView.text`, so `lastTextReportedByUIKit` is assigned at 6 sites but never read. Residual state from the fix's earlier iteration.
**Suggestion:** Remove the property and its assignments, or retain it only if intended for diagnostics.

### [nit] Live reasoning card is outside the scroll-target cache
**File:** Conduit/Views/ChatView.swift:282
**Issue:** The projection card emits no `ChatRenderedScrollTargets.row` frame and occupies no cache `order`. Bottom-follow still works via the bottom-anchor drift mechanism, but stable-top detection and semantic id-based anchoring cannot locate the card while it streams, and a boundary mid-scroll can shift which row is detected as the top.
**Suggestion:** Optionally emit a synthetic scroll target/frame preference for the active live segment so restore/anchor paths can find it.

### [nit] Scaling fixture still leans on wall-clock pacing
**File:** ConduitTests/LongContextScalingFixtureTests.swift:105
**Issue:** `feedReasoningDeltas` sleeps 55 ms per delta and `settleQuiet` drains via `RunLoop` steps. The `reasoningProjectionPublishes >= 5` sanity guard bounds the vacuous-green risk for the zero-mutation assertions (both failure paths bail loudly), but a starved CI runner could still make the window degenerate.
**Suggestion:** Where feasible prefer a deterministic flush seam (as the sidebar toggle provides in `AppStateReasoningStreamTests`) over sleeps, and keep a held-publish sanity assertion ahead of each zero-assertion.

### [nit] ci-lib.sh terminal-marker check matches any line containing the marker
**File:** scripts/ci-lib.sh:28
**Issue:** `grep -q -E '\*\* TEST (EXECUTE )?(SUCCEEDED|FAILED) \*\*'` matches any line containing the marker text, so a test or tool echoing that string mid-run would grant the finalize grace early and delay timeout enforcement. Separately, the `marker` local at line 68 is declared but never used.
**Suggestion:** Anchor the regex to a complete line (`^[[:space:]]*\*\* TEST ... \*\*[[:space:]]*$`) and drop the unused `marker` local.

### [nit] UI test taps a switch that never became hittable
**File:** ConduitUITests/LoginKeyboardUITests.swift:130
**Issue:** When `hittableDeadline` expires, `settled` stays false but the transition loop still calls `element.tap()`. `.tap()` waits synchronously for hittability and can raise a generic XCTest failure before the assertion's diagnostic accessibility-tree dump, obscuring the real cause.
**Suggestion:** `guard settled else { return false }` before the transition loop, preserving the existing diagnostics for successful settling.

## Verdict
APPROVE

The core fixes are well-designed and unusually well-tested: the revision-gated exactly-once composer writes, the @MainActor/weak-self task discipline, and the projection/settle arbitration preserving chronology, dedup, and mid-turn ordering all hold up. No retain cycles, force unwraps, or new data races were introduced, and the instrumentation is Debug-only. All findings are non-blocking; the same-revision-echo clobbering of a deferred intentional replacement (the only one contradicting a documented guarantee) and the thinking-card expansion reset would benefit from the suggested follow-ups.

<a href="https://opencode.ai/s/u4XS1UJW">New%20session%20-%202026-09-01T21%3A34%3A39.748Z</a>
[opencode session](https://opencode.ai/s/u4XS1UJW)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33561716074)

PR #114, opened 2026-08-30
release: 0.1.9 (build 143)

kaishi00:
Release branch for Conduit 0.1.9 build 142. Refreshed from main at 232af5f (merge d2fc4ed4), picking up #115 voice/STT parity, #116 profile-discovery monotonic fix, #118 paginated transcript hydration, and #119 Cloudflare Access interactive login fix, on top of build 141's #112/#113. Version bump only (0.1.9 / 142, commit 3fcde619); branch CI green (run 33454258927); archived from 3fcde619 and uploaded to TestFlight; tag ios/v0.1.9-build-142. Stays open per release practice (see #91).

PR #124, opened 2026-09-03
fix(tests): deterministic synchronization for performance fixtures

kaishi00:
## Problem

CI's unit-4 lane failed **three consecutive runs** on PR #114 (release/0.1.9) with the same signature, while the byte-identical test content was green on main CI and green locally on the Mac build host:

- `LongContextScalingFixtureTests.testLiveReasoningPublishesAvoidTranscriptSizedWork` — settled Markdown counters inflated 23–45 vs expected 0; once `reasoningProjectionPublishes` measured 4 where the fixture drove 12 deltas
- `TranscriptPerformanceFixtureTests.testFirstAppearanceRendersSettledMarkdownOnce` — total grew 1–4 after a quiet baseline
- (run 3 widened) the streaming-dormancy pair — same counter family

The release branch differed from green main **only** by the two `project.yml` version lines, so content innocence was provable.

## Root causes (verified by measurement, not conjecture)

**1. Scheduler-raced publish cadence.** `feedReasoningDeltas` pumped the run loop 55 ms per delta hoping the production ~50 ms coalescing publish `Task.sleep` fired inside each pump. On loaded runners the publishes coalesced or landed late, so the measured window contained an arbitrary subset of the implied publications.

**2. Two distinct phenomena conflated by raw totals.** A new DEBUG-only repeat-detection ledger (per-source, keyed by the row's markdown source at the existing `MarkdownText` note site) split `settledMarkdownTextBodyEvaluations` into what it had always conflated:
- **first-render/mount noise** — lazy first-mounts of not-yet-rendered rows (raw-total CI failures 1–45), including SwiftUI double body-passes on fresh mounts;
- **at-rest re-renders** — re-evaluation of rows already rendered — with the measured source ring showing these are **viewport-EDGE rows** (`Turn 0/2`, `Turn 796/798` of the 800-row fixture): a growing live card legitimately shifts the bottom-anchored LazyVStack and remounts boundary rows. Sporadic on an idle Mac (1–4), heavier on loaded CI.

## Fix (synchronization + honest invariant — production behavior unchanged)

- **Deterministic publish driver:** fixtures feed deltas synchronously and perform each coalesced publication through the **existing production flush path** — `AppState.flushReasoningPublish()`, internalized (`private` → `internal`) as the test seam (the sidebar already used it) — followed by one deterministic run-loop turn per publish. Zero scheduler dependency; no new production code path.
- **Position-classified dormancy invariant:** fixtures assert that at-rest re-renders never touch an **interior** row (`TranscriptPerf.interiorAtRestRerenders`). Interior re-renders are the cascade signature — the pre-fix nil→resolver sweep and per-publish full-cascade re-rendered every mounted row (hundreds per window, interior rows included) and still fail loudly. Edge-row remounts are O(viewport), bounded by layout churn rather than transcript size, and are tolerated however many a loaded runner produces. This is a *mechanism-derived* invariant, not a fitted count: it directly encodes "work must not scale with transcript size".
- **Data-layer contract untouched:** `reasoningTranscriptMutations == 0`, `transcriptChangedCalls == 0`, prefix-walk/fingerprint-set counters == 0 remain exact assertions (all deterministic; flush-fed driver makes publish count deterministic too).
- **Shared wait helper:** `PerformanceFixtureWait` consolidates the three duplicated quiet-drain loops (`settleUntilCountersQuiet`) and adds `eventually(cap:_:)`. Timeouts are failsafes only; counter conditions decide readiness. No delays multiplied, no retries added, nothing skipped.
- **Negative control:** `TranscriptPerfLedgerContractTests` pins the ledger semantics (first vs duplicate vs pre-window repeat, `reset()` carrying the window ledger into the at-rest ledger, explicit clearing) so a future refactor cannot silently vacuate the assertions. (It caught a wrong expectation in its own first draft before push.)

## Review

Three-model review cycle: 3× APPROVE, zero blockers; consolidated suggestions applied (negative-control test, `>=` readiness gates with exactness in trailing asserts, append-specific fingerprint readiness signal, ledger counters in the quiet-drain set, comment precision). The CI failure on this branch's first push drove the deeper split (pre-window vs in-window) and the source-ring measurement that identified the edge-remount mechanism.

## Verification (commit ac1cf64a, Mac build host, temp worktree, iPhone 17 Pro, CODE_SIGNING_ALLOWED=NO)

- Formerly-flaky `testLiveReasoningPublishesAvoidTranscriptSizedWork`: **10/10 consecutive green** (previously failed 2–3 of 6 idle runs with the naive repeat==0 form)
- Full focused suites (`LongContextScalingFixtureTests` + `TranscriptPerformanceFixtureTests` + `TranscriptPerfLedgerContractTests`): **2× 13/13 green**
- Full `ConduitTests`: **1604/1604, 0 failures**
- `git diff --check` clean; no lint/format config in repo; unit-4's classes are all inside `ConduitTests` (full-suite run covers the lane); hosted CI remains the UI gate

## Exclusions (deliberate)

Other suites with sleeps (`ChatViewFollowCorrectionTests`, `AppStateReasoningStreamTests`, `CompactResumeTranscriptTests`, `SettledMessageIsolationTests`) were audited: none share the failing pattern, none are in the failing lane; `SettledMessageIsolationTests` already uses quiet-drain discipline. Narrow scope kept per task constraints.

No production behavior change: the only non-test diffs are the DEBUG-only ledger/classifier and the documented access-level widening of the existing flush function.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -11010,7 +11010,12 @@ final class AppState: ObservableObject {
     /// Publish any coalesced reasoning into the live projection immediately.
     /// Non-boundary callers (the sidebar closing) keep the segment streaming;
     /// boundary callers need `settleReasoningSegmentIntoTranscript()`.
-    private func flushReasoningPublish() {
+    ///
+    /// Internal (not private) as the deterministic test seam: performance
+    /// fixtures flush each fed delta through this SAME production path, so
+    /// a measured window contains exactly the publications the deltas imply
+    /// instead of racing the ~50 ms coalescing task's scheduler timing.
+    func flushReasoningPublish() {
         reasoningPublishTask?.cancel()
         reasoningPublishTask = nil
         hasScheduledReasoningPublish = false
--- a/Conduit/Services/TranscriptPerformanceInstrumentation.swift
+++ b/Conduit/Services/TranscriptPerformanceInstrumentation.swift
@@ -52,10 +52,30 @@ enum TranscriptPerf {
     /// CONDUIT_PERF_TRACE=1 is set (CI diagnostics), each settled-body
     /// evaluation logs its context so a polluted measurement window can be
     /// traced to the exact rows involved. DEBUG builds only.
+    ///
+    /// For `.settledMarkdownBody` the context is the row's markdown source;
+    /// it also feeds the repeat-detection ledgers behind
+    /// `settledMarkdownPreWindowRepeatEvaluations` (re-evaluation of a row
+    /// that rendered BEFORE the current measurement window opened — the
+    /// cascade signature) and `settledMarkdownWindowDuplicateEvaluations`
+    /// (second+ body pass of a row first rendered INSIDE the window —
+    /// lazy-mount noise).
     @inline(__always)
     static func note(_ event: Event, context: String) {
         #if DEBUG
         note(event)
+        if event == .settledMarkdownBody {
+            if storage.settledMarkdownKnownSources.contains(context) {
+                storage.settledMarkdownPreWindowRepeatBody += 1
+                if storage.recentPreWindowRepeatSources.count < 128 {
+                    storage.recentPreWindowRepeatSources.append(context)
+                }
+            } else if storage.settledMarkdownWindowSources.contains(context) {
+                storage.settledMarkdownWindowDuplicateBody += 1
+            } else {
+                storage.settledMarkdownWindowSources.insert(context)
+            }
+        }
         if event == .settledMarkdownBody,
            ProcessInfo.processInfo.environment["CONDUIT_PERF_TRACE"] == "1" {
             // Constant format string: source content must never be
@@ -93,6 +113,26 @@ enum TranscriptPerf {
 
     static var settledMarkdownTextBodyEvaluations: Int {
         get { read(\.settledMarkdownBody) }
+    }
+
+    /// Settled Markdown body evaluations of sources that had ALREADY been
+    /// evaluated before the current measurement window opened (per the
+    /// repeat-detection ledgers fed by `note(_:context:)`). This is the
+    /// cascade signature the scaling fixtures guard: re-rendering rows that
+    /// were already at rest. Two distinct rows sharing identical markdown
+    /// source count as a repeat for the second row; the fixtures use
+    /// per-row-unique content.
+    static var settledMarkdownPreWindowRepeatEvaluations: Int {
+        get { read(\.settledMarkdownPreWindowRepeatBody) }
+    }
+
+    /// Second and later body passes of a row FIRST rendered inside the
+    /// current measurement window. SwiftUI can legitimately evaluate a
+    /// freshly mounted row's body more than once while sizing it — this
+    /// counter exists so fixture failure messages can distinguish that
+    /// mount noise from at-rest re-renders.
+    static var settledMarkdownWindowDuplicateEvaluations: Int {
+        get { read(\.settledMarkdownWindowDuplicateBody) }
     }
 
     static var selectableTextViewUpdateCalls: Int {
@@ -213,13 +253,66 @@ enum TranscriptPerf {
         set { write(\.lastComposerSelectionAfter, newValue) }
     }
 
+    /// Sources of the most recent pre-window repeat evaluations (bounded
+    /// ring, DEBUG diagnostics only). Lets fixtures classify WHICH rows
+    /// re-rendered: viewport-edge rows under live-card layout churn are
+    /// tolerated; interior rows are the cascade signature.
+    static var recentPreWindowRepeatSources: [String] {
+        #if DEBUG
+        return storage.recentPreWindowRepeatSources
+        #else
+        return []
+        #endif
+    }
+
     // MARK: - Control
 
-    /// Reset all counters to zero.
+    /// Reset all counters and open a fresh measurement window. Sources
+    /// first rendered in the closing window merge into the pre-window
+    /// ledger, so the new window still recognizes every row rendered so
+    /// far as already-at-rest.
     static func reset() {
+        #if DEBUG
+        var fresh = Storage()
+        fresh.settledMarkdownKnownSources =
+            storage.settledMarkdownKnownSources.union(storage.settledMarkdownWindowSources)
+        storage = fresh
+        #endif
+    }
+
+    /// Clear counters AND both repeat-detection ledgers. Test fixtures call
+    /// this in setUp so one test's rendered sources never leak into the
+    /// next test's repeat counting; plain `reset()` (per measurement
+    /// window) always preserves the at-rest ledger.
+    static func resetRenderLedgerForTesting() {
         #if DEBUG
         storage = Storage()
         #endif
+    }
+
+    // MARK: - Fixture classification support
+
+    /// Classifies at-rest re-renders against the transcript's viewport
+    /// edges. `sources` are repeat-evaluation source strings (the ring),
+    /// `transcript` the fixture's message list, `edgeMargin` how many rows
+    /// at either end of the transcript count as viewport-edge territory.
+    ///
+    /// Mechanism: a growing live card / streaming bubble legitimately shifts
+    /// the bottom-anchored LazyVStack and can remount rows AT THE EDGES —
+    /// observed churn only ever touches the first/last few messages. A
+    /// re-render of an INTERIOR row is the per-publish cascade the fixtures
+    /// guard (every mounted row re-evaluating), which stays caught no
+    /// matter how much edge churn a loaded scheduler produces.
+    static func interiorAtRestRerenders(
+        sources: [String],
+        transcript: [ChatMessage],
+        edgeMargin: Int = 8
+    ) -> [String] {
+        let edges = Set(
+            transcript.prefix(edgeMargin).map(\.content)
+                + transcript.suffix(edgeMargin).map(\.content)
+        )
+        return sources.filter { !edges.contains($0) }
     }
 
     // MARK: - Private storage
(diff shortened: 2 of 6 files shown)


kaishi00:
## Summary
This PR makes the performance fixture tests deterministic by flushing reasoning publishes through the existing production path (`AppState.flushReasoningPublish()`, widened `private`→`internal` as a test seam) instead of racing the ~50 ms coalescing `Task.sleep`, and adds a DEBUG-only repeat-detection ledger to `TranscriptPerf` so fixtures can distinguish re-renders of already-rendered rows (the regression) from legitimate lazy first-mounts (the CI flake). It consolidates the duplicated quiet-drain loops into a shared `PerformanceFixtureWait` helper and pins the ledger contract with a new negative-control test.

## Findings

### [minor] Failsafe caps count nominal steps, not wall-clock time
**File:** ConduitTests/PerformanceFixtureWait.swift:52-70, 74-87
**Issue:** Both `settleUntilCountersQuiet` and `eventually` advance `elapsed` by the fixed `step` per pass rather than measuring `Date()` deltas. On a loaded runner each `RunLoop.run(until:)` can overshoot `step` substantially (main-actor job backlog, display-link churn). `elapsed` therefore under-counts real time, so the "15 s failsafe" can take many times that in wall-clock before the test fails — exactly the class of loaded lane this PR targets. (Notably the asymmetry is one-directional: `quietForElapsed` under-counting only makes readiness more conservative, which is safe.)
**Suggestion:** Track `let start = Date()` and measure `elapsed = Date().timeIntervalSince(start)` (and ideally real `quietFor` in the same way) per iteration, so the caps actually bound wall-clock.

### [minor] Repeat ledger is keyed by source *string*, not row identity
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:66-72
**Issue:** Two distinct rows whose markdown source is byte-identical count the second row's first render as a repeat. This is documented and the current fixtures generate per-index-unique content, but the seam is now a permanent landmine: any future fixture (or production duplication like a shared greeting/error footprint rendered twice in one window) silently fails `settledMarkdownRepeatEvaluations == 0` as a "cascade" even though no row was re-rendered. The render cache is also content-keyed, so this ambiguity is real in production content.
**Suggestion:** Consider keying the ledger by row identity (message id) when the context can carry it, or add a fixture-side guard that asserts the mounted source set is unique within a measurement window so a future duplicate-content fixture fails with an actionable message instead of a misleading repeat count.

### [minor] Ledger set grows for the entire process lifetime in DEBUG
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:279
**Issue:** `settledMarkdownKnownSources` is never evicted and only cleared by `resetRenderLedgerForTesting()`. `reset()` deliberately preserves it, so across the whole test-process run every distinct source ever rendered accumulates in memory (full source strings). Bounded by distinct rendered rows, but unbounded over time, and the "bounded by the render cache" doc comment is inaccurate — the render cache evicts, the ledger does not.
**Suggestion:** Acceptable for DEBUG-only diagnostics; either soften the doc comment or bound the set (e.g., cap size) if test-process memory ever matters.

### [nit] `tearDown` reset calls are now dead weight
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:38, ConduitTests/LongContextScalingFixtureTests.swift:36
**Issue:** `tearDown` calls `TranscriptPerf.reset()`, whose ledger preservation can no longer be observed — the next test's `setUp` calls `resetRenderLedgerForTesting()`, discarding everything. Harmless, but the preserved ledger in `tearDown` is noise.
**Suggestion:** Keep as-is, or use `resetRenderLedgerForTesting()` in `tearDown` for intentional symmetry.

## Verdict
APPROVE

The approach is sound: the ledger's content-keyed semantics correctly leave changed-content re-evaluations out of the repeat count (that is legitimate work, not the cascade), the synchronous flush path deterministically drives more publishes than production coalescing would, the negative-control test protects the seam, and all new code is DEBUG-only or test-only with no production behavior change. Findings are minor/nits only.

<a href="https://opencode.ai/s/wM531bIt">New%20session%20-%202026-09-03T21%3A35%3A15.622Z</a>
[opencode session](https://opencode.ai/s/wM531bIt)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33808639939)

kaishi00:
## Summary

This PR makes the transcript performance fixtures deterministic by driving reasoning publishes through the production `AppState.flushReasoningPublish()` flush path (widened `private`→`internal` as the test seam) instead of racing the ~50 ms coalescing `Task.sleep`, and adds a DEBUG-only repeat-detection ledger so fixtures assert a position-classified at-rest-render invariant (interior rows only) rather than raw totals. It consolidates the duplicated quiet-drain loops into `PerformanceFixtureWait` and pins the ledger semantics with a contract test. No production behavior changes.

## Findings

### [minor] Negative-control test doesn't pin the call site that feeds the ledger
**File:** ConduitTests/TranscriptPerfLedgerContractTests.swift:23-67, Conduit/Views/Components/MarkdownText.swift:61
**Issue:** The contract test exercises `TranscriptPerf.note(_:context:)` directly, but the fixtures' new invariant only has teeth if `MarkdownText.normalBody` still passes `context: source`. A future refactor that drops the `context:` argument leaves the contract test green while `recentPreWindowRepeatSources` becomes permanently empty and `interiorAtRestRerenders` returns `[]` — silently vacating every new assertion. `testFirstAppearanceRendersSettledMarkdownOnce` (ConversationTests/TranscriptPerformanceFixtureTests.swift:335-345) is 100% ledger-dependent: the nil→resolver sweep it targets re-evaluates bodies with no UIKit work, so the bounded-work asserts (`selectableTextViewUpdateCalls` etc.) would not catch a re-introduced version of that regression if the ledger goes dark.
**Suggestion:** Add a thin guard in the contract test (or a fixture) that renders one row through the real MarkdownText path and asserts the repeat ledger is fed, or assert a non-zero lower bound on `settledMarkdownPreWindowRepeatEvaluations` in the streaming fixture so an unfed ledger fails loudly instead of passing vacuously.

### [minor] "Viewport edge" classification uses transcript ends, not the mounted-window boundary
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:306-316
**Issue:** `interiorAtRestRerenders` tolerates only the transcript's literal first/last `edgeMargin` rows. The claimed mechanism is that a growing live card remounts rows "at the viewport edges," but the classifier keys on the transcript's extremes, not the LazyVStack's actual mounted-window edges. The PR's measured ring (rows 0/2 and 796/798) is consistent with the fixture sweeping top-then-bottom during the drain, but nothing structurally guarantees a runner won't churn a row near, say, the mounted window's leading edge mid-scroll — which would be classified *interior* and spuriously fail the very tests this PR is fixing. The safeguard could reintroduce the flakiness it removes.
**Suggestion:** Either derive the tolerated set from the actually-mounted row ids if the view exposes that, or keep the fixture pinned to a fixed scroll position (drain to bottom before opening the window) so the transcript ends provably coincide with the mounted edges, and document that precondition.

### [minor] Failsafe measures passes × step, not wall-clock time
**File:** ConduitTests/PerformanceFixtureWait.swift:41-63, 74-87
**Issue:** Both helpers accumulate `elapsed += step` per loop pass rather than measuring `Date()` deltas, and `RunLoop.current.run(until:)` can return after its deadline under main-actor backlog. The 15 s / 10 s caps therefore under-count real elapsed time, so a pathological runner can stretch the failsafe well past its stated bound — the opposite of what a loaded-lane reliability fix should do. (Direction is safe for `quietForElapsed`, only the cap is weakened.)
**Suggestion:** `let start = Date()` and compute `elapsed` (and `quietFor`) from actual time per pass.

### [nit] Repeat ledger grows for the entire process; doc comment is inaccurate
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:270-291, 332
**Issue:** `reset()` deliberately unions window sources into `settledMarkdownKnownSources` and never evicts; only `resetRenderLedgerForTesting()` clears it. Across a full ConduitTests run (1604 tests) every distinct rendered source accumulates in DEBUG memory. The comment "bounded by the same order as the render cache" is misleading — the render cache is bounded/evicting, the ledger is not.
**Suggestion:** Acceptable for DEBUG diagnostics, but either cap the set (e.g., drop oldest beyond N) or amend the comment to say it is unbounded over the test process.

### [nit] `tearDown` uses ledger-preserving `reset()`, immediately discarded by next `setUp`
**File:** ConduitTests/LongContextScalingFixtureTests.swift:36, ConduitTests/TranscriptPerformanceFixtureTests.swift:38
**Issue:** `tearDown` calls `TranscriptPerf.reset()` (which preserves the at-rest ledger), but the next test's `setUp` calls `resetRenderLedgerForTesting()`, discarding everything it preserved. The preserved ledger in teardown is dead weight and reads as if it mattered.
**Suggestion:** Use `resetRenderLedgerForTesting()` in `tearDown` for intentional symmetry, or drop the call.

### [nit] Doc comment claims 12 publishes while acknowledging delta 0 no-ops
**File:** ConduitTests/LongContextScalingFixtureTests.swift:168-177
**Issue:** The comment says the flush-fed driver is deterministic — "one publication per delta … so 12 for this fixture" — while simultaneously stating delta 0's flush no-ops on the unchanged-content guard, which implies 11 publications. The loose `>= 5` floor hides the discrepancy.
**Suggestion:** Rectify the count in the comment or assert the exact number once the ledger micro-test is trusted.

## Verdict

APPROVE

The approach is sound and well-verified: the flush seam routes fixtures through the real production publish path deterministically, the ledgers correctly leave changed-content re-evaluations out of the repeat count, the data-layer invariants remain exact, and the new code is entirely DEBUG-only/test-only with no production behavior change. All findings are minor/nits; no critical, retain-cycle, force-unwrap, or concurrency issues found (all instrumentation and fixture code runs on the main actor, and `AppState` remains `@MainActor`-isolated).

<a href="https://opencode.ai/s/4uKvipX2">New%20session%20-%202026-09-03T22%3A17%3A17.983Z</a>
[opencode session](https://opencode.ai/s/4uKvipX2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33812228772)

PR #125, opened 2026-09-03
fix(tests): tolerate viewport-edge remount footprint in live-row work bounds

kaishi00:
Follow-up to #124 (one commit ahead of what merged there — the amendment landed after that PR merged).

## What

`testStreamingTicksLeaveSettledMarkdownDormant_MarkdownTranscript` failed once on release-branch CI (unit-2) on its **total-work bounds**: `selectableTextViewUpdateCalls` 33 vs ≤30 (~3/tick) and `selectableTextViewTextRebuilds` 21 vs ≤20 (~2/tick). Not the dormancy counters — those (with #124's position classifier) passed everywhere.

## Mechanism

The failing overshoot is exactly **one viewport-edge remount's footprint** (≈3 SelectableTextView updates + ≈1 rebuild): the same legitimate LazyVStack edge churn #124 measured via the source ring (Turn 0/2/796/798). The live-row bounds were the last assertions still assuming zero edge churn.

## Fix

Scale each live-row bound by the number of at-rest re-renders the dormancy classifier already tolerates: `30 + 5 × edgeRemounts` updates, `30 + 5 × edgeRemounts` measurements, `20 + 3 × edgeRemounts` rebuilds. Properties:

- Zero-churn runs keep the **original strict bounds** (30/30/20).
- The cascade floor (every mounted row joining per tick ≈ 2,500–7,500 per window for the 250-row fixture) remains **two orders of magnitude** outside the bound — the acceptance criterion "work bounded to the live row, not the settled transcript" is unchanged.
- Interior-row re-renders still fail the dormancy assertion that runs first, so the allowance only ever applies to classifier-tolerated edge churn.

## Verification (commit 4a49ced, Mac build host, temp worktree, iPhone 17 Pro, CODE_SIGNING_ALLOWED=NO)

- Full focused suites ×3 (`LongContextScalingFixtureTests` + `TranscriptPerformanceFixtureTests` + `TranscriptPerfLedgerContractTests`): 13/13 each, 0 failures
- Full `ConduitTests`: 1604/1604, 0 failures
- `git diff --check` clean

## Summary by CodeRabbit

- **Performance**
  - Improved validation of transcript rendering and long-context behavior.
  - Added more precise detection of unnecessary interior-row re-renders while allowing expected viewport-edge updates.
  - Performance measurements now use deterministic settling and live reasoning updates for more consistent results.

- **Tests**
  - Expanded coverage for repeated rendering, measurement-window behavior, and ledger resets.
  - Replaced timing-sensitive waits with shared condition-based checks to improve test reliability.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -11010,7 +11010,12 @@ final class AppState: ObservableObject {
     /// Publish any coalesced reasoning into the live projection immediately.
     /// Non-boundary callers (the sidebar closing) keep the segment streaming;
     /// boundary callers need `settleReasoningSegmentIntoTranscript()`.
-    private func flushReasoningPublish() {
+    ///
+    /// Internal (not private) as the deterministic test seam: performance
+    /// fixtures flush each fed delta through this SAME production path, so
+    /// a measured window contains exactly the publications the deltas imply
+    /// instead of racing the ~50 ms coalescing task's scheduler timing.
+    func flushReasoningPublish() {
         reasoningPublishTask?.cancel()
         reasoningPublishTask = nil
         hasScheduledReasoningPublish = false
--- a/Conduit/Services/TranscriptPerformanceInstrumentation.swift
+++ b/Conduit/Services/TranscriptPerformanceInstrumentation.swift
@@ -52,10 +52,30 @@ enum TranscriptPerf {
     /// CONDUIT_PERF_TRACE=1 is set (CI diagnostics), each settled-body
     /// evaluation logs its context so a polluted measurement window can be
     /// traced to the exact rows involved. DEBUG builds only.
+    ///
+    /// For `.settledMarkdownBody` the context is the row's markdown source;
+    /// it also feeds the repeat-detection ledgers behind
+    /// `settledMarkdownPreWindowRepeatEvaluations` (re-evaluation of a row
+    /// that rendered BEFORE the current measurement window opened — the
+    /// cascade signature) and `settledMarkdownWindowDuplicateEvaluations`
+    /// (second+ body pass of a row first rendered INSIDE the window —
+    /// lazy-mount noise).
     @inline(__always)
     static func note(_ event: Event, context: String) {
         #if DEBUG
         note(event)
+        if event == .settledMarkdownBody {
+            if storage.settledMarkdownKnownSources.contains(context) {
+                storage.settledMarkdownPreWindowRepeatBody += 1
+                if storage.recentPreWindowRepeatSources.count < 128 {
+                    storage.recentPreWindowRepeatSources.append(context)
+                }
+            } else if storage.settledMarkdownWindowSources.contains(context) {
+                storage.settledMarkdownWindowDuplicateBody += 1
+            } else {
+                storage.settledMarkdownWindowSources.insert(context)
+            }
+        }
         if event == .settledMarkdownBody,
            ProcessInfo.processInfo.environment["CONDUIT_PERF_TRACE"] == "1" {
             // Constant format string: source content must never be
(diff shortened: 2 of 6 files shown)


kaishi00:
## Summary
Follow-up to #124 that makes the performance fixtures deterministic: reasoning-delta feeding now flushes through the production seam (`flushReasoningPublish`), the TranscriptPerf counters gain a DEBUG-only per-source repeat ledger, the three quiet-drain loops are consolidated into `PerformanceFixtureWait`, and the live-row work bounds are scaled by a classifier-tolerated viewport-edge remount allowance.

## Findings

### [minor] Unbounded DEBUG-only ledger retains every distinct markdown source forever
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:332
**Issue:** `settledMarkdownKnownSources` grows without bound in a long-running DEBUG build (every distinct settled-markdown source ever rendered, full content strings, never pruned except by a test-only reset). The doc comment claims it is "bounded by the same order as the render cache," but `MarkdownRenderCache` is an `NSCache` with `countLimit = 256` / `totalCostLimit = 32 MB` that self-evicts (MarkdownText.swift:260-263), whereas this `Set<String>` cannot be evicted without breaking the reset() "already at rest" semantics that the fixtures assert against.
**Suggestion:** Bound the set (e.g., cap distinct entries and only retain edge + recently-rendered sources), or store a small stable hash of each source instead of the full content string, at least for the `knownSources` set used purely for membership repeats.

### [minor] Repeat-source ring can silently drop interior-row evidence on heavy windows
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:70
**Issue:** `recentPreWindowRepeatSources` is capped at 128 entries in chronological append order. The interior classifier (`interiorAtRestRerenders`) is the fixtures' *only* signal for which rows re-rendered; if a window produces >128 at-rest repeats dominated by early edge churn, evidence of a later interior-row repeat falls off the ring and the classification asserts on an incomplete set while `settledMarkdownPreWindowRepeatEvaluations` is non-zero.
**Suggestion:** Keep a bounded *distinct*-source sample (e.g., a `Set` capped by distinct count) or add a fallback assert in the fixtures: when `interiorRerenders.isEmpty` but `atRestRerenders > 0`, trivially pass only if the ring is not at capacity.

### [minor] Edge classification can mask a cascade confined to the transcript edges
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:306 (and ConduitTests/TranscriptPerformanceFixtureTests.swift:221-250)
**Issue:** `interiorAtRestRerenders` classifies any re-render of rows 0..7 or last-8 as "tolerated edge churn," and the allowance scales the graded bounds (`30 + 5×allowance`, `20 + 3×allowance`) from that same classifier. A regression that re-rendered only the bottom-anchored mounted window when it falls entirely inside `suffix(8)` (or the top `prefix(8)`) would be classified as edge churn *and* inflate the allowance in lockstep, so both guards vacate simultaneously. The deep-transcript test has data-layer invariants (zero mutations / `transcriptChanged` / prefix walks) as backstop, but `testStreamingTicksLeaveSettledMarkdownDormant_MarkdownTranscript` has no such invariant — it relies entirely on this render-layer heuristic.
**Suggestion:** Shrink `edgeMargin` to the empirically observed churn extent (the PR cites Turn 0/2 and 796/798, i.e., ~2 rows), and/or add an absolute ceiling on `selectableTextViewUpdateCalls` independent of the allowance so the allowance cannot fully absorb a regression.

### [nit] Live-row static bounds leave zero slack for legitimate variation
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:242
**Issue:** `selectableTextViewUpdateCalls ≤ 30` with allowance 0 is exactly 3 updates × 10 ticks with no headroom; the flake this PR fixes (33 vs ≤30) landed only one stray update above the bound. Any simulator/device variant that renders the live row into one extra `SelectableTextView` per tick (e.g., the growing content crossing a quote/code boundary) re-creates a flake that is not a regression.
**Suggestion:** Keep the original strict bound at zero churn as stated, but add a small constant slack independent of the remount allowance so counting noise does not flake while the classifier still distinguishes the cascade.

### [nit] Environment snapshot read per settled-body evaluation in DEBUG
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:79-84
**Issue:** `ProcessInfo.processInfo.environment["CONDUIT_PERF_TRACE"]` materializes a dictionary copy on every settled-markdown body evaluation in DEBUG builds (the hot path), even when the env var is unset. RELEASE is unaffected (compiled out), so this is not a production cost.
**Suggestion:** Hoist the env check into a lazily cached `static let isPerfTraceEnabled` so the dict copy happens once.

## Verdict
APPROVE

The change is test/instrumentation scoped: production behavior is untouched (the only production edits are widening one method's access and adding DEBUG-only ledger bookkeeping in the body-eval path). No retain cycles, force unwraps, or concurrency hazards introduced — `TranscriptPerf.storage` remains effectively main-thread confined (SwiftUI `body` + `@MainActor` tests), and the `flushReasoningPublish` seam is safe since its cancellation path is synchronous on the main actor. The findings above are robustness notes on the new classification heuristics, not correctness defects in the stated goal; the primary data-layer invariants and the interior-row classifier do catch the actual pre-fix cascade shape.

<a href="https://opencode.ai/s/1eYB2Bm5">New%20session%20-%202026-09-03T23%3A28%3A36.793Z</a>
[opencode session](https://opencode.ai/s/1eYB2Bm5)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33817683423)

kaishi00:
Superseded by a clean delta branch — the amended SHAs made this PR unmergeable against main (same changes, different commits). See #126.

PR #128, opened 2026-09-05
feat(ipad): optional persistent sessions sidebar

kaishi00:
## Summary

Adds an **opt-in persistent sessions sidebar for iPad**: with the new Appearance toggle enabled and the window wide enough, the existing Sessions / Cron / Kanban sidebar becomes a permanent left column beside the conversation instead of a modal drawer. iPhone behavior is untouched, and iPad without the opt-in (or with a narrow window) behaves exactly as today.

### Preference (device-local)

- `@AppStorage("conduit.ipadPersistentSidebar")`, default **off** — existing users keep the drawer.
- Surface: `AppearanceSettingsDetail` → new "Layout" section, shown **only** when `UIDevice.current.userInterfaceIdiom == .pad`. A normal `Toggle`; copy states the width condition ("…when the window is wide enough").
- Entirely device-local; nothing is added to the Hermes profile/config.

### Layout decision (pure + centralized)

`Conduit/Views/SidebarLayout.swift` (new) holds all decision logic so it is testable without hosting UI:

- `SidebarPresentation` — `.drawer` / `.persistent`
- `SidebarLayoutMetrics` — `persistentSidebarWidth = 320`, `minimumChatWidth = 430`, `minimumPersistentLayoutWidth = 750` (single source of truth for the threshold)
- `SidebarLayoutPolicy.resolvePresentation(idiom:prefersPersistentSidebar:availableWidth:)` — `.persistent` only for **pad + opt-in + width ≥ 750**. Full-screen portrait iPad (768) qualifies; Split View halves (≤683), Stage Manager tiles, and Slide Over stay on the drawer. Exactly 750 still leaves the 430 pt minimum chat column; 749 falls back.
- `SidebarLayoutPolicy.shouldPresentDrawerForReturnSurface(persistentSidebarActive:drawerPresented:)` — return-surface presentation rule (below).

MainView measures its window via a background `GeometryReader` → `PreferenceKey` (iOS 17-compatible; `onGeometryChange` is iOS 18+), so rotation, Split View divider drags, and window resizing re-resolve the layout live.

### One SidebarView, one chat shell

- The persistent branch is `HStack(spacing: 0) { SidebarView(presentation: .persistent) | Divider | chat }`; the drawer branch is the existing shell with the existing `.sheet(isPresented: $appState.showSidebar)`.
- The chat `NavigationStack`, backdrop, and toolbar are a single shared computed view — **no duplicated chat toolbar**. The hamburger and left-edge `EdgePanGesture` are conditional inside the shared shell: hidden while the persistent sidebar is visible, restored verbatim in drawer/compact mode. The iPhone path is unchanged line-for-line.
- `SidebarView` gains a `presentation` parameter (default `.drawer`, so every existing call site is untouched). Persistent mode hides the X close button; nothing else about the sidebar is redesigned. A 1 pt `Divider` (extended through the safe areas) is the only added framing; the sidebar's existing backdrop already bleeds edge-to-edge.

### `appState.showSidebar` semantics preserved

`showSidebar`'s `didSet` cancels/flushes streaming and reasoning publication, so it must keep meaning *transient drawer presentation*. The persistent column **never** sets it:

- All six in-sidebar session-opening flows (New Chat, session rows, project sheet rows + new-conversation, Cron recent runs, Cron job run history) now call the new `AppState.dismissSidebarDrawer()` — a guard-then-close helper that is a no-op when the drawer isn't presented. Drawer mode behaves identically; persistent mode leaves the column visible while the conversation opens.
- A transition guard (`.onChange(of: isPersistentSidebarActive)`) dismisses the drawer if the layout flips to persistent while the sheet is open (rotation past the threshold, Split View drag, or enabling the toggle) — otherwise `showSidebar` would stay `true` with no sheet, silently suppressing streaming publication.

### Settings and preferred return surface

- Drawer mode: tap Settings → close drawer → present Settings (existing sequencing, unchanged).
- Persistent mode: tap Settings → sidebar stays visible → the same `SettingsView` sheet presents directly (`isSettingsSheetPresented` set before the snapshot, matching existing precedence semantics).
- `presentPreferredReturnSurfaceIfNeeded` still claims the request exactly once via `claimPreferredReturnSurfacePresentation()` (consumption/deferral semantics untouched in AppState), then consults the policy: an active persistent sidebar consumes the request without opening a redundant drawer; drawer layouts re-present exactly as before.

## Scope exclusions

- No `NavigationSplitView` conversion, no app-wide navigation redesign, no second session list, no session state moved out of `AppState`, no Settings/SidebarView redesign, no Hermes server/profile changes, no phone UI changes, no unrelated cleanup.

## Tests

New `ConduitTests/SidebarLayoutTests` (9 tests, no UI host):

- phone ± preference at any width → drawer
- iPad + preference off at any width → drawer
- iPad + preference on + wide (768/834/1024/1194/1366) → persistent
- iPad + preference on + narrow (0/320/417/512/683/749) → drawer
- threshold identity + exact 750 (persistent) / 749 (drawer) boundary
- return-surface: persistent active → never presents the drawer; drawer layout → presents unless already presented
- `dismissSidebarDrawer()`: closes an open drawer; no-op when closed (persistent mode requires no `showSidebar` activity)

## Verification

- `python scripts/plan-tests.py validate` → OK, `SidebarLayoutTests` discovered; `python -m unittest discover -s scripts/tests` → 73 tests OK.
- `git diff --check` clean (no lint/format tooling in repo).
- Mac build host (bundle → temp worktree at `ef1fd45` → `xcodegen generate` → focused `xcodebuild test`, iPhone 17 Pro simulator, signing disabled): **all 9 `SidebarLayoutTests` passed; all 23 `ChatReturnSurfaceTests` passed** (the drawer/return-surface suite most affected by this area). Several runs exited 65 with `** TEST FAILED **` but **zero assertion failures** — the host's known AURemoteIO coreaudio deadlock killed the app host mid-run; per established practice the started-but-unreported cases were re-run in smaller batches and every one passed (final solo run `TEST SUCCEEDED`).
- Diff reviewed specifically for accidental iPhone-path changes (none — the compact shell's modifiers and order are identical).
- Review gate: 3-model cycle run on this exact diff; reviewer 1 REQUEST CHANGES (transition guard + one NIT — both fixed), reviewer 2 APPROVE (two suggestions rejected as out of scope / mismatched to code; noted in thread). Hosted CI remains the authoritative full-suite gate.

## Summary by CodeRabbit

* **New Features**
  * Added an optional persistent sessions sidebar layout for iPad on sufficiently wide windows.
  * Added a Layout setting to enable or disable the persistent sidebar.
  * Sidebar presentation now adapts between a persistent column and a dismissible drawer based on device and available space.
  * Selecting a session from the sidebar now dismisses the drawer when applicable.

* **Bug Fixes**
  * Prevented drawer state from remaining active when the persistent sidebar is displayed.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -736,6 +736,15 @@ final class AppState: ObservableObject {
             }
         }
     }
+    /// Closes the modal sessions drawer if it is open. Session-opening flows
+    /// inside the sidebar call this unconditionally: in drawer mode it
+    /// dismisses the sheet, while in the iPad persistent-sidebar layout the
+    /// drawer is never presented (`showSidebar` stays false), so this is a
+    /// no-op and the persistent column remains visible.
+    func dismissSidebarDrawer() {
+        guard showSidebar else { return }
+        showSidebar = false
+    }
     @Published var showModelPicker = false
     @Published var showContextSheet = false
     @Published var showWorkspaceSheet = false
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -1371,9 +1371,11 @@ private struct AppearanceSettingsDetail: View {
     @EnvironmentObject private var appState: AppState
     let theme: ThemePreference
     let saveTheme: (ThemePreference) -> Void
+    @AppStorage("conduit.ipadPersistentSidebar") private var iPadPersistentSidebar = false
     @State private var selected: ThemePreference
     @State private var isChangingIcon = false
     init(theme: ThemePreference, saveTheme: @escaping (ThemePreference) -> Void) { self.theme = theme; self.saveTheme = saveTheme; _selected = State(initialValue: theme) }
+    private var isPad: Bool { UIDevice.current.userInterfaceIdiom == .pad }
     var body: some View {
         SettingsDetailContainer {
             ConduitSettingsSection(title: "Theme", symbol: "circle.lefthalf.filled", tint: .conduitAccent) {
@@ -1426,6 +1428,16 @@ private struct AppearanceSettingsDetail: View {
                         .disabled(isChangingIcon)
                         .accessibilityLabel("Use \(choice.title.lowercased()) app icon")
                     }
+                }
+            }
+
+            if isPad {
+                ConduitSettingsSection(title: "Layout", symbol: "sidebar.left", tint: .conduitAura) {
+                    Toggle("Persistent session sidebar", isOn: $iPadPersistentSidebar)
+                        .tint(.conduitAccent)
+                    Text("Keep Sessions, Cron, and Kanban visible beside the current conversation when the window is wide enough.")
+                        .font(.footnote)
+                        .foregroundStyle(.secondary)
                 }
             }
         }.navigationTitle("Appearance")
--- a/Conduit/Views/RootView.swift
+++ b/Conduit/Views/RootView.swift
@@ -38,80 +38,23 @@ struct RootView: View {
 
 struct MainView: View {
     @EnvironmentObject var appState: AppState
+    @AppStorage("conduit.ipadPersistentSidebar") private var prefersPersistentSidebar = false
+    @State private var availableWindowWidth: CGFloat = 0
     @State private var settingsPresentation: SettingsSnapshot?
     @State private var shouldPresentSettingsAfterSidebarDismissal = false
 
+    private var sidebarPresentation: SidebarPresentation {
+        SidebarLayoutPolicy.resolvePresentation(
+            idiom: UIDevice.current.userInterfaceIdiom,
+            prefersPersistentSidebar: prefersPersistentSidebar,
+            availableWidth: availableWindowWidth
+        )
+    }
+
+    private var isPersistentSidebarActive: Bool { sidebarPresentation == .persistent }
+
     var body: some View {
-        NavigationStack {
-            ZStack {
-                ConduitBackdrop()
-                ChatView()
-            }
-            .overlay(alignment: .leading) { EdgePanGesture { appState.showSidebar = true }.frame(width: 25).ignoresSafeArea() }
-            .navigationTitle("")
-            .navigationBarTitleDisplayMode(.inline)
-            .toolbarBackground(.hidden, for: .navigationBar)
-            .toolbar {
-                ToolbarItem(placement: .topBarLeading) {
-                    Button {
-                        appState.showSidebar = true
-                    } label: {
-                        Image(systemName: "line.3.horizontal")
-                            .font(.system(size: 16, weight: .semibold))
-                            .frame(width: 40, height: 40)
-                    }
-                    .conduitGlassControl(cornerRadius: 20, tint: .conduitAccent.opacity(0.10))
-                    .accessibilityLabel("Open sessions")
-                }
-                ToolbarItem(placement: .principal) {
-                    Button {
-                        appState.requestChatScrollToTop()
-                    } label: {
-                        Text(appState.activeSessionTitle)
-                            .font(.subheadline.weight(.semibold))
-                            .lineLimit(1)
-                            .padding(.horizontal, 12)
-                            .padding(.vertical, 7)
-                            .conduitGlassSurface(cornerRadius: 16, tint: .conduitAccent.opacity(0.06))
-                    }
-                    .buttonStyle(.plain)
-                    .accessibilityLabel(appState.activeSessionTitle)
-                    .accessibilityHint("Scroll to top of conversation")
-                }
-                ToolbarItem(placement: .topBarTrailing) {
-                    Button {
-                        Task { await appState.refreshActiveSession() }
-                    } label: {
-                        Image(systemName: "arrow.clockwise")
-                            .font(.system(size: 15, weight: .semibold))
-                            .rotationEffect(.degrees(appState.isChatRefreshing ? 360 : 0))
-                            .animation(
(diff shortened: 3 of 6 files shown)


kaishi00:
## Summary
Adds an opt-in iPad persistent sessions sidebar: with `conduit.ipadPersistentSidebar` enabled and the window ≥750pt wide, `SidebarView` renders as a fixed leading column beside the chat instead of the modal drawer; all layout policy is centralized in a pure, tested `SidebarLayoutPolicy`.

## Findings

### [minor] Threshold math ignores the 1pt Divider
**File:** Conduit/Views/SidebarLayout.swift:21; Conduit/Views/RootView.swift:158
**Issue:** `minimumPersistentLayoutWidth = persistentSidebarWidth + minimumChatWidth` = 750, but the actual chat column in the HStack is `availableWidth − 320 − 1` because of the `Divider()`. At exactly 750 the chat column is 429pt, one point below the declared 430pt `minimumChatWidth`. The PR body's claim ("exactly 750 still leaves the 430pt minimum chat column") and the boundary test (`SidebarLayoutTests.swift:98`) both miss the divider.
**Suggestion:** Include the divider in the formula (e.g. `persistentSidebarWidth + DividerWidth + minimumChatWidth`) or explicitly accept the 1pt slop in a comment.

### [minor] Crossing the 750pt threshold tears down and rebuilds the whole chat shell
**File:** Conduit/Views/RootView.swift:148-166
**Issue:** `sidebarLayoutContent` uses a `@ViewBuilder` branch swap; when the window crosses the threshold (Split View divider drag, Stage Manager resize, rotation), the entire `NavigationStack`/`ChatView` subtree is destroyed and recreated on the other branch. Chat content itself lives in `AppState`, but transitory view state in ChatView (`@State bottomMarkerMaxY`, `viewport`, `renderedScrollContent`, backfill task — ChatView.swift:14-32) is reset, so a user scrolled up in history gets yanked back to bottom by a resize. Separately, `availableWindowWidth` starts at 0 before the first geometry measurement, so opt-in wide iPads render one drawer-frame before flipping to persistent at launch/sign-in; and the cold-launch `.task` (RootView.swift:135-138) may present the drawer once before the measurement lands, only for the `.onChange` guard to dismiss it (drawer flicker on first sign-in).
**Suggestion:** Prefer a stable chat identity across presentations (e.g. keep the chat subtree structurally stable and only swap the outer container/safe-area inset), and gate the persistent decision on a distinct "measured yet" state so width 0 is never treated as a real narrow window.

### [minor] Accessibility & Dynamic Type regression in the persistent column
**File:** Conduit/Views/RootView.swift:156; Conduit/Views/SidebarView.swift:31-87
**Issue:** The persistent column is fixed at 320pt with fixed-height header controls (`frame(height: 48)`, 44pt buttons). With the sidebar previously presented at full-modal/sheet width on iPad, large Dynamic Type sizes will now truncate the workspace/profile header text. Additionally, the sidebar appearing/disappearing gives VoiceOver users no announcement, and the toolbar re-flow (hamburger hiding) changes the navigation order each time the layout flips.
**Suggestion:** Make the header size to fit Dynamic Type (e.g. `.dynamicTypeSize`-aware row heights, allow text to scale within the 320pt column), and mark the persistent sidebar with an appropriate accessibility trait/announcement.

### [nit] Fragile invariant: drawer sheet is still attached in persistent mode
**File:** Conduit/Views/RootView.swift:241
**Issue:** `chatNavigationContent` retains `.sheet(isPresented: $appState.showSidebar)` even when the persistent branch is active. Correctness relies entirely on the three gates (hidden hamburger/edge gesture, return-surface policy) plus the `.onChange` guard keeping `showSidebar == false`. A future caller that sets `showSidebar = true` while persistent would present a drawer over the column and silently suppress streaming publication (the stated hard invariant).
**Suggestion:** Defense-in-depth: gate the sheet attachment on `!isPersistentSidebarActive` as well, or assert in `showSidebar.didSet`.

### [nit] Tautological no-op test
**File:** ConduitTests/SidebarLayoutTests.swift:145-150
**Issue:** `testDismissSidebarDrawerIsNoOpWhenDrawerIsClosed` sets nothing and merely asserts `showSidebar == false` after calling the helper on an already-closed drawer — it cannot detect the claimed guarantee (no flush/state mutation), since `showSidebar` was already `false`. The interesting contract (that the persistent mode performs no winding of `showSidebar`) isn't actually exercised (e.g. assert no streaming-publish cancellation/flush side effect occurred).
**Suggestion:** Assert on an observable side effect, such as that no streaming buffer flush ran, or that `showSidebar` never transitions to `true`.

### [nit] No re-entrancy guard on persistent-mode settings presentation
**File:** Conduit/Views/RootView.swift:257-260
**Issue:** `presentSettingsFromPersistentSidebar` sets `isSettingsSheetPresented` and `settingsPresentation` unconditionally; a rapid double-tap on the gear can set the item twice. Harmless given `.sheet(item:)` semantics, but inconsistent with the guarded drawer path.
**Suggestion:** `guard settingsPresentation == nil else { return }` or route through the same single presentation helper.

## Verdict
APPROVE

Design is sound: the policy is pure and centralized with good test coverage, `showSidebar` semantics and its streaming-flush contract are preserved, the iPhone path is untouched, and explicit attention was paid to the layout-flip transition guard. All findings are minor/nit and none block merging.

<a href="https://opencode.ai/s/1JsZw0pi">New%20session%20-%202026-09-05T10%3A10%3A36.502Z</a>
[opencode session](https://opencode.ai/s/1JsZw0pi)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33959888655)

PR #132, opened 2026-09-05
fix: preserve active session once composer editing begins

kaishi00:
## Summary

After foregrounding Conduit, the visible conversation could be replaced by another session several seconds later — while the user was already typing. This PR makes a genuine composer edit an explicit claim on the visible conversation: any in-flight automatic foreground-return work loses its authority to pick a different session, while transport recovery still proceeds and preserves the current one.

Round 2 (review follow-up): the ownership claim moved from the SwiftUI `onChange(of: text)` layer down to the UIKit delegate boundary in `ComposerPasteTextView` — the earliest authoritative point a text mutation is known to be genuine user input — and the round-1 classification bookkeeping (`hasPendingProgrammaticReplacement`, `ComposerTextChangeSource`) was removed entirely.

## The race

1. The app returns to `.active` with session **A** visible. `handleScenePhase(.active)` mints an automatic-work token and starts the foreground task.
2. The healthy path runs read-only liveness checks first. But if `verifyChatResumeTransportHealth` throws, or the transport is missing, the task falls back to `reconnectForRetry(purpose: .automaticReturn)`; some healthy-path branches (empty local transcript, cross-surface attach) also call `syncSession(purpose: .automaticReturn, …)`.
3. `.automaticReturn` is allowed to choose a session by the configured resume behaviour. With *Continue where I left off* that is the **saved** session pointer, not the one on screen; with *Jump to latest activity* it is the first chat row of the catalog. Either can differ from A — the saved pointer in particular explains the report that it "tends to switch to the same older session".
4. The user starts typing in A while the health check / ticket mint / reconnect is still suspended.
5. The delayed recovery completes, `performSyncSession` selects **B**, and `applyChatResume` swaps `activeSessionId` under the user's cursor.

Sending, navigation, and scroll drags already call `cancelChatResumeRestoration()` and are therefore immune. Composer typing did not, so it was the one common user action that left the automatic-return intent alive.

### Why flaky / slow transport exposes it

On a healthy link the foreground check finishes in well under a second, before anyone can type. A slow health check, a slow dashboard ticket mint, or a socket that only fails after a timeout stretches steps 2–5 across the several seconds the user needs to start typing — exactly the window described in the report. It is not reproducible locally because the local gateway answers instantly; the fix is therefore race-condition hardening rather than a replay of a specific failure.

## The fix

**Ownership is claimed at the UIKit edit boundary.** `ComposerPasteTextView` gains `var onUserEdit: (() -> Void)?`. `Coordinator.textViewDidChange` calls it immediately after its existing `guard isActive, !isApplyingProgrammaticState else { return }` and **before** the binding write:

```text
real UIKit edit → textViewDidChange → noteComposerUserEdit()
                    → automatic return loses authority immediately
                    → binding updates normally

programmatic replacement → revision machinery / isApplyingProgrammaticState
                          → no ownership claim
```

Because every programmatic writer (draft restore, prefill, slash insertion, post-send collapse, failed-submit restore, the deferred IME landing) routes through `replaceComposerText` → `composerRevision` → `performProgrammaticReplacement` under `isApplyingProgrammaticState`, none of them can reach the callback. There is no string comparison and no cross-layer classification bookkeeping left: `replaceComposerText` is back to being solely text + revision, and `ComposerBar.onChange(of: text)` keeps only its original slash-suggestion logic.

**`AppState.noteComposerUserEdit()`** reuses the existing invalidation machinery — it calls `cancelChatResumeRestoration()`, which:

- bumps the coordinator's automatic-cancellation epoch so every in-flight checkpoint (`automaticChatResumeWorkIsCurrent`, including the guard inside `applyChatResume`) sees a stale token;
- demotes `recoverySequence` and any queued reconnect purpose to `.preserveCurrent`, so an armed retry timer fires as `.preserveCurrent` (it re-reads its purpose at fire time);
- lets a reconnect that is *already running* continue: `transportContinuation` hands `.automaticReturn` off to `.preserveCurrent`, and `synchronizeTransportContinuation` re-syncs against the current session;
- a foreground attempt whose health check *fails after* the edit repairs the transport with `reconnectForRetry(purpose: .preserveCurrent)` in both scene-failure branches instead of dropping recovery — the socket is fixed, and the repair resumes the visible session rather than letting the resume policy choose.

**One cancellation per generation (no keystroke churn).** `chatResumeRestorationRequest` is `@Published`, so a per-keystroke cancel would re-render `ChatView` while typing. `noteComposerUserEdit()` therefore (a) acts only when automatic-return work is actually outstanding (scene attempt in flight, armed reconnect timer, live automatic sync/reconnect operation, `.automaticReturn` registered or queued in the recovery sequence, or an unconsumed restoration request) — steady-state typing is a pure no-op with zero published state; and (b) latches after its first cancellation, cleared in `beginAutomaticChatResumeWork()` when the next automatic generation begins — so repeated keystrokes inside one outstanding window publish nothing after the first edit.

Nothing about startup or pre-interaction foreground resume changes: with no composer edit both resume behaviours restore exactly as before (pinned by control tests). Draft isolation, cursor behaviour, IME deferral, and the rest of the bridge are untouched.

## Regression tests

`AppStateChatResumeTests` (existing harness, `ControlledSuspension` seams):

- `testComposerUserEditDuringSuspendedReconnectPreservesVisibleSession` / `…LatestActivity…` — **the modelled race**: A visible, B in the catalog and chosen by `.automaticReturn` (saved pointer for Continue-where-I-left-off, newest row for Jump-to-latest); `reconnectForRetry(.automaticReturn)` suspended at the ticket mint; composer edit while suspended; recovery released. Asserts the final active session is A, only A is resumed, `recoverySequence.currentPurpose == .preserveCurrent`, and the transport recovery still completed (`isConnected`, idle).
- `testSuspendedReconnectWithoutComposerInteractionStillRestoresSavedSession` / `…StillRestoresNewest` — **controls**: same flow with no edit still restores B for both resume behaviours.
- `testForegroundHealthCheckFailureRespectsComposerUserEditOwnership` (in `AppStateForegroundLifecycleTests`, real `handleScenePhase(.active)` task over the fake connected transport) — the **slow health-check window**: the foreground task is suspended inside `verifyTransportHealth`, the user edits, the check then fails. Asserts the fallback repairs the transport (one ticket mint, `isConnected`) while resuming the visible session — never the saved older one — with `.preserveCurrent`.
- `testForegroundHealthCheckFailureWithoutComposerEditStillRestoresSavedSession` — control: the same failure with no edit still falls back to `.automaticReturn` and restores the saved older session.
- `testComposerUserEditDemotesArmedAutomaticReturnReconnect` — an armed `.automaticReturn` retry timer fires as `.preserveCurrent` after an edit.
- `testComposerUserEditLatchesUntilNextAutomaticWorkGeneration` — three edits during one parked-reconnect window produce exactly one published state change.
- `testComposerUserEditWithoutOutstandingAutomaticWorkPublishesNothing` — zero `objectWillChange` emissions from repeated calls in steady state.
- `testComposerUserEditCancelsPublishedRestorationRequest` — the published-restoration-request guard arm.
- `testComposerUserEditDuringInFlightAutomaticReturnSyncDoesNotSelectSession` — an edit while an `.automaticReturn` sync is suspended on its catalog fetch stops the selection; the sync settles without resuming anything.

`ComposerPasteTextViewTests` — ownership coverage now lives at the bridge level (the coordinator is driven directly, per the existing harness):

- `testUserEditInvokesOnUserEditBeforeBindingUpdate` — a genuine `textViewDidChange` claims ownership synchronously, before the SwiftUI binding observes the new text.
- `testProgrammaticTextApplicationDoesNotPublishAsUserEditing` — a delegate delivery under `isApplyingProgrammaticState` neither publishes the binding nor claims.
- `testRepeatedUserEditsInvokeOnUserEditEachTime` — every genuine mutation claims; the per-generation collapse is AppState's latch (pinned by the parked-reconnect test above), deliberately out of the bridge.
- `testDeferredIMEProgrammaticReplacementDoesNotClaimOwnership` — a replacement queued while IME marked text is active lands through the programmatic path with zero claims.
- `testCompositionCommitClaimsOnceWhileDeferredReplacementLands` — the production-realistic interleaving: the composition commit claims exactly once, and the deferred replacement landing at the end of the same callback claims nothing.
- `testInactiveCoordinatorDoesNotClaimOwnershipOnUserEdit` — a delegate delivery racing teardown does not claim.

### Failing against current behaviour

The tests call the new `noteComposerUserEdit()` API, so they cannot compile on `main` as-is. To confirm they model the race rather than the fix, the production guard was temporarily neutered on the build Mac (`guard false` — composer edits ignored, i.e. `main` behaviour) and the four ownership tests were re-run: they fail at the session-swap assertions (B resumed / active, `.automaticReturn` retry fired), while the controls still pass. Restoring the guard turns all of them green.

## Verification

- Mac build host (iPhone 17 Pro simulator), final head (`29e2b69`): 8 suites — `AppStateChatResumeTests` (93), `AppStateForegroundLifecycleTests` (95), `ChatResumeCoordinatorTests` (18), `ChatResumePolicyTests` (10), `ChatReturnSurfaceTests` (23), `ComposerDraftStoreTests` (18), `ComposerPasteTextViewTests` (24), `ComposerReturnKeyTests` (21) — **302 tests, 0 failures**. Full unit suite runs via hosted CI (the local Mac full-suite runner has a known host-level AURemoteIO wedge; per-suite verification covers everything this branch touches).
- Multi-model review, round 2 (three independent reviewers): **approve-with-nits × 3** — unanimous. All rigor checks came back clean (no state exists where `textViewDidChange` would wrongly claim; claim-before-binding ordering is re-entrancy-safe; every programmatic path provably skips the callback; round-1 machinery unchanged). All suggestions were applied as the two extra bridge tests plus doc condensing. Round 1 (request-changes / approve / approve-with-nits) had both WARNINGs fixed in round 1: per-keystroke publish churn (latch) and the health-check-failure recovery drop (`.preserveCurrent` repair).

## Out of scope

- PR #128 / persistent iPad sidebar: no dependency found; untouched.
- Attachment-only composer interaction does not claim ownership; the report and this fix are about typing. Easy follow-up if wanted.
- Why the saved session pointer can lag the visible session in the first place (`setActiveSessionState` clears it when the canonical ID is unresolvable) is not changed here; with this fix it can no longer pull the user away mid-edit regardless.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chat resume behavior when users begin editing during automatic session restoration.
  * Preserves the currently visible conversation instead of switching sessions after a user edit.
  * Improved foreground recovery by reconnecting the transport when automatic restoration is interrupted.
  * User edits made through typing, deletion, paste, autocorrect, or composition are now recognized consistently.


kaishi00:
## Summary
This PR makes a genuine composer edit claim ownership of the visible conversation so that in-flight automatic foreground-return work (reconnect/sync/health-check fallbacks) can no longer swap `activeSessionId` under the user's cursor. It routes edits through the existing invalidation machinery with a one-shot latch to avoid per-keystroke published-state churn, and repairs the transport as `.preserveCurrent` when a health check fails after an edit.

## Findings

### [minor] Programmatic-replacement mark can leak and swallow the next keystroke's ownership claim
**File:** Conduit/Views/Components/ComposerBar.swift:172
**Issue:** `replaceComposerText` sets `hasPendingProgrammaticReplacement` when `oldText != newText`, and only `onChange(of: text)` consumes it. Because SwiftUI state writes in one transaction coalesce to a single `onChange` fire, an A→B→A pair inside one render transaction would leave the mark set with no observer firing (net change is none), so the *next genuine keystroke* gets classified as `.programmaticReplacement` and fails to claim the session — exactly the race this PR fixes. The code comment acknowledges this leak and asserts "no such caller exists," which is true today, but the invariant is implicit and fragile against future callers.
**Suggestion:** Guard the lease, e.g. store the pending value alongside the mark and only treat it as programmatic if the observed `newValue == pendingValue`, or clear the mark at the top of a body evaluation so it can never outlive its transaction. A unit test for the A→B→A coalesced pair would pin the invariant.

### [minor] Successful `.preserveCurrent` repair doesn't cancel an already-armed (demoted) retry timer
**File:** Conduit/Services/AppState.swift:4598, 4614
**Issue:** In both new foreground fallback branches, after an edit invalidates the token the code runs `reconnectForRetry(purpose: .preserveCurrent)` in place of the old silent `return`. If a backoff retry timer was armed for `.automaticReturn` during the same `.active` window (flaky-link multi-cycle scenario), the edit demotes it to `.preserveCurrent`, but a successful repair doesn't `cancelScheduledReconnect()` (that only happens via `settleReconciliationAndPublish`). The demoted timer can still fire and run a redundant full reconnect after the transport is already repaired. Harmless (it preserves the visible session) but wasteful, and it re-runs the transcript sync.
**Suggestion:** On successful completion of a `.preserveCurrent` repair, cancel the scheduled reconnect (while leaving `recoverySequence.currentPurpose` as `.preserveCurrent`, since the tests intentionally assert that).

### [nit] Classification relies on a single-consumer `@State` mark that resets if `ComposerBar` is recreated
**File:** Conduit/Views/Components/ComposerBar.swift:47
**Issue:** `hasPendingProgrammaticReplacement` is view-scoped `@State`. If `ComposerBar` is destroyed/recreated while a programmatic replacement is deferred (the UIKit bridge already defers during active IME composition, ComposerPasteTextView.swift:214), the mark is lost and the replacement's observer fires as `.userEdit`. The write-back paths write identical text back to the binding, so in practice `onChange` doesn't re-fire after a programmatic set; the residual risk is a false *userEdit* classification, which over-cancels rather than under-cancels — so the reported race would not recur, but the intent model becomes imprecise.
**Suggestion:** Acceptable given the safe failure direction; consider relocating the mark to a value that survives view recreation (e.g. alongside `composerRevision`) if strictness is desired.

### [nit] No UI-level wiring test for ComposerBar → AppState
**File:** ConduitTests/AppStateChatResumeTests.swift:3592
**Issue:** The end-to-end seam between `ComposerBar.onChange(of: text)` classification and `AppState.noteComposerUserEdit()` is not exercised by an automated test; only the pure classification functions and the AppState API are pinned. The author correctly notes the UI-level wiring would need UI automation.
**Suggestion:** Fine to defer, but a `XCTExpectFailure`-guarded or view-harness test driving a real keystroke through the UIKit bridge would complete the contract.

Notable strengths: no new force unwraps or retain cycles; the latch's re-arm is correctly bound to `beginAutomaticChatResumeWork()` (the sole token mint site), so "one cancellation per generation" holds across scene-phase and reconnect generations; all new state is `@MainActor`-serialized; the regression tests model the race with real suspension seams and include proper controls.

## Verdict
APPROVE

<a href="https://opencode.ai/s/SYvKzKrw">New%20session%20-%202026-09-05T19%3A45%3A29.613Z</a>
[opencode session](https://opencode.ai/s/SYvKzKrw)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33988009961)

kaishi00:
## Summary
This PR fixes a race where an in-flight automatic foreground-return reconnect/health-check could swap the visible session under the user's cursor. A genuine composer edit now claims ownership of the visible conversation at the UIKit delegate boundary (`ComposerPasteTextView.textViewDidChange` → `AppState.noteComposerUserEdit`), which cancels automatic session-selection intent while still letting transport recovery proceed as `.preserveCurrent`. Round-1 classification machinery was removed; the ownership signal was moved to the bridge; the fix is latched per automatic-work generation to avoid per-keystroke `@Published` churn; regression tests cover both the race and controls.

## Findings

### [minor] Duplicated invalidation block across the two scene-failure branches
**File:** Conduit/Services/AppState.swift:4587 and Conduit/Services/AppState.swift:4608
**Issue:** The `guard scenePhaseAttemptIsCurrent(sceneAttemptID)` / `if !automaticChatResumeWorkIsCurrent(automaticWorkToken)` → `settleReconciliation(token)` → `reconnectForRetry(.preserveCurrent)` sequence is copy-pasted in both the healthy-transport error branch and the missing-transport branch. The two copies must stay in lockstep (one logs before reconnect, the other has no log). A future change to the invalidation contract risks fixing only one site.
**Suggestion:** Extract a shared `continueAutomaticForegroundFallback(sceneAttemptID:workToken:reconciliationToken:) async -> Bool` helper (as CodeRabbit suggested), returning whether the `.automaticReturn` fallback should run and performing the settle + `.preserveCurrent` repair otherwise, then have both branches `guard await ... else { return }`.

### [minor] `.preserveCurrent` repair does not cancel an armed/demoted retry timer
**File:** Conduit/Services/AppState.swift:4597 / 4613 (repair sites), versus 4358
**Issue:** `reconnectForRetry` cancels an armed timer only when `purpose == .automaticReturn` (line 4358). In the new post-edit branches the repair runs as `.preserveCurrent`, so a retry timer that was armed earlier in the same window is left armed (demoted to `.preserveCurrent`). It later fires and runs a redundant full reconnect (mint, catalog reload, transcript sync) after the transport is already repaired. Harmless — it preserves the visible session — but wasteful.
**Suggestion:** On successful completion of the `.preserveCurrent` repair, `cancelScheduledReconnect()` while leaving `recoverySequence.currentPurpose == .preserveCurrent` (the tests assert that). Contrast with the demote-only intent in `testComposerUserEditDemotesArmedAutomaticReturnReconnect`, where no repair happens and the timer must fire — so gate the cancel on the repair actually succeeding.

### [minor] `recoverySequence.currentPurpose == .automaticReturn` arm admits a spurious cancel in one steady state
**File:** Conduit/Services/AppState.swift:1830
**Issue:** The compute guard includes `recoverySequence.currentPurpose == .automaticReturn`, but `currentPurpose` is only reverted to `.preserveCurrent` by `recoverySequence.complete()`, which runs in `settleReconciliationAndPublish` — not on the `reconnectForRetry` failure legs. If an `.automaticReturn` reconnect ends in `requireSignIn` (the mint throws `.signInRequired`, no timer is armed, ops finish), `currentPurpose` stays `.automaticReturn` with nothing outstanding. The first keystroke after signing back in would then pass the guard and perform a spurious `cancelChatResumeRestoration()` + one `objectWillChange` emission, contradicting the documented "steady-state typing is a pure no-op" property, and it demotes the purpose used by the next socket-drop reconnect. The `reconnectTask`/operation arms already cover genuinely in-flight work, making this arm mostly redundant.
**Suggestion:** Either drop the `currentPurpose` tests from the guard (the armed-timer case is already covered by `queuedReconnectPurpose`/`reconnectTask`), or document that "nothing outstanding" implicitly requires `recoverySequence` to have completed, and add a test seeding a stale `.automaticReturn` `currentPurpose` to pin the coupling.

### [nit] Misc doc/testing polish
**File:** Conduit/Services/AppState.swift:1825 and ConduitTests/AppStateChatResumeTests.swift:3652
**Issue:** `automaticChatResumeWorkMayStillSelectSession` has no doc comment (CodeRabbit docstring coverage for touched functions is 58% vs an 80% threshold), and `testComposerUserEditWithoutOutstandingAutomaticWorkPublishesNothing` only seeds the default `.preserveCurrent` sequence, so the stale-`currentPurpose` instance above is unpinned by tests.
**Suggestion:** Add a brief doc comment; consider a control asserting the latch/publish behavior with `recoverySequence` left in a completed-but-automaticReturn state.

No critical or major findings: the ownership claim is placed at the earliest authoritative point, ordering (claim before binding write) is re-entrancy-safe, every programmatic path provably skips the callback, all new AppState state is `@MainActor`-isolated, no force unwraps or retain cycles were introduced (the `onUserEdit` closure captures the environment-owned `appState`, which never references back into the view), and the delegate method is `@MainActor`-annotated by the SDK so the MainActor call is sound under strict concurrency.

## Verdict
APPROVE

<a href="https://opencode.ai/s/J0mMnXCj">New%20session%20-%202026-09-05T21%3A13%3A07.075Z</a>
[opencode session](https://opencode.ai/s/J0mMnXCj)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/33992326551)

PR #136, opened 2026-09-06
feat: complete connection setup assistant round 3

kaishi00:
## Summary

Round 3 replaces the Round 2 `detailsReady` placeholder with a complete, session-only connection setup flow.

- Adds route-specific LAN, Tailscale, and HTTPS reverse-proxy detail entry.
- Builds and validates dashboard URLs through the existing `ConnectionURLPolicy`.
- Collects dashboard credentials and shows a password-safe Review step.
- Hands a typed result back to `LoginView` without connecting automatically.
- Seeds troubleshooting entry from the current LoginView fields and preserves route drafts across Back.

## Validation

- Focused unit suites on the remote Mac: 37 passed.
- Earlier related unit suites: 106 passed.
- Existing setup UI paths passed in the remote run.
- `git diff --check` passed.

The broader UI run encountered simulator `signal kill` crashes, including the new LAN UI case and existing keyboard cases. These were infrastructure crashes rather than XCTest assertion failures; the focused setup cases that completed passed.

The repository's documented multi-model review workflow was not present in this checkout and no callable workflow tool was available. An independent read-only source review was completed; its credential-fidelity finding was fixed and covered by a passing unit test.

Live reachability, authentication probing, discovery, and automatic Connect remain deferred to later rounds.

## Summary by CodeRabbit

- **New Features**
  - Added a guided connection setup wizard with dedicated connection details, credentials, and review steps.
  - Supports LAN, Tailscale, reverse-proxy, and existing dashboard addresses.
  - Existing login details can be reused and edited during setup.
  - Added review and confirmation before applying settings.
  - Added accessible labels for connection and credential fields.

- **Bug Fixes**
  - Improved validation for hosts, ports, schemes, secure transport, and malformed addresses.
  - Sensitive credentials are concealed in diagnostic descriptions.
  - Invalid setup errors are shown during review with recovery options.
  - Cloudflare credentials are cleared when switching dashboard origins.

PR #137, opened 2026-09-06
Preserve session identity during recovery

### LINKED ISSUE #134: Sending a second message shortly after the first reply completes bounces to a new/empty conversation
## Bug Description

Sending a second message shortly after the first turn's reply has fully finished (visibly complete, no streaming/spinner) bounces the app away from the current conversation into what looks like a brand-new, empty conversation view. Waiting longer before sending the second message avoids it.

## Repro

1. Open an existing conversation (a canonical/pinned chat, e.g. a Bot Mode conversation bound to a messaging-platform home channel).
2. Send a message, wait for the reply to fully finish rendering (no spinner/streaming indicator left).
3. Send a second message fairly quickly after that (short gap).
4. The app navigates to what appears to be a new, empty conversation instead of continuing the current one.

## Expected

The second message is appended to the same conversation.

## Actual

App navigates to a different/new conversation view. Reproduces reliably with a short gap between the two sends; a longer pause between them avoids it.

## Workaround found

Reopening the same conversation from the sidebar right before sending the second message avoids the issue — the send succeeds immediately in the correct conversation afterward.

## Server-side verification

Checked the Hermes session store (`state.db`) directly during a repro: the underlying session row was never duplicated (still a single row, message count kept incrementing normally). So this looks like a client-side navigation/state issue, not the server actually creating a second session.

## Possible pointer (not verified as the actual cause)

Skimmed `AppState.swift`'s turn/session ownership handling (the ambiguity-recovery logic around `acceptedRunning` / `acceptedUnsettled` / `turnLifecycleEvidence.revision`, roughly the block with the comment "a newer busy edge keeps that ownership live; a newer settled edge ends it"). This is the area that reconciles submission acceptance against registry/lifecycle evidence, which seems like the right neighborhood for a race between "turn N considered settled" and "turn N+1 submitted", but I have not traced the actual faulty branch — flagging it only as a starting point.

## Environment

- Hermes Agent: v0.21.0 (upstream commit 9dd6634c), self-hosted, native dashboard (port 9119)
- Access: via a public HTTPS domain reverse-proxied to the dashboard (Caddy → Tailscale), so there is a few hundred ms of round-trip latency on top of LAN
- Conversation type: Bot Mode canonical chat bound to a messaging-platform home channel (Feishu), i.e. one of the always-present "default profile" conversations
- Single profile involved in this specific repro

kaishi00:
## Summary

This PR fixes issue #134 and establishes the session-identity architecture it exposed as missing: durable conversation identity, runtime routing identity, and positively-confirmed aliases are now separate concepts with one admission gate, instead of interchangeable session-ID strings.

### Delivered model

- **`ConversationIdentity`** (`Conduit/Services/ConversationIdentity.swift`): `profile`, `durableSessionID` (stable UI/persistence identity, present only when positively established), `runtimeSessionID` (Hermes routing identity; rotation is a transport rebind, never navigation), `acceptedSessionIDs` (positively confirmed aliases only, captured BEFORE recovery replaces the catalog).
- **Resume identity parsing**: `SessionResumeResult.storedSessionId` parses `stored_session_id` / `storedSessionId` / `session_key` separately from the runtime id (nil on legacy gateways). A returned runtime id alone is never durable-identity evidence.
- **Admission gate** (`ConversationIdentityGate.admit`): durable match → accept; confirmed alias → accept; runtime id positively owned by another catalog row → reject; durable key positively labeled as another row's stored id → reject; explicit durable contradiction → reject; legacy request-scoped runtime rebind → accept (historical behavior). A rejected claim is never adopted into conversation-owned state — no `activeSessionId`, selected identity, transcript, scroll canonical, conversation persistence key, composer ownership, presentation cache, or resume-store selected identity; ordering evidence captured at reconcile entry is restored. The refreshed session catalog itself is independent discovery state and is NOT rolled back. Rejection also skips automatic-return reconnect scheduling (deterministic contradiction, not a retryable failure).
- **Runtime-to-durable persistence migration**: when admission establishes a durable key for a runtime-only conversation, the canonical key can move (runtime-solo → stored-solo) before any catalog row carries it. `migrateChatResumePersistenceIfNeeded` now has an admission-evidence path (keyed off `resolvedDurableSessionId` + the reconciliation's accepted alias set, conversation- and profile-scoped) that migrates the viewport snapshot and resume-store entry through `chatResumeCoordinator.migrateSessionIdentity` even without a catalog row — the first durable-establishing resume no longer strands scroll restoration under the runtime key. Works through simultaneous rotation, too.
- **Runtime rebinding**: an admitted rotation records `resolvedDurableSessionId`, projected into the scroll-identity resolver and resume store, so `stored-a: runtime-old → runtime-new` rebinds routing while the selected conversation stays stored-a.
- **Composer ownership**: `ComposerSubmissionContext` carries a durable fence. Exact session-ID matches obey it strictly (equal durable ids, one catalog row positively containing both, or catalog silence — the expected state of a just-established row-less durable key); catalog re-attribution that keeps the runtime string while re-homing the conversation is rejected. Non-exact matches go through the alias path under the same profile/client/epoch/viewport/durable fences, so legitimate `runtime-old → runtime-new` rebinds keep ownership while A→B→A navigation handoffs still invalidate suspended work.
- **Review-finding fixes retained**: preserve-current with no current identity keeps the historical newest-chat / create-on-empty-catalog behavior; the accepted alias set is captured before catalog replacement, so stream events for a forgotten runtime alias buffer against the reconciliation and survive the transcript replacement.

Explicit navigation (sidebar tap, notification, profile switch, new/branch/archive/delete) remains authoritative; deleted conversations are not resurrected. Automatic return (Continue Where I Left Off / Jump to Latest Activity) is pinned unchanged. Design and implementation docs describe what shipped, including declined alternatives (self-referential scroll canonical as durable; labeled-only durable derivation) and known limits.

## Validation

- Full iOS unit suite on the final head (`58daa8b`): **1843 tests, 0 failures** (Mac, iPhone 17 Pro simulator).
- Identity/session suites on the final head: `ConversationIdentityTests`, `ChatResumePolicyTests`, `AppStateChatResumeTests`, `AppStateForegroundLifecycleTests`, `CompactResumeTranscriptTests`, `ChatResumeCoordinatorTests`, `ChatResumeStoreTests` — **311 tests, 0 failures**. New round-2 coverage: runtime→durable persistence migration (plain + rotating), exact-ID/durable-mismatch rejection with fresh-context control, row-less establishment composer survival, A→B→A stale rejection with fresh positive control.
- CI planner validation OK; script tests 74 OK (2 skipped); `git diff --check` clean.
- Three-model review + two delta re-reviews completed; every BLOCKER/WARNING resolved. Declined with documented reasoning: treating a self-referential scroll canonical as durable and dropping the legacy row-id durable derivation (both would resurface #134 or false-reject first labelings).
- **Hosted CI fully green on the final head** (`58daa8b`): CI Gate, Plan & validate, Build test products, unit-1..4, UI tests, Report. (An earlier run on the pre-merge base had hit the documented unit-4 settled-Markdown flake and a UI wedge; both cleared after merging main's CI destination fixes — PR #136 — and re-running.)

Not merged from this session; branch updated only.


kaishi00:
## Summary

This PR fixes issue #134 by capturing the selected conversation identity before a preserve-current recovery replaces the session catalog, then resuming that identity directly when the refreshed catalog omits its runtime alias. It also makes `ChatResumeSessionResolver` stop falling back to the newest chat on preserve-current misses, adds regression tests, and includes design/plan docs.

## Findings

### [severity: major] Preserve-current with no current identity now creates a new conversation instead of resuming the newest chat
**File:** Conduit/Services/ChatResumePolicy.swift:52 and Conduit/Services/AppState.swift:2778
**Issue:** The resolver's new `if purpose == .preserveCurrent { return nil }` early-return fires even when `currentSessionID` is nil (no established identity). Previously, preserve-current with a nil/unmatched `currentSessionID` fell through to `scoped.first(where: { $0.source == .chat })` — resuming the newest chat. Now `selectChatResumeTarget` returns nil and `performSyncSession` falls into the `else` branch (`createAndReconcileSession`), which issues `session.create`. This silently creates a brand-new conversation on states where a user had no current session but existing chats (replaced/deleted active session → `clearActiveSessionIfNeeded` sets `activeSessionId = nil`, a pristine new-chat canvas, first launch before a session is opened). That is a UX and side-effect regression: the app now creates a server session where it previously resumed the most recent one, and it contradicts the spec's own boundary ("create one when a current identity exists" — no fallback change was specified for the no-identity case). Confirmed by walking `performSyncSession`: with `activeSessionId == nil`, `preservedSessionID` is `nil`, so the new `else if let preservedSessionID` branch at AppState.swift:2766 cannot match.
**Suggestion:** Keep the newest-chat fallback when no current identity exists: in the resolver return `scoped.first(where: { $0.source == .chat })` only when `currentSessionID == nil`, and return `nil` (→ preserved-identity resume) only when a non-nil identity failed to match. Add a regression test asserting that preserve-current with `currentSessionID: nil` still selects the newest chat.

### [severity: minor] Accepted identity set is computed from the already-replaced catalog, dropping the runtime alias during the resume window
**File:** Conduit/Services/AppState.swift:2766-2771
**Issue:** The preserved branch passes `acceptedSessionIDs: knownSessionIDs(for: preservedSessionID)`, but `knownSessionIDs` (AppState.swift:6054) searches `sessions + cronSessions` *after* the replacement at line 2721 — the very catalog that omitted the runtime alias. So while `reconcile("stored-a")` awaits `openSession` (whose `resolvedSessionId` is not yet known), stream events addressed to the runtime id `runtime-a` fail `Reconciliation.accepts` in `bufferIfReconciling` (AppState.swift:10982) and are applied directly to the soon-to-be-replaced transcript instead of being buffered/replayed, i.e. they can be dropped. The buffered-delta dedup intersection at AppState.swift:3300-3313 is also intentionally emptied (runtime-a not in the boundary's accepted set), disabling dedup. In the previous matched-target path, `acceptedSessionIDs = Set([target.id] + target.alternateIds)` included `runtime-a` when the catalog had the alias. The window is narrow (a second surface continuing the same turn during recovery), but the regression is real.
**Suggestion:** Capture the accepted ID set (catalog row ids + alternates, or `activeChatScrollSessionIdentity.equivalentSessionIDs`) from the pre-replacement catalog at the same point as `preservedSessionID` (AppState.swift:2674), and pass that to the preserved reconcile.

### [severity: minor] Behavior change to the no-identity path is untested and undocumented
**File:** ConduitTests/ChatResumePolicyTests.swift:5
**Issue:** The new tests only cover "established stored identity, fresh catalog omits alias" and "identity present, catalog empty." There is no test for the newly altered no-identity case (nil `currentSessionID` with a populated catalog), where existing behavior silently changed from "resume newest chat" to "create a session." Test 1 (`testPreserveCurrentUsesEstablishedStoredIdentityWhenCatalogLosesRuntimeAlias`) relies on `activeChatScrollSessionIdentity` having been pre-established by a prior `sessions`/`activeSessionId` assignment, so it does not exercise the plain cold-start/empty-identity path that the resolver edit also affects.
**Suggestion:** Add a test asserting the intended no-identity behavior (whichever is decided — resume-newest or create-new) so the fallback change is deliberate and pinned, not incidental to the `#134` fix.

### [severity: nit] Plan/spec drift: documented broader scope not delivered here
**File:** docs/superpowers/plans/2026-09-06-session-identity.md
**Issue:** The spec promises resume-side identity parsing of `stored_session_id`/`session_key`, identity admission/rebind validation, and composer rebinding under ownership fences — none of which is implemented in this PR (unchecked `- [ ]` items remain in the plan). The delivered change is only the capture + resolver portion. This is acceptable if intentionally staged, but the unchecked "Verify their expected assertion failures against unchanged production sources" item and the untracked spec scope make it look unfinished.
**Suggestion:** Update the plan to mark delivered items and explicitly list the composer-rebind/identity-admission work as a follow-up task, so the tracked scope matches the implementation.

Tests, docs, and the core fix approach (capture identity before catalog replacement, resume it directly, no unrelated-chat fallback) are otherwise sound. No SwiftUI, accessibility, retain-cycle, force-unwrap, or concurrency issues were introduced; all new code is MainActor-isolated with no new capture-strong-self hazards.

## Verdict

COMMENT — The fix is directionally correct and well-tested for the reported race, but the resolver's blanket preserve-current `return nil` silently changes the no-identity cold-start/new-chat behavior to create a fresh conversation, which goes beyond the bug's scope and the spec, and should be gated on identity presence with a test.

<a href="https://opencode.ai/s/BsBy5m8n">New%20session%20-%202026-09-06T12%3A15%3A04.832Z</a>
[opencode session](https://opencode.ai/s/BsBy5m8n)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34032461274)

kaishi00:
## Summary
This PR introduces a conversation-identity model (`ConversationIdentity` + `ConversationIdentityGate`), separates durable vs. runtime session IDs in resume parsing, preserves the selected conversation's identity before catalog replacement, and guards composer submissions with a durable-identity fence so runtime rotation is treated as a rebind rather than navigation.

## Findings

### [severity: minor] Unrecognizable current identity still silently falls to `session.create`
**File:** Conduit/Services/AppState.swift:2819 and Conduit/Services/ChatResumePolicy.swift:59
**Issue:** The resolver now returns `nil` whenever `currentSessionID` is non-nil and unmatched (ChatResumePolicy.swift:59), and the preserved branch (AppState.swift:2819) requires `preservedIdentity` from `captureConversationIdentity`. When `activeSessionId` is non-nil but appears in *neither* the pre-replacement catalog *nor* the settled scroll identity (cold launch where `restoreActiveSessionState` restores `lastSessionID` at AppState.swift:1336 with an empty in-memory catalog and a scroll identity still empty), `preservedIdentity` is nil, so the `else` branch runs `createAndReconcileSession` instead of the pre-PR newest-chat fallback. The no-identity tests only cover the `nil`/empty current-session case, not this non-nil-but-unrecognizable one.
**Suggestion:** Gate the create-vs-newest decision on whether any identity evidence existed, and add a regression test asserting that a non-nil but unrecognizable `currentSessionID` (cold-start covered) resumes the newest chat rather than creating a session.

### [severity: minor] Exact-match composer fast path bypasses the new durable fence
**File:** Conduit/Services/AppState.swift:6741-6743
**Issue:** `isCurrentComposerSubmission` returns `true` immediately when `capturedSessionID == activeSessionId`, skipping `composerDurableIdentityMatches` — the fence added for the alias path. A captured id string that has since been re-labeled in the catalog to a *different* durable conversation passes the fast path even though the durable check would reject it, so the defense-in-depth is only partially applied.
**Suggestion:** Also run `composerDurableIdentityMatches(context)` on the exact-match fast path (it is cheap), or document explicitly why a stable id string can never re-map to another durable conversation.

### [severity: minor] Rejection-path invariants are narrower than documented
**File:** Conduit/Services/AppState.swift:3221-3224
**Issue:** The rejection branch restores `durablePersistedRowIDs` and claims "nothing user-visible may change on rejection," but the catalog was already replaced in `performSyncSession` (line 2744) *before* `reconcile` ran and is not rolled back. On a rejected claim the sidebar shows the refreshed catalog (which omits the selected conversation) while the active transcript and error banner reference the rejected selection — the "mutates nothing" invariant doesn't cover the catalog/sidebar.
**Suggestion:** Either accept and document that the foreground catalog refresh is committed independently of claim admission, or ensure the rejected flow re-resolves/re-selects after settling.

### [severity: nit] `.fault` logging for an expected, user-visible condition
**File:** Conduit/Services/AppState.swift:3218
**Issue:** `sessionCatalogLog.fault` is used for a rejection that is an expected runtime condition (mixed-generation catalogs, foreign runtime ids). `.fault` is for unrecoverable/abnormal conditions and will pollute error-rate dashboards; the user-facing message is also hard-coded inline.
**Suggestion:** Use `.error` (or `.notice`) and centralize the user-facing string so test/recovery paths share one definition.

### [severity: nit] Pre-merge CI flag: docstring coverage 0%
**File:** Conduit/Services/ConversationIdentity.swift (and touched functions in AppState.swift/ChatScrollState.swift)
**Issue:** The CodeRabbit pre-merge check reports 0.00% docstring coverage on the 7 functions touched by the diff; reconciliation of the identity model's semantics is largely buried in comments/commit messages rather than type-level docs.
**Suggestion:** Add brief docstrings to the newly introduced gate/parse/resolver entry points so the durable-vs-runtime rule is self-describing and the check passes.

No retain cycles (no new `self`-capturing closures), force unwraps, SwiftUI/accessibility regressions, or strict-concurrency violations were introduced; all new state is MainActor-isolated and the new model types are value types.

## Verdict
APPROVE — The prior review's major/minor findings are substantively addressed (no-identity fallback restored, accepted identity captured before catalog replacement, mid-resume event survival). Remaining items are edge-case behavior interpretations and documentation/CI nits, none blocking.

<a href="https://opencode.ai/s/ZCS27oCt">New%20session%20-%202026-09-06T16%3A05%3A19.207Z</a>
[opencode session](https://opencode.ai/s/ZCS27oCt)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34044265043)

kaishi00:
## Summary
This PR establishes a durable-vs-runtime conversation-identity model (`ConversationIdentity` + `ConversationIdentityGate`), captures the selected conversation's identity before preserve-current recovery replaces the catalog, resumes it directly when the refreshed catalog omits its alias, adds a composer durable-identity fence so runtime rotation is a rebind rather than navigation, and ships substantial regression coverage (12 AppState lifecycle/resume tests, 8 gate/parse-resolver tests) plus updated docs.

## Findings

### [severity: minor] Persisted scroll state is not migrated when the durable canonical has no catalog row
**File:** Conduit/Services/ChatScrollState.swift:579-584 and Conduit/Services/AppState.swift:2054-2063
**Issue:** A runtime-only conversation (canonical = `runtime-solo`) that acquires its durable key on resume gets its canonical changed to `stored-solo`. `migrateChatResumePersistenceIfNeeded` (AppState.swift:2054) `return`s early when no catalog row matches the new canonical key (`guard ... else { return }`), and `ChatResumeStore.snapshot(for:)` (ChatResumeStore.swift:88-91) matches exact `ChatScrollSessionKey`s without consulting equivalent IDs. So the snapshot saved under the runtime key is never moved to the durable key, and `reconciliationSettled` (ChatResumeCoordinator.swift:187) reads `nil` for the new key and restores `.latest` instead of the saved position. This is the bounded regression CodeRabbit flagged as merge risk: a resumed conversation can lose its restored scroll position exactly on the first labeled (durable-establishing) resume, in the window before the catalog gains a row for the durable id.
**Suggestion:** In `migrateChatResumePersistenceIfNeeded`, when no catalog row matches the new canonical key, fall back to migrating from `current.canonicalSessionKey`/active/persisted keys when the new canonical is an established `resolvedDurableSessionId` (which is itself positive identity evidence), e.g. migrate `runtimeKey` → `canonicalKey` based on the scroll identity's equivalent set rather than requiring a catalog row.

### [severity: minor] Exact-match composer fast path bypasses the durable fence
**File:** Conduit/Services/AppState.swift:6741-6743
**Issue:** `isCurrentComposerSubmission` returns `true` immediately when `capturedSessionID == activeSessionId`, skipping `composerDurableIdentityMatches`. The defense-in-depth durable check is only applied on the alias path. If the catalog re-labels the single stable id string to a different durable conversation (the exact scenario the fence was added for), a stale capture passes on the string match alone.
**Suggestion:** Run `composerDurableIdentityMatches(context)` on the exact-match path as well (it's cheap and `canonicalSessionID(for:)` is already consulted at capture time), or explicitly document why a stable id can never re-map to a different durable conversation.

### [severity: minor] Rejection-path comment overstates the "nothing mutates" invariant
**File:** Conduit/Services/AppState.swift:3221-3231
**Issue:** On gate rejection the code restores `durablePersistedRowIDs` and comments "nothing user-visible may change on rejection," but the catalog was already replaced in `performSyncSession` (AppState.swift:2744) before `reconcile` ran and is not rolled back; the rejection also sets `errorMessage`. A rejected claim therefore commits the refreshed sidebar catalog (possibly omitting the still-selected conversation) while the transcript/active id keep the old selection — plus an error banner. The claim in `ConversationIdentityGate`'s doc and the body text ("no active id, transcript, scroll identity, presentation cache, resume store, or composer ownership") is accurate, but the stronger comment at 3221-3222 is misleading.
**Suggestion:** Either scope the comment to the state actually preserved (selection/transcript/persistence) and document that the foreground catalog refresh is committed independently of admission, or re-resolve/re-select after a rejection so the sidebar stays consistent with the rejected selection.

### [severity: nit] `.fault` severity for an expected, user-visible rejection
**File:** Conduit/Services/AppState.swift:3218
**Issue:** `sessionCatalogLog.fault` is used for a contradiction/foreign-ownership rejection, which is an expected runtime condition (mixed-generation catalogs, colliding runtime ids), and `.fault` pollutes error-rate dashboards. The user-facing string is also hard-coded inline while tests re-derive it.
**Suggestion:** Use `.error`/`.notice`, and centralize the user-facing message so the test/recovery paths share one definition.

### [severity: nit] Composer handoff test lacks a positive control
**File:** ConduitTests/AppStateForegroundLifecycleTests.swift:142-145
**Issue:** `submitComposer` with the stale `handoffContext` can return `false` for reasons other than the viewport-generation fence, so the single negative assertion doesn't prove the fence is what rejected the submission.
**Suggestion:** After reopening A, capture a fresh `composerSubmissionContext()`, submit, and assert it succeeds and routes to `runtime-stored-a`, keeping the stale-context assertions unchanged.

### [severity: nit] CI docstring-coverage gate
**File:** Conduit/Services/ConversationIdentity.swift (and touched functions in AppState.swift/ChatScrollState.swift)
**Issue:** The pre-merge check reports docstring coverage below threshold on the newly introduced entry points; the durable-vs-runtime semantics are currently documented via prose/comments rather than type-level docstrings.
**Suggestion:** Add short docstrings to `ConversationIdentityGate.admit`, the resume-parsing sites, and the resolver entry points.

No retain cycles, force unwraps, SwiftUI/accessibility regressions, or strict-concurrency violations were introduced: the new model types are Sendable value types, the gate is a nonisolated pure function, and all new AppState state is MainActor-isolated.

## Verdict
COMMENT — The core fix is correct and well-tested: the prior major/minor findings (no-identity fallback, pre-capture accepted aliases, mid-resume event survival) are resolved in the current head, and the admission gate's ordering is sound. The remaining items are bounded edge cases and nits; the scroll-persistence migration gap is the most actionable and is worth addressing before merge since it directly affects the new durable-identity path.

<a href="https://opencode.ai/s/LDgWW85d">New%20session%20-%202026-09-06T16%3A33%3A06.444Z</a>
[opencode session](https://opencode.ai/s/LDgWW85d)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34045744861)

kaishi00:
## Summary
This PR introduces a durable-vs-runtime conversation identity model (`ConversationIdentity`, `ConversationIdentityGate.admit`), separates `stored_session_id` from the runtime id in resume parsing, captures the selected conversation's identity *before* recovery replaces the catalog, resumes omitted/rebound conversations directly, adds a composer durable-identity fence, and ships extensive regression coverage plus docs.

## Findings

### [severity: minor] Row-backed runtime-only conversations still construct a self-referential durable id, producing the exact false contradiction the design declined
**File:** Conduit/Services/AppState.swift:6227 and Conduit/Services/AppState.swift:2814
**Issue:** `captureConversationIdentity` sets `durableSessionID = row.storedSessionId ?? row.id` when the selected id matches a catalog row, and the target path builds `targetIdentity` with `durableSessionID: target.storedSessionId ?? target.id`. When a conversation has a catalog row whose id is its runtime id and no stored label (a runtime-only conversation that the catalog already lists), the runtime id becomes the "durable" identity. On the first labeled resume the gateway returns a genuinely distinct stored key, and `ConversationIdentityGate.admit` reaches `.failure(.durableContradiction(selected: runtime-id, returned: stored-key))` (ConversationIdentity.swift:101). This is exactly the "self-referential canonical treated as durable → false contradiction on first labeled resume" case the PR body says was deliberated and declined for the scroll-identity path (AppState.swift:6232-6241 leaves durable `nil` for that path) — but the row path reintroduces the same trap unguarded. All establishment tests (`testRuntimeOnlyConversationEstablishesDurableKeyFromResume`, the two migration tests, AppStateChatResumeTests.swift:262/397/434) use catalogs that do *not* contain the selected row, so this path is untested. A rejected claim here shows the error banner and blocks durable adoption for the same conversation that a row-less capture would have established.
**Suggestion:** Mirror the scroll-identity reasoning in the row path: only set a row's durable id when `storedSessionId != nil`, leaving it `nil` for unlabeled rows so the response can positively establish it. Add a test for "catalog row present, no stored label, resume reveals a stored key" asserting admission and durable establishment.

### [severity: minor] Rejection-path comment overstates "nothing user-visible may change" while the sidebar catalog and error banner do change
**File:** Conduit/Services/AppState.swift:3250-3266
**Issue:** On rejection the code restores `durablePersistedRowIDs` and claims ordering evidence is conversation-owned state that "may not" be consumed, but `sessions`/`cronSessions` were already replaced at AppState.swift:2768-2769 before `reconcile` ran and are not rolled back. The refined catalog (which may omit the still-selected conversation) stays committed in the sidebar while the active transcript, `activeSessionId`, and scroll identity remain pinned to the pre-replacement selection, plus a hard-coded `errorMessage` (AppState.swift:3263). The gate/reconcile doc comments accurately scope this ("independent discovery state ... NOT rolled back"); the inline "Rejection must not consume it" ordering-evidence wording is defensible, but the user-visible contract is split across proccedural states.
**Suggestion:** Either document explicitly that the foreground catalog refresh is committed independently of claim admission (sidebar may show a catalog that omits the rejected selection), or after a rejection re-resolve/re-select from the settled catalog so sidebar and transcript converge.

### [severity: nit] `.fault` log severity for an expected, user-visible rejection
**File:** Conduit/Services/AppState.swift:3247
**Issue:** `sessionCatalogLog.fault` is used for contradiction / foreign-ownership rejections, which are expected runtime conditions (mixed-generation catalogs, colliding runtime ids), and the user-facing string at AppState.swift:3263 is hard-coded inline while tests re-derive it. `.fault` inflates error-rate dashboards for a non-crash condition.
**Suggestion:** Use `.error` (or `.notice`), and centralize the user-facing message so production and test recovery paths share one definition.

### [severity: nit] New identity entry points remain at low docstring coverage
**File:** Conduit/Services/ConversationIdentity.swift:83 (and the touched resolver/capture entry points in AppState.swift/ChatScrollState.swift)
**Issue:** The pre-merge check still reports docstring coverage far below threshold (15.56%) on the 45 functions touched. The durable-vs-runtime rule and admission ordering are largely encoded in comments and prose rather than type-level docs.
**Suggestion:** Add short docstrings to `ConversationIdentityGate.admit`, `captureConversationIdentity`, the resume-parse site in HermesClient, and the resolver entry points.

No retain cycles (no new self-capturing closures), force unwraps, SwiftUI/accessibility regressions, or strict-concurrency violations were introduced: `ConversationIdentity`/`ResumeIdentityClaim` are Equatable value types, `ConversationIdentityGate.admit` is a nonisolated pure function, and all new AppState state is MainActor-isolated. The prior major/minor findings (no-identity newest-chat fallback, pre-capture accepted aliases, mid-resume event survival, exact-match durable fence, scroll-persistence migration, composer handoff positive control) are all addressed in the current head and covered by tests.

## Verdict
COMMENT — The core #134 fix is correct, well-reasoned, and well-tested. Remaining items are an untested edge-path false contradiction in the row-backed durable fallback (worth addressing since it contradicts the design's own declined-alternative rationale) plus documentation/logging nits; none block the primary recovery path.

<a href="https://opencode.ai/s/nSseqFDz">New%20session%20-%202026-09-06T17%3A54%3A58.290Z</a>
[opencode session](https://opencode.ai/s/nSseqFDz)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34050013254)

PR #17, opened 2026-08-09
Support partial chat selection and restore scroll position

kaishi00:
## What changed

- Enable native partial text selection across textual chat content, including streaming responses, while preserving existing copy controls and interactive content.
- Restore the user's chat position when returning to the app or switching sessions.
- Follow new messages only when the user was already near the latest message.
- Use semantic, profile-qualified scroll anchors that remain stable across reconciliation, runtime session ID changes, duplicate messages, and streaming updates.
- Add focused regression coverage for selection and scroll restoration edge cases.

## Validation

- `xcodebuild test`: 243/243 tests passed.
- Simulator install and launch succeeded.
- `git diff --check` passed.
- Build metadata remains 0.1.2 (120).

Manual authenticated Hermes UI acceptance remains to be performed.

Closes #2

PR #139, opened 2026-09-06
Harden conversation identity across session boundaries

kaishi00:
# Harden conversation identity across session boundaries

Follow-up to #137 (merged as b3fa7c3). #137 established the conversation identity model — durable identity, runtime identity, confirmed aliases, profile — and the resume admission gate. This pass closes the remaining places that interpreted raw session-ID strings independently, so there is **one consistent definition of conversation identity throughout Conduit**: future features should not need to know whether a string is a stored ID, runtime ID, alias, or current routing ID.

This is **not** a redesign of #137; `ConversationIdentity`, the admission gate, and its tests are preserved intact.

## 0. Round 2: split-brain, delete-of-pending-restoration, durable-owned presentation

Round 2 closes the correctness gaps found in review, without touching the architecture:

- **No split-brain.** `recordAuthoritative(...)` is the explicit fresh-evidence API: an admitted resume, an `active_list` row, or a verified create/branch REBINDS a conflicting historical mapping, because the app has already accepted that routing into conversation-owned state. When nothing else established the durable, the resume TARGET itself is the durable candidate (a stored-addressed resume positively names it; runtime-addressed resumes degenerate to self-mappings the index skips). After every admitted resume: index, selected conversation, and reconciliation agree. `record(...)` remains the low-authority path (historical aliases, notification payloads) and never overwrites.
- **Deletion invalidates pending in-memory restoration.** `ChatResumeCoordinator.removeSessions` now clears `pendingSessionKey` (+ fallback-selection state) and published `pendingRestoration` belonging to the deleted identities — the stale generation stops being current, so completion/abandon cannot resurrect it — unfreezes the viewport, and bumps the automatic-work epoch. Strictly scoped: pending work for other conversations is untouched.
- **Durable-owned presentation persistence.** Once the durable id is established, the durable key is the ONLY persistent presentation key: `consolidateUnderDurableKey` migrates runtime-keyed records into it at admission (freshest wins; nothing lost) and retires the alias keys, while `durableOwnedPresentationIDs` collapses all write id-lists to the durable key so flushes can never recreate alias copies. A runtime id the gateway later re-attributes elsewhere therefore carries no trace of the previous conversation — timestamps, attachments, tool metadata, or pending decision cards. Migrated approval cards have their embedded session id rewritten to the durable id, so an approve tap dispatches to the durable session even after later rotations. Merge lookups remain tolerant supersets.
- **`session.active_list` feeds the index.** All three registry probe sites record rows (runtime → stored, authoritative) with the profile captured before the await and a currency fence (profile + live client identity) so superseded connections cannot write. Strictly observational: rows never mutate the selected conversation, and healthy-foreground observe-don't-resume is untouched.
- **Notification conflict semantics pinned.** A stale dual-ID push against fresher live truth is rejected by admission and poisons nothing; decision cards persist only under the push-named id until the open succeeds.

## 1. Shared confirmed identity index (`ConversationIdentityIndex`)

New profile-scoped component answering: *what durable conversation is runtime-a positively known to belong to?* It is not a second session state machine — `ConversationIdentity` remains the ownership value.

**Accepted evidence only** (each tagged with a source for diagnostics: `catalog`, `resume`, `active_list`-class registries, `create`, `branch`, `notification`):
- admitted resume results (the rebind/establishment point in the reconcile);
- explicitly labeled catalog rows (`storedSessionId` + id set), committed at both catalog publish points;
- verified create/branch responses (the summary built from the same response);
- notification payloads that explicitly carry both identities.

**Never evidence:** catalog ordering, newest activity, timestamps, title equality, string resemblance, profile co-membership.

**Conflict behavior — explicit authority table.** Callers pick the API by authority, never by convenience:

| Evidence | API | May establish? | May rebind? |
|---|---|---|---|
| historical confirmed alias | `record` | yes | no |
| dual-ID notification payload | `record` | yes | no |
| admitted `session.resume` | `recordAuthoritative` | yes | **yes** |
| authoritative catalog snapshot | `recordCatalogIdentity` | yes | yes (first claim per snapshot) |
| `session.active_list` row | `recordAuthoritative` | yes | **yes** |
| verified create result | `recordAuthoritative` | yes | yes (created conversation) |
| verified branch result | `recordAuthoritative` | yes | yes (branched conversation) |

The rule behind the table: when the app has already ACCEPTED fresh routing evidence into conversation-owned state, the index must not disagree — a rebind wins and the displaced mapping is logged; historical observations and unverified payload claims never overwrite a fresher confirmed mapping, they surface a conflict. Suspended operations are unaffected either way because they hold capture-time `acceptedSessionIDs` and can never gain ownership through the index.

**Scope & lifecycle:** mappings are keyed by normalized profile — one profile's runtime IDs can never establish ownership in another. `removeAll()` at the `prepareChatResumeForConnection(to:)` server-change boundary; `removeSessionIDs` on explicit delete. Catalog omission never clears mappings. Logout/re-login on the same server keeps mappings (same server identity).

## 2. Notification routing

`ConduitNotificationTarget` now separates `sessionId` (runtime) from `durableSessionID` (parsed from a new `stored_session_id` / `session_key` **top-level routing field** in the conduit payload; **backward compatible** — older payloads parse exactly as before). `NotificationSessionResolver.resumableSessionID` is replaced by `route(target:catalog:identityIndex:profile:)` with an explicit evidence hierarchy:

1. **explicitDurable** — payload carries the durable id → route there, no catalog needed;
2. **catalogAlias** — live catalog row positively contains the runtime id;
3. **confirmedAlias** — the shared index fills a stale/omitted catalog;
4. **legacyRuntime** — no positive durable evidence: the notification's own runtime id is resumed directly (historical behavior), and nothing else. **An unknown runtime ID is never reinterpreted as a durable conversation — there is no newest-chat/ordering/similarity fallback.**

Decision cards remain functional and are recorded under all routed identities (durable + runtime). `decision.session_key` stays the `approval.respond` answer key and is deliberately **not** promoted to routing identity (a `stored_session_id`/`session_key` precedence test pins the top-level parsing). The payload's dual identity is committed to the index **only after the open succeeds** — a failed resume (e.g. the durable conversation was deleted server-side) leaves no dead mapping — and a disagreement with the confirmed mapping is logged by id at the call site.

*Notifier note:* the relay/gateway today sends only `session_id`; the iOS side is forward-compatible. When the notifier adds `stored_session_id` to the routing payload, routing upgrades automatically without an app release on the parser side (both spellings accepted).

## 3. Server-change contract (cross-server leak fix)

`prepareChatResumeForConnection(to:)` now also clears `SessionYoloStore` overrides and the identity index — previously YOLO overrides were keyed only by (profile, session id), so Server A `default`/`abc` could leak its override into Server B `default`/`abc`. Resume store, presentation cache, titles, pins, and review cache were already cleared there (verified; unchanged).

## 4. Delete revocation

Explicit deletion now revokes the conversation's identity: index mappings, resume/scroll snapshots (plus the profile's last-selected pointer when it names the deleted conversation), and cached presentation — including pending decision cards — under **every** alias. Archive keeps identity (the conversation still exists; product semantics unchanged). Branch responses index the branch's own ids only; source and branch never alias.

## 5. Voice event identity

Voice assistant events carry runtime routing ids. Matching now uses a confirmed alias set (`beginVoiceTurn` capture + `extendAssistantSessionIDs(_:ofConversationContaining:)` on admitted rebinds) instead of raw equality with the turn's captured id, so a mid-voice-turn runtime rotation no longer silently drops the assistant's speech. Two guards keep ownership tight: the extension applies only when the reconciled conversation's accepted ids POSITIVELY overlap the turn's captured ids (a reconcile belonging to a different conversation can never inject its runtime — a vector all three review models flagged), and extensions are ignored when no turn is active, so nothing leaks into the next turn.

## 6. Audits that found no defect (pinned instead)

- **Rename** (`SessionRenameOperation`): existing guards are correct under the durable/runtime model — active-runtime RPC only when `activeSessionID` positively matches the row, durable PATCH fallback otherwise, `ContextChanged` on client swap, profile fences on application. Added a rotation pin: a confirmed new runtime that the stale row doesn't list must rename via the durable id, never the unverified runtime string.
- **Stream-event acceptance**: `reconciliation.accepts` + `eventBelongsToActiveSession` already implement the alias contract (#137); untouched, covered.
- **Composer drafts**: in-memory, equivalence-driven migration verified via new contract tests — a draft follows `runtime-old → runtime-new` rotation, does **not** follow a re-attributed runtime string into another conversation, and is profile-scoped.
- **Compact resume (#106)** and **observational foreground (#17)**: untouched; existing suites (`CompactResumeTranscriptTests`, `AppStateForegroundLifecycleTests`) cover them.

## Test plan

- New `SessionIdentityContractTests`: cross-component identity matrix (routing hierarchy, profile scoping, rotation ownership retention, re-attribution rejection, catalog omission, payload parsing, draft equivalence).
- New `ConversationIdentityIndexTests`: evidence, conflict, authoritative catalog overwrite, scoping, revocation.
- Extended: `AppStateChatResumeTests` (dual-identity notification opens, runtime-only + confirmed alias, unknown-runtime safety pin, index commit through the notification path, server-change isolation, delete revocation), `MessageNormalizerTests` (resolver route API), `VoiceConversationControllerTests` (rebind alias, no-leak), `SessionYoloStoreTests`, `ChatResumeStoreTests`, `SessionPresentationCacheTests`, `SessionRenameTests`.
- Focused suites green locally on the Mac build box; full unit suite + planner validation (`plan-tests.py validate`) + script tests green. Hosted CI green on this branch.

## Review gate

Round 1 — three independent reviewer models (2× APPROVE-WITH-NITS, 1× REQUEST-CHANGES). All flagged defects fixed before push:

- **Voice extension conversation-blindness** (3/3 flagged): adopted the positive-overlap guard + a dedicated cross-conversation non-leak test.
- **Catalog commit ordering** (1/3): `recordCatalogIdentity` is now first-claim-wins within one snapshot (the committed catalog can contain stale cached rows merged after fresh ones, and routing reads that same order), so the index can never pin a re-attribution the resolver would not route to.
- **Notification evidence timing** (2/3): explicit-durable payload evidence is committed only after the open succeeds, with conflict logging at the call site.
- **Resolver hardening** (1/3): catalog alias lookup skips foreign-profile rows and returns trimmed ids.
- **Rejected finding** (documented): one reviewer proposed that `legacyRuntimeRebind` admissions must not record index evidence. Rejected with evidence — the admission gate's `.success` IS the positive adoption (`context?.acceptedSessionIDs.insert` runs for the same result), and the legacy rebind path is precisely how runtime rotation manifests on gateways that do not echo stored ids; excluding it would break rotation routing. The third reviewer independently verified this hook cannot record a contradicted mapping (all rejections return early).

Round 2 (the split-brain / delete / durable-owned-presentation commit) — two focused reviewers, both REQUEST-CHANGES; every Major fixed before push:

- **Write collapse was not durable-only**: `durableOwnedPresentationIDs` returned `aliases + [durable]` when the durable id wasn't already in the list, letting flushes recreate retired runtime keys. A write now lands on the durable key alone once established (the contract test pins the stricter rule).
- **Stale push decision bled into the claimed durable**: decision cards were persisted before the open proved the route. Cards now persist only under the push-named identity before the open; a successful open's consolidation migrates them to the durable key.
- **`active_list` had no currency fence**: in-flight rows from a superseded connection could repopulate the index scope a server change just cleared. All three probe sites fence on the committed server identity (the exact `prepareChatResumeForConnection` boundary) plus profile. A client-identity fence was deliberately rejected — the foreground suite pins that a client replacement (even re-addressed) must still be accepted.
- **Migrated approval cards kept a stale routing id**: consolidation rewrites an approval's embedded session id from a retired runtime alias to the durable id, so an approve tap dispatches to the durable session even after later rotations (the first implementation mutated a local copy without writing it back — the new regression test caught it).

## Not done deliberately

- No persistence schema for the index (in-memory by design; the notification path refreshes the catalog first, and the index exists to bridge within-run catalog churn — persisted mappings would need cross-launch invalidation semantics this pass deliberately avoids).
- Notifier-side payload emission of `stored_session_id` lives in the notifier repo; this PR ships the forward-compatible contract.

## Summary by CodeRabbit

- **New Features**
  - Improved notification routing to open the correct conversation using durable or confirmed session identities.
  - Preserved pending approvals and conversation presentation when session identifiers change.
  - Voice conversations now continue safely across confirmed session-identifier rotations.

- **Bug Fixes**
  - Prevented stale or rejected notifications from appearing in the wrong conversation.
  - Deleting a conversation now clears its saved state and cancels pending restoration.
  - Server changes now clear outdated session overrides and identity mappings.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -3124,7 +3124,8 @@ final class AppState: ObservableObject {
         automaticWorkToken: ChatResumeAutomaticWorkToken? = nil,
         automaticSyncOperationID: UUID? = nil,
         requiredViewportTransitionGeneration: UInt64? = nil,
-        historySourceUnavailable: Bool = false
+        historySourceUnavailable: Bool = false,
+        presentationMigrationSessionIDs: Set<String> = []
     ) async -> Bool {
         guard automaticChatResumeWorkIsCurrent(
             automaticWorkToken,
@@ -3345,8 +3346,14 @@ final class AppState: ObservableObject {
                     // now, and the alias keys are retired so a later
                     // re-attribution of a runtime id to a different
                     // conversation can never inherit this presentation.
+                    // `presentationMigrationSessionIDs` are notification
+                    // presentation sources — untrusted as identity, so they
+                    // were never added to the accepted set; now that
+                    // admission HAS succeeded they are legitimate migration
+                    // sources for the pending decision cards they carried.
                     let runtimeAliases = acceptedSessionIDs
                         .union([result.sessionId, sessionId])
+                        .union(presentationMigrationSessionIDs)
                         .subtracting([admittedDurable])
                     sessionPresentationCache.consolidateUnderDurableKey(
                         profile: profile,
@@ -3400,10 +3407,18 @@ final class AppState: ObservableObject {
                 // Desktop keeps its live projection during an active turn. Seed
                 // the same durable presentation details first so the completed
                 // portion of a backgrounded turn does not lose its timestamps.
+                // Durable-owned: writes land on the durable key only, so the
+                // alias keys consolidation just retired stay retired.
+                let presentationDurableID = context?.resolvedDurableSessionId
+                    ?? referenceIdentity.durableSessionID
+                    ?? sessionId
                 sessionPresentationCache.save(
                     transcript.messages,
                     profile: profile,
-                    sessionIDs: [sessionId, result.sessionId, transcript.resolvedSessionId].compactMap { $0 }
+                    sessionIDs: Self.durableOwnedPresentationIDs(
+                        [sessionId, result.sessionId, transcript.resolvedSessionId].compactMap { $0 },
+                        durableSessionID: presentationDurableID
+                    )
                 )
             }
             // A compact resume ships no persisted transcript, so the REST rows
@@ -6485,7 +6500,8 @@ final class AppState: ObservableObject {
 
     private func openSession(
         _ sessionId: String,
-        reusing viewportTransitionGeneration: UInt64?
+        reusing viewportTransitionGeneration: UInt64?,
+        presentationMigrationSessionIDs: Set<String> = []
     ) async -> Bool {
         guard let client else { return false }
         let previousTurnState = turnState
@@ -6532,7 +6548,8 @@ final class AppState: ObservableObject {
             token: token,
             acceptedSessionIDs: acceptedSessionIDs,
             conversationIdentity: openedIdentity,
-            requiredViewportTransitionGeneration: transitionGeneration
+            requiredViewportTransitionGeneration: transitionGeneration,
+            presentationMigrationSessionIDs: presentationMigrationSessionIDs
         )
         guard chatViewportTransitionIsCurrent(generation: transitionGeneration) else {
             return false
@@ -6633,7 +6650,14 @@ final class AppState: ObservableObject {
         }
         let opened = await openSession(
             route.resumeTargetID,
-            reusing: transitionGeneration
+            reusing: transitionGeneration,
+            // The push-named runtime id is a PRESENTATION MIGRATION SOURCE,
+            // never an identity alias: a pending decision card recorded
+            // under it before the open must be promoted into the admitted
+            // durable conversation now that admission succeeded. On a
+            // rejected open nothing migrates (the hook only runs on
+            // admission success).
+            presentationMigrationSessionIDs: [requestedID]
         )
         // Commit the payload's dual identity as positive evidence only once
         // the open actually succeeded — a failed resume (e.g. the durable
@@ -6651,11 +6675,41 @@ final class AppState: ObservableObject {
                 )
             }
         }
+        if !opened, let decision = target.decision,
+           let evictionKey = Self.pendingDecisionEvictionKey(for: decision) {
+            // The claim was rejected: evict the pre-open card from the
+            // push-named runtime key. Without this, the stale card would
+            // still sit under that runtime id and a LATER legitimate open
+            // of the true owner would promote it into the wrong durable
+            // conversation.
+            sessionPresentationCache.removePendingDecision(
+                key: evictionKey,
+                profile: activeProfile,
+                sessionIDs: [requestedID]
+            )
+        }
         guard notificationOpenAttemptIsCurrent(
             id: notificationAttemptID,
             transitionGeneration: transitionGeneration
         ) else { return false }
         return opened
+    }
+
+    /// The stable decision key a push-delivered card was recorded under
+    /// (`approval:<sessionKey>` / `clarify:<requestId>`) — used to evict a
+    /// pre-open card when its open is rejected. Batch clarifies share the
+    /// scalar key shape via their relay request id.
+    private static func pendingDecisionEvictionKey(
+        for decision: PendingDecisionPayload
+    ) -> String? {
+        switch decision {
+        case .approval(let sessionKey, _, _):
+            return "approval:\(sessionKey)"
+        case .clarify(let requestId, _, _):
+            return "clarify:\(requestId)"
+        case .clarifyBatch(let requestId, _):
+            return "clarify:\(requestId)"
+        }
     }
 
     /// Caches a push-delivered decision card so the resume merge can restore
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -1025,6 +1025,7 @@ final class AppState: ObservableObject {
     private var projectsRequestGeneration = 0
     private let sessionPresentationCache: SessionPresentationCache
     private let sessionYoloStore: SessionYoloStore
+    private let conversationIdentityIndex: ConversationIdentityIndex
     private var sessionYoloWriteRevision: UInt64 = 0
     private var sessionYoloWriteRevisions: [ChatScrollSessionKey: UInt64] = [:]
     /// Sessions whose user-initiated YOLO write is awaiting its RPC, tracked
@@ -1267,6 +1268,7 @@ final class AppState: ObservableObject {
         chatResumeLifecycleOperations: ChatResumeLifecycleOperations = .live,
         sessionPresentationCache: SessionPresentationCache = .shared,
         sessionYoloStore: SessionYoloStore? = nil,
+        conversationIdentityIndex: ConversationIdentityIndex? = nil,
         presentationCacheDebounceSuspension: (@Sendable (Duration) async throws -> Void)? = nil
     ) {
         self.presentationCacheDebounceSuspension =
@@ -1275,6 +1277,7 @@ final class AppState: ObservableObject {
         self.defaults = defaults
         self.sessionPresentationCache = sessionPresentationCache
         self.sessionYoloStore = sessionYoloStore ?? SessionYoloStore(defaults: defaults)
+        self.conversationIdentityIndex = conversationIdentityIndex ?? ConversationIdentityIndex()
         self.chatResumeCoordinator = chatResumeCoordinator
             ?? ChatResumeCoordinator(store: ChatResumeStore(defaults: defaults))
         self.recoverySequence = recoverySequence
@@ -1995,6 +1998,12 @@ final class AppState: ObservableObject {
         defaults.removeObject(forKey: reviewSummaryCacheKey)
         defaults.removeObject(forKey: knownProfilesKey)
         clearSessionPresentationCache()
+        // Identity evidence and per-session overrides are keyed only by
+        // (profile, session id); without this clear they would leak between
+        // Hermes servers whose strings collide. Same boundary that clears
+        // the resume store, titles, pins, and review cache.
+        conversationIdentityIndex.removeAll()
+        sessionYoloStore.clearAllOverrides()
         return true
     }
 
@@ -2767,6 +2776,9 @@ final class AppState: ObservableObject {
             }
             sessions = allSessions.filter { $0.source != .cron }
             cronSessions = allSessions.filter { $0.source == .cron }
+            // Labeled rows are positive identity evidence; commit them so
+            // notification routing survives a later catalog omission.
+            conversationIdentityIndex.recordCatalogIdentity(allSessions, profile: profile)
 
             guard automaticChatResumeWorkIsCurrent(
                 automaticWorkToken,
@@ -3277,6 +3289,31 @@ final class AppState: ObservableObject {
                     || referenceIdentity.durableSessionID == established {
                     context?.resolvedDurableSessionId = established
                 }
+                // Commit the admitted result as positive evidence in the
+                // shared identity index: the returned runtime id (and the
+                // requested id, when it differs) positively route to the
+                // conversation's durable identity.
+                let admittedDurable = context?.resolvedDurableSessionId
+                    ?? referenceIdentity.durableSessionID
+                if let admittedDurable, !admittedDurable.isEmpty {
+                    for admittedRuntimeID in [result.sessionId, sessionId] {
+                        conversationIdentityIndex.record(
+                            runtimeID: admittedRuntimeID,
+                            durableID: admittedDurable,
+                            profile: profile,
+                            source: .resume
+                        )
+                    }
+                }
+                // A live voice turn captured the pre-rebind runtime; the
+                // admitted alias keeps its assistant stream flowing — but
+                // only when the reconciled conversation IS the voice turn's
+                // conversation (positive id overlap), never another one.
+                voiceConversationController.extendAssistantSessionIDs(
+                    [result.sessionId],
+                    ofConversationContaining: referenceIdentity.acceptedSessionIDs
+                        .union([sessionId])
+                )
             }
             reconciliation = context
             refreshActiveChatScrollSessionIdentity(isReconciling: true)
@@ -3676,6 +3713,19 @@ final class AppState: ObservableObject {
             }
 
             let storedID = created.storedSessionId ?? runtimeSessionID
+            // The create response's runtime/stored semantics are verified
+            // here (the summary is built from the same response), so the
+            // pair is positive identity evidence.
+            if let runtime = ChatScrollIdentityNormalization.sessionID(runtimeSessionID),
+               let durable = ChatScrollIdentityNormalization.sessionID(storedID),
+               runtime != durable {
+                conversationIdentityIndex.record(
+                    runtimeID: runtime,
+                    durableID: durable,
+                    profile: activeProfile,
+                    source: .create
+                )
+            }
             let summary = SessionSummary(
                 id: storedID,
                 alternateIds: [runtimeSessionID, created.storedSessionId]
@@ -5973,6 +6023,9 @@ final class AppState: ObservableObject {
             )
             sessions = allSessions.filter { $0.source != .cron }
             cronSessions = allSessions.filter { $0.source == .cron }
+            // Labeled rows are positive identity evidence; commit them so
+            // notification routing survives a later catalog omission.
+            conversationIdentityIndex.recordCatalogIdentity(allSessions, profile: profile)
             if let activeSessionId { updateActiveSessionTitle(for: activeSessionId) }
             if let activeClient {
                 Task { [weak self] in
@@ -6168,9 +6221,14 @@ final class AppState: ObservableObject {
                 method: "DELETE"
             )
             guard profile == activeProfile else { return false }
+            let deletedSessionIDs = Set([session.id] + session.alternateIds)
             sessionYoloStore.clearOverride(
                 for: profile,
                 sessionIDs: [session.id] + session.alternateIds
+            )
+            revokeDeletedConversationIdentity(
+                sessionIDs: deletedSessionIDs,
+                profile: profile
             )
             removeSessionFromLiveCatalog(session)
             archivedSessions.removeAll { sessionMatches($0, session) }
@@ -6186,6 +6244,21 @@ final class AppState: ObservableObject {
 
     func isSessionMutationInFlight(_ session: SessionSummary) -> Bool {
         sessionMutationID == session.id
+    }
+
+    /// Explicit deletion revokes the conversation's identity: index
+    /// mappings, scroll/resume state, and cached presentation (with any
+    /// pending cards) must not survive under any of its aliases.
+    func revokeDeletedConversationIdentity(sessionIDs: Set<String>, profile: String) {
+        conversationIdentityIndex.removeSessionIDs(sessionIDs, profile: profile)
+        chatResumeCoordinator.removeSessions(
+            profile: profile,
+            sessionIDs: Array(sessionIDs)
+        )
+        sessionPresentationCache.removeSessions(
+            profile: profile,
+            sessionIDs: Array(sessionIDs)
+        )
     }
 
     private func encodedSessionID(_ sessionID: String) -> String {
@@ -6435,9 +6508,17 @@ final class AppState: ObservableObject {
             return false
         }
         let requestedID = target.sessionId
-        let resumableID = NotificationSessionResolver.resumableSessionID(
-            for: requestedID,
-            in: sessions + cronSessions
+        // The fresh catalog was just published (its labeled rows are already
+        // committed to the identity index), so resolution runs through the
+        // evidence hierarchy: explicit durable id from the payload, catalog
+        // alias, confirmed index alias — and only then the legacy raw
+        // runtime resume. An unknown runtime id is never reinterpreted as
+        // some other conversation's durable id.
+        let route = NotificationSessionResolver.route(
+            target: target,
+            catalog: sessions + cronSessions,
+            identityIndex: conversationIdentityIndex,
+            profile: activeProfile
         )
         // A decision raised while the app was backgrounded is delivered as a
         // structured payload on the notification (the one-shot gateway stream
@@ -6445,12 +6526,32 @@ final class AppState: ObservableObject {
         // stored IDs so the upcoming resume's `merge` restores the card
         // regardless of which identity the gateway resumes against.
         if let decision = target.decision {
-            recordNotificationDecision(decision, sessionIDs: [resumableID, requestedID])
+            var routedIDs = [route.resumeTargetID, requestedID]
+            if let durable = route.durableSessionID {
+                routedIDs.append(durable)
+            }
+            recordNotificationDecision(decision, sessionIDs: routedIDs)
         }
         let opened = await openSession(
-            resumableID,
+            route.resumeTargetID,
             reusing: transitionGeneration
         )
+        // Commit the payload's dual identity as positive evidence only once
+        // the open actually succeeded — a failed resume (e.g. the durable
+        // conversation was deleted server-side) must not leave a mapping
+        // that dead-routes future notifications.
+        if opened, let durable = route.durableSessionID {
+            if let conflict = conversationIdentityIndex.record(
+                runtimeID: requestedID,
+                durableID: durable,
+                profile: activeProfile,
+                source: .notification
+            ) {
+                sessionCatalogLog.fault(
+                    "Notification identity conflict: runtime \(requestedID, privacy: .public) keeps confirmed durable \(conflict.confirmedDurableID, privacy: .public); payload claimed \(durable, privacy: .public)"
+                )
+            }
+        }
         guard notificationOpenAttemptIsCurrent(
             id: notificationAttemptID,
             transitionGeneration: transitionGeneration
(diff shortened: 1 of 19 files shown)


kaishi00:
## Summary
This PR hardens conversation identity across session boundaries: it adds a profile-scoped `ConversationIdentityIndex` (accepted-evidence-only runtime→durable mappings), rewires notification routing through an explicit evidence hierarchy, clears identity/YOLO state at the server-change boundary, revokes identity on delete, and makes mid-turn voice rebinds follow confirmed aliases instead of raw id equality.

## Findings

### [minor] Logout does not clear the new identity index or YOLO overrides — cross-account retention on a shared server
**File:** Conduit/Services/AppState.swift:2005-2006 (and the absence in `disconnect()` at AppState.swift:2424-2486)
**Issue:** `prepareChatResumeForConnection(to:)` clears `conversationIdentityIndex` and `sessionYoloStore` overrides on a server change, and `disconnect()` clears the resume store, titles, pins, and presentation cache — but `disconnect()` does **not** clear the identity index or YOLO overrides. Two distinct accounts on the same Hermes server (sharing a profile name and colliding session-id strings) would inherit the previous account's runtime→durable mappings and YOLO choices. The same (profile, session id) key space that justifies the server-boundary clear applies here too, and the existing `disconnect()` already treats logout as a harder boundary than a server change for the resume store.
**Suggestion:** Call `conversationIdentityIndex.removeAll()` and `sessionYoloStore.clearAllOverrides()` in `disconnect()` alongside `chatResumeCoordinator.clearResumeState()`, or document explicitly why logout preserves the index while clearing the resume store.

### [nit] `.fault` severity for the designed "earlier row in snapshot wins" case
**File:** Conduit/Services/ConversationIdentityIndex.swift:148-150
**Issue:** `recordCatalogIdentity` emits `conversationIdentityLog.fault` whenever a snapshot contains a conflicting row that the first-claim-wins guard then intentionally ignores. The comment on line 121 explicitly says the committed catalog "can contain cached rows merged after fresh ones", so this is a designed, handled path — yet it surfaces in crash/diagnostic logs as a fault on every routine load where the guard swallows a stale row.
**Suggestion:** Log `.info`/`.debug` when the conflict is resolved by `claimedInSnapshot` (the "earlier row in this snapshot won" branch) and reserve `.fault` for the genuine cross-snapshot "routing identity changed" case.

### [nit] `ConduitNotificationTarget.id` omits `durableSessionID`
**File:** Conduit/Services/PushNotificationService.swift:23
**Issue:** The `Identifiable.id` is built from `profile + sessionId + type` only. Two payloads with the same runtime id/type but different explicit durable ids collide in `id`, which is consumed as a diagnostics event key at ConduitApp.swift:106. Routing (`clearPendingTarget`/`handleFailedNotificationRoute`) uses full Equatable equality, so this is diagnostics-only, but the composed event id can't tell the two targets apart.
**Suggestion:** Include `durableSessionID ?? ""` in `id` so the diagnostics surface distinguishes the targets.

### [nit] Unused `activeList` evidence source
**File:** Conduit/Services/ConversationIdentityIndex.swift:49
**Issue:** `EvidenceSource.activeList` is declared but never recorded anywhere; the reconciling/resume paths always use `.resume`. Dead vocabulary implies the "active_list registries" evidence claim in the PR body isn't actually wired to a producer.
**Suggestion:** Either wire the registry-probe path to record `.activeList` evidence, or drop the case until a caller exists.

### [nit] Docstring ordering claim vs. actual catalog order
**File:** Conduit/Services/ConversationIdentityIndex.swift:120-122 and AppState.swift:2777-2781 / 6024-6028
**Issue:** The comment asserts "routing (`catalog.first`) reads that same order". `recordCatalogIdentity` iterates `allSessions` (the merged, interleaved chat+cron order, `retainedActiveTurn` first), while the resolver reads `sessions + cronSessions` (all chat rows before any cron rows). These orders are only identical when chat and cron rows don't interleave across the split. A runtime id present in both a chat and a cron row with different stored ids (very contrived) could be pinned by the index to a row the resolver would never have matched first.
**Suggestion:** Iterate `sessions + cronSessions` (or call `recordCatalogIdentity` twice with the split arrays) so the index's first-claim-wins order provably matches the resolver's, and correct the docstring.

## Verdict
APPROVE

The core design — accepted-evidence-only recording, conflict surfacing, profile scoping, `@MainActor` isolation throughout, first-claim-wins catalog commits, and the positive-overlap guard on voice rebinds — is sound and well-tested. I found no critical or major defects; no new force-unwraps, retain cycles, or data races were introduced.

<a href="https://opencode.ai/s/mTu83jSB">New%20session%20-%202026-09-06T20%3A24%3A58.474Z</a>
[opencode session](https://opencode.ai/s/mTu83jSB)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34057858175)

kaishi00:
## Summary
This PR hardens conversation identity across session boundaries: a profile-scoped `ConversationIdentityIndex` centralizes accepted runtime→durable evidence, notification routing resolves through an explicit evidence hierarchy (explicit durable → catalog alias → confirmed alias → legacy runtime), the server-change boundary also clears identity/YOLO state, explicit deletion revokes identity across all aliases, and mid-turn voice rebinds follow confirmed aliases instead of raw id equality. Extensive test coverage (4 new/extended suites, ~2700 lines) backs the change.

## Findings

### [minor] First-claim-wins catalog commit order does not match the resolver's read order
**File:** Conduit/Services/AppState.swift:2803, 6118; Conduit/Services/ConversationIdentityIndex.swift:120-122
**Issue:** `recordCatalogIdentity` iterates `allSessions` (`[retainedActiveTurn] + loadedSessions` in merged order), while `NotificationSessionResolver.route` reads `catalog.first` over `sessions + cronSessions` (all chat rows before any cron rows). These orders are only identical when chat and cron rows never interleave across the split. For a runtime id present in both a chat row and a cron row with different stored ids (contrived, but the model explicitly allows it), the index can pin the durable the resolver would never have matched first — the docstring's "routing (`catalog.first`) reads that same order" claim is inaccurate.
**Suggestion:** Iterate `sessions + cronSessions` (or call `recordCatalogIdentity` with the split arrays in resolver order) so the index's first-claim-wins order provably matches the resolver, and correct the docstring.

### [minor] `.fault` severity on designed, handled paths
**File:** Conduit/Services/ConversationIdentityIndex.swift:208-210; 122-125
**Issue:** `recordCatalogIdentity` logs `conversationIdentityLog.fault` for the conflict it then deliberately swallows via `claimedInSnapshot` — the comments (line 121) describe this as the designed merged-cache case — so every routine catalog load where a stale row loses surfaces as a fault in crash/diagnostic logs. Additionally, a notification conflict is fault-logged twice: `record()` at the index (line 122) and again at the call site (AppState.swift:6647).
**Suggestion:** Log the swallowed "earlier row in this snapshot won" branch at `.info`/`.debug` and reserve `.fault` for the genuine "routing identity changed" cross-snapshot case; log the notification conflict once (keep the call-site log, drop the index-level one or make it `.notice`).

### [minor] Decision card cached only under `requestedID` — not migrated when the payload runtime isn't a catalog alias of the durable row
**File:** Conduit/Services/AppState.swift:6613-6650
**Issue:** In the `explicitDurable` case, the card is cached under `requestedID` only, on the assumption that the admitted resume's `consolidateUnderDurableKey` migrates it to the durable key. That migration only covers ids in `acceptedSessionIDs` (= the resume row's `id` + `alternateIds`). If the payload's runtime id is not an alias of the durable row in the freshly loaded catalog, the card stays under a runtime key that is never retired and is not read by the durable-keyed resume merge — the answerable card is lost, and the orphaned runtime key is exactly the retention the consolidation is designed to retire.
**Suggestion:** After a successful open, record/consolidate the `requestedID` runtime key explicitly (include it in the post-open migration's alias set), or have `openNotificationTarget` pass the payload's runtime id into the admission's consolidation. A targeted test with a payload runtime absent from the catalog row's aliases would pin this.

### [minor] `disconnect()` does not clear the identity index or YOLO overrides
**File:** Conduit/Services/AppState.swift:2446-2508
**Issue:** The server-change boundary (`prepareChatResumeForConnection(to:)`, line 2027-2028) clears both because they are keyed only by `(profile, session id)`. `disconnect()` clears the resume store, titles, pins, and presentation cache but not the index/YOLO. Two different accounts on the same Hermes server with colliding profile/session strings would inherit the previous account's mappings. The PR documents this as intentional ("same server identity"), which is defensible, but the same key-space rationale that justified the server-change clear applies here.
**Suggestion:** Either clear `conversationIdentityIndex` and `sessionYoloStore` in `disconnect()` (treating logout as the harder boundary, consistent with the resume store), or document the account-mixing rationale explicitly in the `disconnect()` body.

### [nit] `ConduitNotificationTarget.id` omits `durableSessionID`
**File:** Conduit/Services/PushNotificationService.swift:23
**Issue:** `id` is built from `profile + sessionId + type`. Two payloads with the same runtime id/type but different explicit durable ids collide; `id` feeds the diagnostics event key at ConduitApp.swift:106. Routing uses full Equatable equality so this is diagnostics-only.
**Suggestion:** Append `durableSessionID ?? ""` to `id`.

### [nit] `SessionPresentationCache.removeSessions` key-construction inconsistency with `key(_:_:)`
**File:** Conduit/Services/SessionPresentationCache.swift:574-590
**Issue:** `removeSessions` builds keys as `normalized(profile) + "|" + trimmedID`, while `key(profile:sessionID:)` (used by every write path) must use the same composition. If the caller passes an untrimmed id whose trimmed form matches a stored key, removal works — consistency appears correct here — but the duplicated logic invites drift. Minor at most; consider reusing `key(profile:sessionID:)` truncated by a fixed prefix, or asserting untrimmed ids.
**Suggestion:** Reuse the canonical `key(_:_:)` helper in `removeSessions`.

## Verdict
COMMENT

The architecture is sound and well-tested: `@MainActor` isolation is consistent across the index, resolver, coordinator, and stores; no new force-unwraps, retain cycles, or data races were introduced; and the previously flagged defects (pending-restoration deletion leak, unused `.activeList` source, test `UserDefaults` force-unwrap) are resolved in the current HEAD. All remaining findings are minor or nit severity — the catalog ordering mismatch and the explicit-durable card-migration gap are latent, contrived-case correctness or reliability issues worth pinning with the suggested targeted tests, but they do not block merge.

<a href="https://opencode.ai/s/YPgJqPs8">New%20session%20-%202026-09-06T22%3A23%3A11.830Z</a>
[opencode session](https://opencode.ai/s/YPgJqPs8)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34063806260)

kaishi00:
## Summary
This PR hardens conversation identity across session boundaries: it adds a profile-scoped `ConversationIdentityIndex` (accepted-evidence-only runtime→durable mappings), rewires notification routing through an explicit evidence hierarchy (explicit durable → catalog alias → confirmed alias → legacy runtime), clears identity/YOLO state at the server-change boundary, revokes identity on delete, and makes mid-turn voice rebinds follow confirmed aliases instead of raw id equality.

The previously-flagged defects are resolved in the current HEAD: pending-restoration invalidation on delete (`ChatResumeCoordinator.removeSessions`), the `nil` `UserDefaults(suiteName:)` test trap (now `XCTUnwrap`), and the confirmedAlias decision-card migration gap (threaded as `presentationMigrationSessionIDs` in commit 5775f69). Remaining findings are minor/nit.

## Findings

### [minor] `disconnect()` does not invalidate the identity index or YOLO overrides — cross-account retention on a shared server
**File:** Conduit/Services/AppState.swift:2446-2508 (contrast `prepareChatResumeForConnection` at 2027-2028)
**Issue:** The server-change boundary clears the index and YOLO overrides precisely because they are keyed only by (profile, session id). `disconnect()` clears the resume store, titles, pins, and presentation cache but not the index or YOLO overrides. Two different accounts on the same Hermes server whose profile/session strings collide would inherit the previous account's runtime→durable mappings; the `confirmedAlias` routing fallback can then route a notification to a stale durable id the fresh catalog does not name. The PR documents this as intentional ("same server identity"), which is defensible, but the same key-space rationale that justified the server-change clear applies at the harder logout boundary.
**Suggestion:** Call `conversationIdentityIndex.removeAll()` and `sessionYoloStore.clearAllOverrides()` in `disconnect()` alongside `clearResumeState()`, or document the account-mixing rationale explicitly in the `disconnect()` body.

### [minor] First-claim-wins catalog commit order can diverge from the resolver's read order
**File:** Conduit/Services/AppState.swift:2803, 6131; Conduit/Services/ConversationIdentityIndex.swift:188-220
**Issue:** `recordCatalogIdentity` iterates `allSessions` (interleaved chat+cron order, `retainedActiveTurn` first), while `NotificationSessionResolver.route` scans `sessions + cronSessions` (all chat rows before any cron rows). For a runtime id present in both a chat row and a cron row carrying different stored ids (contrived, but the dual-row model explicitly allows it), the index's first-claim-wins pins the mapping the resolver would never have matched first — the docstring's "routing (`catalog.first`) reads that same order" claim (ConversationIdentityIndex.swift:120-122) is inaccurate.
**Suggestion:** Iterate `sessions + cronSessions` (or call `recordCatalogIdentity` with the split arrays in resolver order) so the index's first-claim order provably matches the resolver, and correct the docstring.

### [minor] `gatewayAnnouncesPendingGate` suppresses ALL cached pending approvals when the gateway announces any single one
**File:** Conduit/Services/SessionPresentationCache.swift:310-318
**Issue:** The old loop restored each cached pending approval not already present in `merged` (per-decision-key dedup). The new branch `break`s out of the entire loop when the gateway transcript contains *any* pending approval under a supplied id. If a conversation legitimately holds two pending approval cards (distinct decision keys), the second card is silently dropped on resume even though it is not announced by the gateway. The "single gate per conversation" invariant is asserted in the comment; if that invariant ever weakens, the all-or-nothing suppression loses cards that the previous per-decision logic would have kept.
**Suggestion:** Key the suppression to the specific announced decision keys rather than an all-or-nothing per-conversation flag, or add a test pinning the multi-pending-approval case.

### [nit] `.fault` severity on designed, handled paths (and a double-fault on notification conflict)
**File:** Conduit/Services/ConversationIdentityIndex.swift:122-125, 208-210; Conduit/Services/AppState.swift:6673-6675
**Issue:** `recordCatalogIdentity` emits `conversationIdentityLog.fault` for the "earlier row in this snapshot won" branch it then deliberately swallows — a designed, routine path that surfaces as a fault on ordinary catalog loads. Additionally, a dual-ID notification whose claim conflicts with a confirmed mapping is fault-logged twice: once in `record()` at the index and once at the `openNotificationTarget` call site.
**Suggestion:** Log the swallowed "earlier row in this snapshot won" branch at `.info`/`.debug`, reserve `.fault` for the genuine cross-snapshot "routing identity changed" case, and keep only the call-site fault for notification conflicts.

### [nit] `ConduitNotificationTarget.id` omits `durableSessionID`
**File:** Conduit/Services/PushNotificationService.swift:23
**Issue:** `id` is composed of `profile + sessionId + type`. Two payloads with the same runtime id and type but different explicit durable ids collide; the id feeds the diagnostics event key at ConduitApp.swift:106. Routing uses full `Equatable` equality, so this is diagnostics-only.
**Suggestion:** Append `durableSessionID ?? ""` to `id`.

### [nit] Force-unwrap in freshest-marker comparison
**File:** Conduit/Services/SessionPresentationCache.swift:744
**Issue:** `aliasMarker > freshestContributedMarker!` is logically safe (short-circuit `||` on the preceding `== nil` check) and the only force-unwrap introduced in the diff, but it is exactly the kind of implicit-unwrap pattern reviewers flag.
**Suggestion:** Use an optional binding (e.g. `if var freshest = freshestContributedMarker, aliasMarker > freshest { freshest = aliasMarker }`) for defensiveness.

## Verdict
COMMENT

The architecture is sound and well-tested: `@MainActor` isolation is consistent across the index, resolver, coordinator, and stores; no new retain cycles, data races, or accessibility regressions were introduced; and the substantive defects from prior review rounds (delete pending-restoration leak, confirmedAlias card migration, test `UserDefaults` force-unwrap, `.activeList` dead vocabulary) are resolved. All remaining findings are minor or nit — the logout-retention, catalog-order mismatch, and multi-approval suppression are latent/contrived correctness or reliability questions worth pinning with targeted tests, but none block merge.

<a href="https://opencode.ai/s/DM1L5KcZ">New%20session%20-%202026-09-07T00%3A46%3A25.696Z</a>
[opencode session](https://opencode.ai/s/DM1L5KcZ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34070804997)

kaishi00:
## Summary
This PR centralizes conversation identity across session boundaries: it adds a profile-scoped `ConversationIdentityIndex`, rewires notification routing through an explicit evidence hierarchy (explicit durable → catalog alias → confirmed alias → legacy runtime), commits `active_list` registry rows and verified create/branch/resume results, makes persisted presentation durable-owned (with alias retirement), clears identity/YOLO state at the server-change boundary, revokes identity on delete, and lets mid-turn voice rebinds follow confirmed aliases. Update covers `ConversationIdentityIndex`, `PushNotificationService`, `SessionPresentationCache`, `ChatResumeCoordinator`/`Store`, `VoiceConversationController`, and 4 new/extended test suites.

## Findings

### [minor] Logout does not clear the identity index or YOLO overrides — cross-account retention on a shared server
**File:** Conduit/Services/AppState.swift:2446 (contrast `prepareChatResumeForConnection(to:)` at AppState.swift:2027-2028)
**Issue:** The server-change boundary clears `conversationIdentityIndex` and `sessionYoloStore` precisely because both are keyed only by (profile, session id). `disconnect()` clears the resume store, titles, pins, and presentation cache but not the index or YOLO overrides. Two different accounts on the same Hermes server whose profile/session strings collide would inherit the previous account's runtime→durable mappings; the `confirmedAlias` fallback could then route a notification to a stale durable id the fresh catalog does not name. The PR documents same-server-identity retention as intentional, which is defensible, but the same key-space rationale that justified the server-change clear applies at this harder boundary.
**Suggestion:** Call `conversationIdentityIndex.removeAll()` and `sessionYoloStore.clearAllOverrides()` in `disconnect()` alongside `clearResumeState()`, or document the cross-account mixing rationale explicitly in the `disconnect()` body.

### [minor] Catalog commit order can diverge from the resolver's read order
**File:** Conduit/Services/ConversationIdentityIndex.swift:188 (and the docstring at 182); AppState.swift:2801, 6131; PushNotificationService.swift:198
**Issue:** `recordCatalogIdentity(allSessions)` iterates `allSessions` (retained-active-turn-first, merged chat+cron order) while `NotificationSessionResolver.route` scans `sessions + cronSessions` (all chat rows before any cron rows). For a runtime id present in both a chat row and a cron row carrying different stored ids (contrived, but the dual-row model explicitly allows it), the index's first-claim-wins pins the mapping the resolver would never have matched first, and the docstring's "routing (`catalog.first`) reads that same order" claim is inaccurate.
**Suggestion:** Iterate `sessions + cronSessions` (or call `recordCatalogIdentity` with the split arrays in resolver order) so the index's first-claim order provably matches the resolver, and correct the docstring.

### [minor] `gatewayAnnouncesPendingGate` suppresses *all* cached pending cards when the gateway announces any single one
**File:** Conduit/Services/SessionPresentationCache.swift:302-331
**Issue:** The old loop restored each cached pending approval not already present in `merged` (per-decision-key dedup). The new branch `break`s out of the whole loop as soon as the gateway transcript contains *any* pending approval under a supplied id. If a conversation legitimately holds two pending approval cards under distinct decision keys, the second is silently dropped on resume even though the gateway never announced it. The "single gate per conversation" invariant is asserted in the comment; if it weakens, this all-or-nothing suppression loses cards the per-decision logic would have kept.
**Suggestion:** Key the suppression to the specific announced decision keys rather than an all-or-nothing per-conversation flag, or add a test pinning the multi-pending-approval case.

### [nit] `.fault` severity on designed, handled paths; notification conflict fault-logged twice
**File:** Conduit/Services/ConversationIdentityIndex.swift:122-125, 208-210; Conduit/Services/AppState.swift:6673-6675
**Issue:** `recordCatalogIdentity` emits `conversationIdentityLog.fault` for the stale-row branch it then deliberately swallows via `claimedInSnapshot` — a routine, designed path on ordinary catalog loads. Additionally, a dual-ID notification whose claim conflicts with a confirmed mapping is fault-logged once in `record()` at the index and again at the `openNotificationTarget` call site.
**Suggestion:** Log the swallowed "earlier row in this snapshot won" branch at `.info`/`.debug`, reserve `.fault` for the genuine cross-snapshot "routing identity changed" case, and log the notification conflict once (keep the call-site log).

### [nit] `ConduitNotificationTarget.id` omits `durableSessionID`
**File:** Conduit/Services/PushNotificationService.swift:23
**Issue:** `id` is composed of `profile + sessionId + type`. Two payloads with the same runtime id and type but different explicit durable ids collide; the id feeds the diagnostics event key at ConduitApp.swift. Routing uses full `Equatable` equality, so this is diagnostics-only.
**Suggestion:** Append `durableSessionID ?? ""` to `id`.

### [nit] Force-unwrap in freshest-marker comparison
**File:** Conduit/Services/SessionPresentationCache.swift:744
**Issue:** `freshestContributedMarker!` is logically safe (short-circuited by the preceding `== nil` check) and is the only force-unwrap introduced in the diff, but it is exactly the implicitly-unwrapped pattern reviewers flag.
**Suggestion:** Rewrite as `if var freshest = freshestContributedMarker, aliasMarker > freshest { freshest = aliasMarker }`.

## Verdict
COMMENT

The architecture is sound and well-tested: `@MainActor` isolation is consistent across the index, resolver, coordinator, and stores; the previously-flagged substantive defects (delete pending-restoration resurrection, confirmedAlias decision-card migration, test `UserDefaults` force-unwrap, unused `.activeList` source, voice extension conversation-blindness) are resolved in the current HEAD; no new retain cycles, data races, or accessibility regressions were introduced. All remaining findings are minor or nit — the logout-retention, catalog-order mismatch, and multi-approval suppression are latent/contrived correctness or reliability questions worth pinning with targeted tests, but none block merge.

<a href="https://opencode.ai/s/85pHz7rW">New%20session%20-%202026-09-07T01%3A29%3A41.192Z</a>
[opencode session](https://opencode.ai/s/85pHz7rW)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34073128212)

PR #138, opened 2026-09-06
feat(setup): test the proposed connection before accepting it

kaishi00:
## Round 4: test the proposed connection before accepting it

Rounds 1–3 built the guided Connection Setup wizard (failure taxonomy → wizard → route-specific entry, credentials, URL synthesis, Review, typed handoff). Round 4 closes the remaining gap: **the wizard can now prove the proposed settings actually work before the user accepts them**, with per-stage diagnosis instead of a generic failure banner.

```
Details → Credentials → Test Connection → Review (only after success) → Use These Settings
```

### Probe architecture — orchestration, not reimplementation

`ConnectionSetupProbe` (the only `ConnectionSetupTesting` conformer in production) drives the **existing `NativeAuthClient` unchanged** — same URL construction, transport policy, redirect policy, Cloudflare header application, cookie rules, and status classification. No new endpoints, no second auth implementation.

| Stage | Copy (running → success) | What proves it |
|---|---|---|
| 1 · server | `Checking server…` → `Dashboard reachable` | The provider-discovery request (`GET /api/auth/providers`) returning any HTTP response — transport errors classify from `URLError` codes via the existing classifier |
| 2 · dashboard | `Checking dashboard…` → `Hermes dashboard found` | The same response names a password-capable provider — the **exact** `supports_password` check the normal login performs (now one shared helper, `HermesProviderCheck`, used by both `LoginView` and the probe). 200 alone is never success; a 5xx at discovery proves transport but fails this stage as `.dashboardUnavailable`; a non-Hermes 200 fails as `.unexpectedServerResponse` |
| 3 · authentication | `Authenticating…` → `Login successful` | The full native `connect()`: password login **plus** the ws-ticket mint — the same milestone normal login reaches before it would commit anything |

"Login successful" therefore means what it means everywhere else in Conduit: not just a 200 on password-login, but a usable session artifact.

### Side-effect protections

The probe is a pure validation operation:

- `NativeAuthConnection.commitCookies()` is **never called** — the ticket and transaction cookies are discarded; the shared cookie store is untouched (asserted by test).
- No Keychain writes, no `AppState` mutation, no `HermesClient`/websocket connection, no `lastDashboardURL` update, no session restore, no screen changes, no dismissal.
- **One user-requested test = one discovery request + one password-login attempt + one ticket mint** (asserted by exact request counts). No automatic retries — the rate-limit model is untouched.
[...]

PR #142, opened 2026-09-07
feat(setup): open Connection Setup from Settings for the current connection

kaishi00:
## Round 5: Connection Setup from Settings

Follow-up to #131 → #136 → #138 (rounds 1–4). **Not a setup-system redesign**: this makes the completed Connection Setup Assistant available from Settings for an already-connected user, seeded from the current configuration, running the unchanged Round-4 staged test.

> An already-connected user can open Settings, launch Connection Setup prefilled from their current Hermes configuration, run the staged connection test, inspect diagnostics, optionally edit/test replacement settings, and leave without disrupting the active session unless they explicitly choose to apply a change.

### Settings files changed

| File | Change |
|---|---|
| `Conduit/Views/AuxiliaryViews.swift` | Settings → Connection section gains **Connection Setup** ("Test, troubleshoot, or change your connection") next to the existing Gateway row. Presents the real `ConnectionSetupView` as a sheet via a per-tap seed (`sheet(item:)`, so Keychain reads stay off the render path). Post-apply confirmation alert shown only when the apply plan actually changed something. |
| `Conduit/Services/ConnectionSetupApplication.swift` (new) | Pure apply-plan policy + seeding policy (below). |
| `Conduit/Services/ConnectionFailure.swift` | `ConnectionHelpDestination.currentConnection` — the semantic Settings entry destination. |
| `Conduit/Services/ConnectionSetupFlow.swift` | Entry mapping, `initialPath`, empty-credential continuation (Settings entry only), interactive-aware acceptance (`acceptedResult`), `testedSettingsUnchanged`. |
| `Conduit/Services/ConnectionSetupDraft.swift` | `testConfiguration()` — address + optional credentials. |
| `Conduit/Views/ConnectionSetupView.swift` / `ConnectionSetupForm.swift` | Probe target uses `testConfiguration()`; entry-aware Review copy and Done/Use-these-settings label; address preview on the test screen for the current-connection entry; username/password rows render honestly for credential-less interactive acceptance. |
| `Conduit/Services/AppState.swift` | Inert DEBUG UI-test connected stub (`-CONDUIT_UI_TEST_CONNECTED_DASHBOARD`) + reconnect suppression while it is active (`#if DEBUG` only). |

### Source of current connection data

- **URL**: `appState.connection?.baseUrl` (the **active** connection) falling back to `appState.lastDashboardURL` (last configured). The two concepts stay distinct; applying never writes to the live connection.
- **Credentials**: `KeychainHelper.loadCredentials()`, used only when the record's base URL matches the current address (policy-normalized on both sides).
- **Cloudflare Access**: `KeychainHelper.loadCloudflareAccess(for: currentURL)` — the existing origin-matched Keychain record. No new store, no new persistence.

### Exact Settings navigation

```
Settings (sheet)
└── Connection
    ├── Gateway          (existing; shows current dashboard + status)
    └── Connection Setup → sheet: existing ConnectionSetupView
        → staged test (or prefilled details → credentials → test)
        → Review → Done (unchanged) / Use These Settings (changed)
```

### How current values are seeded

`ConnectionSetupSeeding.wizardCredentials(for:saved:)` returns the saved username/password only for a record saved for **exactly this URL** (policy-normalized comparison). A **Face ID-protected** record surrenders its username only — the wizard asks for the password again rather than bypassing the biometric gate. The Cloudflare token is passed as the same-origin inherited probe token (Round-3 origin rechecks still apply; never displayed or persisted by the wizard).

### Credential handling

- Nothing is persisted by opening, editing, testing, cancelling, or dismissing. The **only** write path is an explicit **Use These Settings** on Review, through the pure `ConnectionSetupApplication.plan`:
  - remembers the dashboard URL when it changed;
  - **replaces** saved credentials only when a record already existed (preserving its Face ID preference) and values differ — it **never creates** credential persistence the user never opted into;
  - clears saved credentials only when the applied URL moved and the applied config has no credentials (a stale password would otherwise be sent to the new address);
  - rewrites the Cloudflare token **only same-origin** (path-only moves); never copies it cross-origin, never deletes the working origin's token.
- Applying to the **same URL** with a credential-less result (the interactive outcome) is a deliberate no-op — that outcome is not a verified replacement for the working credentials.

### Interactive-auth behavior

Both-credential-fields-empty is the interactive-auth signature: the wizard opens **directly on the staged test** (details screen underneath, so Back reaches an editable address), and `draft.testConfiguration()` lets provider discovery run without credentials. Review acceptance stays strict: the address-only handoff is authorized **only** by the current `requiresInteractiveSignIn` outcome; full native success still requires real credentials; every other validation failure stays a failure. From Settings, a browser-sign-in deployment tests to "Browser sign-in required" and offers plain **Done** — no username/password login request, no ticket mint.

### Active-session safety

The wizard still has no `AppState` mutation capability (architectural guarantee — the probe commits nothing, and `perform(appState:)` only writes the saved/login configuration: defaults URL + Keychain). Opening/editing/testing/failing/cancelling/dismissing cannot disconnect the client, replace the connection, change the session, trigger reconnect, or reset chat state. A failed experiment plans nothing. LoginView's entry behavior is unchanged (strict credential gate, normal Use-these-settings handoff, same handoff semantics).

### Changed settings

`test → apply → live session stays intact → new settings take effect on the next explicit reconnect` (app relaunch / next sign-in). No automatic disconnect or reconnect; after a changed apply a single confirmation says exactly that.

### Tests & results

New: `ConnectionSetupSettingsTests` (20 tests: entry shapes, Face ID withheld seed, unbuildable address fallback, exact-URL preservation, unchanged-Done gating to the Settings entry, cancel safety, interactive completion without credentials, credential-less continuation after editing, seeding normalization, apply-plan policy incl. cross-origin token and no-create rules, redaction). New: `ConnectionSetupSettingsUITests` (2 deterministic UI paths on the inert DEBUG connected stub — no real server). Updated: destination mapping pin.

Mac-verified (worktree at this exact HEAD):

- focused: ConnectionSetupSettings/Flow/Test/Draft + LoginCloudflareHandoff + NativeAuthClient — **161/161** ✓
- full `ConduitTests` — **1927/1927, 0 failures** ✓ (baseline 1907 → +20)
- new UI tests locally — **2/2** ✓ (hosted CI UI lanes exercise them again)

### Hosted CI

**Green** — [CI run 34078500111](https://github.com/kaishi00/hermes-conduit/actions/runs/34078500111): Plan & validate, Build test products, unit-1..unit-4, UI tests (incl. the 2 new Settings paths), CodeRabbit, and opencode-review all pass. Test-planner tooling untouched (`scripts/plan-tests.py validate` / `scripts/tests` unaffected — no lane or planner changes).

### Intentionally deferred

- An explicit "Reconnect Now" button (no existing safe connect-with-new-credentials operation outside LoginView; the safer Round-5 contract is apply-now / connect-on-next-reconnect, per spec).
- Any lighter Settings-only test path (exercising the real Round-4 probe is the point).
- The spec's out-of-scope list: no Bonjour/QR/Tailscale-API/multi-server/credential-vault/etc.

## Round 5.1 pre-merge compatibility fix

**The bug:** an actively-connected native-password user who chose NOT to save credentials seeds an empty draft. Round 5 treated "both credential fields empty" as an interactive-auth signature and opened the staged test directly — where provider discovery correctly found a password-capable dashboard, and the probe then called `client.connect(username: "", password: "")`: a real, rate-limited login attempt with empty credentials against a connection that may already be known to work.

**The semantic model:** credentials are **available** or **unavailable** — `ConnectionSetupResult.hasUsableCredentials` (presence-only, non-empty after trimming; values are never trimmed or rewritten, and `ConnectionSetupDraft.result()` now reuses the same check). Credential absence is NOT an authentication mode; provider discovery remains the sole authority.

**Staged-state changes:** new typed `ConnectionSetupStageState.requiresCredentials` + `ConnectionSetupTestEvent.requiresCredentials(_)` emitted at the authentication stage after server and dashboard succeed. It is not a success, not a failure, not `.authenticationRejected`, not rate-limited, and not the interactive outcome; it seals no revision, never advances to Review, and leaves `canUseSettings == false` — no Done/ready/"Login successful" claims.

**Request counts for password-capable + missing credentials (pinned by tests):** exactly **1 provider-discovery request, 0 password-login requests, 0 ws-ticket requests**. Whitespace-only credentials behave identically. The interactive path is discovery-only as before, and the full native path remains 1/1/1.

**Recovery UX:** the test screen shows ✓ Dashboard reachable, ✓ Hermes dashboard found, ○ Credentials required with the explanatory notice and **Enter Credentials** as the primary action (no retry, no failure presentation, no rate-limit warning — no login attempt occurred). It routes to the existing credentials step with the URL preserved; leaving the test step invalidates the partial result, so Continue runs a clean full staged test.

**Face ID:** unchanged — a protected record still seeds username only and lands on the credential entry screen; no biometric weakening, and no native login can occur before both credentials are present.

**Active-session safety:** unchanged — the probe still commits nothing; no Keychain/defaults writes outside the explicit Apply path.

**UI-test Minor (CodeRabbit):** both settings-row taps now use the scroll-aware `tapVisible` helper; a third deterministic UI test drives the credentials-required path via the new DEBUG-only `auth:credentialsRequired` stub.

**Verification (Mac, worktree at pushed HEAD):** focused ConnectionSetupSettings/Test/Flow/Draft + NativeAuthClient + LoginCloudflareHandoff — 160/160; full `ConduitTests` on the main-merged branch — 2001/2001, 0 failures; all 11 Connection Setup UI tests pass locally; `git diff --check` clean. Review gate on the fix: **APPROVE + 2× APPROVE_WITH_NITS, zero BLOCKER/WARNING findings** (hygiene nits applied: full count pins on the whitespace path, reducer monotonicity test, copy-scan exemption comment, probe doc, marker/MARK tidy-up). Branch merged with latest main (#139) — merge was conflict-free.

### Review gate

3-reviewer cycle (2× REQUEST_CHANGES + 1× approve-with-nits) → every finding fixed or declined with rationale (declines: same-URL credential keeping — no-strand invariant; Cloudflare old-token cleanup — single-record Keychain update in place; unverified typed creds under the interactive outcome — LoginView handoff parity). Follow-up re-review of the fix commits: **APPROVE_WITH_NITS**, both nits applied. No open BLOCKERs or WARNINGs.

## Summary by CodeRabbit

* **New Features**
  * Added a Connection Setup option in Settings for updating an existing connection without interrupting the active session.
  * Existing connection details and eligible credentials are prefilled in the setup flow.
  * Added support for browser-based sign-in when credentials are not required.
  * Added clearer handling and recovery when credentials are required.
  * Shows whether settings changed and confirms when updates are applied.

* **Bug Fixes**
  * Improved validation and credential handling during connection testing and setup.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2227,7 +2227,29 @@ final class AppState: ObservableObject {
 
     // MARK: - Connection management
 
+#if DEBUG
+    /// UI-test-only connected state: a snapshot connection with no client and
+    /// no transport. Reconnect paths refuse to run while it is active (see
+    /// `reconnectForRetry`), so the stubbed session is inert by construction
+    /// and Settings-UI tests never touch a network or a real dashboard.
+    private static func uiTestConnectedStub() -> HermesConnection? {
+        let arguments = ProcessInfo.processInfo.arguments
+        guard let index = arguments.firstIndex(of: "-CONDUIT_UI_TEST_CONNECTED_DASHBOARD"),
+              index + 1 < arguments.count else { return nil }
+        return HermesConnection(baseUrl: arguments[index + 1], ticket: "ui-test-stub")
+    }
+#endif
+
     func loadSavedConnection() {
+        #if DEBUG
+        if let stub = Self.uiTestConnectedStub() {
+            connection = stub
+            isConnected = true
+            isConnecting = false
+            showLogin = false
+            return
+        }
+        #endif
         if let credentials = KeychainHelper.loadCredentials() {
             Task { await restoreSavedCredentials(credentials) }
         } else if let saved = KeychainHelper.loadConnection() {
@@ -4539,6 +4561,11 @@ final class AppState: ObservableObject {
     }
 
     func reconnectForRetry(purpose requestedPurpose: ChatResumeSyncPurpose) async {
+        #if DEBUG
+        // The UI-test stubbed session has no transport to restore; every
+        // automatic or explicit reconnect is a deterministic no-op for it.
+        guard Self.uiTestConnectedStub() == nil else { return }
+        #endif
         guard let savedConnection = connection else { return }
         let purpose = beginChatResumeRecovery(purpose: requestedPurpose)
         let automaticWorkToken = purpose == .automaticReturn
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2243,6 +2243,7 @@ final class AppState: ObservableObject {
     func loadSavedConnection() {
         #if DEBUG
         if let stub = Self.uiTestConnectedStub() {
+            lifecycleLog.notice("UI-test connected stub active: no transport will be created and reconnects are inert")
             connection = stub
             isConnected = true
             isConnecting = false
--- a/Conduit/Services/ConnectionFailure.swift
+++ b/Conduit/Services/ConnectionFailure.swift
@@ -194,6 +194,10 @@ enum ConnectionHelpDestination: Equatable, Hashable, Identifiable, CaseIterable 
     case network
     case tls
     case cloudflare
+    /// Round 5: the Settings entry for an already-connected user. Skips the
+    /// first-run readiness questions and seeds the wizard from the current
+    /// configuration instead.
+    case currentConnection
 
     var id: Self { self }
 }
--- a/Conduit/Services/ConnectionSetupApplication.swift
+++ b/Conduit/Services/ConnectionSetupApplication.swift
@@ -29,7 +29,12 @@ struct ConnectionSetupApplication: Equatable, CustomStringConvertible, CustomDeb
     /// credentials DO exist, and the tested URL moved: the applied
     /// configuration cannot use them, and keeping them would send a stale
     /// password to the next reconnect against the new address. The live
-    /// session is untouched either way.
+    /// session is untouched either way. A credential-less apply to the SAME
+    /// URL deliberately keeps the saved record: that shape is the interactive
+    /// sign-in outcome for the current dashboard, which is not a verified
+    /// replacement for the working native credentials — deleting them here
+    /// would strand the still-connected configuration (Round-5 acceptance:
+    /// a failed or inconclusive experiment must be harmless).
     let clearsSavedCredentials: Bool
     /// Same-origin rewrite of the inherited Cloudflare Access token to the
     /// new normalized URL string (path-only moves). A cross-origin apply
@@ -99,8 +104,9 @@ extension ConnectionSetupApplication {
         var rewrite: CloudflareAccessRewrite?
         if let access = savedCloudflareAccess, access.isConfigured,
            LoginCloudflareHandoff.sameOrigin(currentURL, newURL) {
+            let currentOrigin = (try? ConnectionURLPolicy.normalizedBaseURL(currentURL)) ?? currentURL
             let newOrigin = (try? ConnectionURLPolicy.normalizedBaseURL(newURL)) ?? newURL
-            if newOrigin != currentURL {
+            if newOrigin != currentOrigin {
                 rewrite = CloudflareAccessRewrite(access: access, origin: newOrigin)
             }
         }
@@ -140,10 +146,14 @@ enum ConnectionSetupSeeding {
         for dashboardURL: String,
         saved: DashboardCredentials?
     ) -> (username: String, password: String)? {
-        guard let saved,
-              saved.baseURL == dashboardURL.trimmingCharacters(in: .whitespacesAndNewlines) else {
-            return nil
-        }
+        guard let saved else { return nil }
+        // Both sides compare policy-normalized, so a trailing-slash or
+        // default-port spelling of the same address still seeds. Anything
+        // that fails normalization compares trimmed-raw (fail closed).
+        let wanted = (try? ConnectionURLPolicy.normalizedBaseURL(dashboardURL))
+            ?? dashboardURL.trimmingCharacters(in: .whitespacesAndNewlines)
+        let stored = (try? ConnectionURLPolicy.normalizedBaseURL(saved.baseURL)) ?? saved.baseURL
+        guard stored == wanted else { return nil }
         return (saved.username, saved.requiresFaceID ? "" : saved.password)
     }
 }
--- a/Conduit/Services/ConnectionSetupApplication.swift
+++ b/Conduit/Services/ConnectionSetupApplication.swift
@@ -74,11 +74,12 @@ extension ConnectionSetupApplication {
         savedCredentials: DashboardCredentials?,
         savedCloudflareAccess: CloudflareAccessCredentials?
     ) -> ConnectionSetupApplication {
-        let newURL = result.serverURL.trimmingCharacters(in: .whitespacesAndNewlines)
-        let currentURL = currentDashboardURL.trimmingCharacters(in: .whitespacesAndNewlines)
-        // Both addresses are policy-normalized upstream (the wizard's result
-        // by ConnectionSetupAddressBuilder, the current one at save time), so
-        // string equality is exact equality of scheme, host, port, and path.
+        // Normalize both sides before comparing so a non-canonical spelling
+        // of an unchanged address can never plan a spurious rewrite.
+        let newURL = (try? ConnectionURLPolicy.normalizedBaseURL(result.serverURL))
+            ?? result.serverURL.trimmingCharacters(in: .whitespacesAndNewlines)
+        let currentURL = (try? ConnectionURLPolicy.normalizedBaseURL(currentDashboardURL))
+            ?? currentDashboardURL.trimmingCharacters(in: .whitespacesAndNewlines)
         let urlChanged = newURL != currentURL
 
         let hasResultCredentials =
@@ -152,7 +153,8 @@ enum ConnectionSetupSeeding {
         // that fails normalization compares trimmed-raw (fail closed).
         let wanted = (try? ConnectionURLPolicy.normalizedBaseURL(dashboardURL))
             ?? dashboardURL.trimmingCharacters(in: .whitespacesAndNewlines)
-        let stored = (try? ConnectionURLPolicy.normalizedBaseURL(saved.baseURL)) ?? saved.baseURL
+        let stored = (try? ConnectionURLPolicy.normalizedBaseURL(saved.baseURL))
+            ?? saved.baseURL.trimmingCharacters(in: .whitespacesAndNewlines)
         guard stored == wanted else { return nil }
         return (saved.username, saved.requiresFaceID ? "" : saved.password)
     }
--- a/Conduit/Services/ConnectionSetupApplication.swift
+++ b/Conduit/Services/ConnectionSetupApplication.swift
@@ -1,0 +1,149 @@
+//
+//  ConnectionSetupApplication.swift
+//  Conduit
+//
+//  Round 5: the persisted-configuration effect of applying a tested
+//  ConnectionSetupResult from the Settings current-connection entry.
+//  Deliberately a pure plan: the policy — update the saved/login
+//  configuration WITHOUT touching the active connection, never create
+//  credentials that were never saved, never copy a Cloudflare service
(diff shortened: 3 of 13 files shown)


kaishi00:
## Summary
Adds a Settings entry ("Connection Setup") that opens the existing Connection Setup wizard seeded from the current connection's URL and saved credentials, plus a pure apply-plan (`ConnectionSetupApplication.plan`) that writes only saved/login configuration without touching the live session, and a DEBUG-only connected stub for UI tests.

## Findings

### [minor] Spurious apply plan (and confirmation alert) when the stored record's baseURL spelling differs from the normalized tested URL
**File:** Conduit/Services/ConnectionSetupApplication.swift:93-99
**Issue:** `plan()` builds `DashboardCredentials(baseURL: newURL, ...)` where `newURL` is policy-normalized, then compares `replacement != saved` against the raw Keychain record. If the saved record's `baseURL` differs only by spelling (trailing slash, default port) from the active connection's URL, this equality fails and the plan becomes non-empty — a `saveCredentials` rewrite plus the "Settings applied" alert — even though the user changed nothing. This contradicts the PR's stated promise ("alert shown only when the apply plan actually changed something").
**Suggestion:** Compare the credential URLs policy-normalized (reuse the `ConnectionSetupSeeding`-style normalization) when deciding whether `replacement` differs, so a spelling-only difference never plans a write.

### [minor] Both-fields-empty seed opens directly on the staged test, assuming interactive auth — a native dashboard without a saved password gets an empty-credential probe attempt
**File:** Conduit/Services/ConnectionSetupFlow.swift:221-232
**Issue:** `initialPath` treats "both credential fields empty" as the interactive-auth signature and opens the wizard straight on `.connectionTest`. But that shape also occurs for a *native* deployment where credentials were never saved (the user types them at each login). In that case the probe runs stage 3 with empty credentials against the user's own server — the same "never spend an empty-credential login attempt" the code carefully avoids for the username-withheld seed (line 59 test). The failure then bounces the user to Edit Credentials, so it's a bad first impression, not a safety issue.
**Suggestion:** Keep the behavior but document the assumption, or gate the direct-to-test launch on something less ambiguous (e.g., a stored indicator that the dashboard is interactive), or at minimum update the doc comment to acknowledge the native-but-unsaved false positive.

### [minor] Entry is shown even when not connected and can seed an empty URL, producing a dead-end details form
**File:** Conduit/Views/AuxiliaryViews.swift:629-644
**Issue:** The "Connection Setup" row is always rendered. If `appState.connection` is nil and `lastDashboardURL` is still empty (signed-out user), `currentDashboardURL` returns `""`, the seeded draft has `usesExistingAddress == false` and no access method, and the wizard lands on a "Choose a connection method" details screen with LAN-style host/port fields and no way to proceed sensibly.
**Suggestion:** Gate the row (or the seed) on a non-empty `currentDashboardURL`, or fall back to the normal question flow for an empty seed.

### [nit] `ConnectionSetupSeed.id` is the URL string
**File:** Conduit/Views/AuxiliaryViews.swift:603-609
**Issue:** `sheet(item:)` identifies seeds by URL. It works today because the item always resets to nil on dismiss before a new seed is set, but two seeds with the same URL (e.g., a future re-seed while presented) would be treated as identical even if credentials differ. Fragile and slightly surprising for an `Identifiable` value.
**Suggestion:** Use a UUID or a monotonically increasing counter as the id.

### [nit] Duplicate accessibility hint on settings rows
**File:** Conduit/Views/AuxiliaryViews.swift:721-734
**Issue:** Both row styles apply `.accessibilityHint(detail)` while the detail text is already a visible `Text` inside the label, so VoiceOver reads the subtitle twice.
**Suggestion:** Drop the hint and rely on the label, or set hint only where the detail is truncated/hidden.

## Verdict
APPROVE

<a href="https://opencode.ai/s/acdT41gZ">New%20session%20-%202026-09-07T03%3A07%3A39.306Z</a>
[opencode session](https://opencode.ai/s/acdT41gZ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34078499970)

kaishi00:
## Summary
Adds a "Connection Setup" row to Settings that opens the existing Connection Setup wizard seeded from the current connection (URL + optionally saved credentials/Cloudflare token), runs the staged probe, and applies changes through a pure, unit-tested `ConnectionSetupApplication.plan` that never touches the live session. Includes a `requiresCredentials` partial outcome so credential absence never produces an empty-credential login attempt, plus DEBUG-only UI-test stubs and 20 unit + 3 UI tests.

## Findings

### [minor] Apply-plan credential equality compares a normalized URL against the raw stored record
**File:** Conduit/Services/ConnectionSetupApplication.swift:93-99
**Issue:** `replacement` is built with `newURL` (policy-normalized), then `replacement != saved` compares it against `saved.baseURL` as stored in Keychain. Commit 86067d3 normalized both sides of `urlChanged` and the seeding comparison, but this equality is still normalized-vs-raw. Today the only writers of the record (`LoginView.swift:505`, `perform` itself) store the normalized form, so the practical risk is limited to legacy/foreign-written records — but the exact "spelling-only difference plans a spurious write" class that commit was meant to kill resurfaces for those.
**Suggestion:** Normalize `saved.baseURL` (fail-closed to trimmed-raw, as `ConnectionSetupSeeding.wizardCredentials` already does) before comparing, or fold the equality into that same normalization helper.

### [minor] "Connection Setup" seed can be an empty URL, producing a dead-end details form
**File:** Conduit/Views/AuxiliaryViews.swift:629-644 and :701-703
**Issue:** The row is always rendered and enabled. For a signed-out user with no remembered dashboard, `currentDashboardURL` is `""`, `usesExistingAddress` is false with `accessMethod == nil`; `initialPath` falls back to `[.connectionDetails]`, where the form renders the Tailscale branch with `methodTitle` "Choose a connection method" and Continue always fails `methodRequired`. There is no way to proceed sensibly.
**Suggestion:** Disable/hide the row (or seed) when `currentDashboardURL` is empty, or fall back to the normal first-run question flow.

### [minor] Duplicate `.connectionTest` node in the back-path after the credentials-required recovery
**File:** Conduit/Services/ConnectionSetupFlow.swift:506-521 (with :554-558 and :332)
**Issue:** On the current-connection route the path is `[.connectionDetails, .connectionTest]` and contains no `.loginCredentials`. `editAfterFailedTest(.loginCredentials)` hits the `else` branch and appends it, then `submitCredentials` calls `advance(to: .connectionTest)`, appending a second test node: `[details, test, creds, test, ...]`. Cosmetic, but Back history then revisits two identical test screens, and Back from the fresh test lands on credentials rather than details.
**Suggestion:** Insert the credentials step *before* the existing `.connectionTest` node when remediating to credentials (remove the subrange from `.connectionTest` onward, then advance to credentials), instead of appending.

### [nit] `ConnectionSetupSeed.id` is the URL string
**File:** Conduit/Views/AuxiliaryViews.swift:603-609
**Issue:** Identifiable identity is the URL, so two seeds for the same address are indistinguishable. Safe today only because the item always resets to nil on dismissal before a new tap re-seeds; a future re-seed while presented (or a duplicate URL trigger) would silently no-op.
**Suggestion:** Use a UUID or monotonic counter as `id`.

### [nit] Duplicate accessibility hint on settings rows
**File:** Conduit/Views/AuxiliaryViews.swift:713-723 and :725-735
**Issue:** `.accessibilityHint(detail)` is applied while `detail` is already a visible `Text` inside the combined label, so VoiceOver reads it twice (this pre-dates changes here but was extended to the new row).
**Suggestion:** Drop the hint and rely on the label, or keep the hint only where the detail is truncated/hidden.

## Verdict
APPROVE

No critical or major findings. Concurrency/retain-safety is sound (`perform` is `@MainActor`, sheet completion runs on main, `testTask` is generation-guarded and cancelled on disappear, no retain cycles, no new force unwraps). The credentials-required outcome correctly keeps the probe side-effect-free, and the active-session safety contract holds: the only persistence path is the explicit apply plan. The nominals above are polish items on an otherwise careful PR.

<a href="https://opencode.ai/s/Uv0ktGzT">New%20session%20-%202026-09-07T04%3A36%3A52.929Z</a>
[opencode session](https://opencode.ai/s/Uv0ktGzT)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34083681447)

Issue #148, opened 2026-09-08
Multiple saved dashboards — switch between Hermes servers without re-login


pstarkgit:
## Feature Description

Conduit currently stores **one** dashboard connection. Switching between hosts (this Mac, a VPS, a second machine) means Disconnect / change URL / sign in again, which drops the previous session cookies and origin-bound credentials.

I run several Hermes dashboards and want to keep them all saved in the app and switch with a tap.

## Motivation

Same phone, several backends: home Mac, a remote box, maybe a work instance. Desktop already has **Settings → Gateways**. On iOS the workaround is re-typing the URL and password every time, which is easy to get wrong (9119 vs another host, Tailscale IP vs LAN).

Profiles on **one** dashboard are not a substitute — those are agents on the same machine, not different machines.

## Proposed Solution

A small saved-server registry, one **active** at a time:

- **Add server:** URL + dashboard username/password (optional Cloudflare Access), same login flow as today.
- **Switch:** pick another saved host; rebuild sessions/chat/settings for that origin. Do not mix transcripts, tickets, or cookies across hosts.
- **Remove / sign out** one host without wiping the others.
- Persist the list in Keychain (URLs and credentials are secrets). Scope cookies, Cloudflare tokens, and Face ID saves **per origin**, the way the current single-origin Keychain items already try to.

Today `KeychainHelper` is a single blob (`hermes-conduit.connection.v1` plus one credentials / Cloudflare record). That would need to become a list + `activeServerID`, with identity (label) per entry.

## Alternatives Considered

- Keep disconnecting and re-entering the URL. Works, bad on a phone.
- Point everything at one dashboard. Not how the installs are laid out.

## Additional Context

Happy to test a TestFlight with two Tailscale `:9119` dashboards.

PR #150, opened 2026-09-08
fix(voice): retire speech ownership at the server-replacement boundary

kaishi00:
## Summary

Runtime hardening required before saved multi-server switching (#148). **This PR does not implement issue #148 itself** — no server registry, no Keychain namespacing, no switching UI, no push/notification/approval/Kanban changes.

**The ownership bug:** Voice Conversation and Read Aloud own speech transports that are independent of `HermesClient`. A `HermesVoiceGateway` binds one dashboard bridge + base URL + profile, and each spoken response opens its own ticket-minted speech websocket (`/api/audio/speak-stream` over a bare `URLSessionWebSocketTask`). When Conduit explicitly connected to a *different* Hermes server, `connect(with:)` swapped the client — but a voice conversation mid-utterance, an in-flight speech drain, or a playing Read Aloud kept the old gateway and kept streaming against (and playing audio from) the previous server. Only `disconnect()` stopped them, and its destructive logout semantics (credential/cookie/ticket erasure) are exactly what a future server switch must not trigger.

## The boundary chosen

`AppState.prepareChatResumeForConnection(to:)` is the single authoritative *"this connection belongs to a different Hermes server"* decision — it returns `true` only when the normalized server identity changes, and `connect()` (its only production caller) invokes it before the incoming bridge/client installs. Its identity-changed branch now calls the new `retireSpeechOperationsForServerReplacement(previousIdentity:identity:)`:

- `voiceConversationController.stop()` — generation bump makes every parked continuation (capture resume, speech drain, barge-in, provider tests) inert;
- `messageReadAloudController.setGateway(nil)` — the controller's pinned Option-A semantics perform the single authoritative stop (a nil outgoing gateway means nothing can be live, since an operation cannot start without one);
- `voiceConversationController.setGateway(nil)` + `readAloudGatewayBridge = nil` — the outgoing server's bridge can never be handed to a later operation;
- `showVoiceSheet = false` — the sheet closes as if the conversation had been explicitly ended.

Read Aloud is retired **exactly once** — teardown flows through the controller's own ownership semantics rather than duplicated stop calls from `AppState`.

**Deliberately NOT logout:** no credential, cookie, ticket, Cloudflare, or preference state is touched. If the incoming activation fails, the retirement is not rolled back — the outgoing server's speech is not valid merely because the replacement failed.

## Same-server semantics (deliberate, and pinned by a control test)

Speech survives same-server reconnects, exactly as the existing recovery model expects:

- `prepareDashboardBridge(for:)` **reuses** the bridge when the normalized URL is unchanged, so gateways stay bound to a live bridge across reconnects.
- Scheduled reconnects (`reconnectForRetry`) don't even call `prepareChatResumeForConnection`; silent re-auth calls `connect()` with the same origin → identity unchanged → `false` → no retirement.
- `switchProfile` (same server) never touches the boundary.
- The retirement also covers the stale-gateway-after-signout-then-login-elsewhere hole: signing into a *different* origin after `disconnect()` (which stops speech but leaves the gateway reference) clears that reference too.
- `testSameServerIdentityCallPreservesActiveSpeech` pins the same-server behavior (including the default-port folding case); the harness's seeding call pins that a first identity call with no stored identity retires nothing.

## setGateway semantics — pinned explicitly

- `MessageReadAloudController.setGateway` (**Option A**, pre-existing behavior, now documented): any swap stops the in-flight operation. One short single-message stream fully bound to its gateway, so a swap mid-operation is always an ownership change.
- `VoiceConversationController.setGateway` (**Option B**, now documented): swaps affect future operations only. Capability refreshes deliberately install fresh-*but-equivalent* instances (same bridge/server/profile) mid-conversation, so pointer inequality is not an ownership signal; connection replacement stops the controller explicitly at the boundary. The forbidden sequence (operation on A → `setGateway(B)` → operation continues against A's server) cannot occur: the boundary stops the conversation before any new gateway is installed.

## Provider-test audit (spec §3)

`runTranscriptionTest` / `runSpeechTest` are **server-bound**: they build their gateway via `makeVoiceGateway()` (current bridge + base URL + profile) and share the controller's `operationGeneration` / `isVoiceSessionActive` / `isProviderTestRunning` ownership model. The boundary retires them identically — covered by `testServerReplacementCancelsInFlightTTSSpeechTest`. One residual window is accepted and documented: a transcription/speech HTTP request already in flight when the boundary fires still reaches the outgoing server, but its result is discarded by the post-await generation guard and nothing mutates UI, capture, or playback. That is the same class as the explicitly out-of-scope Kanban HTTP-commit concern (no await exists between the last fence and the request start, so a pre-check would be dead code; cancelling in-flight URL requests would require new gateway cancellation plumbing = scope creep).

## Tests

`ConduitTests/AppStateServerReplacementSpeechTests.swift` — 10 deterministic tests (continuation-gated mock streams, an `InterruptGate` with entry/released markers, explicit-date VAD samples, fixed route policy; no wall-clock ordering, no physical audio hardware):

1. voice conversation A→B stopped, capture retired, gateway cleared, sheet closed
2. read aloud A→B stopped, standalone ownership released, stream cancelled
3. parked assistant-speech drain continuation settles inert (no capture restart, no old-gateway reopen)
4. parked barge-in continuation released *after* the switch cannot resurrect the microphone
5. late A callbacks cannot touch a fresh B voice conversation (session-ID + not-awaiting fences)
6. late A stream failure cannot disturb a fresh B read aloud
7. playback ownership retired once per owner; a B operation claims it normally afterwards
8. replacement identity persists with no transport; a repeat same-identity call resurrects nothing (failed-activation semantics)
9. same-server control: identical + default-port-folded identities leave live speech untouched
10. in-flight TTS provider test is retired and reports failure, never success

The B-operation tests install the incoming connection's bridge first, modeling the post-replacement capability refresh — this also pins the retirement's `readAloudGatewayBridge = nil` line (removing it swaps the mock for a real A-bound gateway and fails the tests).

## Verification

Mac (ios-mac, iPhone 17 Pro simulator), focused runs:

```
@ 7111ae1:  ** TEST SUCCEEDED ** — 361 passed, 0 failed (11 regression suites + the new suite)
@ 97b356e:  ** TEST SUCCEEDED ** —  93 passed, 0 failed (follow-up re-run: the 5 voice/read-aloud suites)
```

Suites: the new `AppStateServerReplacementSpeechTests` (10) plus regression: `VoiceConversationControllerTests`, `MessageReadAloudControllerTests`, `AppStateReadAloudTests`, `VoiceAudioSessionCoordinatorTests`, `VoiceSpeakerSafeBargeInTests`, `VoiceBargeInRoutePolicyTests`, `ConnectionRepairTests`, `AppStateChatResumeTests`, `AppStateForegroundLifecycleTests`, `AppStateVoiceCapabilityTests`, `HapticsVoiceIsolationTests`.

Follow-up (`97b356e`): Read Aloud retirement collapsed to the single authoritative teardown through the controller's Option-A `setGateway` (tests now pin `stopCount` exactly), and the TTS provider-test regression test gained a test-only idempotent backstop so the regression it detects can never hang the runner.

## Review

Three-model review cycle ran before push. Fixed from review: test 5/7 post-replacement gateway-currency setup (would have failed unconditionally), test 9 setup ordering (mutual exclusion killed the voice state it asserted), direct `isGatewayAttached` assertions (new observability seam), bounded settle for test 10, hermetic route policy, `.private` log identities. Rejected with rationale: a pre-`transcribe` `isCurrent` check (no await between the fence and request start on the main actor — dead code), and stopping the voice conversation in `performSignInRequired` (pre-existing gap unrelated to server replacement — flagged as a fast-follow).

## Non-goals (unchanged)

No #148 registry/persistence/UI, no notification server identity, no approval/clarify fencing, no Kanban changes, no cookie-mirror namespacing, no Voice UI/VAD/audio-session-policy changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling when switching servers so active voice conversations, read-aloud playback, and speech tests are stopped cleanly.
  - Prevented delayed callbacks from previous server connections from affecting new speech operations.
  - Preserved ongoing speech when reconnecting to the same server, including equivalent server addresses and default ports.
  - Improved cleanup after unsuccessful server activation to ensure the replacement server remains selected.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -20,6 +20,10 @@ private let sessionYoloLog = Logger(subsystem: "com.milim.conduit", category: "S
 /// resume, and how prompt submissions were classified. Ids only — never
 /// prompt text, credentials, or message content.
 private let lifecycleLog = Logger(subsystem: "com.milim.conduit", category: "TurnLifecycle")
+/// Server-replacement speech retirement: which speech operations were live
+/// when the outgoing server's ownership was revoked. Ids and booleans only —
+/// never prompt text, credentials, or speech content.
+private let speechOwnershipLog = Logger(subsystem: "com.milim.conduit", category: "SpeechOwnership")
 
 typealias ChatResumeReconnectCancellation = @MainActor () -> Void
 typealias ChatResumeReconnectExecutor = @MainActor (ChatResumeSyncPurpose) async -> Void
@@ -2003,6 +2007,7 @@ final class AppState: ObservableObject {
         defaults.set(identity, forKey: chatResumeServerIdentityKey)
         guard let previousIdentity, previousIdentity != identity else { return false }
 
+        retireSpeechOperationsForServerReplacement(previousIdentity: previousIdentity, identity: identity)
         chatResumeCoordinator.clearResumeState()
         cancelOwnedAutomaticOperations()
         activeAutomaticChatResumeWork = nil
@@ -2041,6 +2046,51 @@ final class AppState: ObservableObject {
         conversationIdentityIndex.removeAll()
         sessionYoloStore.clearAllOverrides()
         return true
+    }
+
+    /// Runtime-only speech retirement at the server-replacement boundary.
+    ///
+    /// Voice Conversation and Read Aloud own speech transports that are
+    /// independent of `HermesClient`: a gateway binds one dashboard bridge and
+    /// base URL, and each spoken response opens its own ticket-minted speech
+    /// websocket. Those transports would otherwise keep streaming against —
+    /// and playing audio from — the outgoing server after a new connection
+    /// becomes authoritative, because replacing the client or the gateway
+    /// reference does not terminate an operation that already holds the old
+    /// gateway. This mirrors the voice teardown `disconnect()` performs, with
+    /// the operation generations doing the rest: every parked continuation
+    /// (capture resume, speech drain, playback completion) re-checks its
+    /// generation and turns inert, so nothing the outgoing server owned can
+    /// resurrect capture or playback.
+    ///
+    /// Deliberately NOT logout: no credential, cookie, ticket, or preference
+    /// state is touched here — only the outgoing connection's runtime speech
+    /// ownership. If the incoming connection fails to activate, the retired
+    /// operations stay retired; the outgoing server's speech is not valid
+    /// merely because the replacement failed. Same-server reconnects never
+    /// reach this path: `prepareDashboardBridge(for:)` keeps the bridge (and
+    /// therefore the gateways) when the normalized URL is unchanged, and the
+    /// capability refresh intentionally keeps equivalent gateways alive.
+    private func retireSpeechOperationsForServerReplacement(previousIdentity: String, identity: String) {
+        let voiceWasLive = voiceConversationController.hasLiveVoiceSession
+        let readAloudWasActive: Bool
+        if case .idle = messageReadAloudController.state {
+            readAloudWasActive = false
+        } else {
+            readAloudWasActive = true
+        }
+        speechOwnershipLog.info(
+            "Server replacement \(previousIdentity, privacy: .public) -> \(identity, privacy: .public): retiring speech ownership (voiceLive=\(voiceWasLive ? "yes" : "no", privacy: .public), readAloudActive=\(readAloudWasActive ? "yes" : "no", privacy: .public))"
+        )
+        voiceConversationController.stop()
+        messageReadAloudController.stop()
+        // A swapped gateway must never hand a later operation to the outgoing
+        // server's bridge; the incoming connection rebuilds both references
+        // from its own bridge (capability refresh, read-aloud assignment).
+        voiceConversationController.setGateway(nil)
+        messageReadAloudController.setGateway(nil)
+        readAloudGatewayBridge = nil
+        showVoiceSheet = false
     }
 
     private static func normalizedChatResumeServerIdentity(_ baseURL: String) -> String? {
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2083,12 +2083,15 @@ final class AppState: ObservableObject {
             "Server replacement \(previousIdentity, privacy: .private) -> \(identity, privacy: .private): retiring speech ownership (voiceLive=\(voiceWasLive ? "yes" : "no", privacy: .public), readAloudActive=\(readAloudWasActive ? "yes" : "no", privacy: .public))"
         )
         voiceConversationController.stop()
-        messageReadAloudController.stop()
+        // Read Aloud teardown flows through the controller's own pinned
+        // Option-A semantics: replacing a non-nil gateway performs the single
+        // authoritative stop. A nil gateway means nothing can be live — an
+        // operation cannot start without one.
+        messageReadAloudController.setGateway(nil)
         // A swapped gateway must never hand a later operation to the outgoing
         // server's bridge; the incoming connection rebuilds both references
         // from its own bridge (capability refresh, read-aloud assignment).
         voiceConversationController.setGateway(nil)
-        messageReadAloudController.setGateway(nil)
         readAloudGatewayBridge = nil
         showVoiceSheet = false
     }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -2080,7 +2080,7 @@ final class AppState: ObservableObject {
             readAloudWasActive = true
         }
         speechOwnershipLog.info(
-            "Server replacement \(previousIdentity, privacy: .public) -> \(identity, privacy: .public): retiring speech ownership (voiceLive=\(voiceWasLive ? "yes" : "no", privacy: .public), readAloudActive=\(readAloudWasActive ? "yes" : "no", privacy: .public))"
+            "Server replacement \(previousIdentity, privacy: .private) -> \(identity, privacy: .private): retiring speech ownership (voiceLive=\(voiceWasLive ? "yes" : "no", privacy: .public), readAloudActive=\(readAloudWasActive ? "yes" : "no", privacy: .public))"
         )
         voiceConversationController.stop()
         messageReadAloudController.stop()
--- a/Conduit/Voice/MessageReadAloudController.swift
+++ b/Conduit/Voice/MessageReadAloudController.swift
@@ -100,6 +100,14 @@ final class MessageReadAloudController: ObservableObject {
     /// The active stream belongs to the gateway that opened it. A replaced or
     /// cleared gateway (disconnect, profile change) invalidates the in-flight
     /// operation, so any replacement stops it.
+    ///
+    /// Pinned Option A semantics, deliberately different from
+    /// `VoiceConversationController.setGateway`: a read aloud operation is one
+    /// short single-message stream fully bound to its gateway, so a swap
+    /// mid-operation is always an ownership change and stops playback here.
+    /// The voice conversation instead survives same-server capability
+    /// refreshes that install fresh equivalent instances, so its swap is
+    /// future-operations-only and connection replacement stops it explicitly.
     func setGateway(_ gateway: VoiceGatewayService?) {
         guard activeGateway !== gateway else { return }
         activeGateway = gateway
--- a/Conduit/Voice/VoiceConversationController.swift
+++ b/Conduit/Voice/VoiceConversationController.swift
@@ -168,6 +168,12 @@ final class VoiceConversationController: ObservableObject {
         isVoiceSessionActive || isProviderTestRunning
     }
 
+    /// Observability seam: whether a gateway reference is currently installed.
+    /// The server-replacement boundary clears this reference; tests pin the
+    /// clearing through here because a stale gateway's failure mode (reaching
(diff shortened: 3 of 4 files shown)


kaishi00:
I have a complete picture. Let me write the review.

## Summary
This PR hardens the server-replacement boundary by retiring all speech ownership (Voice Conversation `stop()`, Read Aloud `stop()`, gateway/bridge reference clearing, and sheet dismissal) when `prepareChatResumeForConnection(to:)` detects a normalized server identity change, so no parked voice/read-aloud/provider-test continuation can keep streaming against the outgoing Hermes server. It adds documentation pinning the two `setGateway` semantics and 10 deterministic regression tests.

## Findings

### [minor] Redundant second `stop()` on the read-aloud playback
**File:** Conduit/Services/AppState.swift:2085-2091
**Issue:** `retireSpeechOperationsForServerReplacement` calls `messageReadAloudController.stop()` and then `messageReadAloudController.setGateway(nil)`. Because `MessageReadAloudController.setGateway` (path Option A, MessageReadAloudController.swift:111-114) invokes `stop()` itself on any swap, a non-nil outgoing gateway causes the read-aloud playback service to be `stop()`'d twice during a single retirement. The playbacks are idempotent so this is harmless today, and the test at AppStateServerReplacementSpeechTests.swift:287 deliberately asserts with `>` to absorb it — but it is an accidental double-teardown (the voice side, by contrast, is exactly `+1` at line 286).
**Suggestion:** Reorder to clear the reference before the explicit stop (i.e. `setGateway(nil)` then `stop()`), or drop the explicit `stop()` for read-aloud and rely on `setGateway(nil)` — so the retirement performs a single, deliberate teardown per owner and the `>`/ `+1` asymmetry in the tests disappears.

### [minor] `disconnect()` does not clear the Voice Conversation gateway, so the "stale-gateway" coverage is asymmetric
**File:** Conduit/Services/AppState.swift:2768-2773
**Issue:** The new boundary clears both controllers' gateways, but the "disable logout semantics" path that this mirrors (`disconnect()`) only clears the read-aloud gateway (`messageReadAloudController.setGateway(nil)` + `readAloudGatewayBridge = nil`); it never calls `voiceConversationController.setGateway(nil)`. The PR claims parity with the voice teardown `disconnect()` performs, but in fact the new code is a strict superset. That is the right behavior, but the two paths now diverge, so the stale-voice-gateway-after-signout hole may re-appear through the `disconnect()` route if speech is retired there without clearing the voice gateway at some future point.
**Suggestion:** This is acceptable as-is (the boundary is the deliberately-narrower fix). Worth a short comment or a follow-up so the divergence between `disconnect()` and `retireSpeechOperationsForServerReplacement` is intentional and documented, rather than a latent gap.

### [nit] Silent `showVoiceSheet = false` discards an in-progress conversation without user feedback
**File:** Conduit/Services/AppState.swift:2093
**Issue:** A live voice conversation (mid-listening, thinking, or speaking) is terminated and its sheet closed implicitly and without warning when the connection moves to a different server. This matches the intended boundary semantics and the PR documents it, but from a UX standpoint the user unexpectedly loses an in-flight turn with no indication beyond the sheet disappearing.
**Suggestion:** Optionally surface a lightweight notice when the retirement closes an active voice session versus an idle one (the `voiceWasLive` value is already captured at line 2075 for the log), so the transition is observable to the user. Not required for the fix.

### [nit] Log interpolation layout is dense / relies on interpolation of non-`String` into a formatted value list
**File:** Conduit/Services/AppState.swift:2082-2084
**Issue:** The `speechOwnershipLog` statement mixes `privacy:`-annotated interpolations with `.public` booleans in one long line; it's correct (both `previousIdentity`/`identity` are `String` and sanitized with `.private`) but the single 200+ char line is hard to scan and easy to regress in a follow-up.
**Suggestion:** Consider splitting the metadata into a structured `logger.info(_:privacy:)` message or multiple interpolations for readability. Cosmetic only.

## Verdict
APPROVE

The change is well-scoped, correctly placed at the single authoritative server-identity decision point (`prepareChatResumeForConnection`), runs entirely on the `@MainActor`, introduces no new closures (so no retain cycles), and introduces no force-unwraps or new concurrency hazards. The generation-bump `stop()`/re-check design correctly makes parked continuations inert across the boundary, the same-server control test pins the non-overly-broad behavior, and the test harness is deterministic (no hardware, no wall-clock ordering, continuation-gated doubles). All findings above are minor to nit and do not block the change.

<a href="https://opencode.ai/s/UEEE7UiE">New%20session%20-%202026-09-08T21%3A25%3A20.003Z</a>
[opencode session](https://opencode.ai/s/UEEE7UiE)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34280386991)

kaishi00:
I now have a full picture. The PR description's review comment findings (double read-aloud stop, hang risk) were both addressed by the follow-up commit `97b356e`. The code is `@MainActor`, uses no force unwraps, and introduces no closures/new retain cycles.

Let me consider the ordering concern: the boundary calls `voiceConversationController.stop()` (bump generation) then `setGateway(nil)`. This is correct — parked continuations check generation after the bump and go inert. `setGateway(nil)` being Option B (future-only) means nothing is live to grab the nil gateway. Good.

One residual note consistent with the PR's own documented caveat: the HTTP in-flight provider test request reaching the outgoing server (accepted scope). And the boundary is not rolled back on failed activation (deliberate, documented).

Now let me write the structured review reflecting the current HEAD state.

## Summary
This PR retires all speech ownership (Voice Conversation `stop()`, Read Aloud authoritative teardown, gateway/bridge reference clearing, and voice-sheet dismissal) at the single authoritative server-identity boundary `prepareChatResumeForConnection(to:)`, preventing parked voice/read-aloud/provider-test continuations from streaming against an outgoing Hermes server. It pins the two `setGateway` Option A/B semantics, adds an `isGatewayAttached` observability seam, and adds 10 deterministic regression tests.

## Findings

The two substantive issues raised in earlier review rounds have both already been resolved in the current HEAD (`97b356e`):
- The redundant double read-aloud `stop()` (from the explicit `messageReadAloudController.stop()` + `setGateway(nil)`) is now collapsed to a single authoritative teardown flowing through `setGateway(nil)` — verified in `AppState.swift:2088-2091` and pinned by the now-exact `XCTAssertEqual(stopCount, baseline + 1)` in the tests.
- The potential test hang on `await speechTest.value` is bounded by the idempotent backstop `cancel()` at `AppStateServerReplacementSpeechTests.swift:406`.

### [minor] In-flight provider-test HTTP request still reaches the outgoing server
**File:** Conduit/Services/AppState.swift:2051-2100
**Issue:** The retirement stops provider tests at the operation-generation/cancellation level, but a transcription/speech HTTP request already in flight when the boundary fires still reaches the outgoing server. The result is discarded by the post-await generation guard, which is safe, yet this is a deliberate, documented residual window. This is acceptable and consistent with the explicitly out-of-scope Kanban HTTP-commit concern, but it is a real (if tolerable) data-egress gap to the previous server at a server-switch boundary where the whole point is severing contact.
**Suggestion:** Accept as documented, or track as a fast-follow requiring gateway cancellation plumbing. Not blocking.

### [minor] No rollback on failed replacement activation is a subtle, asymmetric behavior
**File:** Conduit/Services/AppState.swift:2059-2060
**Issue:** If the incoming (`identity`) connection fails to activate after this boundary retires the outgoing speech, the retirement is not rolled back. The design rationale (outgoing speech is not valid merely because replacement failed) is sound and documented, but it means a failed server-switch attempt silently kills an in-progress voice/read-aloud turn with no recovery. This is a deliberate product decision rather than a defect.
**Suggestion:** Document the no-rollback policy as user-facing behavior, and consider a lightweight notice when `voiceWasLive`/`readAloudWasActive` was true for observability. Not blocking.

### [nit] Log interpolation is a dense single line
**File:** Conduit/Services/AppState.swift:2082-2084
**Issue:** The `speechOwnershipLog.info` call mixes privacy-annotated `.private` identities with `.public` booleans in one ~200-char line. It is correct (identities are `String` and sanitized), but hard to scan and easy to regress.
**Suggestion:** Split into structured metadata or multiple interpolations. Cosmetic.

No retain cycles: `retireSpeechOperationsForServerReplacement` introduces no closures beyond existing ones; it runs on `@MainActor`; the `captureEventsTask` already uses `[weak self]` at `VoiceConversationController.swift:96`. No new force unwraps or implicitly unwrapped optionals are introduced. No SwiftConcurrency/data-race hazards introduced; the boundary is single-threaded on the main actor and the generation-bump stop/re-check pattern correctly makes parked continuations inert. Accessibility/lifecycle: `showVoiceSheet = false` is a legitimate state reset on deliberate teardown, not an `onAppear` side effect; no changes to UI labels/traits/dynamic type.

## Verdict
APPROVE

The change is well-scoped to the authoritative identity-change decision, runs entirely on the main actor, adds no closures (so no retain cycles) or force unwraps, and the deterministic test harness (no hardware, no wall-clock ordering, continuation-gated doubles, bounded settle backstop) pins both the replacement and same-server-preservation semantics. Prior review findings have been resolved; the remaining items are minor/nit and non-blocking.

<a href="https://opencode.ai/s/HZSx9vbm">New%20session%20-%202026-09-08T23%3A12%3A34.643Z</a>
[opencode session](https://opencode.ai/s/HZSx9vbm)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34289492888)

PR #151, opened 2026-09-09
fix(decisions): fence native approval/clarify completions to the owning client

kaishi00:
## Summary

B2 runtime hardening required before saved multi-server switching (#148). **This PR does not implement issue #148 itself** — no server registry, no Keychain namespacing, no switching UI, and it does **not** solve push/relay server identity (that is B3, a separate prerequisite). It fences native approval/clarify response completions to the exact Hermes client that initiated them.

**The bug (from the multi-server recon):** `AppState.respondToApproval(...)` and the gateway branch of `AppState.respondToClarify(...)` were the only unfenced post-`await` state commits of their kind in `AppState`. They captured the current `HermesClient`, awaited the RPC, and then mutated decision-card/message state — but because `AppState` is an actor, a connection replacement can re-enter while the RPC is suspended. A stale continuation from outgoing client A would then mark whatever decision state is current as approved/rejected/answered/error — or surface A's failure banner — even when server B collides on every string id (`profile = default`, `session = default`, same message/request/question ids). Every other long-lived operation in `AppState` already uses this fence; these two paths simply never got it.

## The fix

Both paths snapshot ownership **before** the first re-entrant await:

```swift
let profile = activeProfile
...await client.respondToApproval/respondToClarification...
guard profile == activeProfile, self.client === client else { return }
```

- **Both continuations are fenced** — success and failure alike (stale errors must not produce "Failed to respond" state or a banner on B).
- **Stale continuations return silently.** No compensating mutations, no restoring pending, no logging of decision content. The `.submitting` state the outgoing connection set before the await is intentionally left alone — the server-change boundary (PR #150's `prepareChatResumeForConnection` + this same fence family) rebuilds outgoing state.
- **ABA is covered by pointer identity.** A1 → B → A2 (fresh `HermesClient`, same URL) does not revive A1's completion: `HermesClient.connection` is a `let` and every `self.client` write swaps the whole object, so `===` can never pass wrongly. A dedicated test pins this with URL-matched A1/A2.
- **Same-client behavior is byte-for-byte unchanged** (verified by direct pre/post diff and controls): submitting → approved/rejected by choice; failure → `.error` with the existing texts and banner; clarify success → per-question `remaining` semantics; clarify failure → error-restore; expired (4009) → teardown. Same-server navigation away still legitimately completes.
- **No cancellation plumbing** — a wire mutation that already reached server A completing there is accepted; the invariant is only that A's completion cannot mutate B's local state.

## Relay clarifies are deliberately NOT fenced

`respondToClarify` branches on `PendingDecisionPayload.relayRequestPrefix` before the client guard: relay-backed cards are answered through the notifier relay registration, not the active `HermesClient`, so client fencing is meaningless there — and fencing it on client identity would break answering from a freshly-resumed push while the gateway client reconnects. A control test proves the relay branch executes with **no** client installed and reports the relay's own outcome (never the gateway-unavailable error). Cross-server push routing and relay continuation ownership are exactly prerequisite **B3**, not this PR.

## Tests

`ConduitTests/AppStateDecisionFenceTests.swift` — 10 deterministic tests driving the **real `HermesClient` RPC** through the fake WebSocket (fakes shared with `ClarifyBatchStateTests`, promoted to internal): the operation parks between send and response, the client is replaced, then the parked continuation is released. Gates only — no sleeps.

1. stale approval success after A→B cannot mutate B's state (B rebuilds colliding ids first)
2. stale approval failure after A→B: no error card, no banner
3. stale clarify success after A→B: question untouched
4. stale clarify failure after A→B: no error, no banner
5. colliding identifiers: tests 1–4 model B state with literally identical profile/session/message/request/question ids — client identity is the only discriminator
6. same-client approval success control (approved)
7. same-client approval deny control (rejected)
8. same-client approval failure control (`.error` + banner text)
9. same-client clarify success control (answered + remaining semantics)
10. A→B→A ABA: URL-matched fresh client does not revive A1
11. relay clarify control: routes and reports without any `HermesClient` (registration cleared + restored around the test)

Also promoted `ClarifyGate`/`ClarifyFakeSocket`/`ClarifyFakeTransport` from file-private to internal (no symbol collisions; only these two suites use them).

## Verification

Mac (ios-mac, temp worktree @ `266dbfe`, iPhone 17 Pro simulator), focused run:

```
** TEST SUCCEEDED ** — 352 passed, 0 failed (10 fence tests + 342 regression)
```

Suites: the new `AppStateDecisionFenceTests` plus `ClarifyBatchStateTests` (batch clarify lifecycle through the new fences), `AppStateChatResumeTests` (resume + pending-decision promotion), `SessionIdentityContractTests` (decision routing contracts), `MessageNormalizerTests` (payload normalization), `SessionPresentationCacheTests` (decision-card presentation).

## Non-goals (unchanged)

No saved servers/registry/`activeServerID`, no Keychain namespacing or migration, no server picker, no push server identity or relay changes (B3), no cookie partitioning, no Kanban/voice changes, no generic cancellation framework, no changes to PR #150's speech-retirement behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented delayed approval and clarification responses from updating decision cards after a client or profile switch.
  - Improved reliability during reconnects, including cases where a previous connection is replaced and later restored.
  - Preserved expected behavior for responses from the currently active connection and for relay-based clarification requests.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -9506,6 +9506,13 @@ final class AppState: ObservableObject {
             return
         }
         let question = activity.questions[questionIndex]
+        // Ownership snapshot for the gateway-owned respond below — same fence
+        // as `respondToApproval`: a completion (success or failure) from a
+        // replaced client must not touch whatever decision state is current,
+        // even under colliding ids. The relay branch above is deliberately
+        // NOT fenced: it is owned by the relay registration, not by this
+        // HermesClient (push server identity is a separate prerequisite).
+        let profile = activeProfile
         // The wire answer for a multi-select question must be the array form
         // Hermes' batch parser accepts; typed custom text arrives as a bare
         // string and is wrapped here so every multi-select path is uniform.
@@ -9525,6 +9532,7 @@ final class AppState: ObservableObject {
                 // respond shape every gateway generation accepts.
                 questionId: question.isSyntheticID ? nil : question.id
             )
+            guard profile == activeProfile, self.client === client else { return }
             applyClarifyResponseOutcome(
                 outcome,
                 requestId: requestId,
@@ -9532,6 +9540,7 @@ final class AppState: ObservableObject {
                 answer: wireAnswer
             )
         } catch {
+            guard profile == activeProfile, self.client === client else { return }
             if Self.isExpiredPromptError(error) {
                 // Older gateways report expiry as RPC 4009 instead of the
                 // typed `expired` status — same teardown either way.
@@ -10899,12 +10908,23 @@ final class AppState: ObservableObject {
             cacheMessagePresentation()
             return
         }
+        // Ownership snapshot: the completion below may only land while this
+        // exact client (and profile) still owns AppState. A client/server
+        // replacement while the RPC is suspended must turn the stale
+        // continuation inert — success and failure alike — even when profile,
+        // session, and message ids collide across the two connections. This
+        // is the same fence `loadProjects`/`synchronizeTransportContinuation`
+        // use; only client ownership changed here, so no epoch beyond the
+        // pointer identity is needed.
+        let profile = activeProfile
         do {
             try await client.respondToApproval(sessionId: current.sessionId, choice: choice)
+            guard profile == activeProfile, self.client === client else { return }
             guard let updatedIndex = messages.firstIndex(where: { $0.id == messageId }) else { return }
             messages[updatedIndex].approval?.status = choice == "deny" ? .rejected : .approved
             cacheMessagePresentation()
         } catch {
+            guard profile == activeProfile, self.client === client else { return }
             guard let updatedIndex = messages.firstIndex(where: { $0.id == messageId }) else { return }
             messages[updatedIndex].approval?.status = .error
             messages[updatedIndex].approval?.choice = nil
--- a/ConduitTests/AppStateDecisionFenceTests.swift
+++ b/ConduitTests/AppStateDecisionFenceTests.swift
@@ -1,0 +1,367 @@
+//
+//  AppStateDecisionFenceTests.swift
+//  Conduit
+//
+//  Regression coverage for stale-client fencing on native Hermes decision
+//  responses: an approval/clarify respond that was sent through HermesClient
+//  A may complete on A after client B has become authoritative (the wire
+//  mutation already reached A — that is fine), but its continuation must not
+//  mutate AppState. Client identity is the fence, deliberately, so identical
+//  profile/session/message ids across two connections cannot leak state.
+//
+//  The fake socket/transport (shared with ClarifyBatchStateTests) parks the
+//  real RPC between send and response, making "operation in flight across a
+//  client replacement" deterministic without sleeps.
+//
+
+import XCTest
+@testable import Conduit
+
+@MainActor
+final class AppStateDecisionFenceTests: XCTestCase {
+
+    // MARK: - Fixtures
+
+    private func approvalFixture(status: ApprovalActivity.Status = .pending) -> ChatMessage {
+        ChatMessage(
+            id: "approval-msg",
+            role: .approval,
+            content: "Run the deploy?",
+            timestamp: "1",
+            approval: ApprovalActivity(
+                sessionId: "default",
+                command: "deploy",
+                description: "Run the deploy?",
+                choices: nil,
+                allowPermanent: false,
+                smartDenied: false,
+                status: status,
+                choice: nil,
+                error: nil
+            )
+        )
+    }
+
+    private func clarifyFixture(requestId: String = "req-gw") -> ChatMessage {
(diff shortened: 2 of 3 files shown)


kaishi00:
No collisions. My analysis is complete. Let me note one discrepancy worth flagging: the PR body asserts "because `AppState` is an actor" but it's actually a `final class: ObservableObject`. This doesn't change the fence's correctness (in fact an `ObservableObject` with `@Published` is even more vulnerable to post-await mutation, which is exactly why the fence matters), but the description is inaccurate. Now let me write the review.

## Summary
This PR fences the post-`await` completion paths of native approval (`respondToApproval`) and gateway-owned clarify (`respondToClarify`) responses so a stale continuation from a replaced `HermesClient` cannot mutate current decision/message state, even when profile/session/message/request/question ids collide across two connections. It adds 10 deterministic tests plus a relay control, and promotes four test fakes from file-private to internal.

## Findings

### [minor] PR description inaccurately calls AppState an actor
**File:** Conduit/Services/AppState.swift:402
**Issue:** The PR body states the fence is needed "because `AppState` is an actor." `AppState` is actually `final class AppState: ObservableObject` with `@Published` properties (line 402). This is not a code defect — the fence is arguably even more important precisely because `AppState` is not actor-isolated, so re-entrant mutation is unsynchronized — but the description misstates the concurrency model and could mislead reviewers / future maintainers relying on it.
**Suggestion:** Correct the PR description to say AppState is a `@MainActor` ObservableObject (or similar), not an actor.

### [nit] Relay decisions are intentionally unfenced — worth an explicit comment near the relay branch (already documented)
**File:** Conduit/Services/AppState.swift:9509-9515
**Issue:** Not a bug. The fence comment is placed *after* the relay branch (`return` at 9498), explaining why the relay path is unfenced. Since the comment is a few lines below the guarded `guard let client` and the relay check, a future edit could reorder or overlook it.
**Suggestion:** Non-blocking; the rationale is already captured adjacent to the code. No action required.

### [minor] Success-path fences could mask real post-await state loss on same client
**File:** Conduit/Services/AppState.swift:9535, 10922
**Issue:** The fence guards on `self.client === client`. If a user navigates away and the client is *torn down* (set to `nil`/replaced) rather than swapped, the completion returns silently without restoring the `.submitting` state it set pre-await. The PR states `.submitting` is deliberately left for the server-change boundary (`prepareChatResumeForConnection`) to rebuild, and same-server navigation away legitimately completes. This is documented, but a corollary is that if a user navigates away and back to the *same* client object, an in-flight completion will land and mutate state after navigation. This matches the stated intent (only client ownership, no epoch beyond pointer identity), so it is consistent, not a regression.
**Suggestion:** Acceptable as designed; consider a follow-up note that navigation-back-to-same-object is a known boundary gap if it is ever meaningful (e.g. re-entering a session mid-flight). Non-blocking.

### [nit] `let profile = activeProfile` snapshot reads a `@Published private(set)` after earlier await-free reads — fine, but note ordering
**File:** Conduit/Services/AppState.swift:9515, 10919
**Issue:** The profile snapshot is taken at different points in the two functions relative to the await: in `respondToApproval` it is taken after `guard let client else`, and in `respondToClarify` after `let question =`. In `respondToClarify`, `question`/`questionIndex` were resolved before the await and are not re-validated against `messages` after the fence passes (they are used via `applyClarifyResponseOutcome`, which re-finds the message by `requestId`). This is the pre-existing pattern and the fence correctly prevents cross-client mutation.
**Suggestion:** No change needed; noting for completeness.

### [nit] Test fake promotion scope is correct
**File:** ConduitTests/ClarifyBatchStateTests.swift:1149,1205,1239
**Issue:** Promoting `ClarifyGate`/`ClarifyFakeSocket`/`ClarifyFakeTransport` from `private` to `internal` was verified to have no symbol collisions across the test target. This is safe and reasonable for reuse by `AppStateDecisionFenceTests`.
**Suggestion:** No action.

## Verdict
APPROVE

The fence logic is consistent with the established "`self.client === client` + profile snapshot" pattern used throughout `AppState` (`loadProjects`, `synchronizeTransportContinuation`, transport continuations at lines 7242/7262/7312, etc.), correctness holds under ABA because `HermesClient.connection` is a `let` and every `self.client` write swaps the whole object, both success and failure paths are fenced, and the tests deterministically gate the real RPC across a client replacement. No retain cycles, force unwraps, or new concurrency hazards introduced. The only substantive comment is a misstatement in the PR description (actor vs. ObservableObject), which does not affect the code.

No `critical` or blocking `major` findings.

<a href="https://opencode.ai/s/QuVOlBL4">New%20session%20-%202026-09-09T00%3A36%3A22.533Z</a>
[opencode session](https://opencode.ai/s/QuVOlBL4)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34295641165)

PR #152, opened 2026-09-09
ci: parallel per-class UI test shards with targeted retry

kaishi00:
## Problem

The UI lane was the dominant PR bottleneck. On a **green** run (34295641307, PR #151), CI spent **30m34s** in the single `UI tests` job while every unit lane finished in ≤ 8 minutes, and the whole run took **~37.5 minutes**. The architecture made the tail worse: all six `ConduitUITests` classes ran in **one opaque `xcodebuild test-without-building` invocation** under a single suite-level watchdog of **2861s (~48 min)** and a 65-minute outer job ceiling, with suite-wide native retry (`-retry-tests-on-failure -test-iterations 2`). One flaky or hung UI test could hold the entire lane hostage for the better part of an hour before diagnostics appeared.

## Root cause

UI coverage was concentrated into one serial execution unit with suite-level timeout and retry semantics, so unrelated tests shared the cost of a single failure, and failure attribution (which class hung?) required an expensive post-timeout isolation pass over the whole suite.

## Fix

This PR applies the same sharding philosophy the unit lanes already use:

- **Parallel UI shards.** `scripts/plan-tests.py` now balances UI classes into their own dynamic lanes with the existing longest-processing-time-first algorithm over the per-class timing history. Lane count is derived, not hard-coded: `clamp(ceil(total_predicted / 480s), 3, 4)`, capped by class count. Today's ~22 min of cumulative UI work lands on **3 shards** with **5% predicted imbalance**.
- **Per-class `xcodebuild` execution.** Within a shard, every UI class runs as its **own** `test-without-building` invocation from the same shared build products — no rebuilds. Hangs are attributable to exactly one class, each invocation gets a per-class log and result bundle (`class-<class>-a<N>…`), and one broken XCTest process can no longer obscure the rest of the lane. The simulator is booted once per shard so only the first class pays the cold-boot overhead.
- **Per-class watchdogs.** Planned per class from timing history: `max(420s, ceil(estimate × 3.0))`. The floor carries the fixed hosted-runner invocation overhead (~2.5 min before the first test + ~1 min finalizing, measured; see run #500 history for why it must not be lower); the 3× multiplier gives slow-but-legitimate classes headroom. Estimates come from the EWMA timing history (outlier-clamped), so one anomalous run cannot inflate a budget. New/unseen classes bootstrap at the conservative 20s default estimate.
- **Targeted one-shot retry.** A failed class is retried **once**, and only that class. A retry pass is reported as a **runner-level flake** (`retried_classes` in the lane result + FLAKE WARNING in the step summary; both attempt bundles kept) — never silently converted into a clean pass. A retry that rescues an **infrastructure** wedge (exit nonzero, zero failing tests) is reported separately (`infra_recovered_classes`) so honest "every test passed on its first attempt" reporting is preserved. A second timeout names the hung class (`hung_class`), fails the lane, and records later classes as `not_diagnosed` instead of running them on a contaminated simulator. Unclassifiable results still fail the lane without retry (unchanged policy). Unit lanes are untouched: same batched invocation, same native retry, same isolation behavior (their runner flow only moved into a function).
- **Timing history stays per class.** Per-attempt extractions are folded up by a new `merge-parts` subcommand (last attempt per class wins, so a green lane records passing-attempt durations); `update-timing-history.py` and the main-only EWMA cache job are unchanged and now naturally receive per-class UI timings, which is what lets the planner re-balance shards from real runtimes.

### Before / after

```text
Before (measured, run 34295641307):
  UI wall time:            30m34s (healthy/green run)
  Total PR CI:             ~37.5 min
  Cumulative UI work:      1310.4s predicted (~21.8 min), 6 classes, 1 job
  Worst-case hang detection: up to 2861s (~48 min) suite-level watchdog
  Retry scope:             entire suite re-iterated inside one invocation
  Failure attribution:     post-timeout isolation pass over all classes

After (planner output for the same estimates, pinned by tests):
  UI shards:               3 parallel lanes, predicted 426.4-448.4s each (5.0% imbalance)
                             ui-1: ConnectionSetupTestConnectionUITests (999s watchdog)
                                   + SelectionObserverUITests (420s)
                             ui-2: ConnectionSetupSettingsUITests (891s)
                                   + LoginKeyboardUITests (420s)
                             ui-3: ConnectionRepairUITests (698s)
                                   + ConnectionSetupUITests (648s)
  Per-class timeout:       max(420s, 3 x estimate) - hang detection bounded at
                           7 min for most classes, ~16.6 min for the slowest
  Retry scope:             failed class only, exactly once
  Hung-class attribution:  immediate (per-class watchdog), no isolation pass
  Healthy UI wall clock:   governed by the slowest shard (~7.5 min of tests)
                           + per-class invocation overhead and job setup
```

### Measured on this branch (run 34310855569)

First attempt — all three shards ran concurrently from 04:31:41Z:

```text
Plan & validate   4m30s    success
Build products    2m17s    success
UI - ui-1        13m20s    success  (TestConnection + SelectionObserver)
UI - ui-2        14m58s    success  (Settings + LoginKeyboard)
UI - ui-3        25m10s    FAILED   (Repair ok; Setup flaked; includes its
                                     one targeted class retry)
Units (4 lanes)   <=14m    success  (one lane queued ~5m by the hosted
                                     macOS concurrency cap)
```

- **Healthy UI wall clock: 14m58s** (slowest shard) vs **30m34s** before - the
  actual UI test work in the slowest shard is ~7.5 min, right in the target
  range; the rest is the per-class invocation overhead this architecture
  deliberately pays (automation-session startup, app install, xcresult
  finalization per class, plus artifact download/setup) so hangs are
  attributable per class.
- **Flake containment, demonstrated live:** ui-3's
  `ConnectionSetupUITests.testWizardAdvancesThroughQuestionsToTailscaleBranchAndBack`
  ("Copying must confirm to the user") failed twice on the first attempt. The
  failing test and assertion were identified from the shard log in minutes,
  ui-1/ui-2 were untouched and already green, and the `--failed` rerun
  replayed **only ui-3** (21m02s, success). Under the old architecture the
  same flake would have cost a suite-wide retry inside the 48-minute
  watchdog lane. The test passes on main (run 34298383219) and passed on the
  rerun, i.e. a transient exposed by the restructure, not a product
  regression from this PR (no app/test code changes).
- **Timing drift made visible:** ConnectionRepairUITests took 530s actual vs
  its 232.5s stale estimate (2.3x) and still finished within its 751s
  per-class watchdog. Per-class observations now feed the EWMA history, so
  budgets and balance self-correct after merge.

## Safety

- **Product behavior is unchanged.** The diff touches only `.github/workflows/ci.yml`, `docs/CI.md`, and `scripts/**` (CI tooling + tests). No Swift sources, no `project.yml`, no app behavior.
- **UI coverage was not removed.** All six UI test classes still run on every PR — partitioned across shards, each exactly once (planner validation fails on any gap/duplicate).
- **Shared build products are still reused.** Every UI shard consumes the same `build-products` artifact via `test-without-building`; nothing is rebuilt per class or per shard, and the `.xctestrun` portability audit is unchanged.
- **Unit-test behavior is unchanged.** The unit lane body is byte-identical (wrapped in a function); all pre-existing unit state-machine cases pass unchanged.
- Watchdogs are **more** aggressive, not looser: the old 48-minute suite watchdog is replaced by per-class budgets; job ceilings remain emergency-only bounds that cover the worst in-script recovery path.

## Verification

- `python3 scripts/plan-tests.py validate` clean on the real tree (99 unit + 6 UI classes; 3 balanced UI shards).
- `python3 -m unittest discover -s scripts/tests`: **99 tests pass** (planner, merge/aggregate, EWMA, workflow contract, mocked extraction-failure path; the two bash-wrapper tests skip on Windows by design and were run directly).
- Full bash state-machine suite (`scripts/tests/test_lane_runner.sh`) run on **macOS/bash-3.2** (the lane runner's real target): **94 assertions, all passing**, including new UI cases: all-pass (one invocation per class, no native retry flags), flaky-recovery (exactly one retried class, others never re-run, bundles preserved, merge produces no phantom failures), double-failure (lane fails, later classes still run), double-hang (culprit named, later classes `not_diagnosed`), infra-recovery, unclassified (no green retry), and per-class budget isolation (a 2s-budget class killed while a 300s-budget class runs).
- Workflow contract tests pin the UI matrix shape (`fail-fast: false`, planner-driven `ui-matrix`, `--class-timeouts`, and no suite-wide `--iterations`).
- Multi-model review gate (3 reviewers × 2 rounds + a delta round): all findings fixed; final delta verdict APPROVE. Notable fixes from review: merged failures are per-class state (no phantom failures on green lanes), nan/inf timing rejection, class-count-scaled job ceilings, and infra-recovery reporting/bundle preservation.

## Summary by CodeRabbit

* **CI Improvements**
  * UI tests now run in dynamically balanced parallel lanes with class-level timeouts and targeted retries.
  * CI continues unaffected UI tests after persistent infrastructure failures and clearly identifies unrecoverable or undiagnosed classes.
  * Per-class results, timings, artifacts, and infrastructure recovery details improve troubleshooting.
  * Polling intervals can be configured for more responsive test execution.
  * Invalid timing data is rejected during test planning.

* **Documentation**
  * Updated CI documentation explains UI lane behavior, failure handling, timeouts, artifacts, and reporting.

## Hardening pass (post-architecture-review, same PR)

- **Persistent infrastructure failures no longer suppress the shard.** A UI
  class that exits nonzero twice with zero failing tests (no hang, results
  classifiable) is recorded as a `persistent_infra_classes` entry and fails
  the lane, but the remaining classes still run after a clean simulator
  reset. Stops remain for: double watchdog timeout (hang), unclassifiable
  results, and any erase/reboot recovery that cannot be trusted
  (`reset_and_boot_simulator` now returns whether an erase-gated recovery
  completed).
- **`plan-tests.py` is the single authority for UI watchdog budgets.** The
  duplicated runner-side floor/multiplier fallback (and its workflow env
  knobs) is removed; the runner fails fast unless `--class-timeouts` covers
  every assigned class.
- Trivial debt: unused local dropped from `ci-lib.sh`, stale comments
  swept, docs updated. No application Swift code, no test implementations,
  no unit-lane, balancing, flake-reporting, or CI-Gate changes.

Verified: 100 Python tooling tests, 112-assertion bash state machine
(including new persistent-infra-continue, recovery-failure-stop, and
watchdog-table-rejection cases) on macOS/bash 3.2, planner validation with
3 balanced lanes and full watchdog coverage.


kaishi00:
Now let me analyze a potential bug I noticed: in `run_ui_lane`, the `ui_budget_for` for a retried class — after a timeout (status1=124) the sim is erased/rebooted and re-run. But there's a subtle issue with the **`estimate_for` return value being used in `ui_budget_for` even when planner provided class_timeouts** — no, planned takes precedence.

Let me focus on a genuine correctness concern. In `run_ui_lane`, when `status1` is non-zero non-124 and `FAIL_COUNT1==0`: this is treated as "infra-error". But `FAIL_COUNT1==0` means extraction ran but found zero failing tests. But what if the extraction **failed** (schema) making `FAIL_COUNT1=-1`? That's handled separately (fails lane). And what if extraction returned 0 failures BUT the exit code was non-zero because the process was killed by infrastructure? Correctly classified. Good.

Now, let me look at an actual concern worth surfacing. Consider **`ui_job_timeout_min` uses `lane_timeout_s` = sum of per-class budgets, and `2 * lane_timeout_s`**. But the worst-case per class is attempt1 + attempt2 twice, each up to its own budget = sum over classes of 2×budget_i = 2×sum = 2×lane_timeout. Correct.

Let me check the **`ISOLATION_BUDGET_S` default = TIMEOUT_S** for unit lanes, and the unit lane's `run_isolation` — unchanged behavior (PR claims byte-identical). OK.

Let me look for a subtle issue in the retry classification at line 647-660. When `status1` = 124: it erases & reboots. When `status1` = non-124 infra (FAIL_COUNT1==0): erases & reboots. When test-failures: no erase. This is right.

Now a genuine concern: **Test-failure retry without simulator reset.** If attempt 1 had actual failing tests (not infra), the retry runs in the same simulator without reset. That's the intended "targeted retry" - fine.

Let me now consider a real bug: **`record_attempt` and `serialize_attempts` for class-level failures.** Wait, when a retried class's attempt 2 has test failures and `LANE_FAILED=1`, but subsequent classes run. At end, `finish_lane "fail"`. Good.

Let me look at whether there's a **deadline accounting issue once the whole lane's retry of a class is done — if a class's watchdog exceeds the JOB timeout-minutes**. The job ceiling is calculated in the planner to cover worst case, so if the per-class budget sequencing is within sum, it's covered.

Now, one thing I want to verify: **does the UI lane boot the simulator but never call `wait_for_destination_device`?** In run_ui_lane: `bounded_run 60 xcrun simctl shutdown all; reset_and_boot_simulator 0`. `reset_and_boot_simulator 0` resolves UDID, boots, bootstatus. Good. But there's no `wait_for_destination_device` gate. Historically unit lanes don't need it because build job... Actually the build job doesn't call wait_for_destination_device either. It's only referenced in ci-lib, maybe unused now. Not in scope of this PR concern.

Let me look at the docs/CI.md to confirm the claimed semantics, and check for the "flat vs per-class recovery" logic. Actually, I have enough to assess.

Let me identify concrete findings. Given the extensive testing, I'll focus on:

1. **A potential minor issue: `estimate_for`/`count_failures` use of `$RESULT_DIR` during retried class with `FAIL_COUNT2` reading the `parts/detail-$cls-a2.json` but the obs file path uses `parts` dir while det uses `parts` too.** Fine.

2. Let me double check the **retried class case count in merged observations** — a retried (a1 fail, a2 pass) class: obs-a1 has the duration (last attempt a1), obs-a2 overwrites (last attempt a2). a2 is passing duration. Good, because partition sorted by (class, attempt), a2 processed last. Good.

Let me check a real subtle bug I want to confirm: **`_attempt_index` regex `-a(\d+)\.json$` vs observation filename `observations-Foo-a1.json`.** The stem removal regex in `_part_sort_key`/`_part_class` is `-a\d+\.json$`. For `observations-Foo-a1.json`, stem = `observations-Foo`, attempt=1. `_part_class` → strip `^(observations|detail)-` → `Foo`. Good. But what about a class named e.g. `observations-A`? No. Class names are Swift test classes; fine.

Now here's a genuine edge: **a detail part filename `detail-Foo-a1.json` vs a class literally starting with `observations-` or `detail-`** — not realistic.

Let me settle on findings. Given the extensive review already applied and passing tests, the code is high quality. Let me identify a few genuine, defensible issues:

**Finding A (minor):** In `run_ui_lane`, the per-class attempt-2 retry of an **infrastructure failure** reuses the *same* `budget` (`ui_budget_for`), but the erase+reboot + reuse of the full per-class budget means the retry gets the full budget again — fine, that's intended.

**Finding B (minor):** Honestly the code is quite solid. Let me look for a truly defensible finding about the `not_diagnosed` for the *hung* class's attempt-2 timeout. When a class hangs twice, `mark_remaining_not_diagnosed` marks classes *after* it. But the failing class itself is recorded as `class-retry/2/timeout`. Good.

**Finding C:** Let me reconsider the classifier at line 613: `if [ "$status1" -ne 0 ] && [ "$status1" -ne 124 ]`. The `else` at 624 assigns `FAIL_COUNT1=""`. But what about when `status1` is, say, 137 (SIGKILL propagation)? That's non-zero non-124 → goes into classification branch → extract → count_failures. If deaths caused no failures recorded (0), treated as infra. OK.

Let me check one more thing: **the `HUNG_CLASS` report and lane-result `hung_class` — in the double-hang path, `finish_lane "timeout"` is called and `HUNG_CLASS` set. But `finish_lane` writes lane-result with hung_class. Good.**

Let me look at the docs to double check a claim about retries preserving bundles. Actually let me note the more substantive issue: **when a class's attempt 1 has test-failures and attempt 2 passes, the a1 bundle is kept (grep in RETRIED_LINES) — but `extract_bundle` for a2 wrote obs-a2. On retry-pass path `continue`, the a1 bundle also remains until finish_lane cleanup which keeps it. Good.**

Hmm, wait — there's actually a subtle bug. In the **infra-recovered** path (attempt 1 infra-error, attempt 2 pass), the code writes `cls >> INFRA_RECOVERED_LINES`, and in `finish_lane` the cleanup checks `INFRA_RECOVERED_LINES` for keeping the bundle. Good.

But consider the **retry-pass of a class that had a test-failure on a1** — `RETRIED_LINES` gets the class. But the *observations* merged uses a2 (passing). Good for timing.

Now, let me identify a real correctness gap. **Consider a UI class whose attempt 1 hits an infra-error (FAIL_COUNT1==0), and the retry ALSO fails with 0 failures (FAIL_COUNT2==0, non-124 non-timeout) → line 712 `infra-error` → `mark_remaining_not_diagnosed` → `finish_lane "error"`.** This stops the whole lane even though remaining unrelated classes could still run. For unit lanes, double infra failure fails the lane too (line 585). Consistent policy. But for UI, the *intent* claimed is "remaining classes still run" for test failures but not for infra. This is a documented policy choice. Not a bug.

I think the strongest findings here are minor. Let me also verify there are no Swift files in the diff — confirmed, only workflow/scripts/docs.

Let me produce the review. Given no Swift, I'll note the Swift focus areas don't apply and evaluate the bash/python CI logic. I'll give an APPROVE with a couple of minor/nit findings.

Let me verify one concrete concern about **`merge_observation_parts` case-count attribution and the retried class double-count in `counts.cases`** — since last-wins per class, no double count. Good.

Let me also double-check: `merge_parts` returns EXIT_SCHEMA (3) if the parts dir has no obs AND no det files, but `finish_lane` treats a non-zero merge-parts as a *warning* (line 352-353: `|| echo warning`), so it gracefully falls back. Good — never fails the lane.

One more: **the merged observations `counts.cases` could exceed reality if a class twice** — no, last-wins.

Let me finalize. I'll flag:
- Minor: In `merge_detail_parts`, `retried` (flaky) accumulation concatenates across attempts/classes but including a retried class's a1 failures could double-report in aggregate's flaky rendering — actually aggregate reads `res["flaky"]` which is from detail merged `retried`. A retried class contributes its a1 (attempts_count>1 if a1 itself shows >1 native... but UI classes run with iterations=1 so no native retry, each test has exactly 1 attempt). So `flaky`/`retried` would be empty for UI. The FLAKE WARNING comes from `retried_classes` instead, which is separate. Fine.

Let me be honest in the review: this is a well-hardened PR. I'll APPROVE with minor notes.

Actually, let me look for one genuine potential bug more carefully — the **`--class-timeouts` budget not being applied consistently between planner floor (420) and runner fallback formula**. Consistent (both max(420, ceil(est*3))). But note the runner fallback only triggers when planner didn't provide — can't happen from CI since workflow always passes it and validates. Fine.

Let me check the **`ui_budget_for` uses `estimate_for "$1"` — but the planner supplies `class_timeouts`, so `ui_budget_for` returns planned. Good. However the retry (attempt 2) uses the SAME budget as attempt 1. For a timeout-then-retry, both get full budget — good.

Wait, I found a real minor issue: In `run_ui_lane`, after a class HUNG on attempt 2, it calls `finish_lane "timeout"` inside the loop. But before that, at line 690-692 it records the class as timeout and marks remaining as not_diagnosed. But it does NOT set `HUNG_CLASS` before... yes it does, line 689. Good.

I'm confident this is a strong PR. Let me write the review with APPROVE and a couple of minor observations.

Let me reconsider whether any finding rises to major. The one thing that gives me pause: **the double-failure `infra-error` at line 712 stops the lane and marks remaining classes not_diagnosed**, meaning a persistent infra wedge on one class silently disables remaining UI coverage (they're reported not_diagnosed, not run). The PR frames this as intentional ("a contaminated simulator cannot produce misleading secondary failures"). But for the *infra* double-fail case (not a hang), the simulator is NOT necessarily contaminated — the failure is env/infra, not a hang. So stopping all remaining classes on a non-hang infra failure loses coverage for the whole shard unnecessarily. That's a legit coverage-safety concern worth raising as minor/major. But it is documented behavior and the classes are at least reported not_diagnosed. I'll raise it as minor→ arguably major. Given a wedge is uncorrelated per class, re-running later classes post-erase could rescue coverage. I'll flag as minor with suggestion.

Let me finalize the review.</think>This PR is bash/Python CI tooling with zero Swift changes, so the Swift-specific focus areas (retain cycles, concurrency, SwiftUI) don't apply. I reviewed the workflow, lane runner, planner, and timing-extraction logic for correctness.

## Summary
Parallelizes the UI test lane into per-class `xcodebuild` shards with per-class watchdogs and a single targeted class retry, sharing build products and preserving timing history per class. The diff touches only CI tooling (`.github/workflows/ci.yml`, `scripts/**`, `docs/CI.md`).

## Findings

### [minor] Double infra-failure on one class silently disables the rest of the shard's coverage
**File:** scripts/ci-test-lane.sh:712
**Issue:** When a UI class fails both attempts with `FAIL_COUNT2 == 0` (a persistent infrastructure/runtime wedge, *not* a hang), the runner calls `mark_remaining_not_diagnosed` and `finish_lane "error"`, abandoning every remaining class in the shard as `not_diagnosed`. Unlike the double-hang case (where a contaminated simulator legitimately argues against continuing), an infra wedge is per-class/uncorrelated — there is no reason the later classes cannot be erased and re-run, and skipping them silently drops real UI coverage for the whole shard on a non-hang failure.
**Suggestion:** Mirror the test-failure double-fail path (set `LANE_FAILED=1` and `continue`) after an erase+reboot, so remaining classes still execute and the wedge is reported rather than disabling coverage. At minimum, document why a non-hang infra wedge justifies skipping the lane's remaining healthy classes.

### [minor] `estimate_for` fallback recursion in `ui_budget_for` is redundant/divides responsibility
**File:** scripts/ci-test-lane.sh:202-214
**Issue:** `ui_budget_for` reimplements the planner's watchdog formula (`max(floor, ceil(est×mult))`) as a fallback, but the CI workflow always supplies `--class-timeouts` from the planner and hard-fails if it is missing (lines 82-85). The fallback is only reachable for hand-run lanes, so the two formulas must be kept in sync manually (runner here uses `UI_CLASS_TIMEOUT_MIN_S=420`/`3.0`; the planner uses `420`/`3.0`). If they diverge, a hand-run lane's behavior silently mismatches the planned contract, and the duplication isn't pinned by the existing workflow-contract test.
**Suggestion:** Either delete the fallback and require `--class-timeouts` unconditionally for `--kind ui`, or add a test that pins `ui_budget_for`'s default floor/multiplier to the planner constants so the two cannot drift.

### [nit] `already =` unused / dead local in `run_with_deadline`
**File:** scripts/ci-lib.sh:67
**Issue:** The diff adds `local ... marker` among the locals but `marker` is never assigned or used, and the pre-existing `grace`/`remaining` reset logic re-reads the poll timeout from the env on every iteration. Minor dead-code attribution for the poll-cadence feature.
**Suggestion:** Drop the unused `marker` local to keep the function tidy.

### [nit] Retried-class case counts are last-wins but `attempts`/`retried` accumulate across all parts
**File:** scripts/extract-test-timings.py:365-396
**Issue:** `merge_detail_parts` concatenates `attempts` and `retried` across every attempt part of a class while making `failures` last-part-wins. For a UI class retried from a test failure, the merged `attempts` retains the a1 failed attempt alongside the a2 pass — this is intentional for flake reporting, but it means the merged detail carries conflicting attempt chains for the same class that downstream consumers must reconcile. Consistent, but worth an explicit note in the docstring that `attempts` is an append log while `failures` is state.
**Suggestion:** Document this asymmetry (it is intentional) so future maintenance doesn't "fix" it into a double-report bug.

## Verdict
APPROVE

The change is well-designed and unusually well-tested (99 Python + 94 bash assertions, contract tests pinning the matrix shape and CLI contract). Sharding policy, per-class watchdog budgets, hang attribution, and basket preservation are sound, and timing extraction is correctly kept best-effort so it can never fail a lane. The findings above are minor; none block merge.

<a href="https://opencode.ai/s/FcxkZ4vW">New%20session%20-%202026-09-09T04%3A25%3A45.225Z</a>
[opencode session](https://opencode.ai/s/FcxkZ4vW)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34310855562)

kaishi00:
Review complete. VERDICT: **APPROVE** — two minor findings and two nits, all report-fidelity/cosmetic, none blocking.

<a href="https://opencode.ai/s/N1owREWi">New%20session%20-%202026-09-09T09%3A06%3A56.941Z</a>
[opencode session](https://opencode.ai/s/N1owREWi)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34332765731)

PR #153, opened 2026-09-09
test: stabilize copy confirmation and CI UI launches

kaishi00:
## Stack dependency

**Stacked on #152. Do not merge this PR before #152.**

The branch is exactly one commit (`a484f15`) on top of #152 head `4418384`. It targets `main` temporarily so the normal PR CI workflow runs; after #152 merges, this diff collapses to the stabilization commit only.

## Problems

Two recurring failures surfaced once #152 made UI flakes attributable by class/test:

1. `ConnectionSetupUITests.testWizardAdvancesThroughQuestionsToTailscaleBranchAndBack` races the intentionally transient 2-second **Copied** label. On hosted runners XCTest can spend longer than that in post-tap idle/snapshot work before it starts querying the confirmation.
2. `ConnectionRepairUITests` is being blamed for a failure that occurs before Repair logic runs. The failing artifacts show the first `XCUIApplication.launch()` timing out in Xcode's debugger-backed simulator launcher (`IDELaunchiPhoneSimulatorLauncher` / `DebuggerVersionStore`). After a clean retry, the first launch can fail again while a subsequent launch on the same warmed simulator succeeds. Deleting or weakening the named Repair test would just move the failure to whichever test launches first.

### Durable accessibility confirmation, unchanged visual UX

`AskHermesPromptView` still shows the visual **Copied** label for exactly 2 seconds. It now also keeps the successful copy as durable accessibility state on the Copy Prompt button (`accessibilityValue == "Copied"` for the prompt actually copied).

The UI test asserts that semantic state with `XCTNSPredicateExpectation` instead of racing the transient label.

This does **not** lengthen the toast or add sleeps.

### CI-only no-debug test scheme

Adds a separate shared `ConduitCI` scheme whose test action has `debugEnabled: false` and makes `ci-build-for-testing.sh` use it by default (with `SCHEME` still overridable).

The normal local `Conduit` scheme is unchanged, so developer Cmd+U/debugger behavior stays intact. Hosted XCUITests do not need LLDB attached; the experiment removes Xcode 26.x's debugger-backed simulator-launch path that is producing the observed `Timed out while launching application via Xcode` failures.

No Repair assertions or coverage are removed or relaxed.

## Evidence behind the Repair diagnosis

From #152 run `34332769496`, `ui-3`:

- attempt 1: the first Repair test fails at `app.launch()` with `Timed out while launching application via Xcode`; a later launch wedges until the class watchdog;
- attempt 2 after simulator erase: the first Repair launch fails the same way again;
- later Repair launches on that same warmed simulator proceed, including one that emits a DTX signal warning and still passes;
- extracted result: 5 Repair tests, only the first test failed, at launch time before the Repair assertions ran.

That makes broad DTX/debugger-warning matching unsafe and makes deleting the first test ineffective. The no-debug CI scheme is intentionally a narrow source-level experiment.

## Guardrail

Adds a stdlib CI contract test pinning:

- `ConduitCI` exists;
- its test action has `debugEnabled: false`;
- it contains both unit/UI test targets;
- `ci-build-for-testing.sh` defaults to `ConduitCI` but remains overridable;
- the normal local scheme is not replaced.

## Validation target

The PR CI run should prove:

1. XcodeGen accepts/generates `ConduitCI` and build-for-testing succeeds.
2. Existing unit/UI coverage still executes through the shared `.xctestrun` artifact.
3. The Copy Prompt test no longer races the 2-second label.
4. `ConnectionRepairUITests` no longer hits the debugger-backed Xcode app-launch timeout.

If #4 still reproduces with debugging disabled, this experiment should be reverted rather than papered over; the next fallback would be bounded per-test XCTest execution time, not weaker Repair assertions or larger class watchdogs.

## Summary by CodeRabbit

* **Accessibility**
  * Copy Prompt now provides a persistent “Copied” status to assistive technologies, accurately reflecting the currently displayed prompt.

* **Bug Fixes**
  * Prevented reused prompt views from incorrectly reporting that a different prompt was copied.

* **Tests**
  * Updated automated coverage to verify the persistent copy status and CI build configuration.

* **Chores**
  * Improved continuous integration builds with a dedicated, configurable build scheme.

### DIFF
--- a/Conduit/Views/AskHermesPromptView.swift
+++ b/Conduit/Views/AskHermesPromptView.swift
@@ -16,6 +16,11 @@ struct AskHermesPromptView: View {
 
     @State private var copied = false
     @State private var copyCount = 0
+    // Keep the successful copy available as durable accessibility state even
+    // after the intentionally short visual confirmation disappears. Tying it
+    // to the prompt prevents a reused view from claiming a different prompt
+    // was copied.
+    @State private var lastCopiedPrompt: String?
 
     var body: some View {
         VStack(alignment: .leading, spacing: 12) {
@@ -33,12 +38,14 @@ struct AskHermesPromptView: View {
                 Button {
                     UIPasteboard.general.string = prompt
                     copied = true
+                    lastCopiedPrompt = prompt
                     copyCount += 1
                 } label: {
                     Label("Copy Prompt", systemImage: copied ? "checkmark" : "doc.on.doc")
                         .font(.footnote.weight(.semibold))
                 }
                 .accessibilityIdentifier("setup.copy-prompt")
+                .accessibilityValue(Text(lastCopiedPrompt == prompt ? "Copied" : ""))
 
                 if copied {
                     Text("Copied")
--- a/ConduitUITests/ConnectionSetupUITests.swift
+++ b/ConduitUITests/ConnectionSetupUITests.swift
@@ -172,9 +172,18 @@ final class ConnectionSetupUITests: XCTestCase {
         XCTAssertTrue(copyPrompt.waitForExistence(timeout: 5), "Copy Prompt must appear on the No path")
         if !copyPrompt.isHittable { app.swipeUp() }
         copyPrompt.tap()
-        XCTAssertTrue(
-            app.staticTexts["setup.copied-confirmation"].waitForExistence(timeout: 3),
-            "Copying must confirm to the user"
+        // The visual "Copied" label is intentionally transient (2s), which
+        // is shorter than XCTest's post-tap idle/snapshot work on hosted
+        // runners. Assert the button's durable accessibility state instead:
+        // it proves the copy action happened without racing the visual toast.
+        let copied = XCTNSPredicateExpectation(
+            predicate: NSPredicate(format: "value == %@", "Copied"),
+            object: copyPrompt
+        )
+        XCTAssertEqual(
+            XCTWaiter.wait(for: [copied], timeout: 3),
+            .completed,
+            "Copying must confirm to accessibility after the visual toast expires"
         )
 
         let continueButton = app.buttons["setup.continue"]
--- a/project.yml
+++ b/project.yml
@@ -59,3 +59,23 @@ targets:
         GENERATE_INFOPLIST_FILE: YES
     dependencies:
       - target: Conduit
+
+# CI gets its own shared scheme so local Cmd+U / debugger behavior stays
+# unchanged. UI tests do not need LLDB attached in hosted CI, and Xcode 26.x's
+# debugger-backed simulator launcher is the source of the intermittent
+# DebuggerVersionStore / "Timed out while launching application via Xcode"
+# failures seen on GitHub runners. PosixSpawn keeps the same tests and app
+# binary while removing that debugger-only launch path.
+schemes:
+  ConduitCI:
+    build:
+      targets:
+        Conduit: all
+        ConduitTests: [test]
+        ConduitUITests: [test]
+    test:
+      config: Debug
+      debugEnabled: false
+      targets:
+        - ConduitTests
+        - ConduitUITests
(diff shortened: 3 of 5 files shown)


kaishi00:
I have all the context I need. Let me write the review.

## Summary
This PR stabilizes two flaky CI UI failures: it makes copy confirmation assertable via durable accessibility state (instead of racing the 2-second "Copied" toast), and adds a debugger-disabled `ConduitCI` Xcode scheme used by CI's build-for-testing to avoid Xcode 26's flaky debugger-backed simulator-launch path, with a Python contract test as a guardrail.

## Findings

### [minor] UI test no longer verifies the visual "Copied" toast
**File:** ConduitUITests/ConnectionSetupUITests.swift:175
**Issue:** The test previously asserted `staticTexts["setup.copied-confirmation"]` (the transient visual toast) appeared after tapping Copy Prompt. That assertion is now replaced entirely by the accessibility-value predicate, so the rendered toast `Label`/`Text` is no longer verified by any UI test (only indirectly by unit tests of `runCopiedConfirmation`'s cancellation logic, not the rendered view). The toast is a user-facing confirmation, so dropping all UI-level verification of it is a small coverage regression.
**Suggestion:** Acceptable as an intentional trade-off given the documented race, but consider a follow-up that verifies the toast appears within a short window on top of the durable-state assertion (e.g., a brief/sub-second check) so the visual confirmation stays covered without reintroducing the flake.

### [minor] Durable "Copied" accessibility state never expires and can go stale
**File:** Conduit/Views/AskHermesPromptView.swift:48
**Issue:** `lastCopiedPrompt` is set on tap and never reset. Because the value is durable, the Copy Prompt button keeps `accessibilityValue == "Copied"` indefinitely, even after the pasteboard content is replaced elsewhere or much later — so a user revisiting a copied prompt hears "Copied" inaccurately. The `lastCopiedPrompt == prompt` guard correctly handles the cross-prompt reuse case, but it does not address staleness over time. This is documented as intentional, but it's a real (if minor) accessibility-accuracy trade-off.
**Suggestion:** Acceptable by design; if desired, scope the durable value to a bounded lifetime (e.g., clear `lastCopiedPrompt` on the same `runCopiedConfirmation` tick that clears `copied`), while noting the UI test asserts within 3s so it would still pass.

### [nit] Contract test uses brittle raw string matching instead of parsing YAML
**File:** scripts/tests/test_ci_scheme_contract.py:26
**Issue:** The assertions rely on exact substring/indentation patterns (e.g., `"  ConduitCI:\n"`, `"  Conduit:\n    test:\n"`) against `project.yml`. This is whitespace-sensitive and could silently pass or fail if the spec is reformatted or reordered without semantic change (or miss a real change that happens not to match the literal strings). It still guards the core intent, but it's a weak contract.
**Suggestion:** Acceptable as a lightweight guardrail; consider parsing `project.yml` (e.g., `pyyaml`) to assert on the `schemes` structure rather than literal text, if the repo already vendors a YAML parser. Not blocking.

### [nit] Blank `accessibilityValue` modifier produces an empty value token
**File:** Conduit/Views/AskHermesPromptView.swift:48
**Issue:** When `lastCopiedPrompt != prompt`, the button gets `.accessibilityValue(Text(""))`, attaching an empty value. In some VoiceOver configurations this can add a superfluous empty value to the "Copy Prompt" button's announcement. Harmless but slightly noisy.
**Suggestion:** Conditionally apply the modifier only when the copy matches (e.g., via a computed `accessibilityValue` binding or `.accessibilityValue(lastCopiedPrompt == prompt ? "Copied" : "")`) — equivalent result, marginally cleaner. Not blocking.

No retain cycles, force unwraps, data races, or MainActor violations were introduced. The scheme change is correctly scoped to CI (`CONDUITCI` default with `SCHEME` override; local `Conduit` scheme untouched), and the downstream `test-without-building -xctestrun` path means the debugger setting is baked into the shared `.xctestrun` at build time, so the Docker-style xctestrun portability/audit pipeline is unaffected.

## Verdict
APPROVE

<a href="https://opencode.ai/s/wAPS5CBY">New%20session%20-%202026-09-09T15%3A36%3A34.209Z</a>
[opencode session](https://opencode.ai/s/wAPS5CBY)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34371227716)

PR #155, opened 2026-09-10
ci: deterministic connection-setup UI queries, batched UI shards, overhead-aware lane planning

kaishi00:
## CI reliability: deterministic connection-setup UI queries, batched UI shards, overhead-aware lane planning

Scope: **XCUITest reliability + XCTest invocation efficiency.** No application behavior changes; the only app-adjacent surface touched is how UI tests *query* existing accessibility elements (no accessibility instrumentation changed, so VoiceOver behavior is untouched).

### The concrete flaky failure

The most recent green PR run still needed a retry because
`ConnectionSetupSettingsUITests.testInteractiveSignInOutcomeFromSettingsOffersDoneAndDismissesSafely`
failed its first attempt with:

```
Failed to get matching snapshot: Timed out while evaluating UI query.
```

The failure happened while evaluating the authentication-stage-row query — right after the test had already waited for the `setup.test.interactive-ready` terminal marker.

### Root cause (measured, not assumed)

A temporary UI test dumped the full accessibility hierarchy (`app.debugDescription`) at the exact evaluation points, on the Mac, at `origin/main`:

```
Other, identifier: 'setup.test.stage.authentication', label: 'Authentication, browser sign-in required'
StaticText, identifier: 'setup.address-preview', label: 'https://conduit-uitest.example'
```

- Stage rows (an `HStack` collapsed via `.accessibilityElement(children: .ignore)`) are exposed as **`Other`** elements.
- `setup.address-preview` (a `Text`) is a **`StaticText`**.

The old helper — `app.descendants(matching: .any).matching(identifier:).firstMatch` — made every check resolve a snapshot across **every element type in the whole application tree**, and the tests evaluated it with immediate `.exists` checks. During the Review auto-advance the hierarchy is churning; a whole-tree snapshot there is what can starve past Xcode's snapshot timeout. The same broad query was then resolved again for `isHittable`, taps, and label reads.

### The fix (Phase 1)

- `ConnectionSetupSettingsUITests` / `ConnectionSetupTestConnectionUITests`: typed `stageRow()` (`app.otherElements[...]`) and `addressPreview()` (`app.staticTexts[...]`) helpers; `ConnectionRepairUITests`' two identical address-preview lookups fixed the same way.
- Positive existence checks after screen transitions use `waitForExistence(timeout: 5)`; absence assertions stay exactly where a terminal-state marker has already settled the screen (`interactive-ready` / `ready` / `failure` / `credentials-required` / Settings return). No sleeps, no new markers, no weakened assertions, no removed coverage.

**Repeated-test evidence (Mac, iPhone 17 Pro / iOS 26.5, real `test-without-building` invocations, UDID-pinned destination):**

| Suite | Executions | Result |
|---|---|---|
| `testInteractiveSignInOutcome...` (the previously failing test) | **22 consecutive invocations** | 22 pass, ~21-23 s each |
| `ConnectionSetupSettingsUITests` (full class) | 3 × | all pass (88-92 s) |
| `ConnectionSetupTestConnectionUITests` (full class, same helper) | 3 × | all pass (175-177 s) |
| `ConnectionRepairUITests` (full class, same element) | 2 × | all pass (99-101 s) |

`Timed out while evaluating UI query` appears **zero times** across all logs. (No pre-fix local reproduction was attempted — the CI failure is the "before" evidence; the "after" evidence is 30+ consecutive green executions.)

### Healthy-path XCTest invocations (Phase 2)

- **Before:** 4 unit + 6 UI = **10** `xcodebuild` invocations per healthy run (every UI class paid its own test-session startup/finalization).
- **After:** 2 unit + 3 UI = **5** invocations — each UI shard runs all its classes in **one** `test-without-building` invocation (watchdog = sum of the planner's per-class budgets), and the unit suite consolidates to 2 lanes.

**Measured end-to-end on the Mac (real `ci-test-lane.sh` UI shard, 2 classes, real products):** one batched invocation, lane wall **198 s** of which the classes themselves were 143 s — one startup, not two. On hosted runners (run 34349843495) the equivalent 2-class UI shard took 846 s with per-class invocations, and each invocation carries ~210-300 s of fixed cost (unit-lane actual-vs-predicted; the UI per-class floor measured ~3.5 min on run #500) — every dropped invocation saves that fixed cost per shard, per run.

### Failure-path guarantees

- Ordinary UI test failures retry **only the non-passing tests** (any final result ≠ Passed — including Crashed/Skipped residue from aborted runs): exact `ConduitUITests/Class/testMethod` filters derived from the xcresult, falling back to the class when only the class is identifiable. Healthy classes are never re-executed. **A passing retry is still reported as a runner-level flake** (`::warning` naming the retried tests + CI Test Report "FLAKE WARNING" + both attempt bundles kept as artifacts).
- A batch that aborts before a class runs is classified `incomplete` and goes to per-class diagnosis instead of a retry — unexecuted tests can never vanish into a green lane. The same gate guards the exit-0 path: an invocation that returns success while its parseable xcresult has no record of an assigned class diagnoses instead of passing. A green lane reached through diagnosis announces the triggering event instead of blending into a clean pass.
- Batch watchdog timeout / infrastructure wedge: erase the simulator, then per-class diagnosis — per-class watchdogs, one targeted retry, persistent-infrastructure continuation, hung-class naming and lane stop, `not_diagnosed` tail. Unclassifiable failures still fail the lane with no retry.
- Unit lanes keep native `-retry-tests-on-failure -test-iterations` (already method-precise).

### Unit lane count is now measured, not intuited

The planner models `predicted lane cost = invocation_overhead_s + sum(class durations)` and adds a lane only when it buys more than `lane_wall_tolerance_s` (120 s) of modeled wall clock. Defaults are measured: hosted lane actual-vs-predicted in run 34349843495 shows 208-295 s of fixed cost per unit lane against ~96 s of predicted execution in three of the four unit lanes. The hosted 810 s outlier class (`MarkdownRichContentHostedTests`) dominates every possible split, so lanes 3-4 bought nothing — **units go 4 → 2** at identical modeled wall; the UI suite stays at its 3-lane clamp. Bonus from the same data: unit-1 started ~7 minutes after its siblings (hosted-runner concurrency), so fewer lanes also queue less. `target_budget_s` / `ui_target_budget_s` are superseded and removed. The UI job ceiling prices the true reachable worst path (batch + full-sum retry that times out + 2× diagnosis = 4× the budget sum).

### Retry precision, visibility, artifacts, timing history

- Fallback hierarchy: exact failed test → failing class → per-class diagnosis on environment events → **never the whole UI suite** for a test failure.
- Retry passes remain visible: `retried_classes` / attempts chain in `lane-result.json`, `::warning` annotations, FLAKE WARNING rows in the CI Test Report, and both attempt bundles kept on green lanes. PR runs still never write timing history.
- Batch extractions flow through the same `merge-parts` pipeline; per-class diagnosis results supersede batch parts **per class** (wall-clock fold order — and batch failure evidence for classes diagnosis never re-executed survives, so a red lane's report stays complete); a method-filtered retry — batch **or** per-class diagnosis — never contributes its observations to timing history, so a method-only rerun can never shrink a class's duration sample; clean green-lane bundles are still pruned.

### Simulator destination pinning

`build_destination` emits `platform=iOS Simulator,id=<UDID>,arch=arm64` (SIMULATOR_ARCH-overridable; name-based fallback only if the UDID can't be resolved). Verified on hosted CI: run 34425462004 (UDID pin only) still showed one warning per invocation because every Apple Silicon simulator device also registers an x86_64-under-Rosetta candidate — but both candidates were the **same pinned UDID**, so the dangerous cross-device "first match wins" case was already gone; with the arch pin added, run 34426688686 shows **zero ambiguity warnings in all 5 unit/UI lane invocations**. The single remaining informational line is emitted once by the build step's internal destination scan (again the same UDID on both lines — the build makes no device choice).

### ConduitCI / debugger warnings (investigated, no change)

`IDELaunchParametersSnapshot: DebuggerLLDB.DebuggerVersionStore.StoreError error 0` / `no debugger version` were reproduced during **fully passing** local UI runs (exit 0) — they are Xcode 26.x launch-path noise, not related to the snapshot timeout (which was the broad-query cost, now gone). The experimental no-debug `ConduitCI` scheme exists only on the unmerged branch `origin/test/stabilize-repair-copy-ui` (never landed on main); since these warnings occur regardless, that scheme's premise shows no measurable benefit and should not be merged as-is. Nothing needed removing from main.

### Round 3 (CI run 34420978966): Plan-job timeout root cause + fixes

That run's `Plan & validate` job was cancelled at its 10-minute ceiling **before any macOS work started**. The job log shows `test_extract_timings` finishing 7 seconds in, then silence — unittest discovery order puts the lane-runner bash state-machine suite right after it, so that suite consumed the remaining ~10 minutes. No single case hung: per-case `START`/`END` instrumentation (kept in the suite) shows the cost distributed across the new batching recovery cases, each paying many `bounded_run` calls whose poll loop had a fixed 2-second floor (`kill -0` keeps succeeding on an unreaped child), plus a fixed 3 s settle per simulator reset. Measured on macOS: **6:07** for the full suite before the fix.

Fixes:

- `bounded_run` now polls at 0.2 s — identical deadline/kill semantics, no fixed floor per call. Suite on macOS: **3:42** before the two new regression cases, **4:23 with them** (134 assertions), vs 6:07 before.
- The minutes-long CI-tooling meta-suites moved out of `Plan & validate` into a dedicated **`self-test` job** (ubuntu) that runs **concurrently with the macOS build**, so they can never delay planning, the build, or risk another job's timeout. `CI Gate` now requires `--self-test` too.
- The timeout hierarchy is coherent and load-bearing: per-case synthetic-hang watchdogs (≤ 6 s) < the wrapper's subprocess cap (**480 s**; was 900 s — *larger than the old job ceiling*, the invalid contract) < the self-test job's 12-minute ceiling. The plan job skips the bash wrappers via `CONDUIT_CI_SKIP_BASH_WRAPPER_TESTS` and completes in well under a minute.
- Diagnosis method-only retries obey the batch timing rule: the retry's observations part is deleted after extraction (both pass and fail paths) so a method-only rerun can never shrink a class's duration sample; its detail part stays as flake evidence. Regression-tested at the runner level (u15/u16) and the merge level (100 s class vs 5 s method retry → 100 s).
- Batch-failure supersession in `merge_detail_parts` is per class: batch entries survive for classes diagnosis never re-executed (the `not_diagnosed` tail of a lane stopped at a hang) so a red lane's report stays complete.

The batching architecture, typed XCUITest queries, method-precise retry, per-class diagnosis, 2 unit lanes, and overhead modeling are all unchanged — the CI failure was in the harness validating those behaviors, not in the behaviors.

### Verification checklist

1. Previously failing test passes repeatedly — 22 consecutive invocations + class runs, zero failures, zero query timeouts ✔
2. Broad stage-row helper no longer duplicated — typed helpers in both suites (+ Repair's address-preview) ✔
3. No arbitrary sleeps added ✔
4. Stage-row accessibility untouched (query-side change only; VoiceOver labels/identifiers unchanged) ✔
5. Connection Setup UI coverage intact — all tests kept, assertions strengthened to waits ✔
6. Healthy UI shards use fewer Xcode invocations — 6 → 3 (one per shard), proven by a real 2-class shard in one invocation ✔
7. Failed tests isolated and retried precisely — method-level filters, class fallback, diagnosis on environment events ✔
8. Retry passes visible as flakes — annotations (now listing the retried tests), report rows, attempt bundles ✔
9. Destination ambiguity — id-pinned end-to-end; warning removal to be confirmed on CI (not locally reproducible, single runtime) ✔/⏳
10. Unit-lane count from measured startup vs execution — run 34349843495 actuals drive the 240 s default; 4 → 2 ✔
11. Planner regression tests cover the fixed invocation overhead — tolerance rule, outlier consolidation, saturation, bounds, modeled wall, ceiling (4× derivation) ✔
12. CI reporting/artifacts unchanged in usefulness — same lane-result/report/artifact shape, wording updated ✔

**Test suites after this round:** lane-runner state machine **134/134** (macOS bash 3.2, 4:23 with per-case telemetry), Python suite **107/107** (macOS; identical suite green on Windows with the 2 bash wrappers skipped), `plan-tests.py validate` green (99 unit + 6 UI classes exactly once, 2 unit lanes, 3 UI lanes, every UI class carrying an explicit watchdog), no orphaned `sleep`/`bash` processes after the suite.

**Hosted CI on this PR (final head):** run [34426688686](https://github.com/kaishi00/hermes-conduit/actions/runs/34426688686) @ `758e4aa` — every job green, first attempt. Plan & validate **9 s** (was 10m16s cancelled), self-test **4m43s** (concurrent with build), build **3m26s**, unit lanes **6m12s / 6m33s**, UI lanes **14m14s / 14m19s / 14m13s**, overall wall ≈ **21 min** (the prior head's run: 15m40s with no lane retries). Retry visibility proven on real hosted runs: ui-3 in the prior run and ui-1 + ui-3 in this run each needed their targeted batch-retry (`retried_classes` populated, FLAKE WARNING annotations emitted, first-attempt bundles kept) — the previously-failing settings lane passed clean on the first attempt in both runs.

### Round 4 (e2581e5): the two remaining first-attempt UI flakes, fixed at the synchronization contract

The `758e4aa` run was green but needed targeted retries, and both retry-rescued tests had identifiable test-side synchronization holes — fixed directly, no CI policy changes:

1. **Copy Prompt raced its own intentional ephemerality.** The visible "Copied" label lasts ~2 s, and a hosted runner can spend longer than that reaching the first hierarchy snapshot after the tap. `AskHermesPromptView` now also records `lastCopiedPrompt` (keyed to the current prompt, so a reused card can never claim a different prompt was copied) and exposes it as a **durable accessibility value** on the Copy Prompt button; the visible confirmation and its cancellation-aware window are untouched. The UI test asserts `value == "Copied"` on the button via an `XCTNSPredicateExpectation` instead of racing the transient label.
2. **The staged-test walk raced the wizard's push.** `walkToTestScreen` tapped `login.connection-setup` and immediately interacted with the first answer; it now waits for the wizard's own `setup.step-label` to read **"Step 1 of 3"** before the first interaction (polling, same contract as `ConnectionSetupUITests.stepLabel`).

**Stress (Mac, real invocations):** the two previously-flaky methods ran **20× each consecutively — 40/40 first-attempt passes** (wizard 26-27 s, auth-failure 53-54 s), plus full-class runs of both suites, zero `Timed out while evaluating UI query`.

**Final hosted run:** [34432945199](https://github.com/kaishi00/hermes-conduit/actions/runs/34432945199) @ `e2581e5` — all jobs green, and **all three UI shards passed their batch on attempt 1** (`('batch', 'passed')`, empty `retried_classes`, no "passed on its targeted retry" anywhere in the lane logs); unit lanes clean; overall wall ≈ **16 min**. CI-tooling suites unchanged and green (134/134 bash, 107/107 Python).

**Review:** three-model review, two rounds + a focused confirmation pass (all findings fixed; final verdict APPROVE WITH NITS with no open findings beyond a documented extraction-bound nuance absorbed by the ceiling margin).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,15 +3,21 @@ name: CI
 # CI v2 - timing-aware dynamic test lanes for Hermes Conduit.
 #
 # Architecture (docs/CI.md):
-#   plan     (ubuntu, cheap)  validate inventory, run planner regression
-#                             tests, balance test classes into a dynamic
-#                             number of timing-weighted lanes (unit AND UI),
-#                             emit both matrices.
-#   build    (macos, once)    build-for-testing exactly once into a
-#                             workspace-anchored DerivedData path; audit the
-#                             generated .xctestrun for absolute-path
-#                             portability; upload products as an artifact.
-#   unit/ui  (macos, matrix) test-without-building from the downloaded
+#   plan      (ubuntu, cheap)  validate inventory, balance test classes into
+#                              a dynamic number of timing-weighted lanes
+#                              (unit AND UI), emit both matrices. The
+#                              minutes-long CI-tooling meta-suites are NOT
+#                              run here - they must never delay the macOS
+#                              build or risk the job timeout.
+#   self-test (ubuntu)         CI-tooling regression suites (planner,
+#                              lane-runner state machine, destination
+#                              lookup, gate/timing contracts) as their own
+#                              job, concurrent with build.
+#   build     (macos, once)    build-for-testing exactly once into a
+#                              workspace-anchored DerivedData path; audit the
+#                              generated .xctestrun for absolute-path
+#                              portability; upload products as an artifact.
+#   unit/ui   (macos, matrix) test-without-building from the downloaded
 #                             products; measured watchdogs; native flake
 #                             retry (units); one batched invocation per UI
 #                             shard with method-precise retry of only the
@@ -70,6 +76,13 @@ jobs:
             timing-history-v1-
 
       - name: Validate inventory + planner regression tests
+        env:
+          # The minutes-long bash meta-suites (lane-runner state machine,
+          # destination lookup) run in the dedicated self-test job - they
+          # must never delay planning, the macOS build, or risk this job's
+          # timeout (run 34420978966: the suite outlived the 10-minute job
+          # ceiling and GitHub cancelled it mid-run).
+          CONDUIT_CI_SKIP_BASH_WRAPPER_TESTS: "1"
         run: |
           set -euo pipefail
           HISTORY_ARG=""
@@ -111,6 +124,26 @@ jobs:
           name: plan
           path: ci-artifacts/
           retention-days: 7
+
+  self-test:
+    name: CI tooling self-test
+    # No `needs`: the meta-suites test the CI tooling itself, not this run's
+    # plan, so they start immediately and run CONCURRENTLY with the macOS
+    # build instead of serially delaying it.
+    runs-on: ubuntu-latest
+    timeout-minutes: 12
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Run CI tooling regression suites
+        # Full discovery INCLUDING the bash state-machine suites. The
+        # timeout layering must hold: per-case synthetic-hang watchdogs
+        # (<= 6s) < the wrapper's subprocess cap (480s, test_lane_runner.py)
+        # < this job ceiling (12m) - GitHub must never be the first layer to
+        # kill a regression suite.
+        run: |
+          set -euo pipefail
+          python3 -m unittest discover -s scripts/tests -p 'test_*.py' -v
 
   build:
     name: Build test products
@@ -371,7 +404,7 @@ jobs:
   # directly - require "CI Gate" instead (see docs/CI.md).
   ci-gate:
     name: CI Gate
-    needs: [plan, build, unit, ui]
+    needs: [plan, build, unit, ui, self-test]
     if: always()
     runs-on: ubuntu-latest
     timeout-minutes: 5
@@ -384,7 +417,8 @@ jobs:
             --plan "${{ needs.plan.result }}" \
             --build "${{ needs.build.result }}" \
             --unit "${{ needs.unit.result }}" \
-            --ui "${{ needs.ui.result }}"
+            --ui "${{ needs.ui.result }}" \
+            --self-test "${{ needs.self-test.result }}"
 
   timing-history-update:
     name: Update timing history
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,9 +279,9 @@ jobs:
         # shard's classes in ONE batched xcodebuild invocation (watchdog =
         # sum of the planner's per-class budgets; --class-timeouts is the
         # single source of watchdog policy and must cover every assigned
-        # class). A failing batch retries exactly the failed tests - once -
-        # and falls back to per-class diagnosis for timeouts/infrastructure
-        # wedges.
+        # class). A failing batch retries exactly the non-passing tests -
+        # once - and falls back to per-class diagnosis for
+        # timeouts/infrastructure wedges.
         env:
           CONDUIT_PERF_TRACE: "1"
           LANE_NAME: "${{ matrix.lane }}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,8 +13,10 @@ name: CI
 #                             portability; upload products as an artifact.
 #   unit/ui  (macos, matrix) test-without-building from the downloaded
 #                             products; measured watchdogs; native flake
-#                             retry (units); per-class invocations, per-class
-#                             watchdogs and one targeted class retry (UI).
+#                             retry (units); one batched invocation per UI
+#                             shard with method-precise retry of only the
+#                             failed tests and a per-class diagnosis
+#                             fallback for timeouts/infrastructure wedges.
 #   report   (ubuntu, always) aggregate lane results into the step summary.
 #   timing-history-update (main only) EWMA-merge fresh per-class timings into
 #                             the timing-history cache.
@@ -273,11 +275,13 @@ jobs:
       - name: Run UI lane
         # Matrix values enter through the environment, never through bash
         # template interpolation, so even a hostile class name cannot become
-        # code execution in the runner. The lane runner executes each class
-        # as an independent xcodebuild invocation under the planner's
-        # per-class watchdog (--class-timeouts is the single source of
-        # watchdog policy and must cover every assigned class) and retries
-        # only a failed class - once.
+        # code execution in the runner. The lane runner executes all of the
+        # shard's classes in ONE batched xcodebuild invocation (watchdog =
+        # sum of the planner's per-class budgets; --class-timeouts is the
+        # single source of watchdog policy and must cover every assigned
+        # class). A failing batch retries exactly the failed tests - once -
+        # and falls back to per-class diagnosis for timeouts/infrastructure
+        # wedges.
         env:
           CONDUIT_PERF_TRACE: "1"
           LANE_NAME: "${{ matrix.lane }}"
--- a/Conduit/Views/AskHermesPromptView.swift
+++ b/Conduit/Views/AskHermesPromptView.swift
@@ -16,6 +16,12 @@ struct AskHermesPromptView: View {
 
     @State private var copied = false
     @State private var copyCount = 0
+    /// Durable record of which prompt was last copied. The visible "Copied"
+    /// label intentionally disappears after ~2s, but the fact that THIS
+    /// prompt was copied must stay queryable through accessibility (VoiceOver
+    /// users re-reading the control, and automation) — keyed to the prompt
+    /// string so a reused card can never claim a different prompt was copied.
+    @State private var lastCopiedPrompt: String?
 
     var body: some View {
         VStack(alignment: .leading, spacing: 12) {
@@ -33,12 +39,14 @@ struct AskHermesPromptView: View {
                 Button {
                     UIPasteboard.general.string = prompt
                     copied = true
+                    lastCopiedPrompt = prompt
                     copyCount += 1
                 } label: {
                     Label("Copy Prompt", systemImage: copied ? "checkmark" : "doc.on.doc")
                         .font(.footnote.weight(.semibold))
                 }
                 .accessibilityIdentifier("setup.copy-prompt")
+                .accessibilityValue(lastCopiedPrompt == prompt ? "Copied" : "")
 
                 if copied {
                     Text("Copied")
--- a/ConduitUITests/ConnectionRepairUITests.swift
+++ b/ConduitUITests/ConnectionRepairUITests.swift
@@ -110,7 +110,9 @@ final class ConnectionRepairUITests: XCTestCase {
         tapVisible(repair, in: app)
 
         // No retained failure: the repair opens straight on the staged test.
-        let preview = app.descendants(matching: .any).matching(identifier: "setup.address-preview").firstMatch
+        // The preview is a single Text element (exposed as a static text);
+        // a typed query keeps snapshot evaluation out of the whole tree.
+        let preview = app.staticTexts["setup.address-preview"]
         XCTAssertTrue(preview.waitForExistence(timeout: 5), "Repair did not open on the staged test. Tree:\n\(app.debugDescription)")
         XCTAssertEqual(preview.label, Identity.stubDashboardURL)
         tapVisible(app.buttons[Identity.testRun], in: app)
@@ -138,7 +140,7 @@ final class ConnectionRepairUITests: XCTestCase {
         XCTAssertTrue(repair.waitForExistence(timeout: 10))
         tapVisible(repair, in: app)
 
-        let preview = app.descendants(matching: .any).matching(identifier: "setup.address-preview").firstMatch
+        let preview = app.staticTexts["setup.address-preview"]
         XCTAssertTrue(preview.waitForExistence(timeout: 5))
         tapVisible(app.buttons[Identity.testRun], in: app)
 
--- a/ConduitUITests/ConnectionSetupSettingsUITests.swift
+++ b/ConduitUITests/ConnectionSetupSettingsUITests.swift
@@ -45,7 +45,7 @@ final class ConnectionSetupSettingsUITests: XCTestCase {
         // The passwordless seed opens straight on the staged test with the
         // current URL preserved exactly — no first-run readiness questions,
         // no meaningless password field.
-        let preview = row(app, Identity.addressPreview)
+        let preview = addressPreview(app)
         XCTAssertTrue(preview.waitForExistence(timeout: 5), "Wizard did not open on the staged test. Tree:\n\(app.debugDescription)")
         XCTAssertEqual(preview.label, Identity.stubDashboardURL)
 
@@ -108,7 +108,7 @@ final class ConnectionSetupSettingsUITests: XCTestCase {
         openSettings(app)
         tapVisible(app.buttons[Identity.settingsRow], in: app)
 
-        let preview = row(app, Identity.addressPreview)
+        let preview = addressPreview(app)
         XCTAssertTrue(preview.waitForExistence(timeout: 5), "Wizard did not open on the staged test. Tree:\n\(app.debugDescription)")
         XCTAssertEqual(preview.label, Identity.stubDashboardURL)
 
@@ -121,8 +121,8 @@ final class ConnectionSetupSettingsUITests: XCTestCase {
         XCTAssertTrue(interactiveReady.waitForExistence(timeout: 5))
         XCTAssertTrue(interactiveReady.label.contains("browser-based sign-in"), "Got: \(interactiveReady.label)")
         XCTAssertFalse(app.staticTexts["setup.test.ready"].exists, "Interactive auth must never claim the connection is ready to use")
-        let authRow = row(app, "setup.test.stage.authentication")
-        XCTAssertTrue(authRow.exists)
+        let authRow = stageRow(app, "setup.test.stage.authentication")
+        XCTAssertTrue(authRow.waitForExistence(timeout: 5))
         XCTAssertTrue(authRow.label.lowercased().contains("browser sign-in required"), "Got: \(authRow.label)")
 
         tapVisible(app.buttons[Identity.useSettings], in: app)
@@ -147,7 +147,7 @@ final class ConnectionSetupSettingsUITests: XCTestCase {
         openSettings(app)
         tapVisible(app.buttons[Identity.settingsRow], in: app)
 
-        let preview = row(app, Identity.addressPreview)
+        let preview = addressPreview(app)
         XCTAssertTrue(preview.waitForExistence(timeout: 5), "Wizard did not open on the staged test. Tree:\n\(app.debugDescription)")
         XCTAssertEqual(preview.label, Identity.stubDashboardURL)
 
@@ -158,8 +158,8 @@ final class ConnectionSetupSettingsUITests: XCTestCase {
         // ready claim.
         let notice = app.staticTexts["setup.test.credentials-required"]
         XCTAssertTrue(notice.waitForExistence(timeout: 5))
-        let authRow = row(app, "setup.test.stage.authentication")
-        XCTAssertTrue(authRow.exists)
+        let authRow = stageRow(app, "setup.test.stage.authentication")
+        XCTAssertTrue(authRow.waitForExistence(timeout: 5))
         XCTAssertTrue(authRow.label.lowercased().contains("credentials required"), "Got: \(authRow.label)")
         XCTAssertFalse(app.staticTexts["setup.test.ready"].exists)
         XCTAssertFalse(app.staticTexts["setup.test.interactive-ready"].exists)
@@ -172,7 +172,7 @@ final class ConnectionSetupSettingsUITests: XCTestCase {
         XCTAssertTrue(username.waitForExistence(timeout: 5), "Credentials step did not appear. Tree:\n\(app.debugDescription)")
         let urlField = app.textFields[Identity.urlField]
         XCTAssertFalse(urlField.exists, "The URL is owned by the details step; credentials step only asks for credentials")
-        XCTAssertTrue(app.buttons[Identity.back].exists, "Back returns toward the tested connection")
+        XCTAssertTrue(app.buttons[Identity.back].waitForExistence(timeout: 5), "Back returns toward the tested connection")
     }
 
     // MARK: - Walk helpers
@@ -187,10 +187,20 @@ final class ConnectionSetupSettingsUITests: XCTestCase {
         settings.tap()
     }
 
-    /// Stage rows and the address preview are single combined accessibility
-    /// elements, so query by identifier across element types.
-    private func row(_ app: XCUIApplication, _ identifier: String) -> XCUIElement {
-        app.descendants(matching: .any).matching(identifier: identifier).firstMatch
+    /// The address preview is a single `Text` element, which XCTest exposes
+    /// as a static text (verified by hierarchy inspection).
+    private func addressPreview(_ app: XCUIApplication) -> XCUIElement {
+        app.staticTexts[Identity.addressPreview]
+    }
+
+    /// Stage rows are single combined accessibility elements backed by an
+    /// `HStack` with `children: .ignore`, which XCTest exposes as an Other
+    /// element (verified by hierarchy inspection). A typed query evaluates
+    /// one narrow subtree instead of snapshotting every element type in the
+    /// application, which is what the broad `descendants(matching: .any)`
+    /// lookup paid for on every check.
+    private func stageRow(_ app: XCUIApplication, _ identifier: String) -> XCUIElement {
+        app.otherElements[identifier]
     }
 
     private func tapVisible(_ element: XCUIElement, in app: XCUIApplication) {
--- a/ConduitUITests/ConnectionSetupTestConnectionUITests.swift
+++ b/ConduitUITests/ConnectionSetupTestConnectionUITests.swift
@@ -74,7 +74,7 @@ final class ConnectionSetupTestConnectionUITests: XCTestCase {
         XCTAssertTrue(app.buttons["setup.test.edit-credentials"].waitForExistence(timeout: 5))
         // Retry exists but is never the primary action after rejected
         // credentials.
-        XCTAssertTrue(app.buttons["setup.test.retry"].exists)
+        XCTAssertTrue(app.buttons["setup.test.retry"].waitForExistence(timeout: 5))
 
         tapVisible(app.buttons["setup.test.edit-credentials"], in: app)
         let username = app.textFields["setup.username"]
--- a/ConduitUITests/ConnectionSetupTestConnectionUITests.swift
+++ b/ConduitUITests/ConnectionSetupTestConnectionUITests.swift
@@ -9,6 +9,7 @@ final class ConnectionSetupTestConnectionUITests: XCTestCase {
         static let answerYes = "setup.answer-yes"
         static let methodLan = "setup.method-lan"
         static let detailsReady = "setup.details-ready"
+        static let stepLabel = "setup.step-label"
         static let next = "setup.next"
         static let back = "setup.back"
         static let testRun = "setup.test.run"
@@ -175,6 +176,13 @@ final class ConnectionSetupTestConnectionUITests: XCTestCase {
         if !setup.isHittable { app.swipeDown() }
         setup.tap()
 
+        // The entry tap starts a NavigationStack push: right after the tap,
+        // a snapshot can still describe the login card, so interacting with
+        // the first answer races the transition (a hosted failure waited out
+        // exactly this race). The wizard's own step label is the semantic
+        // "Step 1 is active" marker — wait for it before the first answer.
+        waitForStep(app, "Step 1 of 3")
+
         tapVisible(app.buttons[Identity.answerYes], in: app)
         tapVisible(app.buttons[Identity.answerYes], in: app)
         tapVisible(app.buttons[Identity.methodLan], in: app)
@@ -211,6 +219,32 @@ final class ConnectionSetupTestConnectionUITests: XCTestCase {
     /// lookup paid for on every check.
     private func stageRow(_ app: XCUIApplication, _ identifier: String) -> XCUIElement {
         app.otherElements[identifier]
+    }
+
+    /// Polls until the wizard's step label names `expected`, failing with
+    /// the tree if it never does. Polling (not a single assert) matters
+    /// because during the wizard's push/pop transitions both steps can be
+    /// mounted, so the first snapshot may still describe the outgoing step —
+    /// the same contract as ConnectionSetupUITests.stepLabel.
+    private func waitForStep(
+        _ app: XCUIApplication,
+        _ expected: String,
+        timeout: TimeInterval = 5
+    ) {
+        let label = app.staticTexts[Identity.stepLabel]
+        let deadline = Date().addingTimeInterval(timeout)
+
+        while Date() < deadline {
+            if label.exists, label.label == expected {
+                return
+            }
+
+            RunLoop.current.run(until: Date().addingTimeInterval(0.1))
+        }
+
+        XCTFail(
+            "Expected wizard step \(expected). Tree:\n\(app.debugDescription)"
+        )
     }
 
     private func tapVisible(_ element: XCUIElement, in app: XCUIApplication) {
--- a/ConduitUITests/ConnectionSetupTestConnectionUITests.swift
+++ b/ConduitUITests/ConnectionSetupTestConnectionUITests.swift
@@ -35,9 +35,9 @@ final class ConnectionSetupTestConnectionUITests: XCTestCase {
         // stage plus the ready message. The password is never rendered as
         // text anywhere.
         XCTAssertTrue(app.staticTexts["setup.test.ready"].waitForExistence(timeout: 5))
-        XCTAssertTrue(row(app, Identity.stageServer).exists)
-        XCTAssertTrue(row(app, Identity.stageDashboard).exists)
-        XCTAssertTrue(row(app, Identity.stageAuthentication).exists)
+        XCTAssertTrue(stageRow(app, Identity.stageServer).waitForExistence(timeout: 5))
+        XCTAssertTrue(stageRow(app, Identity.stageDashboard).waitForExistence(timeout: 5))
+        XCTAssertTrue(stageRow(app, Identity.stageAuthentication).waitForExistence(timeout: 5))
         XCTAssertFalse(app.staticTexts["round4-private-fixture"].exists)
 
         // The Round-3 typed handoff is unchanged: fields populate, nothing
@@ -63,15 +63,15 @@ final class ConnectionSetupTestConnectionUITests: XCTestCase {
         // stages remain visible, and the classified credential guidance
         // appears with Edit Credentials as the primary recovery action.
         XCTAssertTrue(app.staticTexts["setup.test.failure"].waitForExistence(timeout: 5))
-        XCTAssertTrue(row(app, Identity.stageServer).exists)
-        XCTAssertTrue(row(app, Identity.stageDashboard).exists)
-        XCTAssertTrue(row(app, Identity.stageAuthentication).exists)
+        XCTAssertTrue(stageRow(app, Identity.stageServer).waitForExistence(timeout: 5))
+        XCTAssertTrue(stageRow(app, Identity.stageDashboard).waitForExistence(timeout: 5))
+        XCTAssertTrue(stageRow(app, Identity.stageAuthentication).waitForExistence(timeout: 5))
         let failure = app.staticTexts["setup.test.failure"]
         XCTAssertEqual(
             failure.label,
             "Hermes rejected that username or password. Check your dashboard credentials and try again."
         )
-        XCTAssertTrue(app.buttons["setup.test.edit-credentials"].exists)
+        XCTAssertTrue(app.buttons["setup.test.edit-credentials"].waitForExistence(timeout: 5))
         // Retry exists but is never the primary action after rejected
         // credentials.
         XCTAssertTrue(app.buttons["setup.test.retry"].exists)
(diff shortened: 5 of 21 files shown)


kaishi00:
Confirmed: the batch path deletes `observations-batch-a2.json` (lines 886, 916) to prevent method-only retry timings from polluting history, but class diagnosis keeps `observations-$cls-a2.json` (lines 1050, 1076) with no equivalent removal. I've verified this is a real inconsistency.

I have enough to write the review. Let me verify one more thing about the `echo` at line 894 malformed output and finalize.

## Summary
This PR hardens CI reliability for XCUITest + XCTest invocation efficiency: narrows UI-test accessibility queries to typed element helpers (fixing `Timed out while evaluating UI query`), consolidates unit/UI lanes via overhead-aware planning, batches each UI shard into one xcodebuild invocation with method-precise retry and per-class diagnosis fallback, pins the simulator destination by UDID, and updates docs. No application code changes.

## Findings

### [major] Method-only retry timings pollute class timing history in diagnosis mode (inconsistent with the batch path)
**File:** scripts/ci-test-lane.sh:1049-1052, 1075-1078
**Issue:** In the batch path, the code explicitly deletes the retry observations part (`rm -f .../observations-batch-a2.json`, lines 886/916) with the comment "the retry reran only the failed METHODS, so its observations carry method-only durations per class - the batch attempt owns class timing, so the retry's observations must not fold into the timing history." The per-class diagnosis path does the same method-precise retry (`DIAG_RETRY_FILTERS` are `-only-testing:target/class/testMethod`, line 1031-1036), but writes and *keeps* `$RESULT_DIR/parts/observations-$cls-a2.json` (lines 1050, 1076). In `merge_observation_parts`, fold order is `(is_batch, stem, attempt)` with last-wins, so the retained `$cls-a2` method-only run folds last and wins, shrinking `$cls`'s recorded duration to only the retried tests' span. This contradicts the docstring/branch claim ("a method-only retry never shrinks a class's timing history") which is only true on the batch path.
**Suggestion:** Mirror the batch handling: `rm -f "$RESULT_DIR/parts/observations-$cls-a2.json"` after extracting detail in the diagnosis retry (both the pass branch at ~1052 and the failure branch at ~1076), so only full-class attempt timings feed the EWMA history. If diagnosis class retries are intended to keep timing, they should not use method-only filters, or the timing claim should be corrected.

### [minor] Malformed warning message string concatenation
**File:** scripts/ci-test-lane.sh:894
**Issue:** The flake warning uses unquoted/adjacent expansions: `...runner-level flake ("$(printf '%s, ' $RETRY_CLASSES)"only the failed tests re-ran), reported...`. This renders as `(ClassA, ClassB only the failed tests re-ran)` — the trailing commas plus no space before "only" make the message read incorrectly, which is visible in the CI report/annotations meant to be the human-readable flake evidence.
**Suggestion:** Build the message cleanly, e.g. `echo "::warning::UI shard $LANE PASSED on its targeted retry - runner-level flake (retried: $(printf '%s, ' $RETRY_CLASSES)only the failed tests re-ran), reported, not hidden"` with proper spacing.

### [minor] `merge_detail_parts` batch-failure suppression depends on class-part presence that isn't always guaranteed on failed lanes
**File:** scripts/extract-test-timings.py:401-404
**Issue:** The new guard deletes the `"batch"` failure key only when a per-class key exists. On a clean diagnosis-fallback *pass* this is correct (class parts supersede). But consider the diagnosis path where a class fails and the lane ends with `mark_remaining_not_diagnosed` (e.g. a hang at line 1067-1073): the remaining classes are recorded via `record_attempt "skipped"` only — no per-class detail part file is written — so no per-class key exists, and the batch failures remain folded in. That happens to be the desired result here, but the coupling between "a part file exists" and "batch results were superseded" is implicit and fragile: if a future refactor writes a class part for a terminal `not_diagnosed` class, the batch failures would be silently dropped even though the lane failed with the batch as the only real evidence.
**Suggestion:** Make the supersession explicit rather than inferred from part-file existence — e.g. base the decision on whether per-class *execution/attempt* records exist in the lane-result (or gate on lane status == pass), so batch failures aren't dropped on failed lanes regardless of part-file shape.

### [nit] Global mutable state in `run_ui_lane` / `run_class_diagnosis`
**File:** scripts/ci-test-lane.sh:829, 854, 867, 872, 898-909
**Issue:** `RETRY_LINES`, `RETRY_FILTERS`, `RETRY_CLASSES`, `status2`, `DIAGNOSIS_REASON`, `batch_budget` are non-`local` globals in the big state machine. This works under the current single-entry flow, but it's fragile: `run_class_diagnosis` reuses `budget`, `status1`, etc. as globals and is called from multiple call sites that then rely on control falling out to `finish_lane`. Any future caller or a `set -e` interaction could leak state across invocations.
**Suggestion:** Declare these `local` in their enclosing functions (bash 3.2 supports `local` arrays), matching the pattern used elsewhere in the script, to make the control flow more robust to reuse.

## Verdict
REQUEST_CHANGES

The UI-test accessibility fix and the overhead-aware planner are sound and well-tested. The one substantive issue is the diagnosis-path timing pollution (major): the method-precise retry observations are pruned on the batch path but not on the per-class diagnosis path, so the fix for that guard is incompletely applied. That, plus the malformed flake-warning output, should be addressed before merge.

<a href="https://opencode.ai/s/jAeX0Ww2">New%20session%20-%202026-09-10T00%3A23%3A09.248Z</a>
[opencode session](https://opencode.ai/s/jAeX0Ww2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34420979101)

kaishi00:
The fold-order logic and batch supersession are well-designed and covered by tests. This addresses the core correctness concern.

Let me do a final check on the `sleep` change in `bounded_run` and the concurrency of the self-test. One thing to verify: the self-test job runs on ubuntu, but the commit message says the suite was measured on macOS. Running the state-machine bash suite on ubuntu (GNU bash 5.x, GNU utils) vs macOS (bash 3.2) — the test suite may behave differently, but the PR notes a Windows run also green. This is fine as a CI-tooling suite.

Let me also verify there are no glaring issues in the Swift UI test changes — force unwraps, retain cycles etc. The Swift files use `XCTAssertTrue`/`waitForExistence` consistently, typed helpers `staticTexts[...]`/`otherElements[...]`, no force unwraps. Testing code, not production. There's a minor inconsistency: `ConnectionRepairUITests` still uses inline `app.staticTexts["setup.address-preview"]` (lines 115, 143) instead of a helper, whereas the PR description claimed "ConnectionRepairUITests' two identical address-preview lookups fixed the same way" — it is fixed to staticTexts (typed) but not extracted into a helper. Not an issue; it IS the typed query. Actually the diff shows +4/-2 for Repair, confirming the fix to typed query. Fine.

Let me quickly check for any force-unwrap in the new/edited Swift test lines.

The Swift test files contain no forced unwraps, no implicitly unwrapped optionals, no Combine or closures needing weak self (they're plain sync XCTest). These are test files — no production retain-cycle/concurrency concerns.

Let me verify the diagnosis `budget` variable shadowing (used as global in run_class_diagnosis but also named local in sum_of_class_budgets — separate functions, fine). There's the author's own [nit] about non-local globals in run_class_diagnosis/run_ui_lane. That remains but is defensible for a single-entry script.

Now let me also double check the one concern about `DIAG_RETRY_LINES` being stale across loop iterations. Line 1042 sets it to empty at end of each iteration. But note: within a single iteration, `DIAG_RETRY_LINES=$(retry_filter_lines "$det1")` runs only in the `else` (test-failures) branch (line 1026). If a prior iteration set it, then this iteration... no, line 1042 resets it every iteration. Good, no stale leak.

But there's a subtle issue: line 1026 `DIAG_RETRY_LINES=$(retry_filter_lines "$det1")` is INSIDE the `else` of the a1_status test (line 1022-1028). For timeout and infra-error paths, `DIAG_RETRY_LINES` retains its value from... the previous iteration, which was reset to "" at line 1042. But wait — line 1032 checks `if [ -n "${DIAG_RETRY_LINES:-}" ]`. On the FIRST iteration, DIAG_RETRY_LINES is unset initially (fresh function call). Fine. After line 1042 sets it to "". So the class-level fallback (line 1040) is correctly used for a full-class retry after timeout/infra. And `DIAG_RETRY_METHOD_FILTERED=0` is set at 1031 before the branch (reset per iteration). Good — so a timeout retry is class-level unfiltered, and its observations ARE kept (correctly, since it's a full-class run the class owns). Consistent.

I have enough to write the review. Let me consolidate the findings.

Key findings:
1. [minor] Batch-bundle cleanup doesn't guard on batch attempt status — CodeRabbit finding #1, still open in current code (lines 509-511). When a batch times out/wedges/incomplete and per-class diagnosis passes all classes on attempt 1, `RETRIED_LINES`/`INFRA_RECOVERED_LINES` are empty, so the batch xcresults are deleted even though the diagnosis path is where batch evidence is most valuable. Although consistent with the green-lane-pruning philosophy for class bundles, batch timeouts can leave partial/ambiguous data that is the only record of the batch event.

2. [minor] Test_lane_runner.sh negative assertions still use exact filter literals (lines 690, 801-802) that wouldn't match a spurious retry's actual stub log line — vacuous-pass risk. CodeRabbit finding #4 not fully resolved.

3. [nit] Unassigned retry class → empty retry_budget (CodeRabbit finding #2) — defensible by startup coverage check but classes from detail JSON aren't validated against CLASSES.

4. [nit] Non-local globals in run_ui_lane/run_class_diagnosis (author's own finding).

Let me finalize. I should note the UI fix is sound and the major correctness concern about diagnosis-retry timing was properly fixed.

Let me confirm findings in code once more for finding 2 - the lines. I'm confident. Writing the review.

Actually, let me reconsider the batch-bundle cleanup severity. The PR description explicitly says "clean green-lane bundles are still pruned" and this is a green lane. The batch bundle of a timed-out run, after reboot + clean diagnosis, there's genuinely nothing left to diagnose. The finding is defensible as a "keep evidence" preference. I'll keep it minor.

Let me write the final review.</think>## Summary
This PR hardens CI reliability for XCUITest + XCTest: narrows UI-test accessibility queries to typed element lookups (fixing `Timed out while evaluating UI query`), batches each UI shard into one xcodebuild invocation with method-precise retry and per-class diagnosis fallback, adds overhead-aware lane planning (unit 4→2), pins simulator destination by UDID, and moves the slow CI-tooling meta-suites into a concurrent self-test job. The app-facing changes are test-only; no production code is touched.

## Findings

### [minor] Batch xcresult bundles are pruned on a green lane even when a batch-level event triggered the diagnosis
**File:** scripts/ci-test-lane.sh:509-511
**Issue:** Bundle cleanup on `finish_lane "pass"` deletes all `batch-a*.xcresult` whenever `RETRIED_LINES` and `INFRA_RECOVERED_LINES` are both empty. But a batch that records `incomplete`/`timeout`/`infra-error` (lines 777, 792, 811, 841) falls to per-class diagnosis, and if every class then passes on attempt 1, neither file is populated (attempt-1 passes in `run_class_diagnosis` at lines 982-987 write to neither). The batch bundles — the only raw evidence of the batch-level wedge/timeout — are then silently removed. This is the CodeRabbit finding that was marked "Addressed in commit c2b64eb" but is still present verbatim.
**Suggestion:** Gate the deletion on the batch attempt status, e.g. also require that no batch attempt recorded `incomplete|timeout|infra-error` before pruning, mirroring the bundled guard in the proposed fix, so batches that forced a diagnosis keep their raw bundles even on a green lane.

### [minor] Negative retry assertions count an exact filter literal the stub would not emit, so they can pass vacuously
**File:** scripts/tests/test_lane_runner.sh:690, :801-802
**Issue:** `batch_invocations 'batch-a2 (filters: BetaUITests/testC())' == "0"` and `'batch-a2 (filters: testC())'` assert no retry ran by matching one exact string. The stub logs `batch-a2 (filters:$methods)` where `$methods` is derived from the real `-only-testing:ConduitUITests/Class/testMethod` filters; a spurious retry for this specific case would produce a different literal (e.g. an `AlphaUITests/testX()` filter, or the class prefix), so a regression that erroneously retried would not match and the count stays 0 — the assertion would still pass. This makes the negative checks non-load-bearing.
**Suggestion:** Count all `batch-a2` invocations instead of one literal, e.g. add `any_batch_retry_invocations()` (grep -c '^batch-a2') and assert that == 0 (u6) and that the u14 trailing count == 0 while keeping the existing `batch-a1` count.

### [nit] Retry classes are not validated against the assigned class set before budgeting
**File:** scripts/ci-test-lane.sh:867-868
**Issue:** `RETRY_CLASSES` (line 867) is derived from `detail-batch-a1.json` via `awk`, and `sum_of_class_budgets $RETRY_CLASSES` calls `ui_budget_for`, which does `exit 2` inside a command substitution if a name has no planned watchdog (lines 217-219). Under this shape, an unexpected class name in the detail JSON would not abort the lane cleanly and could yield an empty/zero `retry_budget`, sending `xcodebuild_test` a zero-second deadline (line 874). The startup guard (lines 226-233) only covers desired `CLASSES`, not names that arrive indirectly from the extracted JSON.
**Suggestion:** Before invoking `sum_of_class_budgets`, verify every name in `RETRY_CLASSES` is present in `CLASSES`; fail the lane with the usual "unclassified" record if any are not (mirrors the `unexecuted_classes` defensive pass at 771).

### [nit] Non-local globals leak across the big state machine
**File:** scripts/ci-test-lane.sh:829, 854, 867, 872, 898-909, 953-956
**Issue:** `RETRY_LINES`, `RETRY_FILTERS`, `RETRY_CLASSES`, `status1`/`status2`, `DIAGNOSIS_REASON`, `budget`, `DIAG_CLASSES_ARR`, `LANE_FAILED` are script-level (not `local`) in `run_ui_lane`/`run_class_diagnosis`. This works under the current single-entry flow but makes the control flow brittle to future call sites or `set -e` interactions reusing stale state.
**Suggestion:** Declare them `local` in their enclosing functions (bash 3.2 supports local arrays), consistent with the pattern used elsewhere (e.g. `sum_of_class_budgets`).

The previously highlighted substantive concern — diagnosis-path method-only retries polluting class timing history (the batch path deletes `observations-batch-a2.json` but the per-class diagnosis path kept `observations-$cls-a2.json`) — **has been resolved**: lines 1060-1062 and 1091-1093 now delete the subset/prune `observations-$cls-a2.json` on method-filtered retries for both pass and fail paths, and `merge_detail_parts` (extract-test-timings.py:409-415) performs per-class batch-failure supersession, both covered by new tests. The malformed flake-warning concatenation (line 894) is also fixed.

## Verdict
APPROVE

The UI-test accessibility fix and the overhead-aware batched planner are sound and thoroughly tested (134 bash + 107 python assertions). The one real correctness concern (diagnosis timing pollution) is fixed and covered. The outstanding items are robustness nits: the batch-bundle pruning on diagnosis-triggering green lanes and the vacuous negative assertions weaken retry/evidence guarantees but do not break the happy path. They are worth addressing in a follow-up but do not block merge.

<a href="https://opencode.ai/s/aBPr6UfS">New%20session%20-%202026-09-10T01%3A27%3A19.124Z</a>
[opencode session](https://opencode.ai/s/aBPr6UfS)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34425460649)

kaishi00:
I have enough verified findings. Let me write the final review.

## Summary
This PR hardens CI reliability for XCUITest and XCTest invocation efficiency: narrows UI-test accessibility queries to typed element helpers (fixing the `Timed out while evaluating UI query` snapshot failures), batches each UI shard into one xcodebuild invocation with method-precise retry and per-class diagnosis fallback, adds overhead-aware lane planning (unit 4→2), pins the simulator destination by UDID, moves the slow CI-tooling meta-suites into a concurrent self-test job, and updates docs. No production code changes (app-facing surface is test query logic only).

## Findings

### [major] Class-only diagnosis retry is treated as method-filtered, deleting the full-class retry timing sample
**File:** scripts/ci-test-lane.sh:1036-1038 (and 1056-1059, 1089-1092)
**Issue:** `DIAG_RETRY_METHOD_FILTERED=1` is set unconditionally whenever `DIAG_RETRY_LINES` is non-empty, but `retry_filter_lines` (line 295) can emit a **class-only** filter `$TARGET/$cls` when XCTest cannot identify a failed method (the `class_only` collection at line 317-323). For such a full-class retry, the subsequent `rm -f "$RESULT_DIR/parts/observations-$cls-a2.json"` guards (lines 1056-1059, 1089-1092) delete the retry's full-class observations. The merged timing-history EWMA then keeps only the failed attempt-1 sample instead of the complete retry sample the class actually produced — contradicting the PR's own rule ("a full-class retry owns class timing"). The regression cases u15/u16 only cover the method-only path, so this path is untested.
**Suggestion:** Set `DIAG_RETRY_METHOD_FILTERED=1` only for filters containing a test component, e.g. `case "$filter" in "$TARGET/$cls/"*) DIAG_RETRY_METHOD_FILTERED=1;; esac`, keeping class-only lines unflagged so a full-class retry's observations are retained. Add a regression case for a class-only retry that verifies its complete retry sample remains in the merged history.

### [minor] Timing-history-update does not gate on the self-test job
**File:** .github/workflows/ci.yml:425-432
**Issue:** The `timing-history-update` job lists `needs: [plan, unit, ui]` and its condition requires only plan/unit success + ui success/skipped. It does not depend on `self-test` nor require `needs.self-test.result == 'success'`. A failed CI-tooling regression (planner, lane-runner, timing-extraction) can therefore still let the shared timing history on `main` be EWMA-merged, corrupting the predictive model that the next runs' lane planning depends on.
**Suggestion:** Add `self-test` to `needs` and require `needs.self-test.result == 'success'` in the `if` condition, mirroring the `ci-gate` job at line 407/421.

### [minor] Retry classes extracted from detail JSON are not validated against the assigned class set
**File:** scripts/ci-test-lane.sh:868
**Issue:** `retry_budget=$(sum_of_class_budgets $RETRY_CLASSES)` feeds classes derived from the batch detail JSON. `sum_of_class_budgets` (line 359) calls `ui_budget_for`, which on an unknown class prints an error and `exit 2` inside a `$(...)` subshell — the exit is swallowed and the retry_budget becomes empty. The startup coverage check (lines 226-233) only validates `CLASSES_ARR`, not names that later arrive from extraction, so an unassigned class could yield a zero/empty deadline into `xcodebuild_test`. The batch-path "empty filters" guard (lines 861-863) doesn't cover this.
**Suggestion:** Before calling `sum_of_class_budgets`, verify every entry in `RETRY_CLASSES` is present in `CLASSES`; fail the lane with the usual unclassified/complete record if any are not.

### [minor] Negative retry assertions match exact filter literals the stub would never emit (vacuous-pass risk)
**File:** scripts/tests/test_lane_runner.sh:690, :802
**Issue:** `assert_eq "no retry after an unclassifiable batch" "$(batch_invocations 'batch-a2 (filters: BetaUITests/testC())')" "0"` (line 690) — this case fails `AlphaUITests` and forces an unclassified/no-retry outcome; if a spurious a2 retry ran it would log `BetaUITests/...` or `AlphaUITests/...`, so the exact literal simply can't match and the count is 0 regardless. Likewise line 802 uses the prefix-less literal `testC()` while `retry_filter_lines` emits prefixed `$TARGET/$cls/testMethod`; the fixed-string grep won't match a real retry line. Both assertions pass vacuously and cannot catch a retry regression.
**Suggestion:** Assert zero retries by counting all `batch-a2` invocations (e.g. a shared `any_batch_retry_invocations() { grep -c '^batch-a2' "$INVOCATION_LOG" || true; }`) instead of one literal filter line, for both u6 and u14.

### [minor] Missing `end_case` before u15/u16 `begin_case` breaks per-case telemetry
**File:** scripts/tests/test_lane_runner.sh:815, :838
**Issue:** `begin_case "ui diagnosis method retry keeps class timing"` (815) and `begin_case "ui failing diagnosis method retry keeps rule"` (838) are not preceded by `end_case`. The prior explicit `end_case` for the active case is at 854 (after u16). `run_case`/`begin_case` (line 113) records START telemetry with no matching END for u15 and u16 — the exact two new regression cases this PR touts as "134/134 with per-case telemetry."
**Suggestion:** Add `end_case` immediately before the `begin_case` calls on lines 815 and 838.

### [nit] Batch xcresult bundles pruned even when a batch-level event triggered the diagnosis
**File:** scripts/ci-test-lane.sh:509-511
**Issue:** The green-lane cleanup deletes `batch-a*.xcresult` whenever `RETRIED_LINES` and `INFRA_RECOVERED_LINES` are both empty. A batch recording `incomplete`/`timeout`/`infra-error` falls to per-class diagnosis; if every class then passes on attempt 1 (written to neither file), the batch's raw bundles — the only evidence of the batch-level wedge — are silently removed.
**Suggestion:** Guard the deletion on the batch attempt status too (only prune when no batch attempt recorded `incomplete|timeout|infra-error`), so diagnosis-triggering batches keep their raw bundles even on a green lane.

### [nit] Workflow contract test asserts only ci-gate args, not the self-test wiring
**File:** scripts/tests/test_ci_workflow_contract.py:102-107
**Issue:** The test exercises `ci-gate.py` argument shape but does not assert that `ci.yml` defines the `self-test` job, that `ci-gate.needs` includes `self-test`, that `-needs.self-test.result` is threaded, or that `ci-gate` uses `if: always()`. A workflow regression could pass silently.
**Suggestion:** Add workflow-level assertions for the `self-test` job definition, the gate's `needs`/`if: always()`, and the `timing-history-update` self-test dependency.

### [nit] Non-local globals in the run_ui_lane / run_class_diagnosis state machine
**File:** scripts/ci-test-lane.sh:829, 854, 867, 872, 898-909, 953-956
**Issue:** `RETRY_LINES`, `RETRY_FILTERS`, `RETRY_CLASSES`, `status1`/`status2`, `DIAGNOSIS_REASON`, `budget`, `DIAG_CLASSES_ARR`, `LANE_FAILED` are script-level rather than `local`. Works under the current single-entry flow but is fragile to future call sites or `set -e` interactions.
**Suggestion:** Declare them `local` in their enclosing functions (bash 3.2 supports local arrays), consistent with `sum_of_class_budgets`.

No issues in the Swift UI-test changes: the typed `staticTexts[...]`/`otherElements[...]` helpers, `waitForExistence(timeout: 5)` assertions, and repeated-`isHittable` tap loop are consistent across all three edited files. No force-unwraps, no retained closures/Combine subscribers (plain synchronous XCTest), and correctness-raising assertion changes (`.exists` → waits) are appropriate.

## Verdict
REQUEST_CHANGES

The core substance of the prior reviews (the extract-test-timings per-class batch supersession, and the method-filtered diagnosis observations pruning) is properly implemented — the round-1 flag at ci-test-lane.sh:868 (unassigned retry class) and round-2 timing-history gating are not actually resolved in HEAD. The highest-impact open issue is the class-only diagnosis retry being mis-flagged as method-filtered, which silently corrupts class timing history for the very failure mode relativization is meant to model, and it has no test coverage. The remaining items are smaller but all degrade either data integrity of the timing model (the plan/UI job's predictive input) or retry-evidence guarantees. Blocking found: the class-only-retry timing bug (major) and the self-test gating gap should be fixed and covered before merge; the rest can be follow-ups.

<a href="https://opencode.ai/s/nEdNNMJc">New%20session%20-%202026-09-10T01%3A45%3A54.904Z</a>
[opencode session](https://opencode.ai/s/nEdNNMJc)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34426687160)

kaishi00:
Confirmed precisely: case u14 begins at 790, and its `end_case` (which would be at ~813, before u15's begin at 815) is **absent**. u15 begins at 815 with no prior `end_case` (the last `end_case` is at 789), and u16 at 838 also no prior `end_case`. The single `end_case` at 854 closes u16. So u15 and u16 lack END telemetry. My finding is accurate.

The review is complete and verified against the actual HEAD. My structured review is delivered above.

## Summary
This PR hardens CI reliability: narrows UI-test accessibility queries to typed element lookups (fixing `Timed out while evaluating UI query`), batches each UI shard into one xcodebuild invocation with method-precise retry and per-class diagnosis fallback, adds overhead-aware lane planning, pins the simulator destination by UDID, and moves CI-tooling regression suites into a concurrent self-test job. App changes are limited to making the Copy-Prompt confirmation durable via an accessibility value; the rest is CI tooling and tests.

## Findings

### [major] Class-only diagnosis retries are wrongly marked method-filtered, so their full-class timing samples are discarded
**File:** scripts/ci-test-lane.sh:1038 (effects at 1061, 1092)
**Issue:** `DIAG_RETRY_METHOD_FILTERED=1` is set unconditionally whenever `DIAG_RETRY_LINES` is non-empty, but `retry_filter_lines` (295-331) legitimately emits a class-only filter `TARGET/cls` (line 327) when XCUITest cannot name the failing test. For such a full-class retry the guards at 1060-1062 and 1089-1093 delete `observations-$cls-a2.json`, dropping a complete full-class timing sample. The merged history keeps only the attempt-1 sample. This contradicts the PR's own stated invariant ("a full-class retry owns class timing") and applies the method-only rule to a path that isn't method-only. Regression cases u15/u16 cover only the method-filtered path, so it is untested.
**Suggestion:** Set `DIAG_RETRY_METHOD_FILTERED=1` only when a filter has a test component (`case "$filter" in "$TARGET/$cls/"*) DIAG_RETRY_METHOD_FILTERED=1;; esac`) and add a class-only diagnosis-retry regression asserting its full-class observations survive.

### [minor] timing-history-update does not gate on the self-test job
**File:** .github/workflows/ci.yml:425-432
**Issue:** `timing-history-update` has `needs: [plan, unit, ui]` and its `if` requires only plan/unit/ui success — no `self-test` dependency. A failing CI-tooling regression (planner/lane-runner/timing-extraction are all exercised in `self-test`) can still be EWMA-merged into the shared `main` timing history that drives future lane planning. The `ci-gate` job correctly gates on self-test (407-421), but the predictive model is not protected.
**Suggestion:** Add `self-test` to `needs` and require `needs.self-test.result == 'success'` in the `if`, mirroring ci-gate.

### [minor] Negative retry assertions can never match, so they pass vacuously
**File:** scripts/tests/test_lane_runner.sh:690, :802
**Issue:** Line 690 asserts `batch-a2 (filters: BetaUITests/testC())` == 0 for a case that fails `AlphaUITests`; a spurious retry would log an `AlphaUITests/...` line the `BetaUITests` literal can't match. Line 802 asserts on `batch-a2 (filters: testC())`, but `retry_filter_lines` always emits a target-prefixed line, so it can never equal that literal either. Both negative assertions are non-load-bearing.
**Suggestion:** Count all `batch-a2` invocations (`any_batch_retry_invocations() { grep -c '^batch-a2' "$INVOCATION_LOG" ... } })` and assert == 0 in both cases.

### [minor] self-test wiring is not locked down by the workflow contract test
**File:** scripts/tests/test_ci_workflow_contract.py:92-110
**Issue:** The ci-gate contract tests only exercise synthetic `ci-gate.py` args and check for a "CI Gate" job; they don't assert that ci.yml defines `self-test`, that `ci-gate.needs` includes `self-test`, or that ci-gate uses `if: always()`. A workflow regression in the self-test wiring would pass silently.
**Suggestion:** Add text-level assertions for the self-test job, the gate's `needs`/`if: always()`, and the timing-history dependency.

### [minor] Missing `end_case` before UI cases 14 and 15
**File:** scripts/tests/test_lane_runner.sh:815, :838
**Issue:** Case u14 begins at 790 but its `end_case` is absent; the last `end_case` before u15 is at 789. u15 (815) and u16 (838) overwrite the active case, and the single `end_case` at 854 closes u16. Cases u14/u15 (confirmed: per-case START/END telemetry gaps) contradict the PR's "134/134 with per-case telemetry" claim and would hide a hang in exactly these two regression cases.
**Suggestion:** Add `end_case` immediately before both `begin_case` calls.

### [nit] Batch xcresult bundles pruned even when a batch-level event forced diagnosis
**File:** scripts/ci-test-lane.sh:509-511
**Issue:** On a green lane, `batch-a*.xcresult` are deleted whenever `RETRIED_LINES` and `INFRA_RECOVERED_LINES` are both empty. When a batch records `incomplete`/`timeout`/`infra-error` and every class then passes per-class diagnosis on attempt 1 (written to neither file), the batch's raw bundles — the only evidence of the batch-level wedge — are removed.
**Suggestion:** Also guard on the batch attempt status (only prune when no batch attempt recorded `incomplete|timeout|infra-error`).

### [nit] Extracted retry classes not validated against the assigned class set before budgeting
**File:** scripts/ci-test-lane.sh:867-868
**Issue:** `retry_budget=$(sum_of_class_budgets $RETRY_CLASSES)` feeds classes parsed from batch detail JSON; `ui_budget_for` `exit 2`s inside the command substitution (362) for an unknown name, yielding an empty `retry_budget` while the lane continues, so `xcodebuild_test` (874) can receive a zero/empty watchdog. The startup coverage check (226-233) only validates assigned `CLASSES_ARR`.
**Suggestion:** Verify every `RETRY_CLASSES` entry against `CLASSES` before summing; fail the lane (unclassified) if any are absent.

### [nit] Non-local globals in the run_ui_lane / run_class_diagnosis state machine
**File:** scripts/ci-test-lane.sh:829, 854, 867-868, 872, 898-909, 953-956
**Issue:** `RETRY_LINES`, `RETRY_FILTERS`, `RETRY_CLASSES`, `status1`/`status2`, `DIAGNOSIS_REASON`, `budget`, `DIAG_CLASSES_ARR`, `LANE_FAILED` are script-level rather than `local`. Works under the current single-entry flow but is fragile to reuse or `set -e` interactions.
**Suggestion:** Declare them `local` in their enclosing functions (bash 3.2 supports local arrays), consistent with `sum_of_class_budgets`.

Confirmed fixed at HEAD: `merge_detail_parts` per-class batch supersession (extract-test-timings.py:409-415), the malformed flake-warning string (894), and the batch-path observations pruning (886).

The Swift changes are clean: typed `otherElements`/`staticTexts` helpers, `waitForExistence` replacing `.exists`, and `waitForStep` polling (avoids the NavStack race). The one app change (AskHermesPromptView `lastCopiedPrompt`) is a synchronous `@State` mutation with no closures, force unwraps, or concurrency concerns. Only nit: `.accessibilityValue(... ? "Copied" : "")` attaches an empty value by default.

## Verdict
REQUEST_CHANGES

The UI-determinism fix and batched runner are sound and heavily tested, and the batch-path findings are resolved. But the class-only diagnosis-retry bug sits in the timing-history data-integrity category the PR explicitly guarantees, remains reachable, and has no test coverage; the main-branch timing-history gating gap and the vacuous negative assertions / missing `end_case` further degrade the value of the regression suite the PR touts. The major item should be fixed and covered before merge; the rest are follow-up material.

<a href="https://opencode.ai/s/l4s0oeEZ">New%20session%20-%202026-09-10T03%3A21%3A32.887Z</a>
[opencode session](https://opencode.ai/s/l4s0oeEZ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34432943212)

PR #159, opened 2026-09-11
feat(voice): wire continuousConversation preference to post-TTS relisten

kaishi00:
## Summary

Continuous conversation already existed before this PR: after assistant TTS finished, `VoiceConversationController.drainSpeechQueue` always called `startListening()` again. The bug was that the stored `VoiceProfilePreferences.continuousConversation` flag was dormant — default `true`, never read at runtime, and never exposed in Voice settings.

This PR wires that existing preference into the existing loop.

### Behavior

- **Continuous Conversation ON (default `true`)**: preserves current behavior. After a completed user → assistant exchange, Conduit automatically listens again. Existing users should see no change unless they disable the preference.
- **Continuous Conversation OFF**: the Voice session and sheet stay open, but Conduit does **not** automatically open the next listening turn. Settled as `.idle` with capture paused (not user-paused). The user starts the next turn with the existing Listen control.

### What this does not change

- Route safety is unchanged (speaker/receiver/CarPlay remain fail-safe half-duplex; proven headset routes keep full-duplex barge-in).
- Pause Microphone semantics remain authoritative.
- Stop phrases, goodbye/idle teardown, wake-word, Siri, CarPlay remain out of scope / future work.
- Safety/recovery restarts stay ungated: empty transcript, spoken stop phrase, acoustic barge-in, manual Interrupt, and speech-stream cancellation after the assistant finished still re-listen.

### Architecture

- `VoiceConversationController` remains the single voice runtime; it receives preferences only via `setProfilePreferences`.
- Persistence reuses the existing profile-scoped blob (`conduit.voice.preferences.v1.<gateway>.<profile>`); a write to `continuousConversation` does not clobber other fields.
- Custom `init(from:)` uses `decodeIfPresent` so a missing key decodes as ON; explicit `init()` restores default construction after the custom decoder removed the synthesized zero-arg initializer.
- Profile scoping is preserved through `loadVoiceProfilePreferences` / `saveVoiceProfilePreferences` / `refreshVoiceCapabilities`.

### Commits

1. `4419577` — feature wiring + tests
2. `e291d5c` — restore `VoiceProfilePreferences()` zero-arg initializer (compile fix)
3. `a70c3d8` — fix one-argument `XCTAssertEqual` in a new test (compile fix)

### Tests

[...]

PR #160, opened 2026-09-11
feat(voice): configurable spoken Stop and End Conversation controls

kaishi00:
## Summary

Configurable spoken Voice controls, profile-scoped and editable in Settings:

- **Stop phrases** (existed before this PR — semantics unchanged, now editable): a whole-utterance match consumes the phrase locally, cancels the current Hermes turn, stops TTS, keeps the Voice session open, and relistens. This remains a recovery/control restart even when Continuous Conversation is OFF (PR #159 left spoken-stop recovery ungated; preserved).
- **End Conversation phrases** (new, separate category, defaults `goodbye` / `bye` / `end conversation` / `that's all`): a whole-utterance match consumes the phrase, cancels any active Hermes turn through the same interrupt seam as Stop, and closes the Voice session completely — through the **existing Close teardown path** (`AppState.closeVoiceConversation()`: persist mute → `endVoiceSession()` → sheet dismissal), reached via a minimal `onEndConversation` seam on the controller. No second teardown implementation, no relisten afterward, independent of Continuous Conversation.

Defaults for Stop are unchanged (`stop`, `stop talking`, `be quiet`).

## Command recognition boundary

No new speech-recognition architecture: commands are recognized only where the existing transcription path completes an utterance (`finishUtterance`). No wake-word detector, no keyword spotting during raw capture, no second recognizer.

- **Full-duplex/headset routes:** unchanged — acoustic barge-in interrupts the assistant, the transcribed `goodbye` then matches and closes Voice.
- **Speaker-safe routes (speaker/receiver/CarPlay/unknown):** remain half-duplex. The microphone is NOT reopened during TTS to hear commands; the existing manual Interrupt control stays the way to speak a command during playback on those routes (pinned by tests).

## Matching rules

`VoiceSpokenCommands` is the single shared canonicalization/matching path for both lists. Deliberately conservative:

- whole-utterance matching only — `"I said goodbye to them"` never matches `goodbye`, `"can you stop talking about that"` never matches `stop talking`;
- normalization (case, surrounding whitespace/punctuation) applies to **both** the transcript and the configured phrases;
- no substring, fuzzy, semantic, or regex matching;
- conflict precedence: a phrase present in both lists deterministically takes the stronger action (**End Conversation wins** — end list is checked first).

## Persistence

Both lists live in the existing profile-scoped blob `conduit.voice.preferences.v1.<gateway>.<profile>`:

- `spokenEndConversationPhrases` added to `VoiceProfilePreferences`; older blobs without the key decode successfully and receive the new defaults (backward compatible);
- `outputMuted`, `continuousConversation`, `continueWakeConversation`, `spokenStopPhrases`, `transcriptionMode` are untouched by phrase edits;
- edits go through the PR #159 shared mutation path (`updateActiveProfileVoicePreferences`), so live mute is preserved only when `hasLiveVoiceSession` makes the controller authoritative;
- empty lists are valid and disable that spoken-command category — defaults are never silently restored;
- save-time canonicalization trims, drops blanks, and de-duplicates by the same normalized form used at runtime (display capitalization preserved).

## Generation / stale async work

After a spoken End Conversation closes Voice, the Close-path teardown (`stop()`) bumps the operation generation and cancels in-flight tasks *before* the Hermes interrupt await, so: late transcription continuations cannot submit; stale assistant events cannot restart playback or capture; speech-drain completion cannot relisten; continuous conversation cannot resurrect the session. Existing operation-generation / speech-drain revision fences are reused — no parallel lifecycle token system.

## Settings UI

New **Spoken Controls** section in Voice settings with compact list editors for both categories (view / add / edit / delete, empty allowed). Copy explains the difference ("Cancel the current response and keep Voice open." vs "Close the Voice conversation completely.") and that matching is whole-utterance.

## Explicitly unchanged

`VoiceAudioSessionCoordinator`, `VoiceBargeInRoutePolicy`, route classification, half-duplex eligibility, VAD thresholds, pre-roll, barge-in thresholds, capture suspension rules, Siri/App Intents, CarPlay, wake-word, idle timeout, fillers, provider TTS, AVAudioSession architecture.

## Tests

TDD (tests committed first; red build confirmed, then implementation):

- new `VoiceSpokenCommandMatchingTests` (defaults, backward-compatible decode, normalization both sides, no-substring, empty-list disables, list canonicalization);
- new `VoiceConversationSpokenEndCommandTests` (consume-and-close via the real Close seam, no-submit/no-transcript-entry, relisten-free close on continuous ON and OFF, conflict precedence, stop-still-relistens-when-OFF, barge-in→goodbye without stale relisten, speaker-safe half-duplex + manual Interrupt→goodbye, user-pause authority, late-assistant-event fencing, empty lists submit normally);
- new `AppStateVoiceSpokenPhraseTests` (persistence without clobbering, profile scoping, live-mute preservation with and without a live session, empty-list round-trip);
- regression: `ContinuousConversationPreferenceTests`, `AppStateContinuousConversationPreferenceTests`, `VoiceConversationControllerTests` (incl. existing spoken-stop coverage), plus the full `ConduitTests` unit suite.

Background/disconnect/server-replacement teardown behavior is pinned by the existing suites (`setForegroundActive` stop, server-replacement retirement) and is untouched.

## Review gate

Round 1 (three-model, on the feature): 2× REQUEST-CHANGES → fixed, 1× APPROVE. Accepted and fixed:

- **BLOCKER (2/3 reviewers): self-cancellation of the turn interrupt.** The Close teardown's `stop()` cancelled the very utterance task that then awaited the Hermes interrupt; `HermesClient.rpc` throws `CancellationError` for cancelled tasks, so the turn-cancel RPC would abort exactly when a live Hermes turn exists. Fixed by clearing the utterance-task handle before the teardown (close-first fencing preserved); a fixture assertion pins the interrupt running uncancelled.
- Typographic apostrophe (U+2019 — a common ASR emission) is folded in the shared canonicalization so the default `that's all` matches regardless of which apostrophe the recognizer returns.
- Phrase editor hardening: value-identity rows, punctuation-only drafts refused at commit, delete-while-editing target fix, editor re-seed identity, one preference decode per render.

Declined with reasoning: reordering to interrupt-first (would reintroduce the stale-continuation reopen race the close-first ordering exists to fence); caching pre-canonicalized phrase lists (matching runs once per completed utterance over a tiny list; caching adds invalidation coupling); a dedicated drain-race-after-End test (reaching the transcription path always cancels an active drain first — barge-in/manual Interrupt own that fence, and `stop()`'s drain fencing is already pinned deterministically by `testGenerationFenceBlocksLateCompletionAfterStopWithContinuousOff`).

Round 2 (three-model, on cleanup commit 60f386c): **3× APPROVE, no blocking findings.** The cleanup canonicalizes the phrase editor's seeding (legacy blobs with duplicates cannot collide with value-identity rows), guards the phrase setters on `isConnected` (same policy as `setContinuousConversation` — no writes to the orphaned "disconnected" gateway namespace), and removes dead edit-index bookkeeping in `deletePhrase`. One non-blocking suggestion (returning `Bool` from the phrase setters for failure-signaling parity) declined: the simpler disconnected policy was the explicit product decision for this surface.

## Verification

- macOS/Windows-side: full unit suite run on the Mac build box (result recorded in comments).
- Physical-device acceptance (real speaker/headset/CarPlay routes): **outstanding** — to be verified on device after merge.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -13109,6 +13109,10 @@ final class AppState: ObservableObject {
     /// canonicalized (trimmed, de-duplicated, blanks dropped) at this
     /// boundary — the single write authority for phrase lists.
     func setSpokenStopPhrases(_ phrases: [String]) {
+        // Same connected-state policy as `setContinuousConversation`: the
+        // preference key is gateway-qualified, and a write while
+        // disconnected would land in the orphaned "disconnected" namespace.
+        guard isConnected else { return }
         updateActiveProfileVoicePreferences {
             $0.spokenStopPhrases = VoiceSpokenCommands.canonicalizedPhraseList(phrases)
         }
@@ -13118,6 +13122,7 @@ final class AppState: ObservableObject {
     /// `setSpokenStopPhrases`; an intentionally emptied list persists as
     /// empty and disables that spoken-command category.
     func setSpokenEndConversationPhrases(_ phrases: [String]) {
+        guard isConnected else { return }
         updateActiveProfileVoicePreferences {
             $0.spokenEndConversationPhrases = VoiceSpokenCommands.canonicalizedPhraseList(phrases)
         }
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -811,6 +811,9 @@ final class AppState: ObservableObject {
         },
         interrupt: { [weak self] in
             await self?.interruptForVoice()
+        },
+        onEndConversation: { [weak self] in
+            self?.closeVoiceConversation()
         }
     )
     /// Manual per-message read aloud for completed assistant responses.
@@ -13099,6 +13102,27 @@ final class AppState: ObservableObject {
         return true
     }
 
+    /// Saves an edited spoken Stop phrase list for the active profile and
+    /// reapplies it to the live controller. Goes through the shared Voice
+    /// preference mutation path so live mute is preserved exactly per
+    /// `hasLiveVoiceSession` authority (PR #159 semantics). Entries are
+    /// canonicalized (trimmed, de-duplicated, blanks dropped) at this
+    /// boundary — the single write authority for phrase lists.
+    func setSpokenStopPhrases(_ phrases: [String]) {
+        updateActiveProfileVoicePreferences {
+            $0.spokenStopPhrases = VoiceSpokenCommands.canonicalizedPhraseList(phrases)
+        }
+    }
+
+    /// Saves an edited spoken End Conversation phrase list. Same contract as
+    /// `setSpokenStopPhrases`; an intentionally emptied list persists as
+    /// empty and disables that spoken-command category.
+    func setSpokenEndConversationPhrases(_ phrases: [String]) {
+        updateActiveProfileVoicePreferences {
+            $0.spokenEndConversationPhrases = VoiceSpokenCommands.canonicalizedPhraseList(phrases)
+        }
+    }
+
     /// Loads the active profile's preference blob, applies `mutate`, and
     /// reapplies it to the live controller.
     ///
@@ -13307,6 +13331,13 @@ final class AppState: ObservableObject {
             return VoiceProfilePreferences()
         }
         return preferences
+    }
+
+    /// The active profile's persisted Voice preferences, for display surfaces
+    /// (settings editors seed from this). Mutations go through the dedicated
+    /// setters, never by writing this value back.
+    var activeProfileVoicePreferences: VoiceProfilePreferences {
+        loadVoiceProfilePreferences(profile: activeProfile)
     }
 
     private func saveVoiceProfilePreferences(_ preferences: VoiceProfilePreferences, profile: String) {
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -509,6 +509,8 @@ struct SettingsView: View {
             )
         case .voice:
             if let bridge = appState.dashboardTicketBridge {
+                // One decode per render, not one per seeded field.
+                let voicePreferences = appState.activeProfileVoicePreferences
                 VoiceSettingsRoute(
                     bridge: bridge,
                     profile: snapshot.profile,
@@ -521,8 +523,8 @@ struct SettingsView: View {
                     transcriptionMode: appState.voiceTranscriptionMode,
                     appleSpeechAvailability: appState.appleSpeechAvailability,
                     continuousConversation: appState.continuousConversationEnabled,
-                    spokenStopPhrases: appState.activeProfileVoicePreferences.spokenStopPhrases,
-                    spokenEndConversationPhrases: appState.activeProfileVoicePreferences.spokenEndConversationPhrases,
+                    spokenStopPhrases: voicePreferences.spokenStopPhrases,
+                    spokenEndConversationPhrases: voicePreferences.spokenEndConversationPhrases,
                     setVoiceEnabled: { enabled in
                         await appState.setVoiceEnabled(enabled)
                     },
--- a/Conduit/Views/AuxiliaryViews.swift
+++ b/Conduit/Views/AuxiliaryViews.swift
@@ -521,6 +521,8 @@ struct SettingsView: View {
                     transcriptionMode: appState.voiceTranscriptionMode,
                     appleSpeechAvailability: appState.appleSpeechAvailability,
                     continuousConversation: appState.continuousConversationEnabled,
+                    spokenStopPhrases: appState.activeProfileVoicePreferences.spokenStopPhrases,
+                    spokenEndConversationPhrases: appState.activeProfileVoicePreferences.spokenEndConversationPhrases,
                     setVoiceEnabled: { enabled in
                         await appState.setVoiceEnabled(enabled)
                     },
@@ -529,6 +531,12 @@ struct SettingsView: View {
                     },
                     setContinuousConversation: { enabled in
                         appState.setContinuousConversation(enabled)
+                    },
+                    setStopPhrases: { phrases in
+                        appState.setSpokenStopPhrases(phrases)
+                    },
+                    setEndConversationPhrases: { phrases in
+                        appState.setSpokenEndConversationPhrases(phrases)
                     }
                 )
             } else {
--- a/Conduit/Views/VoiceSettingsView.swift
+++ b/Conduit/Views/VoiceSettingsView.swift
@@ -477,12 +477,14 @@ struct VoiceSettingsView: View {
                 initialPhrases: spokenStopPhrases,
                 onChange: setStopPhrases
             )
+            .id(spokenStopPhrases)
             SpokenPhraseListEditor(
                 title: "End Conversation Phrases",
                 purposeText: "Close the Voice conversation completely.",
                 initialPhrases: spokenEndConversationPhrases,
                 onChange: setEndConversationPhrases
             )
+            .id(spokenEndConversationPhrases)
         }
     }
 
@@ -645,8 +647,10 @@ private struct SpokenPhraseListEditor: View {
                     .font(.caption)
                     .foregroundStyle(.secondary)
             }
-            ForEach(phrases.indices, id: \.self) { index in
-                phraseRow(index)
+            // Entries are pairwise distinct after canonicalization, so the
+            // value itself is a stable identity across deletes and dedupes.
+            ForEach(phrases, id: \.self) { phrase in
+                phraseRow(phrase)
             }
             HStack(spacing: 8) {
                 TextField(editingIndex == nil ? "Add a phrase" : "Edit phrase", text: $draft)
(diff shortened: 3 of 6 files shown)


kaishi00:
## Summary
Adds profile-scoped, Settings-editable spoken "Stop" and "End Conversation" phrase lists, routing an End phrase through the existing Voice Close teardown (`AppState.closeVoiceConversation()`), with shared whole-utterance canonicalization in `VoiceSpokenCommands`.

## Findings

### [minor] `deletePhrase`'s edit-target adjustment is dead code
**File:** Conduit/Views/VoiceSettingsView.swift:710
**Issue:** `deletePhrase` carefully adjusts `editingIndex` when deleting the edited row or one before it (lines 715–720), but it then calls `save(updated)` (line 723), and `save` unconditionally resets `draft = ""` and `editingIndex = nil` (lines 730–731). The branch-by-branch "resolve or drop the pending edit" logic has no effect — every delete simply drops the in-progress edit. The comment describes behavior the code does not implement, and the review-hardening commit's "delete-while-editing target fix" is a no-op.
**Suggestion:** Either let `save` take a parameter controlling whether to clear the editor state (`save(updated, clearEditor:)`), or remove the `editingIndex` bookkeeping from `deletePhrase` and update the comment to state the pending edit is discarded.

### [minor] Phrase setters omit the `isConnected` guard used by sibling setters
**File:** Conduit/Services/AppState.swift:13111
**Issue:** `setSpokenStopPhrases` / `setSpokenEndConversationPhrases` call `updateActiveProfileVoicePreferences` unconditionally, whereas `setContinuousConversation` (line 13098) guards `isConnected`. When `connection` is nil, `voicePreferencesKey` falls back to `connection?.baseUrl ?? "disconnected"` (line 13255), so edits made while disconnected are written under the disconnected namespace and silently lost once a real gateway connects.
**Suggestion:** Add `guard isConnected else { return }` (or return a `Bool`) to both setters, matching `setContinuousConversation`.

### [minor] Idempotent row identity depends on data that is never canonicalized before display
**File:** Conduit/Views/VoiceSettingsView.swift:652
**Issue:** `ForEach(phrases, id: \.self)` is only safe if entries are pairwise distinct, but `initialPhrases` is seeded directly from `activeProfileVoicePreferences.spokenStopPhrases`/`spokenEndConversationPhrases`, which are decoded raw from persistence. The write path canonicalizes, but legacy/externally-written blobs (or the pre-existing `spokenStopPhrases` field) can contain duplicate or non-canonical entries, producing duplicate `id`s and undefined row identity. The comment assumes canonicalization that display-side data is not guaranteed to have.
**Suggestion:** Canonicalize with `VoiceSpokenCommands.canonicalizedPhraseList(initialPhrases)` when seeding the editor, or key rows by a stable per-entry UUID.

### [minor] New defaults silently activate a destructive control for existing profiles
**File:** Conduit/Voice/VoiceTypes.swift:136
**Issue:** Older blobs without `spokenEndConversationPhrases` decode to the new defaults (line 138), so every existing profile gets an active End-Conversation category on upgrade without opting in. Because the defaults include very common utterances (`goodbye`, `bye`), users can close the Voice session by simply saying "bye" post-upgrade. This is a product/behavior change, not just a decoding detail.
**Suggestion:** Confirm this is intended; if not, default to an empty list on decode for pre-existing blobs (e.g., distinguish "key absent" and seed empty) and surface the defaults only for fresh profiles.

### [nit] Stored `initialPhrases` is unused after init; `initialPhrases` + `.id()` is a fragile reset pattern
**File:** Conduit/Views/VoiceSettingsView.swift:623
**Issue:** `initialPhrases` is stored but only consumed in `init` to seed `@State`, and `spokenControlsSection` forces view recreation via `.id(spokenStopPhrases)` (lines 480, 487). Any external list change recreates the editor and discards an in-progress `draft`/`editingIndex`. The pattern works today but is brittle and hides state-ownership intent.
**Suggestion:** Drive the editor from a `@Binding`/observed source, or use an explicit `.onChange(of: initialPhrases)` to re-seed state instead of `.id()`, and drop the unused stored property.

## Verdict
COMMENT

No force unwraps, retain cycles, or MainActor violations were found in the added code; the weak-self closures and generation/speech-drain fencing are consistent with the surrounding design. The findings above are non-blocking cleanups, with the dead-code edit handling and the disconnected-write path being the most worthwhile to fix.

<a href="https://opencode.ai/s/ONPPDLm2">New%20session%20-%202026-09-11T10%3A12%3A44.791Z</a>
[opencode session](https://opencode.ai/s/ONPPDLm2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34587994693)

kaishi00:
## Summary
Adds profile-scoped, Settings-editable spoken Stop and End Conversation phrase lists, with a new `VoiceSpokenCommands` canonicalization/matching helper and an `onEndConversation` seam that routes an End utterance through AppState's existing Close teardown.

## Findings

### [minor] Sibling phrase editors can receive identical SwiftUI identities
**File:** Conduit/Views/VoiceSettingsView.swift:480, 487
**Issue:** `SpokenPhraseListEditor` for Stop and for End Conversation are siblings in the same `ConduitSettingsSection` container and are each given an explicit `.id()` derived from their phrase arrays. Explicit `.id` values are expected to be unique among siblings; when both lists are equal — most plausibly when a user empties **both** lists to disable both categories (`[] == []`) — the two editors collide. SwiftUI identity collisions can cause one editor to be treated as the other, drop re-renders, or share reset behavior, which is exactly the state-management path this feature relies on.
**Suggestion:** Namespace the ids by category, e.g. `.id("stop-\(spokenStopPhrases)")` and `.id("end-\(spokenEndConversationPhrases)")`.

### [minor] New End Conversation defaults silently enable a session-closing control for existing profiles
**File:** Conduit/Voice/VoiceTypes.swift:136-138
**Issue:** A pre-existing blob with no `spokenEndConversationPhrases` key decodes to the new defaults (`goodbye`, `bye`, `end conversation`, `that's all`). Every existing profile therefore gains an active End-Conversation category on upgrade without opting in, and `bye` is a very common utterance. This is a behavioral change for existing users, not merely a decode fallback.
**Suggestion:** If this is intended (the PR body implies it is), call it out in release notes. Otherwise distinguish "key absent" from "key present" and seed an empty list for legacy blobs, surfacing defaults only for fresh profiles.

### [minor] Phrase display reads the active profile, but the route is built from the settings snapshot profile
**File:** Conduit/Views/AuxiliaryViews.swift:513, 516
**Issue:** `voicePreferences` is read via `appState.activeProfileVoicePreferences` (keyed by `activeProfile`), while `VoiceSettingsRoute` is constructed with `profile: snapshot.profile`. The snapshot is captured when Settings is presented (`makeSettingsSnapshot`), so if `activeProfile` ever changes while Settings is open, the phrase editors read/write one profile (active) while the rest of the voice settings page addresses another (snapshot). The setters also key off `activeProfile`, so the two can diverge silently.
**Suggestion:** Source the phrase blob from the same profile the route renders (`loadVoiceProfilePreferences(profile: snapshot.profile)`), or assert/derive that the settings route is always the active profile.

### [nit] `initialPhrases` is stored but only used in `init`; `.id()`-driven re-seeding is fragile
**File:** Conduit/Views/VoiceSettingsView.swift:623, 638, 480-487
**Issue:** `initialPhrases` is retained as a stored property that is never read after initialization; state is seeded once and only re-synced by recreating the whole view through `.id()`. Any parent re-render that recomputes the array resets the editor, discarding an in-progress `draft`/`editingIndex` and dropping keyboard focus. The intent (external list changes re-seed the editor) would be clearer and less brittle with an explicit observation.
**Suggestion:** Drop the stored `initialPhrases` and re-seed via `.onChange(of: initialPhrases)` (or drive the editor from a `@Binding`), removing the reliance on `.id()`.

### [nit] Production controller exposes a test-only preferences accessor
**File:** Conduit/Voice/VoiceConversationController.swift:186-191
**Issue:** `activePreferences` was added solely so tests can read back the applied blob. Exposing internal preference state on the production type is a small API-surface smell.
**Suggestion:** Constrain it to test builds (`#if DEBUG`) or have `setProfilePreferences` tests observe via an existing seam.

### [nit] Disconnected phrase edits silently no-op with no editor feedback
**File:** Conduit/Services/AppState.swift:13111-13129
**Issue:** Both setters `guard isConnected else { return }` but return `Void`, so `SpokenPhraseListEditor.save` mutates its local `@State` and shows the new list even though nothing was persisted. If the voice route can ever render while disconnected (e.g. a retained `dashboardTicketBridge`), the UI will display a phrase list that silently isn't saved and will revert on reload.
**Suggestion:** Return a `Bool` (matching `setContinuousConversation`) and have the editor surface a "not connected" state, or gate the editor itself on connectivity.

## Verdict
COMMENT

No force unwraps, retain cycles, or MainActor violations found in the added code: the `[weak self]` closures (`submit`, `interrupt`, `onEndConversation`), the `utteranceTask = nil` self-cancellation fix, and the close-first generation/speech-drain fencing are consistent with the surrounding design. The findings above are non-blocking; the duplicate `.id` identity and the legacy-defaults behavior are the most worth addressing before merge.

<a href="https://opencode.ai/s/TbqxGcA9">New%20session%20-%202026-09-11T10%3A29%3A30.184Z</a>
[opencode session](https://opencode.ai/s/TbqxGcA9)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34589374766)

PR #158, opened 2026-09-10
fix(voice): harden Siri App Intent pending-launch lifecycle

kaishi00:
## Summary

Harden the Siri/App Intent entry point so "Talk to Conduit" either starts the requested voice conversation **exactly once** after Hermes is ready, or **fails visibly** — and never leaves an immortal pending request that fires minutes later on an unrelated reconnect.

**Acceptance invariant:** Once a Siri request has been superseded, cleared, successfully routed, failed, or expired, no async work created for that request can ever mutate a later pending request or surface a stale result.

This PR is intentionally small and limited to the Siri/App Intent launch + pending-intent lifecycle. `VoiceConversationController` remains the single voice runtime. Composer still calls `openVoiceConversation` directly and does not enqueue through this store.

### Old Siri behavior
- `StartVoiceConversationIntent` used only `openAppWhenRun = true` (deprecated on iOS 26).
- `perform()` enqueued a `PendingVoiceIntent` into an in-memory store.
- The root scene gated routing on `appState.isConnected`. A request received while disconnected stayed pending forever and could open Voice much later when connectivity returned.

### New foreground execution API / compatibility
- Deployment target stays **iOS 17.0**.
- iOS 26+: `supportedModes = .foreground(.immediate)`.
- iOS 17–25: `openAppWhenRun` retained as a **boolean literal** (App Intents metadata requires a literal), marked deprecated at the iOS 26 boundary.
- Foreground-first only; `perform()` only records the request.

### Ownership claims (invariant)
- `enqueue` bumps a monotonic `ownershipGeneration` and `revision`.
- `takeClaim()` returns `{ generation, intent }`.
- After any suspension (`openVoiceConversation`), the router revalidates `isClaimCurrent`. Stale completions return `.superseded` and **do not** requeue, fail, or publish `errorMessage`.
- `requeueDeferred(claim)` restores only a still-current claim into an empty slot; never over a newer enqueue; **no revision bump** (no `.task(id:)` hot loop).
- Clear/supersede invalidates the generation, fencing old deadline waiters from replacement requests.

### Authoritative deadline (backstop)
- Siri launches stamp a **30s** budget at enqueue: wall-clock `Date` for tests/identity + **monotonic `ContinuousClock.Instant`** for the actual wait (immune to wall-clock steps).
- The deadline waiter peeks a claim, sleeps on ContinuousClock until that claim’s elapsed deadline, then calls `expireClaimIfCurrent` — **never** re-enters generic readiness (which could return `.waiting` again).
[...]

PR #161, opened 2026-09-11
feat(voice): suspend Voice on background instead of hard teardown

kaishi00:
## Summary

Changes Voice background behavior from **hard teardown** to **logical suspension + safe restoration**. Previously, backgrounding Conduit fully closed an open Voice conversation; now it releases all local runtime ownership while preserving the logical conversation, and a conservative identity-validated restoration reopens it on foreground return — never with a live microphone.

## Previous behavior (verified call path at base `113c2ae`)

- **ScenePhase delivery:** `RootView` → `AppState.handleScenePhase(_:)`.
- **`.background`:** turn-staleness/freshness bookkeeping → `voiceConversationController.setForegroundActive(false)` → `VoiceConversationController.stop()` (operation-generation bump, all tasks cancelled, capture/transcriber/playback stopped, speech drain torn down, explicit user-pause flag cleared, `isVoiceSessionActive = false`, `expectedAssistantSessionIDs = []`, state `.idle`) → `messageReadAloudController.setForegroundActive(false)` → **`showVoiceSheet = false`** (sheet dismissed exactly like a user Close) → reconnect timer cancelled, caches flushed, reconciliation invalidated.
- **`.inactive`** (overlay dips, pre-lock): the same full `stop()` — the sheet stayed but the session was dead (this also silently broke post-dip assistant-event ownership until the next fresh open).
- **`.active`:** `setForegroundActive(true)` (flag only, nothing restarted); the scene task re-establishes the transport via `reconnectForRetry(.automaticReturn)`. The user had to reopen Voice from scratch through `openVoiceConversation(_:)`.
- **Other teardown boundaries:** explicit Close / swipe / spoken End Conversation → `closeVoiceConversation()`; server replacement → `retireSpeechOperationsForServerReplacement` (stop + gateways nil + sheet dismissal); `disconnect()`; `setVoiceEnabled(false)`; `switchProfile` (programmatic only).

## New behavior

**Backgrounding is not the same thing as the user ending Voice.**

- **Suspension** (`handleScenePhase` `.background` with the Voice sheet open): `VoiceConversationController.suspendRuntimeForLifecycle()` — a shared `releaseRuntimeResources()` primitive (extracted from `stop()`; no second teardown) releases capture, transcription, playback, speech drain/stream, barge-in monitoring, and every in-flight task behind the existing operation-generation bump. Suspended runtime state explicitly rejects capture events (`isRuntimeSuspended`) — `hasLiveVoiceSession` remains logical ownership, never a claim that capture is live. The logical conversation is preserved: in-sheet transcript, session ownership (`expectedAssistantSessionIDs`), `hasLiveVoiceSession`, and an explicit user pause. A Hermes turn in flight is **not cancelled**; its voice-side continuation is retired (same convention as barge-in/manual Interrupt) and remembered as a sticky orphan — the next user submission cancels it through the same interrupt seam as Stop, and the orphan flag survives superseded cancel awaits and consecutive suspensions, so its late events can never alias onto the new turn. `.background` additionally records an AppState-owned `SuspendedVoiceConversation` descriptor (profile, session ID, normalized server identity — logical fields only, no resources/leases/tasks, in memory only) and the sheet presentation **survives** (suspension is not Close).
- **Transient `.inactive` dips are not suspension.** Control Center, incoming-call banners, and the microphone permission prompt dip the scene without backgrounding it. Voice is left completely untouched across a dip: the socket and audio session survive, in-flight turns keep flowing, and a pending first-listen (fresh open → permission dialog) is not silently lost. Suspension happens only on the actual `.background` transition.
- **Restoration runs only AFTER foreground reconciliation, with a fresh capability check.** `handleScenePhase(.active)` no longer restores synchronously — the socket, bridge, and runtime session identity may all be stale. The suspended descriptor is consumed inside the existing scene-phase reconciliation task, after the authoritative transport/session state is established: retained healthy transport → health check + session sync → restore; dead socket → reconnect → restore; reconnect/session/auth failure → **never restored** against stale pre-background state (the `canStartVoiceConversation` predicate requires the post-reconciliation connection).  No parallel reconnect mechanism exists for Voice. When a suspended descriptor is pending, the authoritative `refreshVoiceCapabilities()` runs after reconciliation (and attempt currency is re-checked after that await) so restoration validates against a freshly negotiated capability answer; the live in-session Mute is preserved across that refresh per the PR #159 authority rule (the Mute control only changes controller state, so the reloaded persisted blob must not clobber it), and a superseded or boundary-cleared attempt never reapplies it — a gateway that no longer reports TTS/transcription support fails closed exactly like a fresh open would. Foregrounds without a pending suspension never pay a redundant capability refresh.
- **Restoration validation uses the same capability gate as a fresh open.** `restoreSuspendedVoiceConversationIfNeeded()` requires `canStartVoiceConversation` — live connection, voice enabled, usable transcription AND usable speech, exactly what `openVoiceConversation` requires — plus the suspended conversation's identity: same active profile, and the active runtime session is still the suspended conversation either unchanged or as a legitimate rebind of the same durable conversation (resolved through the existing session catalog/alias machinery — `knownSessionIDs`/`alternateIds`; never Voice-only aliasing). Anything stale fails CLOSED. Cross-profile staleness tears down directly WITHOUT persisting the suspended profile's live mute into the now-active profile's preferences (the same protection `switchProfile` and the fallback-profile adoption apply by persisting the outgoing mute under the outgoing profile before the identity flip). A valid descriptor reinstalls the gateway from the current bridge (fail-closed too if one cannot be built) and settles **open but non-listening**; the user's Listen tap resumes capture. Restoration never starts the microphone: the sheet's fresh-open auto-listen is a one-shot AppState-owned intent (`voiceSheetShouldAutoListen`), armed only by `openVoiceConversation`, disarmed by suspension and every destructive boundary, and consumed exactly once — so even a SwiftUI view-identity recreation of a restored sheet cannot hot-start the mic.
- **Destructive boundaries clear the suspended state:** explicit Close (button, swipe, spoken Goodbye — user intent beats suspension), profile switch, authoritative fallback-profile adoption, server replacement, disconnect, forced sign-out (`performSignInRequired`), and voice disabled.

## Stale-async fencing

Reuses the existing `operationGeneration`, `speechDrainRevision`, and capture-generation fences exclusively — no new token system. The Voice interrupt seam is result-bearing (`interruptForVoice` → `cancelCurrent` → the Hermes cancellation path reports real success/failure): the suspended-orphan cancellation clears the orphan only on a successful, still-current cancellation — a failed cancel retains the orphan, fails the submission with an explanatory message, and a later successful retry submits normally. Barge-in, spoken Stop, and spoken End Conversation intentionally proceed on the attempt regardless of the result. After suspension: late ASR completions cannot submit; late TTS/drain completions cannot play or relisten; late assistant events cannot reopen audio; suspended runtime explicitly rejects capture events (`isRuntimeSuspended`, including stale `.interrupted` events from the released runtime); interruption events are generation-bearing end to end — `AVAudioCaptureService` captures the observed capture generation synchronously on the notifying thread and fences BEFORE tearing anything down, so a delayed interruption belonging to a torn-down generation can never stop the live one, release its lease, or fail the session, and the controller additionally rejects interrupted events for foreign generations; a superseded foreground-reconciliation attempt never consumes the descriptor (only a still-current attempt restores), so a newer `.background` re-arming the descriptor is honored by the successor; and an old restoration cannot override a newer Close/Siri/profile/server boundary. The orphan record survives superseded cancel awaits and consecutive suspensions.

## Interaction contracts pinned by tests

- **Continuous Conversation** governs turn-to-turn continuation only — background/foreground restores non-listening with it ON or OFF.
- **Explicit Close / spoken Goodbye** prevents restoration; background/foreground afterwards does not resurrect Voice.
- **User Pause** survives suspension and is restored paused/open; Listen then genuinely recaptures.
- **Stop phrases** are unchanged (cancel turn, stop TTS, keep open, relisten) and are not reinterpreted as suspension.
- **Route safety** unchanged: speaker-safe stays half-duplex in post-restore turns; `VoiceAudioSessionCoordinator`, route classification, VAD, pre-roll, Read Aloud mutual exclusion untouched.
- **Siri** (PR #158 lifecycle untouched): an explicit Siri/composer launch supersedes stale suspension state — `openVoiceConversation`'s first act on a presented sheet is `closeVoiceConversation()`, which clears the descriptor; Case C (closed → Siri) has no stale state by construction.
- The existing 12-second silence mic pause, idle behavior, and all PR #159/#160 semantics are unchanged.

## Explicit non-goals

No continuous background Voice, no WebSocket keepalive, no background microphone, no new background modes or entitlements, no VoIP/PushKit/CallKit/BackgroundTasks. iOS remains free to suspend the process and drop networking; the transport is re-established by the existing scene-phase reconnect, and restoration tolerates socket-dead, reconnected, rebound-runtime, and sign-out states (fail-closed). **Process death intentionally loses the transient suspension state** (the descriptor is in-memory only) — the durable Hermes conversation remains recoverable through normal app/session behavior. Idle Voice auto-close and VAD/route/provider changes are untouched.

## Tests

TDD — the suite was committed first (red build on the missing API), then implemented. `ConduitTests/VoiceBackgroundSuspensionTests.swift`: 47 deterministic tests across controller-level suspension semantics (identity preservation, pause survival, orphan retirement/cancel, consecutive-suspension orphan retention, late-transcription fencing, post-restore route safety) and AppState-level lifecycle (descriptor record/consume, dip vs background routing, capability-loss fail-closed, explicit Close/Goodbye precedence, profile/server-identity/disconnect/voice-disabled invalidation, signed-out fail-closed `.active` wiring, full post-restore conversation, auto-listen consume-once, mute preservation).

Also run on the Mac build box: `VoiceConversationControllerTests`, `ContinuousConversationPreferenceTests`, `AppStateContinuousConversationPreferenceTests`, the PR #160 spoken-control suites, `PendingVoiceIntentLifecycleTests` (PR #158), `AppStateServerReplacementSpeechTests`, `VoiceAudioSessionCoordinatorTests`, `VoiceSpeakerSafeBargeInTests`, `HapticsVoiceIsolationTests`, `AppStateReadAloudTests`, and the full `ConduitTests` unit suite in planner lanes. Results recorded in the PR comments.

## Review

Three-model review ran twice on this branch. Round 1 (2× REQUEST-CHANGES → fixed, 1× APPROVE): view-identity hot-mic path (fixed via the AppState-owned auto-listen intent), profile-switch stranding, capability-loss fail-closed, same-session event aliasing (fixed via the orphan-cancel), test-double race, `.inactive` coverage. Round 2 (2× REQUEST-CHANGES → fixed, 1× APPROVE): orphan flag clobbered by consecutive suspensions (sticky-OR), profile-switch mute cross-contamination via the deferred `onDismiss` (fixed by persisting the outgoing mute before the identity flip plus a `hasLiveVoiceSession` authority guard on close-time persistence), forced-sign-out immediate teardown. Round 3 (three-model, on the focused correctness pass): restoration deferred to post-reconciliation with a currency guard (superseded attempts never consume the descriptor), fresh-open supersession of stale descriptors, `canStartVoiceConversation` capability parity (an earlier draft had declined a `supportsSpeech` restore gate on the incorrect claim that fresh opens do not require speech — corrected: `voiceUnavailableReason` requires both transcription and speech), `.inactive` reverted to a no-op for Voice, cross-profile fail-closed mute protection plus `adoptAuthoritativeFallbackProfile` teardown, sticky-orphan survival across superseded cancel awaits, and the explicit `isRuntimeSuspended` capture gate (with the un-pause path clearing it).

## Physical-device acceptance: **outstanding**

1. Start Voice on iPhone, speak normally. 2. Lock the device. 3. Unlock and return. 4. Voice returns open with the mic NOT hot. 5. Tap Listen and continue the same conversation. 6. Repeat after switching apps for several minutes. 7. Repeat with network loss while backgrounded. 8. Repeat on AirPods/headset and on speaker. 9. Spoken Goodbye followed by background/foreground does not resurrect Voice. 10. Battery/privacy indicators disappear while suspended.

### DIFF
--- a/Conduit/Services/AppState.swift
+++ b/Conduit/Services/AppState.swift
@@ -810,12 +810,49 @@ final class AppState: ObservableObject {
             return await self.submitVoiceTranscript(transcript)
         },
         interrupt: { [weak self] in
-            await self?.interruptForVoice()
+            guard let self else { return false }
+            return await self.interruptForVoice()
         },
         onEndConversation: { [weak self] in
             self?.closeVoiceConversation()
         }
     )
+    /// Logical identity of a Voice conversation suspended by an app lifecycle
+    /// transition (scene backgrounded / device locked). Identity only: it
+    /// never holds audio resources, gateway or transport references, session
+    /// leases, or task handles. Recorded when the scene deactivates with the
+    /// Voice sheet open, consumed by foreground restoration, and cleared by
+    /// every destructive Voice boundary (explicit Close, spoken End
+    /// Conversation, profile switch, server replacement, disconnect, forced
+    /// sign-out, voice disabled). In memory only — process death
+    /// intentionally loses it.
+    struct SuspendedVoiceConversation: Equatable {
+        let profile: String
+        let sessionID: String
+        let serverIdentity: String?
+    }
+    @Published private(set) var suspendedVoiceConversation: SuspendedVoiceConversation?
+    /// One-shot intent for the Voice sheet's auto-listen on fresh
+    /// presentation. Armed only by `openVoiceConversation` and consumed
+    /// exactly once by the sheet, so a restored suspended conversation can
+    /// never hot-start the microphone when SwiftUI recreates the sheet's
+    /// view identity. Lifecycle suspension and every destructive Voice
+    /// boundary disarm it.
+    internal(set) var voiceSheetShouldAutoListen = false
+
+    func consumeVoiceSheetAutoListen() -> Bool {
+        guard voiceSheetShouldAutoListen else { return false }
+        voiceSheetShouldAutoListen = false
+        return true
+    }
+    /// The profile the voice conversation controller's current session was
+    /// opened for. Close-time mute persistence is keyed on this rather than
+    /// on session liveness alone: an audio interruption or any boundary that
+    /// stops the controller must still let a same-profile Close persist the
+    /// user's mute, while a Close arriving after a profile flip (the sheet's
+    /// deferred `onDismiss`) must not write the outgoing profile's mute into
+    /// the now-active profile's blob.
+    internal(set) var voiceControllerSessionProfile: String?
     /// Manual per-message read aloud for completed assistant responses.
     /// TTS-only: independent of the voice conversation and of STT.
     lazy var messageReadAloudController = MessageReadAloudController(
@@ -2101,6 +2138,11 @@ final class AppState: ObservableObject {
         // from its own bridge (capability refresh, read-aloud assignment).
         voiceConversationController.setGateway(nil)
         readAloudGatewayBridge = nil
+        // Server replacement is a destructive Voice boundary: a suspended
+        // conversation from the outgoing server must never be restored.
+        suspendedVoiceConversation = nil
+        voiceSheetShouldAutoListen = false
+        voiceControllerSessionProfile = nil
         showVoiceSheet = false
     }
 
@@ -2782,6 +2824,9 @@ final class AppState: ObservableObject {
         // never open a stream again, so it must not survive the re-login.
         messageReadAloudController.setGateway(nil)
         readAloudGatewayBridge = nil
+        suspendedVoiceConversation = nil
+        voiceSheetShouldAutoListen = false
+        voiceControllerSessionProfile = nil
         showVoiceSheet = false
         voiceCapabilitySnapshot = .unavailable
         isVoiceEnabled = false
@@ -2949,6 +2994,15 @@ final class AppState: ObservableObject {
         messageReadAloudController.stop()
         messageReadAloudController.setGateway(nil)
         readAloudGatewayBridge = nil
+        // Forced sign-out supersedes any suspended Voice conversation (the
+        // restore-time connection identity check would fail closed anyway);
+        // stop the controller and drop the sheet immediately like
+        // Disconnect does instead of waiting for the next scene cycle.
+        suspendedVoiceConversation = nil
+        voiceSheetShouldAutoListen = false
+        voiceControllerSessionProfile = nil
+        voiceConversationController.stop()
+        showVoiceSheet = false
         projects = []
         supportsProjects = false
         projectsLoading = false
@@ -5206,6 +5260,11 @@ final class AppState: ObservableObject {
             isSceneActive = true
             voiceConversationController.setForegroundActive(true)
             messageReadAloudController.setForegroundActive(true)
+            // Voice restoration deliberately does NOT run here: the socket,
+            // bridge, and runtime session identity may all be stale after a
+            // background suspension. The suspended descriptor is consumed
+            // inside the reconciliation task below, after the authoritative
+            // transport/session state is established (or definitively fails).
             // Consume the background arming even while signed out, so a
             // background → active cycle on the login screen doesn't surface
             // a stale request after the user signs back in.
@@ -5276,13 +5335,13 @@ final class AppState: ObservableObject {
                             // session.
                             self.settleReconciliation(token)
                             await self.reconnectForRetry(purpose: .preserveCurrent)
-                            return
+                        } else {
+                            lifecycleLog.notice(
+                                "Foreground refresh: health check failed (\(error.localizedDescription, privacy: .private)); reconnecting"
+                            )
+                            await self.reconnectForRetry(purpose: .automaticReturn)
+                            self.settleReconciliation(token)
                         }
-                        lifecycleLog.notice(
-                            "Foreground refresh: health check failed (\(error.localizedDescription, privacy: .private)); reconnecting"
-                        )
-                        await self.reconnectForRetry(purpose: .automaticReturn)
-                        self.settleReconciliation(token)
                     }
                 } else {
                     guard self.scenePhaseAttemptIsCurrent(sceneAttemptID) else {
@@ -5292,13 +5351,25 @@ final class AppState: ObservableObject {
                     if !self.automaticChatResumeWorkIsCurrent(automaticWorkToken) {
                         self.settleReconciliation(token)
                         await self.reconnectForRetry(purpose: .preserveCurrent)
-                        return
+                    } else {
+                        lifecycleLog.notice(
+                            "Foreground refresh start: transport=missing-or-unhealthy; reconnecting"
+                        )
+                        await self.reconnectForRetry(purpose: .automaticReturn)
+                        self.settleReconciliation(token)
                     }
-                    lifecycleLog.notice(
-                        "Foreground refresh start: transport=missing-or-unhealthy; reconnecting"
+                }
+                // Voice restoration runs here — after the authoritative
+                // reconciliation above established transport and session
+                // state — and only while this attempt is still current. A
+                // pending suspended conversation forces a capability refresh
+                // first: the pre-background snapshot cannot be trusted across
+                // suspension, and restoration must fail closed when the
+                // refreshed state says a fresh open would be rejected.
+                if self.scenePhaseAttemptIsCurrent(sceneAttemptID) {
+                    await self.refreshCapabilitiesAndRestoreSuspendedVoice(
+                        isCurrent: { self.scenePhaseAttemptIsCurrent(sceneAttemptID) }
                     )
-                    await self.reconnectForRetry(purpose: .automaticReturn)
-                    self.settleReconciliation(token)
                 }
             }
             scenePhaseTask = task
@@ -5333,9 +5404,12 @@ final class AppState: ObservableObject {
                 locallyOwnedInFlightTurn = nil
             }
             foregroundFreshnessCheckArmed = true
-            voiceConversationController.setForegroundActive(false)
             messageReadAloudController.setForegroundActive(false)
-            showVoiceSheet = false
+            // Suspension is not Close: an open Voice conversation releases its
+            // runtime ownership and is recorded for foreground restoration,
+            // and the sheet presentation intentionally survives the
+            // background transition.
+            suspendVoiceConversationForBackground()
             // Drop any armed reconnect timer as well: in-flight cycles abort
             // at their next transportContinuation checkpoint, and foreground
             // activation re-establishes the transport.
@@ -5372,8 +5446,14 @@ final class AppState: ObservableObject {
             // transition too, rather than at its next checkpoint.
             cancelScenePhaseAttempt()
             chatResumeCoordinator.freezeViewport()
-            voiceConversationController.setForegroundActive(false)
             messageReadAloudController.setForegroundActive(false)
+            // A transient .inactive dip (Control Center, an incoming-call
+            // banner, the microphone permission prompt) is NOT a lifecycle
+            // suspension: the socket and audio session survive, in-flight
+            // turns keep flowing, and a pending first-listen must survive the
+            // dip. Voice suspension happens only on the actual .background
+            // transition — a descriptor-less suspension here would tear the
+            // runtime down with no restoration path.
             return nil
 
         @unknown default:
@@ -5383,6 +5463,110 @@ final class AppState: ObservableObject {
 
     private func scenePhaseAttemptIsCurrent(_ id: UUID) -> Bool {
         !Task.isCancelled && scenePhaseAttemptID == id
+    }
+
+    /// Background transition for the Voice conversation. An open Voice sheet
+    /// routes through the controller's lifecycle suspension — runtime
+    /// ownership released, logical conversation preserved — and records the
+    /// suspended descriptor for foreground restoration. A closed Voice sheet
+    /// keeps the existing full-stop semantics. Transient .inactive dips never
+    /// reach this: they leave Voice untouched.
+    private func suspendVoiceConversationForBackground() {
+        // Suspension disarms the fresh-open auto-listen intent: the restored
+        // sheet must never hot-start the microphone, even if SwiftUI
+        // recreates the sheet's view identity while the binding stays true.
+        voiceSheetShouldAutoListen = false
+        voiceConversationController.setForegroundActive(false)
+        guard showVoiceSheet else {
+            voiceConversationController.stop()
+            return
+        }
+        voiceConversationController.suspendRuntimeForLifecycle()
+        suspendedVoiceConversation = SuspendedVoiceConversation(
+            profile: activeProfile,
+            sessionID: activeSessionId ?? "",
+            serverIdentity: Self.voiceSuspensionServerIdentity(connection)
+        )
+    }
+
+    /// Consumes the suspended Voice descriptor on foreground return. Called
+    /// ONLY after the foreground reconciliation task has established the
+    /// authoritative transport/session state (or definitively failed) — never
+    /// synchronously at the scene transition, when the socket, bridge, and
+    /// runtime session identity may all still be stale.
+    ///
+    /// Validation is synchronous against that post-reconciliation state: the
+    /// same authoritative predicate as a fresh `openVoiceConversation`
+    /// (`canStartVoiceConversation`: live connection, voice enabled, usable
+    /// transcription AND speech) plus the suspended conversation's identity —
+    /// same active profile, and the active runtime session is still the
+    /// suspended conversation (unchanged, or a legitimate rebind per the
+    /// session catalog/alias machinery). Anything stale fails CLOSED. A valid
+    /// descriptor reinstalls the gateway and settles open-but-NON-listening;
+    /// capture is NEVER started by restoration (Continuous Conversation
+    /// governs turn-to-turn continuation only, not lifecycle transitions).
+    func restoreSuspendedVoiceConversationIfNeeded() {
+        guard let suspended = suspendedVoiceConversation else { return }
+        suspendedVoiceConversation = nil
+        let conversationStillMatches = suspended.sessionID == activeSessionId
+            || knownSessionIDs(for: suspended.sessionID).contains(activeSessionId ?? "")
+        let identityStillValid = suspended.profile == activeProfile
+            && !suspended.sessionID.isEmpty
+            && conversationStillMatches
+            && connection != nil
+            && suspended.serverIdentity == Self.voiceSuspensionServerIdentity(connection)
+            && canStartVoiceConversation
+            && showVoiceSheet
+            && voiceConversationController.hasLiveVoiceSession
+        guard identityStillValid else {
+            if showVoiceSheet {
+                if suspended.profile == activeProfile {
+                    closeVoiceConversation()
+                } else {
+                    // Cross-profile staleness: the suspended controller's
+                    // live mute belongs to the descriptor's profile. Tear
+                    // down directly — closeVoiceConversation would persist
+                    // that mute into the now-active profile's preferences.
+                    voiceConversationController.stop()
+                    voiceSheetShouldAutoListen = false
+                    voiceControllerSessionProfile = nil
+                    showVoiceSheet = false
+                }
+            }
+            return
+        }
+        // Rebuild the gateway from the current bridge/connection; the
+        // controller keeps its preserved conversation identity and
+        // preferences, settled in .idle.
+        refreshVoiceControllerGateway()
+        // The capability checks above passed, so a gateway that still could
+        // not be built (e.g. the bridge rotated under us) is staleness too:
+        // fail closed instead of restoring an open-but-dead sheet.
+        if !voiceConversationController.isGatewayAttached {
+            closeVoiceConversation()
+        }
+    }
+
+    /// Refreshes Voice capability state when a suspended Voice conversation
+    /// is pending, then validates/restores. The pre-background capability
+    /// snapshot cannot be trusted across suspension, so restoration must run
+    /// against a freshly negotiated answer; `isCurrent` is re-checked after
+    /// that await so a superseded attempt never consumes the descriptor.
+    /// Without a pending descriptor this is a no-op — a plain foreground
+    /// return never pays a redundant capability network refresh here.
+    func refreshCapabilitiesAndRestoreSuspendedVoice(isCurrent: @MainActor () -> Bool) async {
+        guard suspendedVoiceConversation != nil else { return }
+        // The central refreshVoiceCapabilities authority preserves the live
+        // in-session mute for the owning profile during this refresh, so the
+        // restored session keeps the user's Mute state.
+        await refreshVoiceCapabilities()
+        guard isCurrent() else { return }
+        restoreSuspendedVoiceConversationIfNeeded()
+    }
+
+    private static func voiceSuspensionServerIdentity(_ connection: HermesConnection?) -> String? {
+        guard let baseURL = connection?.baseUrl else { return nil }
+        return normalizedChatResumeServerIdentity(baseURL)
     }
 
     private func cancelScenePhaseAttempt() {
@@ -9448,12 +9632,17 @@ final class AppState: ObservableObject {
         }
     }
 
-    func cancelCurrent() async {
-        guard isBusy else { return }
+    /// Cancels the authoritative Hermes turn, reporting whether the
+    /// cancellation succeeded. "Success" also covers nothing being left to
+    /// cancel (the turn already settled server-side).
+    @discardableResult
+    func cancelCurrent() async -> Bool {
+        guard isBusy else { return true }
         let submissionContext = composerSubmissionContext()
-        guard await interruptForReplacement(context: submissionContext) else { return }
-        guard let currentContext = currentComposerSubmissionContextIfOwnedAndAliased(submissionContext) else { return }
+        guard await interruptForReplacement(context: submissionContext) else { return false }
+        guard let currentContext = currentComposerSubmissionContextIfOwnedAndAliased(submissionContext) else { return true }
         await recoverComposerSubmission(using: currentContext)
+        return true
     }
 
     /// Answers one question of a clarification request. Batch questions route
@@ -10669,6 +10858,22 @@ final class AppState: ObservableObject {
             connection = freshConnection
             client = nextClient
             clearPendingDecisionRestorationGuard()
+            // A profile switch ends any live or suspended Voice conversation
+            // from the outgoing profile. Persist the outgoing mute and tear
+            // down BEFORE the identity flip: the sheet's deferred onDismiss
+            // re-enters closeVoiceConversation after the switch, and its
+            // hasLiveVoiceSession guard makes that a no-op so the outgoing
+            // mute can never be persisted under the target profile.
+            if showVoiceSheet {
+                var preferences = loadVoiceProfilePreferences(profile: activeProfile)
+                preferences.outputMuted = voiceConversationController.isOutputMuted
+                saveVoiceProfilePreferences(preferences, profile: activeProfile)
+            }
+            voiceConversationController.stop()
+            showVoiceSheet = false
+            suspendedVoiceConversation = nil
+            voiceSheetShouldAutoListen = false
+            voiceControllerSessionProfile = nil
             // Fence any residual deferred cache write scheduled under the
             // outgoing profile before identities and namespaces change over.
             setActiveProfile(target)
@@ -10846,11 +11051,29 @@ final class AppState: ObservableObject {
     /// filtered and path-prefixed by the explicit profile, so no
     /// cross-profile transcript or cache state leaks in the window before
     /// that reconnect lands.
-    private func adoptAuthoritativeFallbackProfile(_ fallback: String) {
+    ///
+    /// Internal for tests: the fallback adoption is a real destructive Voice
+    /// boundary and the cross-profile fail-closed restore contract is pinned
+    /// through it.
+    func adoptAuthoritativeFallbackProfile(_ fallback: String) {
         guard fallback != activeProfile else { return }
         // A profile change replaces the voice gateway; any in-flight read
         // aloud belongs to the outgoing profile (same as switchProfile).
         messageReadAloudController.stop()
+        // A profile re-home ends any live or suspended Voice conversation
+        // from the outgoing profile. Persist the outgoing mute and tear down
+        // BEFORE the identity flip so it lands in the outgoing profile's
+        // blob, never the fallback's (same contract as switchProfile).
+        if showVoiceSheet {
+            var preferences = loadVoiceProfilePreferences(profile: activeProfile)
+            preferences.outputMuted = voiceConversationController.isOutputMuted
+            saveVoiceProfilePreferences(preferences, profile: activeProfile)
+        }
+        voiceConversationController.stop()
+        showVoiceSheet = false
+        suspendedVoiceConversation = nil
+        voiceSheetShouldAutoListen = false
+        voiceControllerSessionProfile = nil
         flushPendingPresentationCache()
         sessions = []
         cronSessions = []
@@ -12939,9 +13162,10 @@ final class AppState: ObservableObject {
         await submitComposer(text: transcript, attachments: [])
     }
 
-    /// Stops the authoritative Hermes turn when a spoken stop command or
-    /// barge-in wins the race with model generation.
-    func interruptForVoice() async {
+    /// Stops the authoritative Hermes turn when a spoken stop command,
+    /// barge-in, or Voice orphan cancellation wins the race with model
+    /// generation. Reports whether the cancellation succeeded.
+    func interruptForVoice() async -> Bool {
         await cancelCurrent()
     }
 
@@ -13042,7 +13266,16 @@ final class AppState: ObservableObject {
         await service.reload()
         guard profile == activeProfile, bridge === dashboardTicketBridge else { return }
         voiceCapabilitySnapshot = service.snapshot.capability
-        let preferences = loadVoiceProfilePreferences(profile: profile)
+        var preferences = loadVoiceProfilePreferences(profile: profile)
+        // PR #159 authority, enforced centrally: while the controller owns a
+        // live Voice session for THIS profile, its live mute is
+        // authoritative over the persisted blob. Ownership is keyed on the
+        // profile the session was opened for, so a refresh for another
+        // profile never carries a live mute across ownership.
+        if voiceConversationController.hasLiveVoiceSession,
+           voiceControllerSessionProfile == profile {
+            preferences.outputMuted = voiceConversationController.isOutputMuted
+        }
         voiceTranscriptionMode = preferences.resolvedTranscriptionMode
         continuousConversationEnabled = preferences.continuousConversation
         voiceConversationController.setProfilePreferences(preferences)
@@ -13059,6 +13292,9 @@ final class AppState: ObservableObject {
         refreshReadAloudGateway()
         if !enabled {
             voiceConversationController.stop()
+            suspendedVoiceConversation = nil
+            voiceSheetShouldAutoListen = false
+            voiceControllerSessionProfile = nil
             showVoiceSheet = false
         }
         return true
@@ -13190,18 +13426,39 @@ final class AppState: ObservableObject {
             errorMessage = "Hermes could not prepare a voice conversation."
             return true
         }
+        // A fresh user-initiated open supersedes any stale suspension (e.g.
+        // a descriptor retained across sign-out, or one whose reconciliation
+        // task is still in flight): it must never fail-close this new
+        // session on a later scene cycle.
+        suspendedVoiceConversation = nil
         voiceConversationController.setGateway(gateway)
         voiceConversationController.beginVoiceTurn(sessionID: sessionID)
+        voiceControllerSessionProfile = activeProfile
         showSidebar = false
+        voiceSheetShouldAutoListen = true
         showVoiceSheet = true
         return true
     }
 
     func closeVoiceConversation() {
-        var preferences = loadVoiceProfilePreferences(profile: activeProfile)
-        preferences.outputMuted = voiceConversationController.isOutputMuted
-        saveVoiceProfilePreferences(preferences, profile: activeProfile)
+        // Explicit Close (button, swipe, spoken End Conversation) beats any
+        // lifecycle suspension: the conversation is not restorable.
+        suspendedVoiceConversation = nil
+        voiceSheetShouldAutoListen = false
+        // Live-mute persistence obeys the PR #159 authority rule: the
+        // controller's mute wins only for the profile that owns the session.
+        // A same-profile close that arrives after the controller was
+        // already torn down (e.g. audio interruption) still persists; a
+        // close arriving after a profile flip (the sheet's deferred
+        // onDismiss) persists nothing.
+        if voiceConversationController.hasLiveVoiceSession
+            || voiceControllerSessionProfile == activeProfile {
+            var preferences = loadVoiceProfilePreferences(profile: activeProfile)
+            preferences.outputMuted = voiceConversationController.isOutputMuted
+            saveVoiceProfilePreferences(preferences, profile: activeProfile)
+        }
         voiceConversationController.endVoiceSession()
+        voiceControllerSessionProfile = nil
         showVoiceSheet = false
     }
 
--- a/Conduit/Views/RootView.swift
+++ b/Conduit/Views/RootView.swift
@@ -85,7 +85,8 @@ struct MainView: View {
             VoiceConversationSheet(
                 controller: appState.voiceConversationController,
                 profile: appState.activeProfile,
-                onClose: appState.closeVoiceConversation
+                onClose: appState.closeVoiceConversation,
+                shouldAutoListen: { appState.consumeVoiceSheetAutoListen() }
             )
                 .presentationDetents([.large])
                 .presentationDragIndicator(.visible)
--- a/Conduit/Views/VoiceConversationSheet.swift
+++ b/Conduit/Views/VoiceConversationSheet.swift
@@ -11,6 +11,11 @@ struct VoiceConversationSheet: View {
     @ObservedObject var controller: VoiceConversationController
     let profile: String
     let onClose: () -> Void
+    /// One-shot gate armed by a fresh `openVoiceConversation` and consumed
+    /// exactly once. A restored suspended conversation deliberately does NOT
+    /// auto-listen: the gate stays disarmed even if SwiftUI recreates this
+    /// view's identity while the presentation binding survives.
+    let shouldAutoListen: () -> Bool
     var startsListening: Bool = true
 
     @State private var didRequestStart = false
@@ -41,6 +46,7 @@ struct VoiceConversationSheet: View {
         .task {
             guard startsListening, !didRequestStart else { return }
             didRequestStart = true
+            guard shouldAutoListen() else { return }
             await controller.startListening()
         }
         .accessibilityElement(children: .contain)
--- a/Conduit/Voice/AVAudioCaptureService.swift
+++ b/Conduit/Voice/AVAudioCaptureService.swift
@@ -6,6 +6,7 @@
 import AVFAudio
 import Foundation
 import OSLog
+import os
 
 private let voiceAudioLogger = Logger(subsystem: "com.milim.relay", category: "VoiceAudio")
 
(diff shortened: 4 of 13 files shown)


kaishi00:
## Summary
This PR changes Voice background handling from a hard teardown to a logical suspension (`VoiceConversationController.suspendRuntimeForLifecycle()`) plus an identity-validated foreground restoration (`AppState.restoreSuspendedVoiceConversationIfNeeded()`), with a one-shot auto-listen gate so a restored sheet never hot-starts the mic.

## Findings

### [severity: critical] Restoration is less strict than opening: `supportsSpeech` is not validated
**File:** Conduit/Services/AppState.swift:5475-5483 (and PR rationale)
**Issue:** `restoreSuspendedVoiceConversationIfNeeded()` gates on `isVoiceEnabled && selectedTranscriptionIsAvailable`, but never checks `voiceCapabilitySnapshot.supportsSpeech`. Opening Voice does require speech: `openVoiceConversation` (AppState.swift:13311) guards on `canStartVoiceConversation` → `voiceUnavailableReason`, which explicitly returns when `!voiceCapabilitySnapshot.supportsSpeech` (AppState.swift:13105-13107, present at base `113c2ae`). The PR body's stated reason for declining a speech gate ("`voiceUnavailableReason` requires transcription but not speech") is factually wrong for this code. Net effect: a conversation whose TTS provider disappeared while suspended is restored open, even though a fresh open would have been refused; the user gets a live-looking sheet whose assistant replies cannot be voiced.
**Suggestion:** Add `&& voiceCapabilitySnapshot.supportsSpeech` to `identityStillValid`, and add a test mirroring `testCapabilityLossWhileSuspendedFailsClosed` with `supportsTranscription: true, supportsSpeech: false`. If the intent really is text-only Voice, then `voiceUnavailableReason` should be fixed first and the inconsistency removed, not silently tolerated in the restore path.

### [severity: major] A descriptor-less `.inactive` dip disarms auto-listen and nothing re-arms or resumes it
**File:** Conduit/Services/AppState.swift:5398-5425, 5443-5459; Conduit/Voice/VoiceConversationController.swift:249
**Issue:** `.inactive` routes through `suspendVoiceConversationForSceneDeactivation(recordsDescriptor: false)`, which always disarms `voiceSheetShouldAutoListen` and tears down runtime, but records no descriptor. On return, `restoreSuspendedVoiceConversationIfNeeded()` no-ops, so the sheet stays open in `.idle` and capture never resumes. The code comment at AppState.swift:5422-5424 ("the conversation simply resumes foreground-side") is inaccurate: nothing restarts listening. Concretely, the first-time microphone permission prompt is triggered *inside* `startListening()` (`capture.requestPermission()`); if the system alert drives the scene `.inactive`, the pending listen is fenced by the generation bump and the gate is consumed/disarmed, leaving a fresh conversation open but not listening until the user taps Listen. Control Center, notification banners, and incoming-call overlays hit the same path.
**Suggestion:** Distinguish a transient overlay dip from a real background transition (e.g., re-arm the auto-listen intent when no descriptor is recorded, or only suspend/disarm on `.background`), so a dip does not silently drop an in-progress listening turn.

### [severity: minor] Suspension keeps `isVoiceSessionActive == true`, defeating the `captureLive` guard in `handleCaptureEvent`
**File:** Conduit/Voice/VoiceConversationController.swift:380-390, 701-705
**Issue:** `releaseRuntimeResources()` stops capture but `suspendRuntimeForLifecycle()` deliberately does not clear `isVoiceSessionActive`, so `captureLive` computes as `true` while suspended. A late `.level` frame whose generation happens to match (the only remaining fence) is accepted, publishing `microphoneLevel`/running `ingestAudioLevel` as if capture were live. It is benign today only because suspension forces `state == .idle`, which makes `ingestAudioLevel` a no-op and hides the meter — a fragile invariant for a code path whose whole purpose is "no live capture."
**Suggestion:** Add an explicit `isLifecycleSuspended` flag checked in `handleCaptureEvent`, or preserve logical ownership (`hasLiveVoiceSession`) via a separate flag instead of leaving `isVoiceSessionActive` true.

### [severity: minor] Restore validates session identity before the `.active` sync can reassign it, with no re-validation
**File:** Conduit/Services/AppState.swift:5249-5251 vs 5280-5345
**Issue:** `restoreSuspendedVoiceConversationIfNeeded()` runs synchronously before the scene task's `reconnectForRetry(.automaticReturn)`/`syncSession`. It validates `suspended.sessionID == activeSessionId`, but the sync that runs immediately afterward may adopt a different session. If that happens, the restored sheet stays open with `expectedAssistantSessionIDs` still bound to the old session; `extendAssistantSessionIDs` (AppState.swift:3718) only extends on positive overlap, so the new session's assistant events would be dropped.
**Suggestion:** Re-validate (or fail closed) once the scene attempt settles, rather than only at the synchronous restore point.

### [severity: nit] Orphan marker cleared before the interrupt await
**File:** Conduit/Voice/VoiceConversationController.swift:803-807
**Issue:** `suspendedInFlightTurnOrphaned` is set to `false` immediately before `await interrupt()`. If that await is cancelled by a subsequent suspension/close and `isCurrent(generation)` fails, the orphan marker is already gone even though the cancel may not have completed, so the aliasing protection can be lost.
**Suggestion:** Clear the marker only after the interrupt returns successfully (and keep it set on the cancelled path).

### [severity: nit] Overstated/incorrect comments versus behavior
**File:** Conduit/Services/AppState.swift:5422-5424
**Issue:** The comment claims a descriptorless dip "simply resumes foreground-side," which no code does. Combined with the incorrect `supportsSpeech` rationale in the PR description, reviewers may be misled about actual guarantees.
**Suggestion:** Align the comments with the code, or make the code match the stated intent.

## Verdict
REQUEST_CHANGES

The core suspension/restoration design and test coverage are strong, but the missing `supportsSpeech` restore gate is a real open-vs-restore inconsistency (justified by an incorrect premise), and the `.inactive` path's silent loss of the listening turn contradicts its own documentation.

<a href="https://opencode.ai/s/9uKNoE2Q">New%20session%20-%202026-09-11T16%3A46%3A00.150Z</a>
[opencode session](https://opencode.ai/s/9uKNoE2Q)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34623666053)

kaishi00:
## Summary
This PR replaces hard Voice teardown on backgrounding with a logical suspension (`suspendRuntimeForLifecycle`) plus identity-validated foreground restoration, and adds a one-shot AppState-owned auto-listen gate so a restored sheet never hot-starts the mic. The design and test breadth are solid, but I found three functional gaps in the new lifecycle paths.

## Findings

### [severity: major] Restored user-pause resume leaves `isRuntimeSuspended` set, silently killing the microphone
**File:** Conduit/Voice/VoiceConversationController.swift:311-325 (and guard at :715)
**Issue:** `resumeMicrophone()` only clears `isRuntimeSuspended` indirectly via `startListening()` (line 254). When the user had an explicit pause active at suspension time, `isMicrophonePaused == true`, so the restore path takes the pause branch (`capture.resume()` at :313) and never clears the flag. After that, `handleCaptureEvent` returns at `guard !isRuntimeSuspended` (:715) for every `.level` frame: the visible state becomes `.listening`, the meter reads active, but `ingestAudioLevel`/VAD never runs, so the user speaks and nothing happens. This directly contradicts the stated contract "User Pause survives suspension … Listen then genuinely recaptures", and `testSuspensionPreservesExplicitUserPause` only asserts `state == .listening`/`resumeCount`, so it does not catch it.
**Suggestion:** Clear (or re-arm) `isRuntimeSuspended` on every capture restart, including the paused `resumeMicrophone()` branch — e.g. `isRuntimeSuspended = false` after a successful `capture.resume()`, or route both branches through a shared re-arm helper. Add a test that, after a restored pause + `resumeMicrophone()`, emits a capture `.level` event and asserts it reaches VAD/transcription.

### [severity: major] A late `.interrupted` capture event during suspension destroys the preserved logical session
**File:** Conduit/Voice/VoiceConversationController.swift:732-733 (and `failForAudioInterruption` at :915-941)
**Issue:** The `.level` case is fenced by `isRuntimeSuspended`, but `.interrupted` has no generation or suspension guard. `AVAudioCaptureService` registers its interruption observer in `init` and only removes it in `deinit` (AVAudioCaptureService.swift:65-79), and its `stop()` does not cancel the event stream — `handleInterruption` yields `.interrupted` after calling `stop()` (AVAudioCaptureService.swift:344-351). The controller's `captureEventsTask` runs for the controller's lifetime (:133-138). So a phone call / Siri / alarm while Voice is suspended runs `failForAudioInterruption()`, which clears `isVoiceSessionActive`, `expectedAssistantSessionIDs`, and `suspendedInFlightTurnOrphaned` and sets `.failed`. On foreground, `hasLiveVoiceSession` is now false, so restoration fails closed and closes the sheet — the suppressed conversation is discarded in exactly the scenario the feature targets.
**Suggestion:** Gate the `.interrupted` case on the suspended state (e.g. `guard !isRuntimeSuspended else { return }`) and/or the capture generation, and add a test that emits `.interrupted` while suspended and asserts the logical identity survives.

### [severity: major] Restoration defer runs for cancelled/superseded scene attempts and can consume the fresh descriptor
**File:** Conduit/Services/AppState.swift:5293 (defer) with :5395/:5406
**Issue:** `defer { self.restoreSuspendedVoiceConversationIfNeeded() }` runs unconditionally, including when the attempt was superseded. On `.background`, `suspendVoiceConversationForBackground()` records the descriptor (:5395), then `cancelScenePhaseAttempt()` (:5406) cancels the in-flight `.active` task. That cancelled task's defer then consumes and validates the brand-new descriptor while still backgrounded, before any foreground reconciliation. The comment at :5290-5292 ("a superseded attempt's restore is a no-op: newer boundaries clear the descriptor") is inverted for this sequence — the transition that cancelled the attempt is the one that just set the descriptor. Depending on live state at that moment, the early restore either attaches a possibly-stale gateway and leaves nothing for the real foreground return, or fails closed and closes the sheet. No test covers backgrounding while a foreground reconciliation is in flight.
**Suggestion:** Gate the restore defer on attempt currency (e.g. `defer { if self.scenePhaseAttemptIsCurrent(sceneAttemptID) { self.restoreSuspendedVoiceConversationIfNeeded() } }`), or only restore from the `.active` handler. Add a test that parks the foreground reconciliation, backgrounds, and asserts the descriptor is not consumed by the cancelled attempt.

### [severity: nit] Duplicated doc comment on `adoptAuthoritativeFallbackProfile`
**File:** Conduit/Services/AppState.swift:10995-11017
**Issue:** The existing multi-paragraph doc comment is immediately followed by a second added doc paragraph (11013-11017) for the same function, making the header redundant and harder to read.
**Suggestion:** Merge the added "Internal for tests" sentence into the existing doc block or drop the duplicate paragraph.

### [severity: nit] Timing-based test waits undermine the "deterministic" claim
**File:** ConduitTests/VoiceBackgroundSuspensionTests.swift:923-924 (and many similar `try? await Task.sleep` sites)
**Issue:** Many lifecycle assertions are synchronized with fixed `Task.sleep` durations. Under CI load these are the classic source of intermittent failures, which conflicts with the PR's "31 deterministic tests" claim. The gated continuations (`ControlledSuspension`, `InterruptGate`, `waitUntilTranscribing`) are the right pattern and are only used on some paths.
**Suggestion:** Replace the remaining fixed sleeps with continuation/expectation-based synchronization (or `fulfillment(of:)`) for the state transitions being asserted.

## Verdict
REQUEST_CHANGES

The `supportsSpeech`/`canStartVoiceConversation` alignment, `.inactive` dip handling, and cross-profile mute fixes from the earlier rounds look correct at HEAD, but the restored-user-pause dead mic, the unfenced `.interrupted` teardown during suspension, and the superseded-attempt restore race are real open issues in the new lifecycle paths.

<a href="https://opencode.ai/s/85LRk9wF">New%20session%20-%202026-09-11T19%3A15%3A22.877Z</a>
[opencode session](https://opencode.ai/s/85LRk9wF)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34637731677)

kaishi00:
## Summary
Replaces Voice's hard teardown on backgrounding with a logical suspension (`suspendRuntimeForLifecycle`) plus an identity-validated foreground restoration, and adds a one-shot AppState-owned auto-listen gate so restored sheets never hot-start the microphone. The prior-round fixes (speech-capability restore gate, `.inactive` dip handling, paused-resume rearm, `.interrupted` guard, superseded-attempt restore gating, cross-profile mute behavior) are present at HEAD.

## Findings

### [severity: major] Orphaned-turn cancel does not verify success before clearing the alias guard
**File:** Conduit/Voice/VoiceConversationController.swift:828-836; Conduit/Services/AppState.swift:13144-13146, 9617-9623
**Issue:** `finishUtterance` clears `suspendedInFlightTurnOrphaned` immediately after `await interrupt()`, but `interrupt` is `@MainActor () async -> Void` (VoiceConversationController.swift:63) and the AppState binding is `interruptForVoice()` → `cancelCurrent()` → `interruptForReplacement(...)`, which discards its `Bool` result and swallows RPC cancellation failures (sets `errorMessage`, calls `recoverComposerSubmission`, returns `false`). The controller therefore clears the marker and submits a new voice turn even when the old Hermes turn was not actually cancelled. Because `isAwaitingVoiceAssistant` is then set `true` and `expectedAssistantSessionIDs` still contains the same session, the orphan's late `.started`/`.delta` events pass the `receiveAssistantEvent` guard and alias onto the new turn — directly contradicting the PR's "late events can never alias onto the new turn" guarantee. (This is the CodeRabbit finding marked "Addressed in commit 04b48d8"; the signal is still not propagated.)
**Suggestion:** Make the interrupt seam report success (`interrupt: () async -> Bool`, `interruptForVoice()/cancelCurrent() -> Bool`), and only clear `suspendedInFlightTurnOrphaned` when cancellation succeeded and `isCurrent(generation)` still holds. On failure, retain the marker and abort the submission.

### [severity: minor] Interruption events are fenced by state, not by capture generation
**File:** Conduit/Voice/VoiceConversationController.swift:737-744, 955-960
**Issue:** The new `guard !isRuntimeSuspended` correctly drops interruptions while dormant, but `isRuntimeSuspended` is a state flag that is cleared by `rearmRuntimeAfterCaptureRestart()` on the next listen/un-pause. `AVAudioCaptureService.handleInterruption` (AVAudioCaptureService.swift:344-351) stays registered and yields `.interrupted` with no generation (unlike `.level`), and the stream is unbounded, so a queued interruption consumed after rearm calls `failForAudioInterruption()` and tears down the freshly rearmed session. Low probability, but the fence is not tied to the runtime instance it is meant to protect.
**Suggestion:** Give interruption/route events a capture-generation identity and fence on `generation == capture.captureGeneration` (or cancel/recreate `captureEventsTask` at suspension boundaries), matching the `.level` path.

### [severity: minor] `failForAudioInterruption` still duplicates the released-resource teardown
**File:** Conduit/Voice/VoiceConversationController.swift:926-953
**Issue:** `releaseRuntimeResources()` was extracted as the shared teardown primitive, but `failForAudioInterruption` re-implements the same body (generation bump, task/stream cancellation, capture/transcriber/playback stop, detector reset, meter reset). The two can drift; the PR's stated goal is a single teardown seam.
**Suggestion:** Call `releaseRuntimeResources()` first, then apply the logical resets (`isMicrophonePaused = false`, `isVoiceSessionActive = false`, awaiting/orphan/expected-ID clears, `state = .failed(...)`).

### [severity: nit] "Deterministic" tests still synchronize on fixed `Task.sleep`
**File:** ConduitTests/VoiceBackgroundSuspensionTests.swift:890, 981, 986, 1012, 1019 (and similar)
**Issue:** Many lifecycle assertions wait a fixed 150 ms (`try? await Task.sleep(...)`). Under CI load these are the classic intermittent-failure source and undercut the "31 deterministic tests" claim; the gated continuations (`ControlledSuspension`, `InterruptGate`) are the right pattern and are only used on some paths.
**Suggestion:** Replace fixed sleeps with continuation/`fulfillment(of:)`-based synchronization for the state transition being asserted.

### [severity: nit] `suspendedVoiceConversation` is `@Published` but drives no view
**File:** Conduit/Services/AppState.swift:833
**Issue:** The descriptor is only read by `restoreSuspendedVoiceConversationIfNeeded()`; publishing it triggers `objectWillChange` on every scene transition with no SwiftUI consumer. Harmless now, but it needlessly invalidates all `AppState` observers.
**Suggestion:** Drop `@Published` and make it `private(set)` (or `internal(set)` for tests) unless a view will observe it.

## Verdict
REQUEST_CHANGES

The suspension/restoration design is sound and the prior review findings are resolved at HEAD, but the orphan-cancel success signal is still discarded, so the central "no late-event aliasing" guarantee is not actually enforced on cancellation failure.

<a href="https://opencode.ai/s/meERWGLn">New%20session%20-%202026-09-11T19%3A51%3A29.796Z</a>
[opencode session](https://opencode.ai/s/meERWGLn)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34640990932)

kaishi00:
## Summary
This PR replaces hard Voice teardown on backgrounding with a logical suspension (`VoiceConversationController.suspendRuntimeForLifecycle()`) plus identity-validated foreground restoration and an AppState-owned one-shot auto-listen gate. The lifecycle fencing, orphan-cancel success propagation, post-reconciliation currency guard, and capability gate from earlier rounds are present and coherent at HEAD.

## Findings

### [severity: major] Restoration silently discards a live output mute
**File:** `Conduit/Services/AppState.swift:5559` (calls `refreshVoiceCapabilities()`), `Conduit/Services/AppState.swift:13269` (`setProfilePreferences`), `Conduit/Voice/VoiceConversationController.swift:190-193`
**Issue:** `refreshCapabilitiesAndRestoreSuspendedVoice` awaits `refreshVoiceCapabilities()` before restoring. That path loads the persisted blob and calls `voiceConversationController.setProfilePreferences(preferences)`, which unconditionally overwrites `isOutputMuted` with the persisted value. The user's in-session Mute toggle only mutates `controller.isOutputMuted` (the sheet calls `setOutputMuted` directly at `VoiceConversationSheet.swift:116`; nothing persists it at toggle time). So after background → foreground, the conversation is restored but the mute is reverted to whatever was last persisted (usually `false`), and Hermes' next reply becomes audible. This contradicts the PR's stated "an explicit mute survives suspension" contract, and the only mute tests (`testBackgroundWithOpenVoiceSuspendsRuntimeAndRetainsDescriptor`, line 435) assert pre-restore state or use the `foregroundForRestoration()` helper that deliberately bypasses the capability refresh (line 1104-1107), so the real `.active` path is untested.
**Suggestion:** Preserve the live mute across the refresh, e.g. capture `voiceConversationController.isOutputMuted` before `refreshVoiceCapabilities()` and reapply it after (or route the refresh through a helper that honors `hasLiveVoiceSession`, like `updateActiveProfileVoicePreferences`). Add a test that sets mute, backgrounds, runs the real `refreshCapabilitiesAndRestoreSuspendedVoice`, and asserts `isOutputMuted` is still `true`.

### [severity: minor] `.interrupted` capture events are not fenced by capture generation
**File:** `Conduit/Voice/VoiceConversationController.swift:745-752`; `Conduit/Voice/AVAudioCaptureService.swift:344-351`
**Issue:** The `.level` case is fenced by both `capture.captureGeneration` and `isRuntimeSuspended`, but `.interrupted` carries no generation and is fenced only by `isRuntimeSuspended`. `AVAudioCaptureService` registers its interruption observer for the service lifetime and yields `.interrupted` on an unbounded `AsyncStream`, so a queued interruption from the released runtime can still be pending when `rearmRuntimeAfterCaptureRestart()` clears the flag (at `startListening`/`resumeMicrophone`). It would then reach `failForAudioInterruption()`, tearing down the freshly re-armed restored session — the exact failure the suspension guard is meant to prevent. Low probability, but the fence is a bool, not the runtime instance it is protecting.
**Suggestion:** Attach the `captureGeneration` to `.interrupted` (and route/interruption events) and drop events whose generation is stale, or cancel/recreate `captureEventsTask` at suspension boundaries. At minimum, gate on the capture generation as the `.level` path does.

### [severity: nit] `failForAudioInterruption` still duplicates the shared teardown primitive
**File:** `Conduit/Voice/VoiceConversationController.swift:941-968` vs `:421-441`
**Issue:** `releaseRuntimeResources()` was extracted as the single teardown seam, but `failForAudioInterruption` still re-implements the same body (generation bump, task/stream cancellation, capture/transcriber/playback stop, detector/meter reset). The two will drift as the resource set evolves.
**Suggestion:** Call `releaseRuntimeResources()` first, then apply only the logical resets (`isMicrophonePaused = false`, `isVoiceSessionActive = false`, awaiting/orphan/expected-ID clears, `state = .failed(...)`).

### [severity: nit] `suspendedVoiceConversation` is `@Published` but has no view consumer
**File:** `Conduit/Services/AppState.swift:834`
**Issue:** The descriptor is only read by restoration logic; publishing it invalidates every `AppState` observer (`objectWillChange`) on each scene transition with no SwiftUI benefit.
**Suggestion:** Drop `@Published` and make it `private(set)`/`internal(set)` for tests.

### [severity: nit] "Deterministic" tests still synchronize on fixed `Task.sleep`
**File:** `ConduitTests/VoiceBackgroundSuspensionTests.swift:70,76,97,118,145,...`
**Issue:** Many lifecycle assertions wait a fixed 150–250 ms. Under CI load this is the classic source of intermittent failures and undercuts the "deterministic tests" claim; the gated continuations (`ControlledSuspension`, `InterruptGate`, `GatedTranscriptionGateway`) already exist and should be used more broadly.
**Suggestion:** Replace fixed sleeps with continuation/`fulfillment(of:)`-based synchronization for the transitions being asserted.

## Verdict
REQUEST_CHANGES

The core design and the prior-round fixes (speech-capability gate, `.inactive` no-op, paused-resume rearm, `.interrupted` suspended guard, superseded-attempt currency, cross-profile mute protection, orphan-cancel success propagation) are correct at HEAD. The remaining blocker is that the live output mute is reverted on the real restore path by the capability refresh, contradicting the PR's stated preservation contract; the `.interrupted` generation gap is a secondary correctness risk.

<a href="https://opencode.ai/s/3pauUiDD">New%20session%20-%202026-09-11T20%3A51%3A27.327Z</a>
[opencode session](https://opencode.ai/s/3pauUiDD)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34646297833)

kaishi00:
## Summary
Replaces Voice's hard teardown on backgrounding with a logical suspension (`VoiceConversationController.suspendRuntimeForLifecycle()`) plus an identity/capability-validated foreground restoration and a one-shot auto-listen gate, so a restored sheet never hot-starts the microphone.

## Findings

### [severity: major] `.interrupted` capture events are not fenced by capture generation, only by the transient `isRuntimeSuspended` flag
**File:** Conduit/Voice/VoiceConversationController.swift:745-752
**Issue:** The `.level` case is fenced by both `capture.captureGeneration` and `isRuntimeSuspended` (line 724, 728), but `.interrupted` carries no generation (`VoiceTypes.swift:237`) and is guarded only by `isRuntimeSuspended`. `AVAudioCaptureService` yields `.interrupted` on an unbounded `AsyncStream` from an observer that lives until `deinit` (`AVAudioCaptureService.swift:65-79, 344-351`), and the controller's `captureEventsTask` is never recreated at suspension boundaries. `rearmRuntimeAfterCaptureRestart()` clears `isRuntimeSuspended` on the next Listen/un-pause, so an interruption queued from the released runtime can reach `failForAudioInterruption()` and clear `isVoiceSessionActive`/`expectedAssistantSessionIDs`, destroying the restored logical session — the exact failure suspension is meant to prevent. Exposure is small (the drain task usually consumes the event while still suspended), but the fence is a bool rather than the runtime instance it protects.
**Suggestion:** Carry `captureGeneration` on `.interrupted` (and `.routeChanged`) through the async notification hop and drop events whose generation is stale, or cancel/recreate `captureEventsTask` at suspension boundaries. Add a test that emits `.interrupted` after a rearm and asserts the logical session survives.

### [severity: minor] Orphan-cancel failure leaves capture running (microphone stays hot)
**File:** Conduit/Voice/VoiceConversationController.swift:832-851
**Issue:** `state = .thinking` and `beginBargeInMonitoring()` (live capture) run before the orphan-cancel block. When cancellation fails, the path sets `state = .failed(...)` and returns without releasing capture/tasks; `isInputMeterActive` reads inactive for `.failed`, but the engine/tap remain live until the user closes the sheet.
**Suggestion:** Call `releaseRuntimeResources()` (or `stop()`) before settling `.failed` on this branch.

### [severity: minor] Voice restoration is skipped on the reconciliation early-return paths
**File:** Conduit/Services/AppState.swift:5330-5339, 5351-5355
**Issue:** `refreshCapabilitiesAndRestoreSuspendedVoice` only runs after the fall-through at 5362-5373. When `automaticChatResumeWorkIsCurrent` is false (user edited the composer / viewport action during the health check), the task reconnects and `return`s, leaving `suspendedVoiceConversation` and `isRuntimeSuspended` set. The sheet stays open but dormant and the descriptor lingers until a later `.active`, with no capability refresh and a possibly stale gateway. This is unrelated to chat-resume currency, which Voice restoration does not require.
**Suggestion:** Run the Voice restore/consume step on those paths too (or clear the descriptor) rather than gating it on the automatic chat work token.

### [severity: nit] Preserved live mute can be dropped when the descriptor is re-recorded with different identity mid-refresh
**File:** Conduit/Services/AppState.swift:5565-5577
**Issue:** The live mute is reapplied only when `suspendedVoiceConversation == suspended`. If a second `.background` re-records the descriptor with a different `sessionID` while `refreshVoiceCapabilities()` is in flight, the guard fails, and the successor attempt reads the already-clobbered `isOutputMuted` from the controller, losing the user's mute across the refresh.
**Suggestion:** Hold the preserved live mute in AppState keyed to the profile/suspension epoch rather than re-reading it after the refresh.

### [severity: nit] `suspendedVoiceConversation` is `@Published` but has no view consumer
**File:** Conduit/Services/AppState.swift:834
**Issue:** Only restoration logic reads it, yet publishing it invalidates every `AppState` observer (`objectWillChange`) on each scene transition.
**Suggestion:** Drop `@Published`; keep `private(set)` (or `internal(set)` for tests).

### [severity: nit] `failForAudioInterruption` still duplicates the shared teardown primitive
**File:** Conduit/Voice/VoiceConversationController.swift:941-968 vs 421-437
**Issue:** `releaseRuntimeResources()` was extracted as the single teardown seam, but `failForAudioInterruption` re-implements the same body (generation bump, task/stream cancellation, capture/transcriber/playback stop, detector/meter reset). The two can drift as the resource set changes.
**Suggestion:** Call `releaseRuntimeResources()` first, then apply only the logical resets.

### [severity: nit] New lifecycle tests synchronize on fixed `Task.sleep`
**File:** ConduitTests/VoiceBackgroundSuspensionTests.swift:70, 76, 97, 118, 145, 189, 216, 229, 282, 396, 409, 455, 725, 748, 931, 1129-1167
**Issue:** Many assertions wait fixed 150–250 ms, which is the classic CI flake source and undercuts the "deterministic" claim. The gated continuations (`ControlledSuspension`, `InterruptGate`, `waitUntilTranscribing`) already exist but are only used on some paths.
**Suggestion:** Replace remaining fixed sleeps with continuation/`fulfillment(of:)`-based synchronization for the transition being asserted.

## Verdict
COMMENT

Core design and prior-round fixes (speech-capability restore gate, `.inactive` no-op, paused-resume rearm, superseded-attempt currency guard, orphan-cancel success propagation, live-mute preservation) are correct at HEAD and well covered. Remaining items are robustness/edge gaps rather than clearly reachable regressions; the `.interrupted` generation fence is the one I'd fix before or shortly after merge.

<a href="https://opencode.ai/s/8zouP0c3">New%20session%20-%202026-09-11T22%3A07%3A06.400Z</a>
[opencode session](https://opencode.ai/s/8zouP0c3)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34652461854)

kaishi00:
## Summary
Replaces the hard Voice teardown on backgrounding with a logical suspension (`suspendRuntimeForLifecycle`) plus identity/capability-validated foreground restoration and a one-shot auto-listen gate. The design and test breadth are strong, but the new capture-generation fencing has a real defect that disables the audio-interruption teardown path in production, and the orphan-cancel failure path leaves the microphone live.

## Findings

### [severity: critical] Real audio interruptions are always dropped by the `.interrupted` generation fence
**File:** Conduit/Voice/AVAudioCaptureService.swift:372-386, Conduit/Voice/VoiceConversationController.swift:745-756
**Issue:** `handleInterruption` captures the pre-teardown generation, then calls `stop()` — which increments `captureGeneration` — and only afterwards yields the event tagged with the **old** generation:
```swift
let interruptedGeneration = generationForInterruptionObservers   // G
...
guard interruptedGeneration == captureGeneration else { ... }    // G == G
stop()                                                           // captureGeneration -> G+1
continuation?.yield(.interrupted(generation: interruptedGeneration))  // yields G
```
The controller then applies:
```swift
case .interrupted(let generation):
    guard !isRuntimeSuspended else { return }
    guard generation == capture.captureGeneration else { return } // G vs G+1 -> DROP
    failForAudioInterruption()
```
Because `stop()` runs before the yield and the controller reads `capture.captureGeneration` after that teardown, `generation` can never equal the service's current generation. Every genuine phone-call/Siri/alarm interruption is discarded and `failForAudioInterruption()` is effectively dead from the notification path: the engine is torn down by the service, but the controller is left in `.listening` with `isVoiceSessionActive == true` and no capture. Note the asymmetry that proves the mismatch — the route-change failure path yields `captureGeneration` *after* `stop()` (line 408) and still works, while the interruption path yields the pre-stop value. The new service-level test even pins this buggy shape (`AVAudioCaptureServiceGenerationTests.swift:216-222`, `captureGeneration == generationB + 1` but emitted `generationB`), and the controller tests emit `.interrupted` at the *current* generation without a preceding stop, so the doubles mask the production divergence.
**Suggestion:** Make the value the controller will observe as current at handling time match the fence. Either yield the installed generation after teardown (`stop(); continuation?.yield(.interrupted(generation: captureGeneration))`) and update the service test, or remove the service-side `stop()` and let `failForAudioInterruption()` own teardown; or fence in the controller against the generation the service tore down. Add a case that reproduces the service's stop-then-yield ordering through the controller (not just emitted-at-current doubles).

### [severity: major] Failed orphan cancellation leaves the microphone live while the UI reports failure
**File:** Conduit/Voice/VoiceConversationController.swift:836-854
**Issue:** `finishUtterance` sets `state = .thinking` and calls `beginBargeInMonitoring()` (which starts live capture) *before* the orphan-cancel block. When `cancelledOrphan` is false the path does `state = .failed(...); return` without `releaseRuntimeResources()`/`stop()`. Capture/tasks therefore stay live while `isInputMeterActive` computes inactive for `.failed`, so the UI shows a failed/idle state and a hidden meter while the engine, tap, and audio-session lease remain held (the OS mic indicator can stay lit until the user closes the sheet).
**Suggestion:** Call `releaseRuntimeResources()` (or `stop()`) before settling `.failed` on this branch, and add a test asserting capture is released when the orphan cancel fails.

### [severity: minor] Voice restoration is skipped on the reconciliation early-return paths
**File:** Conduit/Services/AppState.swift:5330-5339, 5351-5355, 5369-5373
**Issue:** `refreshCapabilitiesAndRestoreSuspendedVoice` is reached only via the fall-through. When `automaticChatResumeWorkIsCurrent` is false (composer edit / viewport action during the health check), the scene task calls `settleReconciliation` + `reconnectForRetry` and `return`s, leaving `suspendedVoiceConversation` set and `isRuntimeSuspended == true`. The sheet stays open but dormant with no restoration and no visible error until a later `.active` cycle. Voice restoration does not depend on chat-resume currency, so this coupling silently strands the suspended conversation.
**Suggestion:** Run the Voice restore/consume step (or clear the descriptor) on those early-return paths too, independent of the automatic chat-resume token.

### [severity: minor] Live output mute preservation is not robust against an independent capability refresh
**File:** Conduit/Services/AppState.swift:5557-5581, Conduit/Views/RootView.swift:129-131, Conduit/Services/AppState.swift:13267-13292
**Issue:** `refreshCapabilitiesAndRestoreSuspendedVoice` preserves the live mute only around its own `refreshVoiceCapabilities()` call by snapshotting before and reapplying after. But `refreshVoiceCapabilities()` unconditionally calls `setProfilePreferences` (which overwrites `isOutputMuted` from the persisted blob), and RootView has a separate `.task(id: voiceCapabilityRefreshKey)` that invokes it when `isConnected` flips (e.g. socket death on background, reconnect on foreground). A RootView-triggered refresh interleaving at the `await service.reload()` inside the restore's refresh can re-clobber the mute after it was reapplied, reverting to the persisted value — violating the stated "an explicit mute survives suspension" contract. The `guard suspendedVoiceConversation == suspended` also fails if a second `.background` re-records the descriptor mid-refresh, dropping the preserved mute for the successor (the successor then reads the already-clobbered controller value).
**Suggestion:** Make mute authority centralized — e.g. have `refreshVoiceCapabilities`/`setProfilePreferences` honor `hasLiveVoiceSession` (as `updateActiveProfileVoicePreferences` already does), or hold the preserved mute in AppState keyed to the profile/suspension epoch rather than re-reading the controller after the await.

### [severity: nit] `failForAudioInterruption` duplicates the shared teardown primitive
**File:** Conduit/Voice/VoiceConversationController.swift:945-972 vs 421-441
**Issue:** `releaseRuntimeResources()` was extracted as the single teardown seam, but `failForAudioInterruption` re-implements the same body (generation bump, task/stream cancellation, capture/transcriber/playback stop, detector/meter reset). The two will drift as the resource set evolves.
**Suggestion:** Call `releaseRuntimeResources()` first, then apply only the logical resets (`isMicrophonePaused = false`, `isVoiceSessionActive = false`, awaiting/orphan/expected-ID clears, `state = .failed(...)`).

### [severity: nit] `suspendedVoiceConversation` is `@Published` with no view consumer
**File:** Conduit/Services/AppState.swift:834
**Issue:** The descriptor is only read by restoration logic; no production view observes it. `@Published` invalidates every `AppState` observer on each scene transition for no UI benefit.
**Suggestion:** Drop `@Published` and keep `private(set)`/`internal(set)` for tests.

### [severity: nit] New lifecycle tests synchronize on fixed `Task.sleep`, and the capture double diverges from the real service
**File:** ConduitTests/VoiceBackgroundSuspensionTests.swift:70,76,97,118,145,189,216,229,282,396,409,455,729,752,830,935,1133,1138,1164,1171; ConduitTests/VoiceBackgroundSuspensionTests.swift:1427-1428
**Issue:** Many assertions wait fixed 150–250 ms, the classic CI-flake source, undercutting the "deterministic" claim; gated continuations exist but are used only on some paths. Relatedly, `MockCapture.emitInterrupted` yields at the current generation, which is exactly what hides the critical production mismatch above (the real service yields the pre-stop generation).
**Suggestion:** Use continuation/`fulfillment(of:)` synchronization for the asserted transitions, and model the real stop-then-yield generation ordering in the double. Replace `try!`/forced unwraps at lines 979, 1003, 1014-1016 with `XCTUnwrap` and `throws` tests.

## Verdict
REQUEST_CHANGES

The suspension/restoration architecture and most prior-round fixes are sound, but the `.interrupted` generation fence is internally inconsistent and silently disables a real teardown path in production (masked by test doubles), and the orphan-cancel failure path holds the microphone while reporting `.failed`. Both should be fixed before merge.

<a href="https://opencode.ai/s/jy8Ipgjy">New%20session%20-%202026-09-11T22%3A45%3A22.116Z</a>
[opencode session](https://opencode.ai/s/jy8Ipgjy)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34655277961)

kaishi00:
## Summary
This PR changes Voice background handling from a hard teardown to a logical suspension (`suspendRuntimeForLifecycle`) with identity/capability-validated foreground restoration and a one-shot, AppState-owned auto-listen gate so a restored sheet never hot-starts the mic.

## Findings

### [severity: minor] Suspension gate is cleared before the new capture runtime exists, opening a window for a queued interruption to fail a resumed session
**File:** Conduit/Voice/VoiceConversationController.swift:257-259 (guard at 745-756)
**Issue:** `startListening()` calls `rearmRuntimeAfterCaptureRestart()` (`isRuntimeSuspended = false`) *before* `await capture.requestPermission()` and `try capture.startListening(...)`. Until `capture.startListening` advances `captureGeneration`, an `.interrupted(generation:)` event that was queued for the just-released runtime still satisfies `generation == capture.captureGeneration` while `isRuntimeSuspended` is now false, so `failForAudioInterruption()` runs and tears down the session the user just tried to resume (clears ownership, sets `.failed`). The window requires an interruption queued during suspension plus a Listen tap before the consumer drains it, but it is exactly the class of stale-teardown the fence is meant to prevent. (`resumeMicrophone()` is safe because it calls `capture.resume()` before rearming.)
**Suggestion:** Re-arm only after the capture restart succeeds (move `rearmRuntimeAfterCaptureRestart()` to after `try capture.startListening(...)`), or additionally gate `.interrupted` on a live-capture state rather than only the transient bool + service generation.

### [severity: minor] Voice restoration is skipped on the healthy-transport automatic-work guard, stranding the descriptor
**File:** Conduit/Services/AppState.swift:5314-5318
**Issue:** At the top of the `.active` task, if `automaticChatResumeWorkIsCurrent(automaticWorkToken)` becomes false after `verifyChatResumeTransportHealth`, the task settles and `return`s, never reaching `refreshCapabilitiesAndRestoreSuspendedVoice`. The suspended descriptor and `isRuntimeSuspended == true` linger with the sheet open but dormant. The reconnect early-return paths were fixed to fall through (and are tested by `testVoiceStillRestoresWhenAutomaticWorkInvalidatedDuringReconnect`), but this healthy-transport path is not covered. Voice restoration is independent of chat-resume currency.
**Suggestion:** Run the Voice restore/consume step (or clear the descriptor) on this path too, independent of the automatic chat-resume token. Same reasoning applies to the `connection != nil` early return at 5273 for a signed-out foreground.

### [severity: nit] `suspendedVoiceConversation` is `@Published` but has no view consumer
**File:** Conduit/Services/AppState.swift:834
**Issue:** The descriptor is only read by restoration logic; no SwiftUI view observes it. `@Published` therefore fires `objectWillChange` for every `AppState` observer on each scene transition with no UI benefit.
**Suggestion:** Drop `@Published` and keep `private(set)`/`internal(set)` for tests.

### [severity: nit] `failForAudioInterruption` re-implements the shared teardown primitive
**File:** Conduit/Voice/VoiceConversationController.swift:952-979 vs 421-441
**Issue:** `releaseRuntimeResources()` was extracted as the single runtime teardown seam, but `failForAudioInterruption()` still duplicates the same body (generation bump, task/stream cancellation, capture/transcriber/playback stop, detector/meter reset). The two will drift as the resource set evolves.
**Suggestion:** Call `releaseRuntimeResources()` first, then apply only the logical resets (`isMicrophonePaused`, `isVoiceSessionActive`, awaiting/orphan/expected-ID clears, `isRuntimeSuspended`, `state = .failed(...)`).

### [severity: nit] Controller test double does not model the real stop-then-yield generation ordering
**File:** ConduitTests/VoiceBackgroundSuspensionTests.swift:1466-1468
**Issue:** `MockCapture.emitInterrupted(generation:)` defaults to the *current* generation, whereas the real service emits the *post-stop* generation. The service-level test (`AVAudioCaptureServiceGenerationTests`) pins the production ordering, so the critical path is covered, but the controller-level doubles still can't catch a future regression in the service/controller generation contract.
**Suggestion:** Have the double optionally bump its generation before yielding (mirroring `stop()` then yield), or note explicitly that the service test owns that contract.

### [severity: nit] Fixed `Task.sleep` synchronization and force unwraps in new lifecycle tests
**File:** ConduitTests/VoiceBackgroundSuspensionTests.swift:70,76,97,118,145,455,935,1171 (and 979,1003,1014-1016)
**Issue:** Many assertions wait fixed 150–250 ms, the classic CI-flake source and in tension with the "deterministic" claim; the gated continuations (`ControlledSuspension`, `InterruptGate`, `waitUntilTranscribing`) already exist but are used only on some paths. Lines 979/1003/1014-1016 also use `try!`/force unwrap, which crashes the whole test process instead of failing one test.
**Suggestion:** Prefer continuation/`fulfillment(of:)`-based synchronization for the asserted transition, and replace `try!`/`!` with `throws` tests + `XCTUnwrap`.

## Verdict
COMMENT

The core suspension/restoration design is coherent and the previously reported blockers (speech-capability gate, `.inactive` no-op, paused-resume rearm, post-stop interruption generation, orphan-cancel success propagation, live-mute preservation, superseded-attempt currency) appear resolved at HEAD. No new force unwraps or retain cycles were found in production code. The two lifecycle gaps above are robustness issues rather than clearly reachable regressions and are cheap to close.

<a href="https://opencode.ai/s/XNeJDdfh">New%20session%20-%202026-09-11T23%3A39%3A04.008Z</a>
[opencode session](https://opencode.ai/s/XNeJDdfh)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34658787148)

PR #162, opened 2026-09-11
feat(l10n): zh-Hans localization via String Catalogs (resolves #135)

### LINKED ISSUE #135: Feature request: localization support, starting with Simplified Chinese
## Feature Request

Please consider adding localization support, starting with Simplified Chinese (简体中文).

## Why

Hermes Agent itself already has solid Chinese-language support (server-side `display.language: zh` config, Chinese docs, etc.), and a meaningful part of the self-hosted user base runs their agent primarily in Chinese. Conduit being English-only is currently the one link in the chain without Chinese support.

## Current state (checked in the repo)

- No `.lproj` folders and no `Localizable.strings` / String Catalog (`.xcstrings`) anywhere in the project.
- `Info.plist` only sets `CFBundleDevelopmentRegion` to `$(DEVELOPMENT_LANGUAGE)` — no `CFBundleLocalizations` array declaring supported locales.
- UI strings are hardcoded directly in SwiftUI views — a rough grep of just the `Views/` folder turns up 227+ `Text("...")` literals, before counting alerts, buttons, accessibility labels, etc. elsewhere in the app.

So this isn't "drop in a translated `.strings` file" — there's no i18n plumbing yet, meaning it's a real (if mechanical) lift: extracting hardcoded strings into a String Catalog, then translating. Flagging that scope up front so it's not mistaken for a quick fix.

## Ask

- Adopt Xcode's String Catalog (`.xcstrings`) as the localization mechanism (least invasive for a SwiftUI-only codebase using `Text("...")` literals — Xcode can auto-extract most of them).
- Add `zh-Hans` (Simplified Chinese) as the first additional locale.
- I'd be glad to help with translation review/QA on the Chinese strings once the extraction lands, if useful.

vecStardo:
Resolves #135

Adds Simplified Chinese (zh-Hans) localization via native String Catalogs.

## Approach

- **Three catalogs**: `Conduit/Localizable.xcstrings` (1,074 entries), `Conduit/InfoPlist.xcstrings` (Face ID / microphone / speech-recognition prompts), `Conduit/AppShortcuts.xcstrings` (Siri shortcut phrases).
- **Keys are the English source strings.** The development language stays English: en-locale behavior is byte-identical, SwiftUI `LocalizedStringKey` call sites (`Text`, `Button`, `alert`, …) resolve through the catalog with **no code change**, and only true `String`-context literals (error messages, notifications, composer placeholder, …) are wrapped in `String(localized:)`.
- **Config values are never translated.** Server-facing values (`manual`/`smart`/`off`, `project`/`strict`, personality ids like `catgirl`/`noir`) are written to the Hermes profile verbatim; a value→display map localizes only their labels in the settings UI. Unmapped dynamic values fall back to raw.
- **Terminology** is coordinated with the MIT-licensed hermes ecosystem (`hermes-agent`, `hermes-desktop`, `hermes-webui` zh packs), preferring official translations where they exist (e.g. approval → 审批, session → 会话).

## Notable implementation details

- **Int interpolations are pre-formatted** (`String(localized: "pid \(String(pid))")`): iOS 26 applies locale digit grouping to bare `Int` args inside localized strings, which regressed `pid 4242` → `pid 4,242` even in English. Caught by the test suite.
- **Split multi-line prompt literals were merged into complete keys** so translated sentences render in full instead of half-localized fragments.
- Permission prompts and Siri phrases ship as `InfoPlist.xcstrings` / `AppShortcuts.xcstrings`.
- xcodegen picks up the catalogs automatically (no `project.yml` changes needed).

## Testing

- All **2,237 unit + UI tests pass** (en locale — assertions reference English source keys, so they are unaffected by translations).
- Verified on device (iPhone 15 Pro Max, iOS 26): settings, kanban, chat, voice settings all render zh-Hans; no layout truncation observed.

## Screenshots

| Kanban | Settings | Chat settings |
|---|---|---|
|  |  |  |

Happy to adjust terminology or split the PR if you'd prefer it smaller — the catalog and the source wraps are independent once the catalogs land.

## Maintainer integration (appended on top of vecStardo's five commits)

Current `main` (Voice background suspension #161, CarPlay Voice #165) is merged in with a **normal merge commit** (61daa4f) — the contributor's commits are untouched, no rebase, no squash. All surfaces newly introduced by that merge are localized and checker-covered (CarPlay state titles/buttons, capability reasons, audio errors, spoken-phrase editor, kanban status names, setup-wizard guidance, about page).

### In-app App Language (Settings → Appearance → App Language)

- **System Default / English / 简体中文**, global to Conduit (not profile-scoped), persisted across launches.
- **No `AppleLanguages` mutation, no process termination.** A small facility (`Conduit/Services/AppLanguage.swift`) makes both localization halves resolve through the same selection:
  - `AppLocalization.string(…)` — every explicit `String(localized:)` site (710 call sites) is routed through the selected language's bundle (`zh-Hans.lproj` / `en.lproj`); a key missing there resolves back to the **English source string** (keys *are* the source strings), still formatted with its arguments — never an untranslated key, an empty label, or a broken placeholder.
  - the root `.environment(\.locale)` plus a content-identity rebuild makes SwiftUI literal keys re-resolve immediately; frozen `static let` localized tables were converted to computed properties so they re-resolve per use.
- **UI language only.** STT/TTS provider config, speech locales, and Hermes/model language are untouched (pinned by tests). Server-facing values (config values, identifiers, provider/model/personality IDs) remain raw protocol values; only their display labels are translated (`ConfigFieldLocalizationTests` now also pins `display.tool_progress` = `all`/`off`).

### Multilingual built-in spoken Voice controls

- **Stop** gains 停止 / 别说了 / 不要说了; **End Conversation** gains 再见 / 拜拜 / 结束对话 / 就这样吧. Both sets are recognized **regardless of the App Language** (commands match the transcript, never the UI locale), matching stays exact whole-utterance, and the user-editable lists keep working.
- Persisted lists that are *exactly* the previously shipped defaults are migrated to the new built-ins; customized lists — including deliberate removals — are preserved (tested).

### Localization review finished + CI gate strengthened

- All still-valid findings fixed: raw ternary branches (gateway notice, Copy/Done/Off), fragment concatenation (credentials explanation), Memory/Skill labels, singular bulk actions ("Archive 1"/"Delete 1" now use the localized `%lld` key), catalog plural variations own English *and* Chinese grammar (`%lld conversations` keeps "1 conversation"), and the `subtitle:`/section-title String-context labels.
- `scripts/check-l10n-coverage.py` now requires every catalog key to carry a **real zh-Hans unit** (state `translated`, non-empty, printf placeholders matching the key including positional `%1$@` reorderings), scans `AppLocalization.string(` plus `ProgressView`/`ContentUnavailableView`/`Section`/`Menu`/`GroupBox`/`.alert`/`.confirmationDialog` literals, validates the `AppShortcuts`/`InfoPlist` catalogs too, and pins dynamic/ternary sites via an explicit `REGRESSION_KEYS` list. 1,464 call sites resolve; every key has a real translation.

### Verification

- Multi-model review gate (3 independent reviewers): no blockers; all consolidated findings fixed (slash-command language refresh without refetch, stable `SlashCommand` identity, terminology alignment — 档案/语音转文字 — and remaining raw display params).
- Full Conduit test suite executed on the Mac build host at the exact pushed head (results in the PR timeline comment below).

## Summary by CodeRabbit

- **New Features**
  - Added an in-app language setting with System Default, English, and Simplified Chinese options.
  - Added localized app shortcuts and permission descriptions for voice conversations, Face ID, microphone, and speech recognition.
  - Added Chinese voice stop and end-conversation phrases while preserving customized phrase lists.
  - Improved voice conversation restoration across backgrounding, profile changes, and CarPlay.
  - Expanded localization across settings, conversations, connection setup, voice, Kanban, accessibility, and sidebar labels.

- **Quality Improvements**
  - Added automated localization coverage checks and tests for catalog translations and raw configuration values.


kaishi00:
## Summary
Adds Simplified Chinese (zh-Hans) support through three String Catalogs plus `String(localized:)` wrapping of String-context literals, aiming to keep English behavior unchanged. The catalog infrastructure is sound, but one change translates a server-facing config value and the source migration is incomplete, leaving several primary flows half-English.

## Findings

### [severity: critical] `agent.image_input_mode` config value `"auto"` is translated, which corrupts the Hermes profile
**File:** Conduit/Views/AuxiliaryViews.swift:868
**Issue:** `.options([String(localized: "auto"), "native", "text"], defaultValue: String(localized: "auto"))`. Under zh-Hans `String(localized: "auto")` returns `"自动"`. The `.options` control persists the option string verbatim (`save(field, value: .text(option))` at line 1378), so choosing the first entry writes `agent.image_input_mode = "自动"` into the Hermes profile, which Hermes cannot interpret. The localized `defaultValue` also poisons the read path (`textValue` falls back to `"自动"` and `displayLabel` can no longer map it). This directly violates the PR's stated invariant that config values are never translated.
**Suggestion:** Keep values raw: `.options(["auto", "native", "text"], defaultValue: "auto")`. The display map at line 1342 already produces the localized label. Add a test asserting a `.options` selection saves the raw English value under a non-en locale.

### [severity: major] `contextUsed` still uses a bare `Int` interpolation, reintroducing the digit-grouping regression
**File:** Conduit/Views/AuxiliaryViews.swift:32
**Issue:** `String(localized: "\(appState.runtime.contextUsed) tokens")` yields the key `%lld tokens`, but the catalog contains only `%@ tokens` (Localizable.xcstrings:110). The sibling `contextMax` on line 33 was correctly pre-formatted with `String(...)`. Per the PR's own note, iOS 26 groups bare `Int` args, so this row will render `4,242 tokens` and stay untranslated in zh-Hans.
**Suggestion:** Pre-format it like the sibling: `String(localized: "\(String(appState.runtime.contextUsed)) tokens")`, or add a matching `%lld tokens` entry.

### [severity: major] Catalog placeholder mismatch (`%lld` vs `%@`) makes a translation silently dead
**File:** Conduit/Services/KanbanService.swift:38 (catalog Conduit/Localizable.xcstrings:4130)
**Issue:** `status` is a `String`, so the lookup key is `Hermes does not allow %@ as a manual Kanban destination.`, while the catalog defines `...allow %lld...`. The zh-Hans string never applies.
**Suggestion:** Correct the catalog specifier to `%@` (or pre-format the value to match an `%lld` key). Add a code/catalog key audit to catch specifier drift.

### [severity: major] Large areas of String-context copy remain unwrapped and untranslated
**File:** multiple
**Issue:** Strings flowing through `String` parameters/returns or String-typed ternary branches are not localized, and their keys are absent from the catalog:
- Conduit/Views/ConnectionSetupView.swift:642 `answerRow("Yes", ...)` ("Yes" is not in the catalog; "No"/"I don't know" were wrapped); lines 361-363 dashboard-step `explanation`; `supporting:` strings at 435/443/452 (`answerRow`/`methodCard` take `String` and render via `Text(string)`, which does not localize).
- Conduit/Views/VoiceConversationSheet.swift:158,160,187,188 "Transcribing"/"Listen"/"Listening" returned as `String` and passed to `Text`/`Label`.
- Conduit/Services/ConnectionFailure.swift:279-318 nearly all of `userMessage` (only the middle fragment at 308 is wrapped).
- Conduit/Views/ChatView.swift:1500 "Command"; 2068-2070,2494-2495 "ANSWERED"/"EXPIRED"/"APPROVED"/"REJECTED"; 2485 "Rejected".
- Conduit/Views/ConnectionSetupForm.swift:84 "Port", 284 "Done", 350 "Entered".
The result is a half-English onboarding, error, voice, and approval UI under zh-Hans.
**Suggestion:** Wrap each branch/argument in `String(localized:)`, and add a CI coverage check that every String-context literal has a catalog entry (this class of miss is invisible in the en-locale test suite).

### [severity: major] Sentence-fragment concatenation defeats localization
**File:** Conduit/Views/VoiceSettingsView.swift:187,255; Conduit/Views/KanbanViews.swift:1572; Conduit/Services/KanbanService.swift:41; Conduit/Services/ConnectionFailure.swift:306-309; Conduit/Services/NativeAuthClient.swift:60
**Issue:** Sentences are assembled from independently localized fragments, e.g. `String(localized: "Voice on ") + profileDisplayName`, `String(localized: "No ") + (…"transcription"…) + " providers were discovered…"`, `String(localized: "Server current (") + slug + ")"`, and `"The task was created" + identifier + ", but Hermes could not move it to " + targetStatus + …`. Fragment keys (`"No "`, `"Voice on "`, `"Server current ("`) cannot encode correct Chinese grammar/word order, companion fragments are often missing from the catalog, and the Cloudflare message localizes only its middle third (bilingual output).
**Suggestion:** Use one key with placeholders: `String(localized: "Voice on \(profileDisplayName)")`, `String(localized: "No \(kind) providers were discovered for this Hermes profile.")`, `String(localized: "Server current (\(slug))")`.

### [severity: minor] Missing catalog entries for keys actually used
**File:** Conduit/Views/Kanban/KanbanV2Support.swift:453,466; Conduit/Services/KanbanService.swift:49; Conduit/Services/AppState.swift:695
**Issue:** `Unblocked → %@`, `Priority set to %@`, `Could not build a safe Hermes Kanban request (invalid %@)...`, and `%@ %@…` (running composer placeholder) have no catalog entries and will always render English.
**Suggestion:** Add entries with specifiers matching the real interpolation types (`%@` for the String maps), or re-run extraction and assert coverage.

### [severity: minor] Generic empty-option "Default" fallback dropped
**File:** Conduit/Views/AuxiliaryViews.swift:1351-1355, 1375-1381
**Issue:** Previously the `.options` menu rendered `option.isEmpty ? "Default" : option`. Now `displayLabel` maps only `display.personality[""]` to "Default" and otherwise returns the value unchanged, so any other field whose choices include `""` (e.g. server-provided `memory.provider`/`context.engine`) renders a blank row/control.
**Suggestion:** Restore the generic fallback before the map lookup: `if value.isEmpty { return String(localized: "Default") }`.

### [severity: minor] Unannounced English copy change: `ChatReturnSurface.sessions` now reads "Session list"
**File:** Conduit/Services/ChatReturnSurface.swift:14
**Issue:** `.sessions` changed from "Sessions" to `String(localized: "Session list")` with no `en` override, so the English segmented control (AuxiliaryViews.swift:1005) now shows "Session list". This contradicts the PR's "en-locale behavior is byte-identical" claim and is not covered by tests.
**Suggestion:** Keep "Sessions", or add an explicit `en` string unit and call out the copy change in the PR.

### [severity: minor] Committed one-off tooling and generated artifacts
**File:** `.loc_scan.py`, `.loc-scan.json`, `docs/localization/**`, `本地化工作量评估.md`, `scripts/set-sim-english.sh`
**Issue:** ~36k lines of translation/WBS scripts and a 162 KB scan dump (including generated batch results) are added to the repo, increasing noise without shipping value.
**Suggestion:** Keep only the catalogs plus any repeatable extract/verify script referenced from CI; move or drop the intermediate/batch artifacts and the root-level scan output.

Note on the requested focus areas: this diff introduces no retain cycles, no new concurrency/MainActor issues, and no new unguarded force unwraps (the `!` uses are pre-existing and guarded). The `String(localized:)` calls inside `static let` initializers (`WakeModelDescriptor.bundledBilingualPack`, `AppState.builtInSlashCommands`) capture the locale at first access, which is acceptable given iOS language changes require an app restart but is worth being aware of.

## Verdict
REQUEST_CHANGES

The critical `agent.image_input_mode` regression can write invalid values to the Hermes config, and the incomplete String-context wrapping plus the `%lld`/`%@` key mismatch mean zh-Hans is not actually complete in several primary flows.

<a href="https://opencode.ai/s/1aYnsZWS">New%20session%20-%202026-09-11T21%3A10%3A32.599Z</a>
[opencode session](https://opencode.ai/s/1aYnsZWS)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34647939297)

vecStardo:
Thanks for the thorough review — all findings are addressed in 02e6c2a..c33daee.

**[critical] Translated config value**: `.options` values for `agent.image_input_mode` are raw again; localized display names come from the value→display mapping (now extracted as `ProfileConfigValueDisplay`). Added `ConfigFieldLocalizationTests` pinning a raw-value allowlist for every options field so a localized value can't sneak back in.

**[major] contextUsed**: pre-formatted like its sibling; resolves to the existing `%@ tokens` key. **[major] Placeholder mismatch**: renamed the catalog key to the `%@` form the code actually emits. **[major] Unwrapped copy**: wrapped all listed sites (plus the neighboring ones the same sweep surfaced — NativeAuthClient login errors, KanbanService error branches) with zh-Hans entries. **[major] Fragment concatenation**: now single keys with placeholders (`"Voice on %@"`, `"Server current (%@)"`, the task-created error, one shared Cloudflare key); dead fragment keys removed.

**[minor] Missing entries** added; **[minor] Default fallback** restored for empty values; **[minor] "Session list"** reverted to `"Sessions"` so English copy is byte-identical to main again — for context, the distinct key originally existed to give that screen a more specific zh label (会话列表 vs the sidebar's 会话); happy to re-add it via a separate key with an explicit `en` localization if you'd prefer the specific label without changing rendered English. **[minor] tooling artifacts** removed from the PR.

Also added `scripts/check-l10n-coverage.py` to the CI plan job: it extracts every `String(localized:)` / SwiftUI literal key (interpolation skeletons included) and fails when one is missing from the catalog — the same sweep caught a few gaps beyond the review list (e.g. `"Delete phrase %@"`, the sidebar's "N conversations", now a proper plural entry).

kaishi00:
## Summary
Adds Simplified Chinese (`zh-Hans`) support through `Localizable.xcstrings` / `InfoPlist.xcstrings` / `AppShortcuts.xcstrings` plus `String(localized:)` wrapping, and a CI catalog-coverage guard. The earlier critical bug (localized persisted config values) is fixed and verified, but the String-context migration is still incomplete: several user-facing flows render half- or fully-English under `zh-Hans`, and the new CI guard does not catch this class of miss.

## Findings

### [severity: major] Many user-facing `String`-context literals are still not localized
**File:** multiple (see list)
**Issue:** Values that flow through `String` parameters/returns are not localized: `Text(someString)` uses the non-localized `StringProtocol` overload, and computed `String` properties returning literals render English. None of these keys exist in the catalog, so they stay English under `zh-Hans`. Confirmed sites:
- `Conduit/Services/ConnectionSetupTest.swift:145,151-152,159-160,165-166` — `readyMessage`, `interactiveReadyMessage`, `credentialsRequiredMessage`, `repairInteractiveMessage`; consumed via `Text(ConnectionSetupTestState.<...>)` in `ConnectionSetupForm.swift:203,256-257,266` and `ConnectionSetupView.swift:288,293,298`. This is the whole connection-test result copy.
- `Conduit/Services/HermesClient.swift:2430-2431` — `return "Memory"` / `return "Skill"` (the sibling `"User profile"` at 2429 was wrapped).
- `Conduit/Services/AppState.swift:10944` — `"Hermes did not accept that decision."`
- `Conduit/Services/PushNotificationService.swift:313,316` — `"Updating"`, `"Off"` (313/316 siblings `Enabled`/`Notifications denied` were wrapped).
- `Conduit/Views/AuxiliaryViews.swift:189,201,1477-1478` — `"Connected"`/`"Disconnected"`/`"Reconnect"`; `SettingsMetricRow.value` is `String` (AuxiliaryViews.swift:1945) so `Text(value)` never localizes.
- `Conduit/Views/Components/ComposerBar.swift:966,973` — `"Could not prepare \(name) for upload."`, `"Off"`.
- `Conduit/Views/ConnectionSetupForm.swift:73` — ternary with a concrete `String(localized:)` branch forces `String`, so `"Tailscale hostname / address"` is raw (`labeled(...)` takes `String`).
- `Conduit/Voice/VoiceTypes.swift:163,166,248-249` — `"Ready"`, `"Unavailable"`, `"Microphone access is required for voice conversations."`, `"No speech was captured."`.

**Suggestion:** Wrap each in `String(localized:)` and add the corresponding `zh-Hans` entries. Do not rely on the en-placeholder fallback; these keys are absent from the catalog entirely.

### [severity: major] Sentences/grammar are still assembled by concatenating outside localization
**File:** multiple
**Issue:** Even after the partial fixes, several strings are built from independently-localized or raw fragments, which cannot reorder or inflect correctly in Chinese and leaves trailing English:
- `Conduit/Views/ConnectionSetupView.swift:389-390`: `String(localized: "This means ... not Tailscale, ") + "Cloudflare, or Apple credentials."` — the second half is a raw literal absent from the catalog, so the sentence renders bilingual.
- `Conduit/Views/Kanban/KanbanV3ASupport.swift:308`: `String(localized: "Decomposed into \(String(childCount)) task") + (childCount == 1 ? "" : "s")` — appends a raw `"s"`, so `zh-Hans` for >1 renders `…个任务s`.
- `Conduit/Views/KanbanViews.swift:171-178`: `let noun = selectedCount == 1 ? "task" : "tasks"` is interpolated into `"Move %@ selected %@"`, producing e.g. `移动所选 2 个 tasks` for VoiceOver.
- `Conduit/Views/KanbanViews.swift:1131,1140`: singular confirmation titles return raw `"Archive 1"` / `"Delete 1"` while the plural branches are localized.

**Suggestion:** Use a single key per sentence with placeholders (`"This means ... not Tailscale, Cloudflare, or Apple credentials."`, two complete plural keys for the decompose notice, `"Move %@ selected tasks"`/singular variants), and wrap `"Archive 1"`/`"Delete 1"` in `String(localized:)` (or add plural `en` one/other entries).

### [severity: minor] Ternary `Label`/`Text` branches defeat localization of the inactive state
**File:** `Conduit/Views/Components/MarkdownLargeDocumentView.swift:958`, `Conduit/Views/Components/MarkdownText.swift:2020,2030`
**Issue:** `Label(copied ? String(localized: "Copied") : "Copy", ...)`: because one branch is concretely `String`, the whole ternary is `String` and `Label` selects the non-localizing `StringProtocol` overload — the raw `"Copy"` never translates (and `"Copy"` is not in the catalog). Same for `Text(normalizedLanguage == "plain" ? "Code" : normalizedLanguage)`.
**Suggestion:** Wrap both branches (`String(localized: "Copy")`, `String(localized: "Code")`) and add catalog entries.

### [severity: minor] CI coverage guard cannot see the class of miss above, and ignores placeholder type
**File:** `scripts/check-l10n-coverage.py:52-54,125-130`
**Issue:** The guard only matches `String(localized:)` and literal-first-argument SwiftUI initializers. Raw literals passed to `String` parameters/returns (all `String(localized:)-in-a-ternary` cases and `Text(variable)` cases) are invisible, so it reports success while the flows above remain English — i.e. it provides false confidence for the PR's "translations can never silently render English" claim. `catalog_has` also treats any interpolation as `%@` and accepts a `%lld` key, so it cannot catch the exact wrong-specifier/digit-grouping regression this PR fixed.
**Suggestion:** Add a checker for user-facing `String`-returning properties/`String`-parameter call sites (even a heuristics list), and make the placeholder check type-aware (`%lld` for `Int` args) or at least assert a fixed catalog skeleton per code site.

### [severity: minor] Stale/unreachable catalog entries add translator noise
**File:** `Conduit/Localizable.xcstrings`
**Issue:** `"N updated, M failed"` and `"2 updated, 1 failed"` remain in the catalog but only appear in comments (`KanbanViews.swift:1373,1459`); the live code emits `"%@ updated, %@ failed"` etc. Dead keys invite re-translation and mask real coverage.
**Suggestion:** Remove unreachable entries, or add a reverse (catalog→call-site) audit to the guard.

### [severity: nit] Localized values in `static let` capture the locale at first access
**File:** `Conduit/Voice/Wake/WakeModelDescriptor.swift:46-62`, `Conduit/Services/AppState.swift:8828-8850`, `Conduit/Services/ConnectionSetupTest.swift:145-166`
**Issue:** `String(localized:)` inside `static let` initializers is evaluated once and frozen for the process lifetime. iOS restarts the app on language change, so this is acceptable today, but it is a latent trap for any future in-app language switch or dynamic locale tests.
**Suggestion:** Prefer computed static properties (or resolve at point of use) for anything that must track the current locale.

## Verdict
REQUEST_CHANGES

The critical config-value regression is correctly fixed and the catalog placeholder/`%@ tokens` issues were resolved. However, the PR's central claim (complete `zh-Hans` coverage with English behavior preserved) does not hold: multiple primary flows — connection test results, approval errors, push status, kanban confirmations/VoiceOver, markdown copy controls — remain untranslated, and the new CI guard does not detect that class of miss.

<a href="https://opencode.ai/s/ooNNEp4Q">New%20session%20-%202026-09-12T13%3A55%3A04.363Z</a>
[opencode session](https://opencode.ai/s/ooNNEp4Q)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34697702936)

kaishi00:
Maintainer follow-up plan after review:

First, thank you for doing the heavy lifting here. We want to keep #162 as the canonical contribution rather than replace it with a maintainer PR, and preserve the existing contributor-authored commits.

Before merge, we’re going to finish the localization pass on top of this branch:

- integrate current `main` (including the newly landed Voice/CarPlay work) and resolve any localization gaps introduced since this branch diverged;
- fix the remaining verified String-context/pluralization/catalog-coverage findings from the current review;
- add an in-app **App Language** setting under Appearance with **System Default / English / 简体中文** so localization can be QA’d without changing the whole device language;
- make the selected app language apply consistently to both SwiftUI-localized literals and explicit `String(localized:)` paths (no half-switched UI);
- keep UI language independent from STT/TTS/provider/model language settings;
- add localized built-in Voice control aliases (at minimum Stop / Goodbye equivalents) **additively**, so English and Chinese control phrases are both recognized regardless of UI language; user-edited phrase lists remain supported;
- continue treating Hermes/server-facing config values and identifiers as raw protocol values — only their display labels may be localized;
- strengthen localization coverage checks so a catalog key without a `zh-Hans` translation does not count as covered, and extend coverage for the SwiftUI/String-context cases already identified by review;
- add regression tests for the language override and multilingual Voice controls.

Once that combined head is ready, we’ll run the full suite/hosted CI and do the final zh-Hans walk-through.

Authorship note: we intend to preserve the existing contributor commits and add maintainer follow-up commits on top rather than rewriting this PR’s history.

vecStardo:
Sounds good — this works for me. I'll leave the branch to you and stop pushing changes so as not to disrupt your plan (especially with main moving so fast these days). Happy to help on the contributor side if anything comes up — zh phrasing QA, triaging the review findings, or testing the build. Really looking forward to seeing the in-app language setting and the bilingual voice controls land. Thanks for taking this on!

kaishi00:
## Maintainer verification at head `c54d4f2` (this exact SHA)

**Integration**
- Current main (Voice background suspension #161, CarPlay Voice #165) merged in via a **normal merge commit** (`61daa4f`). The five contributor commits (`bd14c45`…`c33daee`) are untouched — no rebase, no squash, no force-push. All maintainer work is appended as new commits by kaishi00.
- Intended merge: a normal GitHub merge commit so both authors' commits retain authorship.

**Localization review (re-verified against current code before fixing)**
- Already fixed in the contributor's later commits (verified, no action): ConnectionFailure partial-localization fragments, KanbanService trailing literal, `image_input_mode` value safety, VoiceSettings profile title + empty-provider variants, ChatReturnSurface `Sessions`, VoiceConversationSheet mic label, SidebarView `1 conversations` (catalog plural variation), ConnectionSetupForm `Port`/`Entered`.
- Fixed in this round: Memory/Skill display labels; the composer gateway-notice / Copy / Done / Off String-ternary branches; the credentials-explanation fragment concatenation (now one key); `Tailscale hostname / address`; singular "Archive 1"/"Delete 1" (localized `%lld` key for all counts); plural grammar moved into catalog variations (`Decomposed into %lld task`, `Move/Assign/More … %lld selected task`); frozen `static let` localized tables → computed; every raw String-context literal surfaced by the audit (voice/audio errors, CarPlay state titles and buttons, kanban status names, setup-wizard guidance, about-page sections, profile-settings subtitles) — 130+ keys added with zh-Hans.
- Obsolete findings skipped with reason: the `docs/localization` tooling and `scripts/set-sim-english.sh` were removed in `19f677d`.

**In-app App Language** — Settings → Appearance → App Language (System Default / English / 简体中文), global, persisted; runtime switch via the new `AppLocalization` facility + root `.environment(\.locale)` and content-identity rebuild. No `AppleLanguages` mutation, no relaunch. English-source fallback for missing translations; UI language never touches STT/TTS/speech/provider configuration (pinned by tests).

**Multilingual spoken Voice controls** — built-in Stop/End phrase lists now include the Chinese commands alongside English in every UI language; exact whole-utterance matching unchanged; stored lists exactly equal to the previously shipped defaults migrate to the new built-ins, customized lists are preserved.

**Checker** — `scripts/check-l10n-coverage.py` now requires a real zh-Hans unit per key (translated state, non-empty, matching printf placeholders incl. positional forms), scans `AppLocalization.string(` and the previously-uncovered SwiftUI APIs, validates `AppShortcuts`/`InfoPlist` catalogs, and pins dynamic sites via `REGRESSION_KEYS`. Local run: **1,464 call sites, all resolve; catalog zh-Hans validation green.** Python unit tests: **32/32**. `plan-tests.py validate`: OK.

**Tests (Mac build host, iPhone 17 Pro simulator, head `c54d4f2`)**
- Full **ConduitTests: 2,373 / 2,373 passed, 0 failures** (includes the new AppLanguage, multilingual spoken-command, and protocol-value suites).
- **ConduitUITests could not initialize on this machine** — the runner died before executing any test with the known local infrastructure wedge (`Timed out waiting for AX loaded notification`, reproduced twice, unrelated to this branch; recorded in earlier PRs as the local Mac UI-runner environment issue). The UI lane is covered by hosted CI.

**Hosted CI**: workflows on this fork are maintainer-gated and are **awaiting maintainer approval** — they have not executed yet, so no CI status is claimed. Nothing merged.

kaishi00:
## Summary
Adds Simplified Chinese (`zh-Hans`) localization through String Catalogs, routes explicit String-context copy through a new `AppLocalization` facility driven by an in-app App Language setting, and adds a CI catalog-coverage checker. The infrastructure is sound, but there are a few concrete defects: one catalog value stores literal escaped-unicode garbage, the root `.id()` language rebuild resets the whole view tree (and re-runs the primary-window claim), and several raw/dead-key sites still render English in `zh-Hans`.

## Findings

### [severity: major] Catalog translation is stored as literal escaped Unicode and will render as `\u91cd…`
**File:** `Conduit/Localizable.xcstrings:9361`
**Issue:** The `zh-Hans` value for `"Rename conversation"` is written as `"\\u91cd\\u547d\\u540d\\u5bf9\\u8bdd"`. That is not JSON-escaped Unicode; after JSON parsing the value is the literal characters `\u91cd\u547d\u540d\u5bf9\u8bdd`. `.alert("Rename conversation", …)` at `Conduit/Views/SidebarView.swift:334` will display backslash-u gibberish in the rename dialog for Chinese users. This is the only such value (verified across all three catalogs). The checker accepts it because it only asserts non-empty/`translated`.
**Suggestion:** Replace with the real characters `重命名对话`, and add a checker assertion that translated values contain no literal `\u` escape sequences.

### [severity: major] Root `.id()` language rebuild resets all view state and re-runs the primary-window claim
**File:** `Conduit/ConduitApp.swift:102-107`
**Issue:** `.id(appLanguage.contentIdentity)` applied to `RootView()` forces a full tear-down/re-create of the entire view hierarchy on every App Language change. This resets every descendant `@State`/`@StateObject` (navigation, sheets, scroll position, composer drafts, Kanban/Voice view models at `Conduit/Views/KanbanViews.swift`, `Conduit/Views/VoiceSettingsView.swift`, etc.), contradicting the `AppLanguageStore` comment "no application state is reset". More seriously, `RootView` owns `@State private var isPrimaryWindow` and its `onDisappear`/`onAppear` call `ConduitWindowClaimKeeper.releaseClaim()` / `claimPrimaryWindow()` (`Conduit/Views/RootView.swift:43-61`). Because the identity change destroys and re-creates `RootView`, the new instance starts with `isPrimaryWindow == false` and re-runs `claimPrimaryWindow()`. If the new `onAppear` runs before the old `onDisappear` (SwiftUI does not guarantee the ordering during replacement), the claim still reads `isClaimed == true`, returns `false`, and the new primary window calls `dismissWindow()` — dismissing the app's main window on a language change. Even in the benign ordering, the claim is released and re-claimed and scene-phase handling is briefly dropped.
**Suggestion:** Don't couple the language refresh to view identity. Prefer observing the locale/language in the views that render localized strings (e.g. a custom `LocalizedStringKey`/environment value, or make `AppLocalization`-backed text computed in `body`), or scope the rebuild below `RootView`. At minimum, move the `ConduitWindowClaimKeeper` claim out of the identity-sensitive view (or guard `claimPrimaryWindow` so a same-window rebuild cannot self-dismiss) before adopting `.id()`.

### [severity: major] String-context copy still unlocalized, and the checker cannot see this class
**File:** `Conduit/Views/Components/ComposerBar.swift:966`; `Conduit/Voice/HermesVoiceConfigurationService.swift:550`
**Issue:** `appState.errorMessage = "Could not prepare \(name) for upload."` is a raw literal (key absent from the catalog) rendered back to the user; the neighboring `Off`/`Extra High`/accessibility strings in the same file were wrapped. Likewise the speed-field help text `"Speech rate multiplier (0.25–4.0); …"` remains raw while every other `help:` in that function was wrapped. Neither key exists in `Localizable.xcstrings`, so both stay English under `zh-Hans`. The coverage checker misses both because it only scans literal first arguments of a fixed initializer allowlist and `AppLocalization.string(`, not assignments to `String` state or `help:` fields.
**Suggestion:** Wrap both with `AppLocalization.string(...)` and add the keys/translations. Extend the checker to cover user-facing `String` state assignments and `help:`/`accessibilityLabel`/`accessibilityHint` literals.

### [severity: minor] Accessibility label catalog key uses the wrong placeholder type and is unreachable
**File:** `Conduit/Views/Components/ComposerBar.swift:640` (catalog `Conduit/Localizable.xcstrings`: `"Delegate agents, %@ active"`)
**Issue:** `appState.activeAgents` is an `Int` (`Conduit/Services/AppState.swift:702`), so the SwiftUI `LocalizedStringKey` runtime key is `Delegate agents, %lld active`. The catalog only defines `Delegate agents, %@ active`, so the lookup misses and the VoiceOver label stays English. The checker passes this because `catalog_has` treats `%@` and `%lld` as interchangeable (`scripts/check-l10n-coverage.py:273-278`), which is exactly the type-blindness that previously allowed the `contextUsed` `%lld`/`%@` regression. (The sibling `Context usage, %lld percent` is correct.)
**Suggestion:** Change the catalog key to `Delegate agents, %lld active` (or pre-format the value and use `%@`), and make the checker emit/validate type-aware skeletons (`%lld` for `Int` interpolations) rather than collapsing every interpolation to `%@`.

### [severity: minor] Checker's comment stripping and placeholder compare have gaps
**File:** `scripts/check-l10n-coverage.py:237-248, 325-343`
**Issue:** `strip_comment_lines` only removes full-line comments, so a trailing `// Text("…")`-style comment can be scanned as a call site. In `placeholders_compatible`, `sorted(key_specs)`/`sorted(value_specs)` at line 333 sort `(position, kind)` tuples where `position` may be `None` or an `int`; a mixed positional/non-positional format (e.g. `%1$@ %@`) raises `TypeError: '<' not supported between 'int' and 'NoneType'`, crashing the gate instead of reporting.
**Suggestion:** Strip trailing comments as well (respecting string literals), and normalize positions before sorting (e.g. sort by `(pos is not None, pos or 0, kind)`) so mixed formats are compared deterministically.

### [severity: minor] Force unwrap `selectedSource!` (guarded, but avoidable)
**File:** `Conduit/Views/SidebarView.swift:273`
**Issue:** `selectedSource == nil ? … : AppLocalization.string("No \(selectedSource!.label) Sessions")` force-unwraps in the else branch. It is logically guarded, but it violates the repo's no-force-unwrap convention and `label` is only safe because the ternary is evaluated lazily.
**Suggestion:** Bind once: `if let source = selectedSource { … "No \(source.label) Sessions" } else { "No Sessions" }`.

### [severity: minor] `bundledBilingualPack` still force-unwraps and now rebuilds per access
**File:** `Conduit/Voice/Wake/WakeModelDescriptor.swift:46-62`
**Issue:** Converting the `static let` to a computed `static var` re-evaluates the whole descriptor (including `URL(string:)!` and all `AppLocalization.string` calls) on every access, including as a default argument in `SherpaWakeWordService.swift:61`. The two `URL(string:)!` force unwraps remain, so a malformed literal traps on first access.
**Suggestion:** Keep checked construction (`URL(string:)` returning `nil` handled explicitly or a `#if DEBUG` assertion), and consider caching the non-localized parts while localizing only `displayName`.

### [severity: nit] Default-phrase migration compares `Set`s, dropping order/duplicates
**File:** `Conduit/Voice/VoiceTypes.swift:90-94`
**Issue:** `migratedDefaultPhrases` compares `Set` values, so a stored list that merely reorders or duplicates the previous defaults is treated as untouched and silently replaced by `current`, discarding the user's ordering.
**Suggestion:** Compare canonicalized arrays in order (and note duplicates) if ordering is a user customization; otherwise document that only set-membership matters.

### [severity: nit] Vacuous test loop in App Language tests
**File:** `ConduitTests/AppLanguageTests.swift:179-190`
**Issue:** `for language in AppLanguage.allCases` never applies `language` to the selection (`AppState.builtInSlashCommands` reads the persisted global), so every iteration asserts against the same language; the test does not actually verify the raw-token invariant "under every App Language".
**Suggestion:** Set `standardDefaults` per iteration (or assert against `AppLocalization.string(..., language:)` directly) so each language is exercised.

## Verdict
REQUEST_CHANGES

The escaped-Unicode catalog value produces visibly broken `zh-Hans` copy, the root `.id()` rebuild reintroduces a window-claim/state-reset hazard for a shipping feature, and the `Delegate agents`/raw-string misses show the new CI gate still cannot catch placeholder-type and non-scan-node regressions. These are bounded fixes but should land before merge.

<a href="https://opencode.ai/s/clpZQmYH">New%20session%20-%202026-09-12T21%3A16%3A53.080Z</a>
[opencode session](https://opencode.ai/s/clpZQmYH)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34719446735)

kaishi00:
## Summary
Adds Simplified Chinese (`zh-Hans`) localization via three String Catalogs, routes explicit String-context copy through a new `AppLocalization` facility driven by an in-app App Language setting, and adds a CI coverage checker. The catalog infrastructure is sound, but the migration is still incomplete, one catalog value is visibly broken, and the root language-rebuild introduces a view-state/window-claim hazard.

## Findings

### [major] `zh-Hans` value is stored as literal escaped Unicode and renders as `\u91cd…`
**File:** `Conduit/Localizable.xcstrings:9371` (key `Rename conversation` at `:9366`; used at `Conduit/Views/SidebarView.swift:334`)
**Issue:** The value is `"\\u91cd\\u547d\\u540d\\u5bf9\\u8bdd"`. In JSON, `\\` decodes to a backslash, so the runtime string is the literal characters `\u91cd\u547d\u540d\u5bf9\u8bdd`, not `重命名对话`. `.alert("Rename conversation", …)` will display backslash-u gibberish to Chinese users. The checker accepts it because it only asserts non-empty/`translated`.
**Suggestion:** Replace with the real characters `重命名对话`, and have the checker reject values containing literal `\uXXXX` sequences.

### [major] Root `.id()` language rebuild resets all view state and can re-run the primary-window claim
**File:** `Conduit/ConduitApp.swift:103`
**Issue:** `.id(appLanguage.contentIdentity)` on `RootView()` forces a full tear-down/re-create of the hierarchy on every language change, resetting every descendant `@State`/`@StateObject` (navigation, sheets, scroll, composer drafts, Kanban/Voice view models) — contradicting `AppLanguage.swift:95`'s "no application state is reset". More seriously, `RootView` owns `isPrimaryWindow` and its `onAppear`/`onDisappear` drive `ConduitWindowClaimKeeper.claimPrimaryWindow()`/`releaseClaim()` (`RootView.swift:43-61`). The re-created instance starts `isPrimaryWindow == false`; if its `onAppear` runs before the old `onDisappear` (SwiftUI does not guarantee ordering during replacement), the claim still reads `isClaimed == true`, returns `false`, and the new primary window calls `dismissWindow()` — dismissing the main window on a language change.
**Suggestion:** Don't couple the language refresh to view identity. Resolve localized text in `body` (or scope the rebuild below `RootView`), and at minimum move the window claim out of the identity-sensitive view so a same-window rebuild cannot self-dismiss.

### [major] Large areas of user-facing `String`-context copy remain unlocalized, and the checker cannot see them
**File:** `Conduit/Services/AppState.swift:2344` (and 52 more `errorMessage = "…"` sites in that file), `Conduit/Voice/HermesVoiceConfigurationService.swift:161-310`, `Conduit/Services/KanbanStore.swift:194`, `Conduit/Views/Components/ComposerBar.swift:966`
**Issue:** Only 7 of 65 `errorMessage = "…"` literals are wrapped in `AppLocalization.string`; the rest render English verbatim (e.g. `"Failed to load sessions: …"`, `"Could not delete this conversation: …"`, `"That conversation belongs to another workspace. Switch profiles to open it."`, `"This Hermes gateway is too old to report voice readiness."`, `"Connect to a Hermes dashboard to use Kanban."`). These are shown directly (`ComposerBar.swift:419` `Text(error)`, `VoiceSettingsView.swift:228`), so primary error flows stay English under `zh-Hans`. The coverage checker scans only `String(localized:)`/`AppLocalization.string(` and literal first-arguments, so assignments/returns to `String` state are invisible and CI reports success.
**Suggestion:** Wrap each user-facing literal with `AppLocalization.string(…)` and add the entries. Extend the checker to cover user-facing `String` state assignments and `String`-returning/parameter sites, or add a reverse audit.

### [minor] `Delegate agents, %@ active` uses the wrong placeholder type and is unreachable
**File:** `Conduit/Views/Components/ComposerBar.swift:640` (catalog key `Delegate agents, %@ active`)
**Issue:** `appState.activeAgents` is `Int` (`AppState.swift:702`), so the runtime `LocalizedStringKey` is `Delegate agents, %lld active`; the catalog defines only `%@`. The lookup misses and the VoiceOver label stays English. The checker accepts it because `catalog_has` treats `%@` and `%lld` as interchangeable.
**Suggestion:** Change the key to `Delegate agents, %lld active` (or pre-format the value and keep `%@`), and make `catalog_has` type-aware.

### [minor] Several remaining raw `String`-context sites defeat localization
**File:** multiple
**Issue:** Verified untranslated (keys absent from the catalog unless noted):
- `Conduit/Views/Components/ComposerBar.swift:990-998` — `modelAccessibilityLabel` builds `"\(model), \(reasoning)…"` with raw ASCII separators, mixing English punctuation with localized clauses.
- `Conduit/Views/Kanban/KanbanTaskDetailView.swift:474` — `AppLocalization.string("Unassigned") + " → default"` (raw suffix).
- `Conduit/Views/Kanban/KanbanV2Support.swift:118` — `effortCapitalized` localizes only `xhigh`; the other valid tokens (`none/minimal/low/medium/high/max/ultra`) fall through `value.capitalized`.
- `Conduit/Views/Kanban/KanbanV2Support.swift:421-422` — activity detail appends raw `"as \(s)"` / `"for \(a)"`.
- `Conduit/Views/Kanban/KanbanV3ASupport.swift:261` — returns raw `"Default"`.
- `Conduit/Views/VoiceSettingsView.swift:513` — raw Apple on-device speech detail.
- `Conduit/Views/Components/VoiceInputLevelMeter.swift:65` — accessibility value returns raw `"Silent"`.
- `Conduit/Views/ChatView.swift:2532` — `Text(active ? "Responding" : "Finishing")`; neither key is in the catalog.
**Suggestion:** Route each through `AppLocalization.string(…)` with complete keys/placeholders (and catalog entries), rather than concatenating fragments.

### [minor] Checker: comment stripping shifts reported line numbers
**File:** `scripts/check-l10n-coverage.py:237-248` (used at `:389`)
**Issue:** `extract_sites` records offsets after `strip_comment_lines()` deletes whole comment lines, but `check()` maps offsets back onto the original source (`line = source.count("\n", 0, offset) + 1`). A comment before a missing key yields an incorrect diagnostic line.
**Suggestion:** Replace removed lines with same-length whitespace so offsets stay stable, and strip trailing comments (respecting string literals) too.

### [minor] Checker can crash on mixed positional/non-positional specifiers
**File:** `scripts/check-l10n-coverage.py:333`
**Issue:** `sorted(key_specs)`/`sorted(value_specs)` sort `(position, kind)` tuples whose first element may be `None` or `int`. A format such as `%1$@ %@` makes Python compare `int` and `None`, raising `TypeError` and crashing the CI gate instead of reporting.
**Suggestion:** Normalize before sorting, e.g. `sorted(specs, key=lambda s: (s[0] is not None, s[0] or 0, s[1]))`.

### [minor] Checker does not fail when a required secondary catalog is missing
**File:** `scripts/check-l10n-coverage.py:396`
**Issue:** `SECONDARY_CATALOGS` (`AppShortcuts.xcstrings`, `InfoPlist.xcstrings`) are required, but a missing file hits `continue` without recording a problem, so deleting a required catalog still passes CI.
**Suggestion:** Record a catalog problem before `continue`.

### [minor] Newly introduced force unwrap in `No Sessions` empty state
**File:** `Conduit/Views/SidebarView.swift:273`
**Issue:** `AppLocalization.string("No \(selectedSource!.label) Sessions")` force-unwraps in the ternary else-branch. Logically guarded, but it violates the repo's no-force-unwrap convention.
**Suggestion:** Bind once: `if let source = selectedSource { … "No \(source.label) Sessions" } else { … }`.

### [minor] `bundledBilingualPack` force-unwraps URLs and rebuilds on every access
**File:** `Conduit/Voice/Wake/WakeModelDescriptor.swift:46-62`
**Issue:** Converting the frozen `static let` to `static var` re-evaluates the whole descriptor (all `AppLocalization.string` calls plus two `URL(string:)!`) on every access, including as a default argument in `SherpaWakeWordService.swift:61`. The force unwraps remain a crash vector if a literal changes.
**Suggestion:** Use checked URL construction (handle `nil`/`#if DEBUG` assert) and cache the non-localized fields while localizing only `displayName`.

### [nit] Default-phrase migration compares `Set`s, discarding ordering/duplicates
**File:** `Conduit/Voice/VoiceTypes.swift:90-94`
**Issue:** `migratedDefaultPhrases` compares canonicalized `Set`s, so a stored list that merely reorders the previous defaults is treated as untouched and silently replaced by `current`, losing the user's ordering.
**Suggestion:** Compare canonicalized arrays in order (documenting duplicate handling) if ordering is a customization.

### [nit] Vacuous App Language test loop
**File:** `ConduitTests/AppLanguageTests.swift:179-190`
**Issue:** `for language in AppLanguage.allCases` never applies `language`; `AppState.builtInSlashCommands` reads the persisted global, so every iteration asserts against the same language. The "under every App Language" invariant is not actually exercised.
**Suggestion:** Set `standardDefaults.set(language.rawValue, forKey: AppLanguageStore.defaultsKey)` inside the loop (or assert via `AppLocalization.string(…, language:)`).

### [nit] Test overwrites/clears the persisted App Language instead of restoring it
**File:** `ConduitTests/VoiceSpokenControlTests.swift:168`
**Issue:** `defer { standardDefaults.removeObject(forKey: AppLanguageStore.defaultsKey) }` always removes the key; if a prior value existed, later tests observe `.system` rather than the original selection.
**Suggestion:** Capture `object(forKey:)` before modifying and restore it in `defer`, removing only when it was originally absent.

### [nit] Newly introduced force unwrap in slash-catalog normalization
**File:** `Conduit/Services/AppState.swift:9148`
**Issue:** `description?.isEmpty == false ? description! : …` force-unwraps under a lazy-guard idiom; safe today but against convention and newly added by this PR.
**Suggestion:** Use `if let description, !description.isEmpty { … }`.

Note: no new retain cycles or Combine subscription leaks were found, and the nonisolated `AppLocalization`/bundle-cache path reads only thread-safe `UserDefaults`/`static let` state, so no data races were identified.

## Verdict
REQUEST_CHANGES

The escaped-Unicode catalog value produces visibly broken `zh-Hans` copy, the root `.id()` rebuild risks dismissing the primary window and drops user state, and a large share of user-facing error copy (53+ raw `errorMessage` literals) remains English while the new CI gate cannot detect that class. These are bounded but should land before merge.

<a href="https://opencode.ai/s/yDdQ6r4l">New%20session%20-%202026-09-12T22%3A40%3A23.689Z</a>
[opencode session](https://opencode.ai/s/yDdQ6r4l)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34723324579)

kaishi00:
## Correctness pass at head `8b3696a` (this exact SHA)

**1. Root identity rebuild removed.** The `.id(contentIdentity)` mechanism is gone. Literal-key SwiftUI text re-renders reactively via the root `.environment(\.locale)`; every view that builds String-context copy now observes `AppLanguageStore` (the established `@ObservedObject` singleton idiom, 69 view structs), so a language change re-renders exactly the localized views — RootView is never recreated, so navigation paths, sheets, scroll state, composer drafts (`ComposerDraftStore`), Voice/Kanban state objects, and the `ConduitWindowClaimKeeper` primary-window claim (which runs only in genuine `onAppear`/`onDisappear`) are untouched. Regression tests pin message/profile/voice-phase/slash-identity/composer-draft survival across System Default → English → 简体中文 → System Default.

**2. Conditional-branch literals fixed.** The zh-Hans simulator smoke caught the class the environment locale cannot reach: `Text(condition ? "A" : "B")` binds the verbatim `Text(String)` overload, so branch literals stayed English. Every such site (sidebar, voice sheet, model picker, login, setup wizard, markdown preview chips, notifications, phrase editor) now resolves each branch through `AppLocalization.string`, and `REGRESSION_KEYS` was pruned to genuinely dynamic lookups so stale entries can't give false confidence.

**3. Placeholder type families enforced.** The checker emits runtime-accurate keys (String-wrapped → `%@`; Int casts/`.count`/count-shaped identifiers → `%lld`) and never normalizes across families — the Delegate-agents class (`accessibilityLabel("Delegate agents, \(Int)…")` vs a `%@` catalog key) is now caught and was fixed at the source. Six Kanban catalog keys were renamed `%lld`→`%@` to match the convention; their stale `en` units (which misformatted English as garbage integers — caught by the full-suite run) were removed, and the checker now validates placeholder compatibility for EVERY language, not just zh-Hans.

**4. Malformed catalog values.** `Rename conversation` carried a literal `\u91cd…` escape as its zh value — fixed to 重命名对话; the checker rejects literal Unicode escapes in any value, and all three catalogs (Localizable/InfoPlist/AppShortcuts) audit clean.

**5. User-facing String-context audit.** 65 raw `errorMessage` assignments (AppState, KanbanStore, ComposerBar, HermesVoiceConfigurationService), the voice speed `help:`, the raw `Delete \(session.title)` accessibility label, and `navigationTitle`/alert-class modifiers are now covered — checker scans `errorMessage =`/`help:`/`purposeText:` assignments plus `.accessibilityLabel`/`.accessibilityHint`/`.accessibilityValue`/`.navigationTitle`. 58 new zh-Hans entries; protocol tokens (cron action verbs, config values, `display.tool_progress` all/off) stay raw and are pinned by a test that actively switches every App Language.

**Verification at `8b3696a`**: full ConduitTests **2,377/2,377 passed** on the Mac build host; checker **1,686 call sites** resolve with type-matched placeholders; Python suite **160 passed** (53 checker tests incl. the full placeholder matrix and escape detection); `plan-tests.py validate` OK. Simulator screenshots: English and zh-Hans both render end-to-end (开始对话 / 模型 / 给 Hermes 发消息…). Interactive tap-through switching remains covered by hosted UI lanes (local XCUITest is AX-wedged) and physical-device QA; hosted CI on this SHA is maintainer-gated and has not executed.

kaishi00:
## Summary
Adds zh-Hans localization via three String Catalogs, an in-app App Language facility (`AppLanguage`/`AppLocalization`), bilingual built-in Voice commands, and a CI catalog-coverage checker. The infrastructure and most of the prior review fixes are in place, but several conditional-branch literals still bind the non-localizing `Text(String)`/`Label(String)`/`Button(String)` overloads and the new CI gate cannot see them.

## Findings

### [severity: major] Conditional-branch literals still bypass localization (claim "every such site" is not met)
**File:** `Conduit/Views/ChatView.swift:2541` (also `Conduit/Views/VoiceConversationSheet.swift:118,276`, `Conduit/Views/SidebarView.swift:289,1153,1159`, `Conduit/Views/VoiceSettingsView.swift:667`, `Conduit/Views/ModelPickerView.swift:369`, `Conduit/Views/ProfilePickerSheet.swift:31`, `Conduit/Views/Kanban/KanbanTaskDetailView.swift:706`, `Conduit/Views/AuxiliaryViews.swift:1131`, `Conduit/Views/Components/MarkdownText.swift:2023`, `Conduit/Views/ChatSupportSheets.swift:229`)
**Issue:** Commit 8b3696a claims it "wraps every conditional-branch copy site", but these remain raw, e.g. `Text(active ? "Responding" : "Finishing")`, `Label(controller.isOutputMuted ? "Unmute" : "Mute", …)`, `Label(action.label.isEmpty ? "Reclaim" : action.label, …)`, `Button(editingIndex == nil ? "Add" : "Save", …)`, `Label(job.enabled ? "Pause" : "Resume", …)`, `Section(pinnedSessions.isEmpty ? "Sessions" : "Recent")`, `Text(displayedTitle.isEmpty ? "Default" : displayedTitle)`. A ternary of string literals infers `String`, so the verbatim overload is selected and `zh-Hans` renders English. `Responding`/`Finishing`/`Unmute`/`Mute`/`Reclaim` are not in the catalog at all; `Add`/`Save`/`Pause`/`Resume`/`Hidden`/`Visible`/`Reorder`/`Done`/`Recent`/`Default`/`Code` exist but are never consulted here. `VoiceConversationSheet.swift:276` (`.accessibilityLabel("\(isUser ? "You" : "Hermes"): …")`) and `ChatSupportSheets.swift:229` (`?? "File"`) have the same problem and are VoiceOver-visible.
**Suggestion:** Resolve each branch through `AppLocalization.string(...)` (or split into separate localized `Text`/`Label` views), and add catalog entries with zh-Hans for the five missing keys. This is the exact class the guard was meant to pin.

### [severity: minor] Coverage checker cannot see the above class of miss
**File:** `scripts/check-l10n-coverage.py:299-323`
**Issue:** `extract_sites` only checks a string literal when it is the *immediate* first token after the initializer (`SWIFTUI_RE` + `if source[i] == '"'`). Raw literals passed through `String` parameters, ternaries, or `??` defaults (`Text(cond ? "A" : "B")`, `navigationTitle(x ?? "File")`) are invisible, so CI reports "OK" while these flows stay English — the same false confidence the checker exists to prevent.
**Suggestion:** Add a heuristic pass for string literals inside ternaries / `??` in known localizable argument/assignment positions, or maintain a small explicit list of the remaining sites until each is wrapped.

### [severity: minor] Comment stripping shifts diagnostic line numbers
**File:** `scripts/check-l10n-coverage.py:285-296, 447`
**Issue:** `strip_comment_lines` deletes whole-line comments, but `check()` maps the stripped source's offsets back onto the original `source` (`source.count("\n", 0, offset) + 1`). Any file with comment lines before a missing key reports a too-low line, sending authors to the wrong place.
**Suggestion:** Replace removed comment lines with equal-length whitespace (or strip while preserving offsets) so reported `file:line` values remain accurate.

### [severity: minor] Required secondary catalogs are not required
**File:** `scripts/check-l10n-coverage.py:452-455`
**Issue:** `SECONDARY_CATALOGS` is documented as required, but a missing `AppShortcuts.xcstrings`/`InfoPlist.xcstrings` hits `continue` without recording a problem, so deleting a required catalog still passes the gate.
**Suggestion:** Record `key_problems[f"{name}: <catalog>"] = ["required catalog file is missing"]` before `continue`.

### [severity: minor] Newly touched lazy force unwrap
**File:** `Conduit/Services/AppState.swift:9148`
**Issue:** `description?.isEmpty == false ? description! : AppLocalization.string("Hermes command")` force-unwraps under a lazy-guard idiom. It is safe today but violates the repo's no-force-unwrap convention and was touched by this PR.
**Suggestion:** Use `if let description, !description.isEmpty { … }`.

### [severity: minor] `bundledBilingualPack` force-unwraps URLs and now rebuilds on every access
**File:** `Conduit/Voice/Wake/WakeModelDescriptor.swift:46-64`
**Issue:** Converting the frozen `static let` to a computed `static var` re-evaluates the whole descriptor (two `URL(string:)!` plus every `AppLocalization.string` call) on each access, including as a default argument in `SherpaWakeWordService.swift:61`. The force unwraps remain a crash vector if a literal changes.
**Suggestion:** Keep checked URL construction (handle `nil` / `#if DEBUG` assert), and cache the non-localized descriptor while localizing only `displayName` at use time.

### [severity: nit] Default-phrase migration compares `Set`s, discarding order/duplicates
**File:** `Conduit/Voice/VoiceTypes.swift:90-94`
**Issue:** `canonicalStored == canonicalPrevious` compares `Set`s, so a stored list that merely reorders or duplicates a previous default is treated as untouched and silently replaced by `current`, dropping the user's ordering.
**Suggestion:** Compare canonicalized arrays in order if ordering is a customization, or document that only membership matters.

### [severity: nit] Vacuous App Language test loop
**File:** `ConduitTests/AppLanguageTests.swift:266-277`
**Issue:** `for language in AppLanguage.allCases` never applies `language`; `AppState.builtInSlashCommands` reads the persisted global, so every iteration asserts against the same language. The stated "under every App Language" invariant is not exercised.
**Suggestion:** Set `standardDefaults.set(language.rawValue, forKey: AppLanguageStore.defaultsKey)` inside the loop (or assert via `AppLocalization.string(…, language:)`).

### [severity: nit] Spoken-control test does not restore a pre-existing selection
**File:** `ConduitTests/VoiceSpokenControlTests.swift:168`
**Issue:** `defer { standardDefaults.removeObject(forKey: AppLanguageStore.defaultsKey) }` always deletes the key; if a prior value existed, later tests observe `.system` instead of the original selection.
**Suggestion:** Capture `object(forKey:)` before modifying and restore it in `defer`, removing only when it was originally absent.

### [severity: nit] Committed generated screenshots
**File:** `.smoke/l10n-smoke-en.png`, `.smoke/l10n-smoke-zh.png`
**Issue:** ~1.5 MB of smoke-test PNGs are committed as artifacts, adding repo noise with no shipping value (previous review asked for artifact cleanup).
**Suggestion:** Drop them from the PR or move them out-of-tree; they can live in the PR description instead.

Note: no new retain cycles, Combine subscription leaks, `MainActor` violations, or unguarded force unwraps were identified; `AppLocalization`/`AppLanguage.current` read only thread-safe `UserDefaults`/`static let` state. The prior critical config-value translation and escaped-Unicode catalog value are correctly fixed.

## Verdict
REQUEST_CHANGES

The core deliverable — complete zh-Hans coverage with a CI guard that prevents English regressions — does not hold: multiple named conditional-branch controls still render English, and the checker structurally cannot detect that class. The fixes are bounded and mechanical.

<a href="https://opencode.ai/s/Bu8gmRdH">New%20session%20-%202026-09-13T01%3A16%3A57.848Z</a>
[opencode session](https://opencode.ai/s/Bu8gmRdH)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34730091284)

PR #164, opened 2026-09-12
feat(signing): enable CarPlay Voice Based Conversation entitlement

kaishi00:
## Summary

Declares the Apple-approved **CarPlay Voice Based Conversation** managed capability for Conduit and proves the full signing contract end-to-end on the Mac build machine.

> **This PR only establishes and verifies CarPlay Voice Based Conversation provisioning. CarPlay application/UI integration will be implemented separately.**

Deliberately boring by design: a later CarPlay implementation failure can now be clearly distinguished from a signing/provisioning failure. **Do not merge yet** (decision left to the maintainer; no release action is bundled).

## Approved entitlement

```text
com.apple.developer.carplay-voice-based-conversation
```

## What changed

One file, two lines — `Conduit/Conduit.entitlements`:

```diff
 <dict>
+    <key>com.apple.developer.carplay-voice-based-conversation</key>
+    <true/>
     <key>aps-environment</key>
     <string>production</string>
 </dict>
```

- **Target receiving it:** `Conduit` (app) only — `project.yml` already wires `CODE_SIGN_ENTITLEMENTS: Conduit/Conduit.entitlements` to the app target; test targets are unaffected. No `project.yml` change, no XcodeGen output change (`Conduit.xcodeproj` is generated and never committed).
- Nothing else: no CarPlay UI/scene/template/navigation code, no `CPTemplateApplicationScene`, no `CPVoiceControlTemplate`, no scene delegate, no background modes, no VoIP/PushKit/CallKit, no lifecycle changes from PR #161.

## Provisioning: what it took (the interesting part)

Apple's approval made the capability **available but not enabled**. Verified empirically before any change was requested of the account: a freshly minted App Store profile (via ASC API) did **not** contain the entitlement. Enabling it proved impossible to automate by design:

- The App Store Connect public API cannot express it — a `POST /v1/bundleIdCapabilities` attempt returned 409 with the server's own complete `capabilityType` enumeration, which contains **no CarPlay value**.
- Xcode CLI automatic signing (`xcodebuild archive -allowProvisioningUpdates`) is blocked on this team: development-profile generation requires at least one registered device, and the team has none (App Store distribution only).
- So the capability was enabled interactively in App Store Connect (Identifiers → `com.milim.relay` → CarPlay → Voice Based Conversation) by the account holder, and a gated re-run then confirmed `GATE_OK`: new profiles carry the entitlement.

Both pinned App Store profiles were then regenerated in place via the ASC API — same names, same team:

| Profile | Before | After |
[...]

PR #165, opened 2026-09-12
feat(carplay): CarPlay Voice V1 over the shared Voice runtime

kaishi00:
## Summary

Implements the first end-to-end CarPlay Voice experience: launching Conduit from CarPlay presents a native `CPVoiceControlTemplate` backed by the **same** Hermes connection, AppState, VoiceConversationController, audio ownership, lifecycle rules, and session identity the iPhone Voice UI uses. No second Voice stack, no separate Hermes client, no CarPlay-specific backend, no duplicate AppState.

> CarPlay is another presentation/lifecycle surface over the existing Voice architecture.

**Base:** current `main` — includes PR #161 (background suspension) and PR #164 (CarPlay Voice entitlement; provisioning is already proven and untouched here).

### 1. Native CarPlay scene with genuine phone + CarPlay coexistence

`Info.plist` sets **`UIApplicationSupportsMultipleScenes = true`** — Apple's contract for this flag is that UIKit creates no more than one scene of ANY kind, and this architecture requires the phone UIWindow scene and the connected `CPTemplateApplicationScene` to exist simultaneously. The CarPlay role pins `UISceneClassName = CPTemplateApplicationScene` explicitly (canonical manifest shape). The foreground role stays SwiftUI-managed with **no application-role configuration declared**.

Because `WindowGroup` would expose user-creatable duplicate iPad windows under multi-scene support, and SwiftUI's singleton `Window` scene is **iOS-unavailable** (macOS 13+/visionOS only — verified in the iOS 26 SDK and docs; the deployment floor is iOS 17), the phone keeps `WindowGroup` and duplicate iPad windows **dismiss themselves** via a process-scoped first-window claim guard in `RootView` (`ConduitWindowClaimKeeper`; only the primary window drives `handleScenePhase`, so a transient duplicate's lifecycle can never suspend a conversation the primary presents). Manifest facts are pinned by `CarPlaySceneManifestTests`; these plist assertions pin the supported CONFIGURATION — runtime coexistence needs the CarPlay Simulator / a vehicle and stays on the physical checklist (CarPlay Simulator is not installed on the build Mac).

`CarPlayVoiceSceneDelegate` is deliberately thin (documented main-thread contract, `MainActor.assumeIsolated`): it forwards `didConnect`/`didDisconnect` to the coordinator.

### 2. Same-AppState invariant (CarPlay-first launch)

`AppStateRuntimeRegistry` (MainActor) is the process-wide home of the single `AppState`. `ConduitApp` adopts it (`@StateObject private var appState = AppStateRuntimeRegistry.shared.appState`) and the CarPlay scene resolves the same instance — **whoever needs it first creates it**. `AppState.init` already performs the authoritative cold-launch bootstrap (saved-connection restore), so a CarPlay-first launch establishes connection readiness without any phone view running. The registry is a reference/lifecycle seam only; it owns zero business logic. `AppStateRuntimeRegistryTests` pin one-instance-across-accesses, the CarPlay-first ordering, and the coordinator binding.

The coordinator also converges disconnect/controls on the **bound** instance (`lastBoundAppState`), so a swapped provider seam can never split surface bookkeeping across two AppStates.

### 3. Voice lifecycle generalizes to presentation surfaces

`AppState` gains `isCarPlayVoiceSurfaceActive` and the derived rule

```text
Voice may own live foreground audio
iff at least one legitimate Voice presentation surface is active
        (hasActiveVoiceSurface = isSceneActive || isCarPlayVoiceSurfaceActive)
```

- `.active` / `.background` feed the controller's surface gate with `hasActiveVoiceSurface` — the phone never *fakes* `.active`; `isSceneActive` still governs all phone-specific transport/chat reconciliation (unchanged).
- `.background` **skips** the PR #161 suspension while CarPlay presents: Voice stays live for the driver (no restoration descriptor is consumed — there is nothing to restore).
- CarPlay disconnect: if the phone scene is active, Voice is untouched; with no surface left, `handleCarPlayVoiceSurfaceRemoved()` routes through the **same** `suspendVoiceConversationForBackground()` contract (sheet-open → suspend + restoration descriptor; sheet-closed → full stop). A disconnect is release, not spoken Goodbye (which remains Close).
- Surface **activation** re-asserts the controller gate (`handleCarPlayVoiceSurfaceActivated`) — no scene-phase event fires while the phone is locked, so the reconnect-with-phone-locked flow (`testReconnectAfterCarPlayOnlyDisconnectReArmsTheListenPath`) re-arms Listen through the activation itself.

### 4. One logical conversation: prepare vs present

`openVoiceConversation` is split:

- `prepareVoiceConversation(profile:startsFreshConversation:)` — the logical acquisition (profile switch, capability refresh, session continuation/creation, gateway install, suspension supersede). Returns `.deferred` ("not connected, stay pending" — preserves the pending-intent `false` contract) or `.handled`.
- The phone flow additionally presents the sheet (`showVoiceSheet`, auto-listen), exactly as before.

Attaching is a **generic** helper, `attachToLiveVoiceConversation()`, used by both surfaces: re-asserts the presentation-surface gate, re-arms the gateway only when detached, and retires the restoration descriptor **only after** the attach succeeds. `prepareVoiceConversation` attaches to an already-live session **before** the generic turn-running rejection when `startsFreshConversation == false` — so the phone composer Voice attaching to a CarPlay-owned conversation preserves the transcript, session identity, and assistant ownership (no `beginVoiceTurn` reset, no session churn, no in-flight-turn cancellation), and the same continuation works while an assistant turn is in flight. A failed re-arm surfaces a `.failed` outcome instead of presenting a dead sheet. Fresh requests (Siri/wake/new-conversation) keep their existing cancel-and-create behavior. `PendingVoiceIntent.Source.carPlay` was deliberately **not** added: CarPlay does not route through the intent store, so it can never inherit Siri's external-launch deadline semantics.

### 5. CarPlay surface: five states, deduped, presentation-gated

`CarPlayVoiceState` = ready / listening / processing / responding / error with the specified controller mapping (`.failed(message)` drops the message — no internals on the display). `CarPlayVoiceStateActivation` suppresses duplicates (`.speaking` ↔ `.muted` collapse), and the coordinator observes only the controller's published `state` — microphone-level events never reach CarPlay. iOS 26.4 `CPVoiceControlState.actionButtons` add exactly one control per state: **Listen** at Ready/Error, **End** while active.

Activation is **presentation-gated**: `activateVoiceControlState` is a documented no-op before the template is presented, and an early activation would poison duplicate suppression (the ignored pre-presentation activation primes the dedupe, freezing the surface on Ready — e.g. connecting while a response is mid-playback). The latest desired state is therefore retained while the template is unpresented; the generation-fenced `setRootTemplate` completion marks presentation exactly once (stale completions from a superseded connection cannot present the new one), activates the retained state if it differs from the default, and only then resumes normal dedupe.

### 6. Controls

End → `AppState.closeVoiceConversation()` (the authoritative teardown — spoken Goodbye semantics untouched). Ready/Error **Listen** → the shared prepare/attach path + `controller.startListening()`; it never mutates the saved Continuous Conversation preference (pinned by test, with the preference persisted as `false`).

### 7. Stale-completion fencing

`connectionGeneration` rotates on every connect/disconnect. Establish/listen continuations (`completeVoiceEstablishment`, `completeListenTurn`) split out so the fence is deterministic: a prepare that was in flight across a disconnect applies the no-surface release contract instead of orphaning an armed session. Regression tests cover stale-establish, stale-reconnect, stale-completion-with-active-phone (conversation preserved), and stale-completion-with-no-surface (released, still restorable).

### 8. Audio

`VoiceAudioSessionCoordinator` remains the single `AVAudioSession` authority — no CarPlay-specific session code. Inspection result: the conversation policy (`.playAndRecord`/`.voiceChat`) already satisfies Apple's CarPlay voice guidance, and `defaultToSpeaker` is a no-op on accessory (car) routes, so **no route-specific mode adjustment is made**. The conservative half-duplex barge-in policy is preserved and now pinned for CarPlay routes (`carAudio` output stays `speakerSafeHalfDuplex` even with a car microphone present), alongside speaker/receiver/Bluetooth regression pins.

## Resolution round (review findings on the first head)

Three deterministic findings on `bbdd585` were fixed in `9061b33` (the current head):

1. **`.handled` now means logical Voice preparation SUCCEEDED.** Every rejection that leaves no usable Voice conversation armed (requested profile unselectable, non-fresh request while a non-Voice turn is running, Voice/STT/TTS capability unavailable, fresh-session creation failure, missing session or gateway) returns `.failed(message)`; the open contract stays router-friendly (`openVoiceConversation` returns true/consumed for `.failed`) but **never presents the Voice sheet and never arms auto-listen**. `.deferred` keeps the not-connected-yet pending contract; Siri fresh-conversation behavior is unchanged.
2. **A foreground phone scene is a Voice surface only while the Voice sheet is presenting**: `hasActiveVoiceSurface = (isSceneActive && showVoiceSheet) || isCarPlayVoiceSurfaceActive`. This closes the privacy case (CarPlay-only listening + foreground phone with the sheet closed + CarPlay disconnect → capture released instead of live audio with no controls) and keeps every legitimate flow working: presenting the sheet re-asserts the gate before auto-listen (`reassertVoiceSurfaceGate()`), sheet-open phone Voice survives a CarPlay disconnect, and the sheet-closed disconnect releases the runtime. Only the primary window drives `handleScenePhase`.
3. **The duplicate iPad window dismisses itself with environment-scoped `dismissWindow()`** — ID-scoped dismissal targets the whole WindowGroup and would have closed the primary window. A source-contract test pins the environment-scoped form; UIKit/iPad runtime behavior stays on the physical checklist (CarPlay Simulator is not installed on the build Mac).

Regression coverage added: prepare-outcome semantics (capability unavailable, missing session, fresh-session failure → `.failed`, no sheet, no auto-listen, no microphone), the surface-gate matrix (sheet open/closed × CarPlay disconnect), the phone re-listen after a closed-sheet background/foreground cycle, and the dismissal source contract.

## Test results (Mac build machine, Xcode 26.6)

- New `CarPlayVoiceTests`: **51/51 passed** — scene manifest (coexistence config), mapping, dedupe, template shape, registry identity, connect/establish, live-attach-without-restart, presentation gating (speak-before-presentation, stale completion vs new connection, presentation failure), phone-open attach (idle + in-flight turn), phone-background × CarPlay matrix, End convergence, Continuous-OFF re-listen, unavailable fail-closed (no microphone), stale-generation fencing, gate re-assert reconnect flow, failed-attach descriptor preservation, route pins, single-window claim.
- Full `ConduitTests`: **2341/2341 passed, 0 failures, exit 0** (includes VoiceConversationControllerTests, VoiceBackgroundSuspensionTests, Continuous Conversation tests, spoken Stop/End tests, PendingVoiceIntentLifecycleTests, VoiceAudioSessionCoordinatorTests, VoiceBargeInRoutePolicyTests).
- Signed Release device archive with the repo's existing manual signing: `ARCHIVE_EXIT=0`; signed entitlements carry `com.apple.developer.carplay-voice-based-conversation = true` (provisioning from PR #164 intact).
- CarPlay Simulator is not installed on the build Mac; runtime scene-coexistence and vehicle verification are deferred with the TestFlight checklist (below) — the plist/UI-test suite proves configuration and launch, not on-the-road coexistence.

## Review

Three-model review gate: 3× REQUEST CHANGES → all findings fixed. The convergent BLOCKER (CarPlay reconnect while the phone is locked never re-asserted the controller gate → Listen silently no-oped) is fixed by `handleCarPlayVoiceSurfaceActivated` with the reconnect regression test; the second BLOCKER-adjacent gap (a prepare in flight across a disconnect could orphan an armed session) is fixed by the fence's no-surface release contract with two dedicated tests. WARNINGs fixed: attach no longer discards the restoration descriptor when the gateway re-arm fails; disconnect/controls converge on the bound AppState; the root template installs before registry resolution; explicit Combine sink cancellation; `setRootTemplate` failures are logged; registry tests no longer spawn async Voice work.

## Scope guard (not in this PR)

No fillers, wake word, CarPlay transcript/history, settings UI, contacts, navigation, communication-entitlement usage, tool-result buttons, iOS-27-only features, or unrelated Voice cleanup. No provisioning changes.

## Physical TestFlight checklist (deferred, accumulates with the next shipped build)

Phone Voice: normal conversation; lock/unlock restoration; mic cold after restoration; network loss while backgrounded; AirPods; speaker; real Siri/phone interruption; spoken Goodbye no-resurrection.
CarPlay: app appears in the CarPlay launcher; cold CarPlay-first launch; wired/wireless connect; vehicle microphone input; assistant audio output; Listening/Processing/Responding transitions; Stop/Goodbye; Continuous Conversation ON/OFF; phone lock while the CarPlay conversation continues; disconnect/reconnect; phone ↔ CarPlay ownership transition.

## Summary by CodeRabbit

* **New Features**
  * Added CarPlay support for voice conversations with driver-friendly states and connection lifecycle handling.
  * Added Listen and End actions where supported.
  * Preserved active conversations when switching between phone and CarPlay.
  * Added shared voice-session management and CarPlay-first startup support.
  * Added safeguards for managing multiple phone windows.
  * Improved voice-surface gating and failure handling to prevent unintended listening.

* **Tests**
  * Added comprehensive CarPlay, lifecycle, routing, template, and shared-session coverage.

* **Documentation**
  * Clarified voice presentation-surface behavior.

### DIFF
--- a/Conduit/CarPlay/AppStateRuntimeRegistry.swift
+++ b/Conduit/CarPlay/AppStateRuntimeRegistry.swift
@@ -1,0 +1,38 @@
+//
+//  AppStateRuntimeRegistry.swift
+//  Conduit
+//
+//  Process-wide home of the single AppState instance.
+//
+//  SwiftUI's `@StateObject` evaluates its factory lazily when the foreground
+//  scene first renders, but a CarPlay-first launch can connect the CarPlay
+//  scene before any phone scene exists. Both surfaces therefore resolve the
+//  AppState through this MainActor registry: whoever needs it first creates
+//  it, and the other surface adopts the exact same instance. The registry is
+//  a reference/lifecycle seam only — it owns no business logic.
+//
+
+@MainActor
+final class AppStateRuntimeRegistry {
+    static let shared = AppStateRuntimeRegistry()
+
+    private var stored: AppState?
+
+    /// The process-wide AppState, created on first access. AppState's own
+    /// initializer performs the authoritative cold-launch bootstrap (saved
+    /// connection restore), so a CarPlay-first launch establishes connection
+    /// readiness without waiting for any phone view to run.
+    var appState: AppState {
+        if let stored { return stored }
+        let created = AppState()
+        stored = created
+        return created
+    }
+
+    internal init() {}
+
+    /// Test isolation only.
+    internal func resetForTesting() {
+        stored = nil
+    }
+}
--- a/Conduit/CarPlay/CarPlayVoiceCoordinator.swift
+++ b/Conduit/CarPlay/CarPlayVoiceCoordinator.swift
@@ -1,0 +1,335 @@
+//
+//  CarPlayVoiceCoordinator.swift
+//  Conduit
+//
+//  MainActor bridge between the UIKit CarPlay scene (scene delegate +
+//  CPInterfaceController) and the process-wide AppState/Voice stack.
+//
+//  Responsibilities are deliberately narrow — a reference, presentation, and
+//  lifecycle seam, NOT a second Voice owner:
+//    • installs the root CPVoiceControlTemplate on connect;
+//    • observes the shared VoiceConversationController's published state and
+//      forwards DEDUPLICATED, mapped states to CarPlay (never mic-level or
+//      transcript content);
+//    • rotates a connection generation so a stale async readiness completion
+//      after disconnect/reconnect can never reopen Voice or touch a dead
+//      interface controller — and applies the no-surface release contract
+//      when a prepare was in flight across a disconnect;
+//    • reports CarPlay surface (de)activation to AppState, which owns the
+//      Voice lifecycle policy (PR #161 rules stay authoritative).
+//
+
+import Combine
+import CarPlay
+import Foundation
+import OSLog
+
+private let carPlayLogger = Logger(subsystem: "com.milim.relay", category: "CarPlayVoice")
+
+/// Seam over `CPInterfaceController` so coordinator behavior is testable
+/// without a vehicle session.
+@MainActor
+protocol CarPlayInterfacing: AnyObject {
+    func setRootTemplate(
+        _ rootTemplate: CPTemplate,
+        animated: Bool,
+        completion: ((Bool, (any Error)?) -> Void)?
+    )
+}
+
+extension CPInterfaceController: CarPlayInterfacing {}
+
+@MainActor
+final class CarPlayVoiceCoordinator {
+    static let shared = CarPlayVoiceCoordinator()
+
+    private(set) weak var interfacing: (any CarPlayInterfacing)?
+    private(set) var template: CPVoiceControlTemplate?
+    /// Monotonic fence: every connect and disconnect rotates the generation;
+    /// async completions captured under an older generation are discarded.
+    private(set) var connectionGeneration: UInt64 = 0
+    private(set) var lastActivatedState: CarPlayVoiceState?
+    /// True once the root template's `setRootTemplate` completion reported
+    /// success for the CURRENT generation. `activateVoiceControlState` is a
+    /// documented no-op before presentation, and calling it early would
+    /// poison duplicate suppression (the pre-presentation activation is
+    /// ignored, then the dedupe never forwards the real transition) — so no
+    /// state is forwarded until this flips.
+    private(set) var isTemplatePresented = false
+    /// Latest desired CarPlay state while the template is not yet presented.
+    /// Retained (not activated); consumed exactly once on presentation.
+    private(set) var pendingPresentationState: CarPlayVoiceState?
+    /// The AppState this coordinator bound to on connect. Disconnect and the
+    /// controls converge on THIS instance (never a fresh provider resolve),
+    /// so a swapped provider cannot split surface bookkeeping across two
+    /// AppStates. Test-only observability; intentionally retained.
+    private(set) var lastBoundAppState: AppState?
+
+    var isConnected: Bool { interfacing != nil }
+
+    /// Test seam; production resolves the process-wide registry.
+    var appStateProvider: @MainActor () -> AppState = { AppStateRuntimeRegistry.shared.appState }
+    /// Production connects always spawn the establish task; tests disable
+    /// this and call `establishVoice(generation:)` directly so async Voice
+    /// establishment is deterministic.
+    var autoEstablishOnConnect = true
+    /// Test seam over template state activation (rate-limited by the system
+    /// template and a no-op before the template is presented).
+    var stateActivator: @MainActor (CPVoiceControlTemplate, CarPlayVoiceState) -> Void = {
+        $0.activateVoiceControlState(withIdentifier: $1.identifier)
+    }
+
+    private var stateObservation: AnyCancellable?
+
+    internal init() {}
+
+    // MARK: - Scene lifecycle
+
+    /// `CPTemplateApplicationSceneDelegate` connect. Installs the root
+    /// template synchronously (before returning from the scene callback),
+    /// binds the shared AppState, and starts async Voice establishment under
+    /// the current generation.
+    func handleConnect(_ interfacing: any CarPlayInterfacing) {
+        connectionGeneration &+= 1
+        let generation = connectionGeneration
+        // Defensive: never two live sinks, even for an unpaired re-connect.
+        stateObservation?.cancel()
+        stateObservation = nil
+        self.interfacing = interfacing
+        lastActivatedState = nil
+        isTemplatePresented = false
+        pendingPresentationState = nil
+
+        // The template install satisfies the scene time budget first; it
+        // needs no AppState. Registry resolution (which may construct the
+        // AppState and run its cold-launch bootstrap) happens right after.
+        installRootTemplate()
+
+        let appState = appStateProvider()
+        lastBoundAppState = appState
+        appState.setCarPlayVoiceSurfaceActive(true)
+        // Re-assert the Voice gate: the phone may be locked/backgrounded with
+        // a gate left false by an earlier CarPlay-only disconnect, and the
+        // driver's next Listen must be able to re-arm capture.
+        appState.handleCarPlayVoiceSurfaceActivated()
+        beginObservingController(appState.voiceConversationController)
+
+        if autoEstablishOnConnect {
+            Task { @MainActor [weak self] in
+                await self?.establishVoice(generation: generation)
+            }
+        }
+    }
+
+    /// `CPTemplateApplicationSceneDelegate` disconnect. Rotates the fence so
+    /// in-flight establishment becomes stale, unbinds presentation, and hands
+    /// the lifecycle decision to the BOUND AppState: an active phone Voice
+    /// surface keeps the conversation; a CarPlay-only conversation releases
+    /// the runtime exactly like the PR #161 background boundary.
+    func handleDisconnect() {
+        connectionGeneration &+= 1
+        interfacing = nil
+        template = nil
+        stateObservation?.cancel()
+        stateObservation = nil
+        lastActivatedState = nil
+        isTemplatePresented = false
+        pendingPresentationState = nil
+
+        let appState = lastBoundAppState ?? appStateProvider()
+        appState.setCarPlayVoiceSurfaceActive(false)
+        appState.handleCarPlayVoiceSurfaceRemoved()
+    }
+
+    // MARK: - Template
+
+    private func installRootTemplate() {
+        guard let interfacing else { return }
+        let generation = connectionGeneration
+        let handlers = CarPlayVoiceActionHandlers(
+            startListening: { [weak self] in self?.startListeningTurn() },
+            endConversation: { [weak self] in self?.endConversation() }
+        )
+        let template = CarPlayVoiceTemplateFactory.makeTemplate(handlers: handlers)
+        self.template = template
+        interfacing.setRootTemplate(template, animated: false) { [weak self] success, error in
+            // MainActor Task hop (never a trapping assumeIsolated): the
+            // completion is expected on the main queue, but a wrong-queue
+            // delivery must degrade to a hop, not crash the process in a car.
+            Task { @MainActor [weak self] in
+                guard let self else { return }
+                // A completion from a superseded connection must never mark
+                // the new connection's template presented.
+                guard self.isCurrent(generation), self.interfacing === interfacing else {
+                    carPlayLogger.notice("stale root-template install completion ignored")
+                    return
+                }
+                guard success, error == nil else {
+                    carPlayLogger.error("root template install failed: \(String(describing: error), privacy: .public)")
+                    return
+                }
+                self.isTemplatePresented = true
+                // The template presents its FIRST state (ready) by default;
+                // activate a retained pending state exactly once when it
+                // differs, then resume normal dedupe against the presented
+                // state. If the factory ever changes its default first state,
+                // keep this coupling in sync.
+                let pending = self.pendingPresentationState ?? .ready
+                self.pendingPresentationState = nil
+                self.lastActivatedState = pending
+                if pending != .ready {
+                    self.stateActivator(template, pending)
+                }
+            }
+        }
+    }
+
+    private func beginObservingController(_ controller: VoiceConversationController) {
+        stateObservation?.cancel()
+        stateObservation = controller.$state
+            .sink { [weak self] state in
+                self?.handleControllerState(state)
+            }
+    }
+
+    /// The single forwarding path from controller state to CarPlay. Internal
+    /// so the duplicate-suppression policy is deterministically testable.
+    func handleControllerState(_ state: VoiceConversationState) {
+        guard isConnected else { return }
+        guard let template else { return }
+        let target = CarPlayVoiceState.map(state)
+        guard isTemplatePresented else {
+            // Pre-presentation: activateVoiceControlState has no effect, so
+            // only RETAIN the latest desired state. Recording it as
+            // activated would suppress the real post-presentation activation
+            // and freeze the surface on the template's default state.
+            pendingPresentationState = target
+            return
+        }
+        guard let activated = CarPlayVoiceStateActivation.activationTarget(
+            lastActivated: lastActivatedState,
+            newState: target
+        ) else { return }
+        lastActivatedState = activated
+        stateActivator(template, activated)
+    }
+
+    // MARK: - Controls
+
+    /// Listen button (Ready/Error states). Reuses the shared prepare/attach
+    /// path — never forces Continuous Conversation on and never creates a
+    /// parallel session decision.
+    func startListeningTurn() {
+        let generation = connectionGeneration
+        Task { @MainActor [weak self] in
+            await self?.performStartListeningTurn(generation: generation)
+        }
+    }
+
+    func performStartListeningTurn(generation: UInt64) async {
+        guard isCurrent(generation), isConnected else { return }
+        let appState = lastBoundAppState ?? appStateProvider()
+        let controller = appState.voiceConversationController
+        var outcome = AppState.VoiceConversationPrepareOutcome.handled
+        if controller.hasLiveVoiceSession {
+            // Attach (re-arm the gateway if the runtime was suspended); a
+            // failed attach settles the surface into the error state instead
+            // of a silent no-op listen.
+            guard appState.attachToLiveVoiceConversation() else {
+                handleControllerState(.failed(""))
+                return
+            }
+        } else {
+            outcome = await appState.prepareVoiceConversation(
+                profile: nil,
+                startsFreshConversation: false
+            )
+        }
+        await completeListenTurn(generation: generation, outcome: outcome)
+    }
+
+    /// End button. Converges on the authoritative Close teardown — no
+    /// parallel CarPlay teardown exists.
+    func endConversation() {
+        (lastBoundAppState ?? appStateProvider()).closeVoiceConversation()
+    }
+
+    // MARK: - Voice establishment
+
+    /// Connect-time Voice establishment. A live conversation is attached
+    /// display-only (no restart, no new session, no listen start — the phone
+    /// may be mid-turn); otherwise the shared prepare path runs and, on
+    /// success, listening starts: the CarPlay launcher tap is the listen
+    /// intent, and Continuous Conversation governs only turn-to-turn
+    /// continuation.
+    func establishVoice(generation: UInt64) async {
+        guard isCurrent(generation), isConnected else { return }
+        let appState = lastBoundAppState ?? appStateProvider()
+        let controller = appState.voiceConversationController
+        if controller.hasLiveVoiceSession {
+            guard appState.attachToLiveVoiceConversation() else {
+                handleControllerState(.failed(""))
+                return
+            }
+            return
+        }
+        let outcome = await appState.prepareVoiceConversation(
+            profile: nil,
+            startsFreshConversation: false
+        )
+        await completeVoiceEstablishment(generation: generation, outcome: outcome)
+    }
+
+    /// Post-prepare continuation of connect-time establishment, split out so
+    /// the stale-completion fence is deterministically testable without
+    /// racing a real in-flight prepare.
+    func completeVoiceEstablishment(
+        generation: UInt64,
+        outcome: AppState.VoiceConversationPrepareOutcome
+    ) async {
+        guard fence(generation) else { return }
+        let appState = lastBoundAppState ?? appStateProvider()
+        let controller = appState.voiceConversationController
+        guard outcome == .handled, controller.hasLiveVoiceSession else {
+            handleControllerState(.failed(""))
+            return
+        }
+        await controller.startListening()
+    }
+
+    /// Post-prepare continuation of the Listen button, split out for the same
+    /// deterministic fence coverage.
+    func completeListenTurn(
+        generation: UInt64,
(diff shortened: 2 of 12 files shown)


kaishi00:
## Summary
Adds a CarPlay `CPVoiceControlTemplate` presentation over the existing `AppState`/`VoiceConversationController` runtime, introducing a process-wide `AppState` registry, a thin scene delegate, coordinated surface lifecycle gating, and stale-completion fencing. The architecture is sound and the weak-self/retain-cycle hygiene is generally good, but the `openVoiceConversation` split and the surface predicate introduce two behavioral regressions plus a CarPlay display-init race.

## Findings

### [severity: critical] `.handled` conflates "prepared" with "failed", so phone Voice now opens the sheet and auto-listens on failure
**File:** Conduit/Services/AppState.swift:13426-13445 (and enum at 13448-13456)
**Issue:** `prepareVoiceConversation` returns `.handled` for every failure path (profile mismatch, `turnState.isRunning`, `!canStartVoiceConversation`, session/gateway failure) — the doc even states `.handled` means "prepared, *or a user-visible failure was raised*". `openVoiceConversation` then treats `.handled` as success and unconditionally does `showSidebar = false; voiceSheetShouldAutoListen = true; showVoiceSheet = true`. Before this PR these failure branches `return true` **before** the presentation flags were set, so no sheet appeared. Now, e.g., opening Voice while a turn is running (the explicit "Stop the current response before starting voice" case) presents the sheet with auto-listen armed; `VoiceConversationSheet.task` consumes the gate and calls `controller.startListening()` (Conduit/Views/VoiceConversationSheet.swift:46-51), potentially starting capture while the assistant is still responding. This directly contradicts the PR claim "exactly as before".
**Suggestion:** Make the outcome distinguish real acquisition from failure (e.g. `case prepared` / `case deferred` / `case failed`), or return a second success flag. `openVoiceConversation` should only set `showVoiceSheet`/`voiceSheetShouldAutoListen` for `prepared`; failures should surface `errorMessage` and return `true` without presenting (old behavior). CarPlay can keep using the boolean-ish check.

### [severity: critical] Phone "active" is treated as a Voice presentation surface, leaving a CarPlay conversation live with a hot mic and no UI after disconnect
**File:** Conduit/Services/AppState.swift:872-874, 13552-13555, 13541-13543; Conduit/CarPlay/CarPlayVoiceCoordinator.swift:125-127, 273-282
**Issue:** `hasActiveVoiceSurface` is `isSceneActive || isCarPlayVoiceSurfaceActive`, and `handleCarPlayVoiceSurfaceRemoved`/`fence` use it to decide whether to release. But the phone *presents* Voice only when `showVoiceSheet` is true. In the supported CarPlay-only flow (`establishVoice` prepares + `startListening` with `showVoiceSheet == false`, see `testCarPlayConnectWithNoLiveVoicePreparesThroughTheSharedPathAndStartsListening`), if the user has the phone foregrounded with the Voice sheet closed and CarPlay disconnects, `handleCarPlayVoiceSurfaceRemoved()` early-returns, leaving the microphone live with no visible control. This is worse on a CarPlay-first launch: `isSceneActive` defaults to `true` (AppState.swift:1101) and no phone scene callback may ever fire, so a disconnect after a CarPlay-only conversation can orphan a live mic indefinitely. The same gap applies to the stale-completion `fence` at CarPlayVoiceCoordinator.swift:273-282.
**Suggestion:** Separate the foreground *gate* (`hasActiveVoiceSurface`, which legitimately must keep the controller startable while the phone is foregrounded) from the *release* decision. Have `handleCarPlayVoiceSurfaceRemoved`/`fence` keep the conversation only when a surface is actually presenting it, e.g. `showVoiceSheet || isCarPlayVoiceSurfaceActive`. Also consider not defaulting `isSceneActive` to `true` before any scene event.

### [severity: major] Initial CarPlay state activation races template presentation and is then deduped away
**File:** Conduit/CarPlay/CarPlayVoiceCoordinator.swift:94-103, 157-167; Conduit/CarPlay/CarPlayVoiceStateMapping.swift:53-62
**Issue:** `handleConnect` calls `installRootTemplate()` (async `setRootTemplate` completion) and then `beginObservingController`, whose `Combine` sink immediately emits the controller's current state. `handleControllerState` sets `lastActivatedState = activated` and calls `stateActivator` synchronously — before the template is presented. The code's own seam comment says activation is "a no-op before the template is presented", so this first activation is dropped while `lastActivatedState` is still recorded. For a mid-turn attach at connect (`.speaking` → `responding`) the template is left on its default state until the next *distinct* mapped state (e.g. back to `.ready`), so CarPlay can display "Ready"/Listen throughout a response. The tests inject a recorder `stateActivator`, so they never exercise the real no-op and cannot catch this.
**Suggestion:** Re-assert the current controller state from the `setRootTemplate` completion (reset `lastActivatedState = nil` and call `handleControllerState`) or seed `lastActivatedState` from `template.activeStateIdentifier` after presentation rather than recording an activation that never took effect.

### [severity: minor] `handleDisconnect` can construct an `AppState` as a side effect
**File:** Conduit/CarPlay/CarPlayVoiceCoordinator.swift:125
**Issue:** `let appState = lastBoundAppState ?? appStateProvider()` runs on every disconnect. If no `AppState` was ever bound, this triggers `AppStateRuntimeRegistry.shared.appState`, whose initializer performs the cold-launch bootstrap during a teardown callback.
**Suggestion:** Guard the no-bound case (`guard let appState = lastBoundAppState else { return }`) or make the provider a non-creating accessor for teardown paths.

### [severity: minor] CarPlay has no Listen/End controls below iOS 26.4
**File:** Conduit/CarPlay/CarPlayVoiceTemplateFactory.swift:38-40
**Issue:** Buttons are only attached under `if #available(iOS 26.4, *)`, while the deployment target is iOS 17.0. On iOS 17–26.3 the driver has no on-screen controls for the shared conversation; only state titles are shown.
**Suggestion:** Confirm this is an intentional platform limitation (and document it), or provide an alternative interaction (e.g. navigation-bar buttons) for pre-26.4.

## Verdict
REQUEST_CHANGES

The two critical regressions (unintended sheet/auto-listen on failure, and a live mic with no presentation surface after CarPlay disconnect) should block merge; the CarPlay initial-state race is a real user-facing defect. No retain-cycle, force-unwrap, or resource-cleanup issues were found in the new code.

<a href="https://opencode.ai/s/1bDjsyOu">New%20session%20-%202026-09-12T13%3A07%3A43.854Z</a>
[opencode session](https://opencode.ai/s/1bDjsyOu)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34695503968)

kaishi00:
## Summary
Adds a CarPlay Voice V1 surface (`CPTemplateApplicationScene` + `CPVoiceControlTemplate`) layered over the existing phone `AppState`/`VoiceConversationController` runtime via a process-wide `AppState` registry, a thin scene delegate/coordinator, presentation-gated state activation, and shared prepare/attach lifecycle logic.

## Findings

### [severity: critical] Failure paths return `.handled`, so the phone still presents the Voice sheet and arms auto-listen after a rejected open
**File:** Conduit/Services/AppState.swift:13488-13530, 13437-13454
**Issue:** `prepareVoiceConversation` returns `.handled` on every rejection branch — profile mismatch (13490), running-turn without fresh (13510), capability unavailable (13517), session-creation failure (13524), missing session/gateway (13529) — while also setting `errorMessage`. `openVoiceConversation` treats `.handled` as success (`case .handled: break`) and unconditionally executes `showSidebar = false; voiceSheetShouldAutoListen = true; showVoiceSheet = true` (13452-13454). Before this PR those branches `return true` *before* touching presentation flags, so no sheet appeared. Now the sheet is presented and `VoiceConversationSheet.task` (Conduit/Voice/VoiceConversationSheet.swift:46-51) consumes the gate and calls `controller.startListening()`. If a gateway is still attached (e.g. a prior session) this can open the microphone on a failed open; otherwise it presents a dead sheet. The doc comment at 13460-13464 even codifies this conflation.
**Suggestion:** Make the outcome distinguish success from rejection (e.g. `case prepared / deferred / failed(String)`, or a separate success flag) and have `openVoiceConversation` only set `showVoiceSheet`/`voiceSheetShouldAutoListen` for the prepared case; rejections should surface `errorMessage` and return `true` without presenting. CarPlay already guards on `hasLiveVoiceSession`, so it can keep its boolean-ish handling. Add a regression test asserting the sheet is not presented (and auto-listen not armed) when preparation is rejected.

### [severity: critical] CarPlay disconnect while the phone is foregrounded but the Voice sheet is closed leaves a live microphone with no UI
**File:** Conduit/Services/AppState.swift:872-874, 13578-13581; Conduit/CarPlay/CarPlayVoiceCoordinator.swift:321-330
**Issue:** The release decision reuses the *runtime gate* `hasActiveVoiceSurface = isSceneActive || isCarPlayVoiceSurfaceActive`. But `isSceneActive` only means the phone scene is foregrounded — not that the phone is *presenting* Voice (`showVoiceSheet`). In the supported CarPlay-only flow (`establishVoice` prepares + `startListening` with `showVoiceSheet == false`), if the phone is foregrounded when CarPlay disconnects, `handleCarPlayVoiceSurfaceRemoved()` early-returns (`guard !hasActiveVoiceSurface`), leaving the controller `.listening` with capture live and no visible control. `fence` (321-330) has the same hole. On a CarPlay-first launch this is worse: `isSceneActive` defaults to `true` (AppState.swift:1101) and no phone scene callback ever fires, so the mic can be orphaned indefinitely. Tests only cover the closed-sheet case after `handleScenePhase(.background)` (`testCarPlayOnlyDisconnectWithClosedPhoneSheetStopsTheRuntime`), so this foreground+closed-sheet path is untested.
**Suggestion:** Separate the gate from the release predicate. Keep `hasActiveVoiceSurface` for `setForegroundActive`, and use a "actually presenting" predicate — e.g. `showVoiceSheet || isCarPlayVoiceSurfaceActive` — in `handleCarPlayVoiceSurfaceRemoved` and `fence`. Note that simply adding `showVoiceSheet` to `hasActiveVoiceSurface` would break a fresh phone open (nothing re-asserts the gate between `showVoiceSheet = true` and `startListening`), so a distinct predicate is needed. Reconsider defaulting `isSceneActive = true` before any scene event.

### [severity: major] Duplicate-window guard dismisses the whole `WindowGroup`, including the primary window
**File:** Conduit/Views/RootView.swift:51
**Issue:** The duplicate branch calls `dismissWindow(id: "conduit-primary")`. The identifier names the shared `WindowGroup`, and per Apple's `DismissWindowAction` semantics `callAsFunction(id:)` dismisses *all* windows in that group — so the duplicate window's `onAppear` closes the primary window too, defeating the first-window claim guard. The claim-keeper unit tests don't exercise the dismissal.
**Suggestion:** Use `dismissWindow()` (no argument) to dismiss only the current duplicate window and leave the primary `WindowGroup` window open.

### [severity: major] No Listen/End controls on iOS 17–26.3 while the deployment target is iOS 17
**File:** Conduit/CarPlay/CarPlayVoiceTemplateFactory.swift:38-40
**Issue:** Action buttons are attached only under `if #available(iOS 26.4, *)`. On iOS 17–26.3 the template presents state titles but has no buttons, yet `CarPlayVoiceCoordinator` auto-starts listening on connect. A driver on those versions has no way to end the active session or start another turn from the car screen.
**Suggestion:** Either document this as an intentional platform limitation (and ensure a workable interaction, e.g. the vehicle voice button, is verified) or provide an alternative control path for pre-26.4 (e.g. a navigation-bar / `CPNowPlaying`-style control), and add a version-gated test/regression note.

### [severity: minor] `handleDisconnect` can construct an `AppState` as a side effect
**File:** Conduit/CarPlay/CarPlayVoiceCoordinator.swift:139 (also 231, 267, 291, 307, 323)
**Issue:** `let appState = lastBoundAppState ?? appStateProvider()` runs on teardown; when nothing was ever bound, the provider creates the registry `AppState` and runs its cold-launch bootstrap during the disconnect callback.
**Suggestion:** Guard the no-bound case (`guard let appState = lastBoundAppState else { return }`) or use a non-creating accessor for teardown/fence paths.

### [severity: nit] `completeVoiceEstablishment` and `completeListenTurn` are identical
**File:** Conduit/CarPlay/CarPlayVoiceCoordinator.swift:286-314
**Issue:** The two methods have byte-identical bodies; future fence/failure changes can diverge silently.
**Suggestion:** Keep both as test seams but delegate to one private implementation.

## Verdict
REQUEST_CHANGES

The two critical items (phone sheet/auto-listen on failed preparation, and a live microphone with no presentation surface after CarPlay disconnect) are user-facing/privacy regressions that should block merge; the root-level `WindowGroup` dismissal and the pre-26.4 control gap are real defects as well. No retain-cycle, force-unwrap, or resource-cleanup issues were found in the new code.

<a href="https://opencode.ai/s/CocdLD6c">New%20session%20-%202026-09-12T16%3A26%3A38.794Z</a>
[opencode session](https://opencode.ai/s/CocdLD6c)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34702209692)

kaishi00:
## Summary
This PR adds a CarPlay `CPVoiceControlTemplate` surface on top of the existing phone `AppState`/`VoiceConversationController` runtime via a process-wide `AppState` registry, a thin scene delegate/coordinator, presentation-gated state mapping, and shared prepare/attach lifecycle logic. The architecture is coherent and the weak-self/force-unwrap hygiene is good, but the new `hasActiveVoiceSurface` gate is overloaded and breaks non-conversation Voice features.

## Findings

### [severity: critical] The new Voice gate disables Voice provider tests and on-device permission requests
**File:** Conduit/Services/AppState.swift:875-886, 5297; Conduit/Voice/VoiceConversationController.swift:244, 473
**Issue:** `isForegroundActive` on the controller is now driven solely by `hasActiveVoiceSurface = (isSceneActive && showVoiceSheet) || isCarPlayVoiceSurfaceActive`. But `isForegroundActive` is not only a listening/audio gate: it also guards `requestOnDeviceTranscriptionPermissions()` and the provider-test entry points `runTranscriptionTest()`/`runSpeechTest()`. Those are invoked from the Settings sheet (`AuxiliaryViews.swift:519-532` via `AppState.runVoiceASRTest`/`runVoiceTTSTest` and `setVoiceTranscriptionMode(.appleOnDevice)`), where `showVoiceSheet` is false and CarPlay is normally absent. After any `.inactive`→`.active`/background return (`handleScenePhase(.active)` sets the gate from `hasActiveVoiceSurface`, which is false with the sheet closed), selecting on-device transcription fails with "Voice permissions can only be requested while Conduit is in the foreground", and both ASR/TTS tests return "Voice tests only run while Conduit is in the foreground". The test suite misses this because it sets `controller.setForegroundActive(true)` directly, bypassing the AppState scene path. This is a regression: previously `.active` set `isForegroundActive = true` unconditionally.
**Suggestion:** Separate "foreground liveness" from "a Voice presentation surface exists". Keep a plain foreground flag for provider tests/permission requests (e.g. `setForegroundActive(isSceneActive || isCarPlayVoiceSurfaceActive)`), and introduce a distinct `isVoicePresentationActive`/conversation gate (`(isSceneActive && showVoiceSheet) || isCarPlayVoiceSurfaceActive`) that only `startListening`/`resumeMicrophone` consult. Add a regression test that runs `handleScenePhase(.background)` → `.active` with the sheet closed and then asserts `runVoiceASRTest()`/on-device permission still succeed.

### [severity: major] CarPlay Voice has no Listen/End controls on iOS 17–26.3
**File:** Conduit/CarPlay/CarPlayVoiceTemplateFactory.swift:38-40
**Issue:** Action buttons are only attached under `if #available(iOS 26.4, *)`, while the deployment floor is iOS 17. On iOS 17–26.3 the driver gets state titles only, yet `CarPlayVoiceCoordinator.establishVoice` auto-starts a listening turn on connect, leaving no on-screen way to end the active session or start another turn. The PR acknowledges this but leaves the surface without usable controls on the vast majority of supported OS versions.
**Suggestion:** Either provide an alternative control path for pre-26.4 (e.g. a navigation-bar/`CPBarButton` or a separate action template) or fail closed/present a clear "controls unavailable" state on those versions, and document the limitation. Pin the chosen behavior with a version-gated test.

### [severity: minor] Teardown paths can construct the AppState as a side effect
**File:** Conduit/CarPlay/CarPlayVoiceCoordinator.swift:139, 254, 323
**Issue:** `lastBoundAppState ?? appStateProvider()` runs in `handleDisconnect`, `endConversation`, and `fence`. If nothing was ever bound (e.g. a disconnect racing a failed connect), this calls `AppStateRuntimeRegistry.shared.appState`, whose initializer performs the cold-launch bootstrap during a teardown callback.
**Suggestion:** Guard the no-bound case (`guard let appState = lastBoundAppState else { ... }`) or use a non-creating accessor on teardown paths.

### [severity: nit] `completeVoiceEstablishment` and `completeListenTurn` are byte-identical
**File:** Conduit/CarPlay/CarPlayVoiceCoordinator.swift:286-314
**Issue:** The two methods have identical bodies; future fence/failure changes can silently diverge between the connect path and the Listen path.
**Suggestion:** Keep both as named test seams but delegate to one private implementation.

## Verdict
REQUEST_CHANGES

The `hasActiveVoiceSurface`→`isForegroundActive` conflation is a real regression that breaks Voice provider tests and on-device permission setup after any foreground return, and it is not covered by the current tests. The pre-26.4 control gap and the teardown-time AppState construction should also be addressed. No retain cycles, force unwraps, or resource-cleanup issues were found in the new CarPlay code.

<a href="https://opencode.ai/s/XkavgqP2">New%20session%20-%202026-09-12T16%3A57%3A40.349Z</a>
[opencode session](https://opencode.ai/s/XkavgqP2)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34706735295)

PR #166, opened 2026-09-13
feat(connection): multi-server dashboards — registry, scoped auth, switching, push ownership (#148)

### LINKED ISSUE #148: Multiple saved dashboards — switch between Hermes servers without re-login

kaishi00:
# Multi-server dashboards: registry, scoped auth, switching, push ownership (#148)

Closes #148. Companion relay PR: kaishi00/hermes-conduit-notifier (dashboard-identity binding — link in the comment below).

## Summary

Conduit can now save, switch between, sign into, sign out of, and remove **multiple Hermes dashboards**. Every dashboard is identified by a stable, app-generated **UUID** — never its URL — and every reusable authentication record (connection ticket, password credentials + Face ID flag, Cloudflare service token, durable cookie mirror) lives in a **dashboard-scoped Keychain record** whose account embeds that UUID, so dashboard A's state can never be read or cleared through dashboard B. The push/relay ownership gap left by PR #151 (blocker B3) is closed end to end with the companion relay change.

## Registry schema

```swift
SavedDashboardRegistry {          // Keychain: hermes-conduit.dashboard-registry.v1
  schemaVersion: Int              // 1
  activeDashboardID: UUID?        // the selected target — set at switch INTENT
  dashboards: [SavedDashboard]    // { id: UUID, label: String, normalizedURL: String }
}
```

Registry data names the user's servers, so it is stored in the **Keychain**, not UserDefaults. Labels are derived from the host (first DNS label, whole IP/IPv6 literals) with duplicate disambiguation; label editing is deliberately omitted (YAGNI). Dashboard adoption dedupes by normalized URL, so two saved entries can never resolve to one server.

## Legacy migration

Runs once per launch before connection hydration; idempotent, retryable, crash-safe:

1. Registry present → nothing to migrate; leftover legacy records are retired lazily.
2. No legacy state → commit an empty registry (clean installs never re-run it).
3. Legacy state → derive the dashboard URL (credentials.baseURL → connection.baseUrl → Cloudflare origin → remembered URL), generate one UUID, **write every dashboard-scoped record FIRST**, commit the **registry LAST**, and only then retire the legacy global records. A crash at any point leaves the old state working and the next launch retries.

The chat-resume server identity is relabeled (URL → UUID) during migration when it names the migrated server, so upgrading triggers **no runtime teardown**. Covered states: full password auth, Face ID-protected credentials, browser-auth (ticket + cookies, no password), Cloudflare, cookies, partial legacy, interrupted migration + retry, already migrated, clean install.

## Switching

**Switch Dashboard is not Disconnect.** `AppState.switchDashboard(to:)`:

1. Marks the target selected **before** connecting — a failed switch leaves it selected with its sign-in/repair surface; no silent fallback to the outgoing server.
2. Runs the authoritative `prepareChatResumeForConnection(to:dashboardID:)` boundary plus a transport-only teardown — the outgoing dashboard's client, sessions, speech ownership, and queued callbacks are retired **even when the switch then fails to authenticate**. Its saved auth, cookies, and registry entry stay intact.
3. Restores through the target's own scoped auth: saved credentials (with the Face ID gate) → ticket/cookie resume → otherwise the sign-in surface seeded with that dashboard's address.

The server-change identity is now the dashboard UUID, so two dashboards sharing an address spelling are still distinct servers, and **A → B → A is always a fresh A generation** — A1's suspended work cannot commit into A2 merely because URL or ids match. Same-dashboard reconnects (same UUID) are never treated as a switch. Switches carry a generation fence: a restore started under an older switch aborts inertly instead of installing its connection underneath a newer one. A rapid A→B→C sequence leaves the newest selection owning the flow.

## Sign out / remove

- **Sign Out of This Dashboard** — clears only that dashboard's reusable auth/session/cookie material (scoped records + live WebKit/Foundation cookies for its origin); metadata and every other dashboard untouched. The Gateway destructive action is now this scoped operation.
- **Remove Dashboard** — removes the entry and clears only its secure state. Removing the **active** dashboard leaves Conduit disconnected with the selection/add flow presented — no auto-connect of another dashboard.
- Installation-wide push registration is deliberately untouched by both.

## Server-scoped state

`prepareChatResumeForConnection` remains the single teardown state machine (no second one added). On identity change it already retires: ChatResume state, ConversationIdentityIndex, session catalogs, SessionPresentationCache, titles/pins, pending decision restoration, projects, streaming/reasoning state, and speech ownership (Voice Conversation, Read Aloud — via `retireSpeechOperationsForServerReplacement`). Kanban is server-scoped by its existing `configure(serverIdentity:)` generation fences. Switch retirement additionally rotates the client epoch, cancels a parked presentation-cache flush, and cancels scheduled reconnects.

## B3: push / relay dashboard ownership

ConduitNotificationTarget parses the relay-stamped `dashboard_id` (opaque UUID; a malformed value is its own fail-closed mode). Routing matrix:

| Push | Routing |
|---|---|
| `dashboard_id` == active dashboard | route normally |
| `dashboard_id` of another **known** saved dashboard | switch/connect it first; the decision card is recorded only after the owning dashboard is active |
| unknown or malformed `dashboard_id` | **fail closed** (never opened, never recorded) |
| no `dashboard_id` (legacy) | ≤1 saved dashboard: route (compatibility); >1: **fail closed** rather than guess |

Pairing creation binds the **active dashboard's UUID** to the relay pairing; the relay persists the binding on the gateway record at claim and stamps every outgoing push from the **authenticated gateway credential** — the event body is never a trust source. One installation still supports multiple paired gateways (one per dashboard). Relay-backed clarify answers can only be recorded under the owning dashboard's session state by construction.

## UI

Settings → Connection → **Saved Dashboards**: per-dashboard rows with live state (Connected / Connecting… / Disconnected / Saved / Sign In required), tap-to-Switch, context actions (Sign Out of This Dashboard, Remove Dashboard with destructive confirmation), and Add Dashboard (hands to the existing login surface — no second auth implementation). Gateway remains the active dashboard's status/config detail. Connection Setup, Repair Connection, and seeding all read/write dashboard-scoped records.

## Localization

Every new user-facing string ships in `Localizable.xcstrings` with a real zh-Hans translation (20 keys): `check-l10n-coverage.py` — **1711 call sites, all resolve; every key zh-Hans-translated with type-matched placeholders**. Dashboard UUIDs, URLs, and wire values are never translated; App Language changes touch no dashboard identity or protocol state.

## Verification

- `python3 scripts/plan-tests.py validate --repo-root . --baseline scripts/test-timings.json` — OK (dynamic lanes, 4.4% / 5.0% imbalance)
- `python3 -m unittest discover -s scripts/tests` — 160 tests OK
- `python3 scripts/check-l10n-coverage.py` — OK (above)
- Focused suites: DashboardSessionIsolationTests (9), SavedDashboardRegistryTests (21), AppStateMultiDashboardTests (16), NotificationDashboardOwnershipTests (11), plus NativeAuth (unit + integration), server-replacement speech, and decision-fence regression suites
- **Full ConduitTests on the Mac build host: 2440 tests, 0 failures** (final commit 3754cec)
- Multi-model review gate (2 reviewer reports + focused re-review) and post-push CodeRabbit findings: 1 BLOCKER + accepted WARNINGs/Minors fixed across three hardening commits (dashboard-owned WebKit/native session isolation, stale-flow fences, pairing gating); see the review-response comment for the fixed/rejected breakdown
- `git diff --check` clean

## Physical-device / TestFlight checklist

- Face ID prompt when switching to a Face ID-protected dashboard
- Browser-auth (interactive Cloudflare/OAuth) sign-in for a second dashboard; cold-launch cookie restore for both dashboards
- Push from server A while connected to B: banner tap switches to A, then opens the decision; unscoped legacy pushes fail closed with ≥2 dashboards
- Live `hermes conduit-push pair` flow against a relay with the companion binding change; `/v1/meta` dashboard_id rendering
- Migration of a real 0.1.x install (single dashboard → registry) without re-login

## Summary by CodeRabbit

- **New Features**
  - Manage multiple saved dashboards from Settings, including switching, signing in or out, and removing dashboards.
  - Dashboard connections, credentials, cookies, and Cloudflare access remain isolated.
  - Notifications open the correct saved dashboard and safely reject unknown or invalid links.
  - Pairing codes are now associated with a specific dashboard.
  - Existing dashboard settings migrate automatically.
- **Bug Fixes**
  - Improved connection switching and restoration to prevent stale results and cross-dashboard session mix-ups.
- **Localization**
  - Added Simplified Chinese translations for multi-dashboard features and connection guidance.


kaishi00:
Companion relay PR: kaishi00/hermes-conduit-notifier#16 — pairing dashboard-identity binding (B3 server side).

kaishi00:
## Post-push review round — fixed vs. rejected

All findings re-verified against current code before acting. Fixes landed in 19d7d81 + 3754cec; verified by 9 focused suites and the full ConduitTests run (2440 tests, 0 failures) at the pushed head.

**Fixed (Conduit):**
1. **Complete cookie/session isolation (the CodeRabbit security Major + review item):** the durable Keychain mirror was scoped, but the live WebKit/native stores were not. Every dashboard now owns its `WKWebsiteDataStore(forIdentifier:)` (iOS 17 deployment target) and its own `HTTPCookieStorage` jar for native password-login cookies (`commitCookies(dashboardID:)`). `restoreNativeCookies` imports only the owning dashboard's jar — never `HTTPCookieStorage.shared` — so sibling-host parent-domain cookies (`Domain=.example.com`) cannot cross dashboards in either direction. Sign-in WebViews run inside the owning dashboard's identified store (per-presentation identity for a brand-new dashboard, resolved at sign-in intent, never at render time). Regression coverage drives the LIVE stores: sibling-host WebKit isolation, cross-jar import refusal, and sign-out clearing the selected-but-disconnected dashboard's owned session.
2. **Stale restore failure paths fenced:** biometric-failure and network-failure branches now write nothing when a newer dashboard switch owns the flow (`showLogin`/`pendingLoginFailure`/`lastConnectionFailure` stay untouched). Deterministic test on the fenced presenter, both stale and current generations.
3. **Ticket renewal fenced:** `reconnectForRetry` captures the owning dashboard UUID + switch generation before minting; after the mint or silent re-auth await, a superseded renewal is rejected before installing, and the renewed connection saves under the CAPTURED UUID. Regression test parks the mint mid-flight across an A→B switch and proves B is untouched and A's stale ticket installs nothing.
4. **Sign-out cookie clearing without a live connection:** the dashboard identity is captured before teardown and the saved `normalizedURL` drives the web/native cleanup when `connection == nil` (test included).
5. **Pairing always dashboard-scoped:** `createPairingCode(dashboardID:)\) now requires a UUID; the Settings action is disabled with an explanatory line when no dashboard is active. Conduit never creates new unscoped pairings (existing ones remain routable via the legacy compatibility policy).
6. **Pluralization:** `%lld dashboards` gained an `en` plural rule (one/other) with zh-Hans preserved under `other`.
7. **Force unwrap removed:** new `registerDashboard(forNormalizedURL:)` non-optional registration API; `adoptDashboard` uses it — no duplicated fallback logic.
8. **Test cleanups (CodeRabbit):** `InMemoryKeychainBackend` imports Security; suite defaults pass through every `loadRegistry` call; `testMigrationRetriesAfterInterruption` now seeds CONFLICTING scoped + legacy state and proves the legacy ticket wins until the registry commits.

**Rejected / adjusted with rationale:**
- `WKWebsiteDataStore.remove(forIdentifier:)` for sign-out cleanup: segfaults inside WebKit's run-loop dispatch on current simulator hosts (crash report: `removeDataStoreWithIdentifierImpl`). Replaced with a full wipe of the identified store's cookie store — same scope (the store is exclusively the dashboard's), stable primitive, covered by the same tests.
- CodeRabbit's generic docstring-coverage suggestions: declined per repo documentation style (constraint-bearing comments, not coverage metrics).

**Companion notifier PR** (kaishi00/hermes-conduit-notifier#16, head 57a277b): literal-null JSON bodies no longer 500; `RelayStore` re-validates `dashboard_id` at the persistence boundary (canonical form, nil-UUID rejection, legacy null/absent kept); new e2e via an `APNS_CAPTURE_PATH` seam proves a plugin-supplied `dashboard_id` can never affect the REAL APNs payload and an explicit-null pairing stays unbound on the wire. Relay suite 47/0; plugin Python suite 58/0.

kaishi00:
## Summary
PR #166 adds a UUID-keyed `SavedDashboardRegistry` with dashboard-scoped Keychain records and one-shot legacy migration, multi-dashboard switching/sign-out/removal in `AppState`, dashboard-owned WebKit/native cookie stores, and dashboard-ownership gating for push/relay notifications.

## Findings

### [severity: major] A cross-dashboard push switch cancels the very task that is performing it
**File:** `Conduit/ConduitApp.swift:115-122` (interacting with `Conduit/Services/AppState.swift:3047` and `:3147`)
**Issue:** The notification route runs inside `.task(id: notificationRouteKey)`, and `notificationRouteKey` embeds `appState.isConnected`. For a push that belongs to another saved dashboard, `openNotificationTarget` calls `await switchDashboard(to:)`, which synchronously calls `retireConnectionRuntimeForDashboardSwitch()` and sets `isConnected = false` (AppState.swift:3147) *before* the first await (biometric/network restore). That flips `notificationRouteKey`, so SwiftUI cancels the in-flight `.task` right when it suspends. The credential-restore path honors cancellation (`NativeAuthClient.perform`'s cancellation handler throws `CancellationError`), and the ticket path is fenced by `transportContinuation`'s `Task.isCancelled` check, so the target never connects; the replacement task then bails on `guard appState.isConnected`. Net effect: the headline "banner tap switches to A, then opens the decision" path is racy/usually dropped, and `pendingTarget` is left unresolved.
**Suggestion:** Decouple routing from connection state. Either drop `isConnected` from `notificationRouteKey` and drive the switch in an unstructured `Task` that is not cancelled by the modifier, or re-trigger the route after the switch completes via a dedicated generation that the switch itself does not mutate. Ensure a superseded/cancelled route either completes the switch or leaves `pendingTarget` in a state that will be retried once `isConnected` becomes true.

### [severity: major] `nativeCookieStorages` is mutated without its lock
**File:** `Conduit/Services/DashboardTicketBridge.swift:100` (declaration/accessor at `:108-132`)
**Issue:** `nativeCookieStorage(for:)` is `nonisolated` and guards `nativeCookieStorages` with `nativeCookieStorageLock`, precisely so it can be called off the main actor. But `clearWebKitSession` (MainActor) does `nativeCookieStorages.removeValue(forKey: dashboardID)` without taking that lock. Any concurrent `nativeCookieStorage(for:)` from a non-main context (the API is advertised as thread-safe for that) races on the dictionary and can crash or corrupt state. `nonisolated(unsafe)` means the compiler will not catch it.
**Suggestion:** Remove through the same lock (or a locked `removeNativeCookieStorage(for:)` helper), e.g. `nativeCookieStorageLock.lock(); defer { unlock }; nativeCookieStorages.removeValue(...)`.

### [severity: minor] `.disabled(row.isActive)` on the row may suppress the context menu
**File:** `Conduit/Views/SavedDashboardsView.swift:124`
**Issue:** The whole row `Button` is disabled for the active dashboard, and the `contextMenu` (Sign Out of This Dashboard / Remove Dashboard) is attached to that same button. In SwiftUI, `.disabled(true)` disables the view's interactions, including context-menu presentation, so the active dashboard's Sign Out/Remove may be unreachable — and Remove is only exposed through this menu (`removeDashboard` has no other call site).
**Suggestion:** Don't disable the interactive container. Gate only the tap action (e.g., ignore taps when active) and keep the `contextMenu` enabled, or move the actions to a visible trailing menu button.

### [severity: minor] Migration tests still read/write `UserDefaults.standard`
**File:** `ConduitTests/SavedDashboardRegistryTests.swift:254,258,311`
**Issue:** `SavedDashboardMigrator.loadRegistry()` defaults to `.standard`, so these tests read `conduit.dashboardURL` / the chat-resume identity from and (inside `execute`) write the relabeled identity into the shared standard domain, which the suite's teardown (which removes only the per-test suite) never cleans. This can leak into other suites and contradicts the PR's claim that "suite defaults pass through every `loadRegistry` call."
**Suggestion:** Pass `defaults: defaults` to the `loadRegistry()` calls at lines 254, 258, and 311.

### [severity: minor] Per-presentation WebKit data store for newly added dashboards is never cleaned up
**File:** `Conduit/Views/LoginView.swift:547`
**Issue:** Adding a dashboard uses a throwaway identifier (`resolveDashboardID(..., registerIfMissing: false) ?? UUID()`), while the durable cookie mirror is captured under the dashboard UUID that `dashboardIDProvider` registers later. The identified `WKWebsiteDataStore` created for that sign-in is left behind (and `remove(forIdentifier:)` is deliberately avoided), so each add-dashboard sign-in leaks a persistent WebKit store.
**Suggestion:** Once the real dashboard UUID is resolved at capture time, clear the temporary store's contents (full cookie wipe as done in `clearWebKitSession`), or reuse the newly registered dashboard's identifier for the WebView when the address is new.

### [severity: minor] Non-string `dashboard_id` is treated as an absent legacy identity
**File:** `Conduit/Services/PushNotificationService.swift:754`
**Issue:** `payload["dashboard_id"] as? String` returns nil for a non-string JSON value (e.g. a number, object, or a stray `null`), which sets `hasMalformedDashboardID = false` and falls through to the legacy route. With ≤1 saved dashboard that routes rather than failing closed, so a malformed wire value can bypass the "never guess" intent that the code explicitly protects for malformed *strings*.
**Suggestion:** Distinguish "key absent" from "key present but not a valid string/UUID" (check `payload.keys.contains("dashboard_id")`) and mark any present-but-unusable value as malformed so it fails closed.

## Verdict
REQUEST_CHANGES

<a href="https://opencode.ai/s/0GodRNyv">New%20session%20-%202026-09-13T11%3A43%3A33.705Z</a>
[opencode session](https://opencode.ai/s/0GodRNyv)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34755120200)

kaishi00:
## Summary
This PR introduces a UUID-keyed multi-dashboard registry with dashboard-scoped Keychain/keychain-backed auth, dashboard-owned WebKit/native cookie stores, switching/sign-out/remove flows, and dashboard-ownership gating for push notifications. The security scoping and migration ordering are thoughtfully designed, but two scheduling/interaction bugs undermine the switch guarantees, and several smaller items remain.

## Findings

### [severity: major] A cross-dashboard push switch cancels the very task performing it
**File:** `Conduit/ConduitApp.swift:115-122`, `Conduit/ConduitApp.swift:131-132`, `Conduit/Services/AppState.swift:3147`
**Issue:** `openNotificationTarget` runs inside `.task(id: notificationRouteKey)`, and `notificationRouteKey` embeds `appState.isConnected`. For a push owned by another saved dashboard, `openNotificationTarget` calls `await switchDashboard(to:)`, which synchronously calls `retireConnectionRuntimeForDashboardSwitch()` and sets `isConnected = false` **before the first suspension**. That flips the task id, so SwiftUI cancels the in-flight task at its first `await`. The credential/ticket restores observe cancellation (`NativeAuthClient`/URLSession throw `CancellationError`), so the switch aborts; the replacement task then bails on `guard appState.isConnected`. The banner-tap "switch to A, then open the decision" path is dropped and `pendingTarget` is left unresolved (retry loop just re-checks `isConnected == false`).
**Suggestion:** Decouple routing from connection state: remove `isConnected` from `notificationRouteKey` (or drive the switch in an unstructured `Task` not owned by the modifier), and re-trigger the route via a generation the switch itself does not mutate. Ensure a superseded/cancelled route leaves `pendingTarget` retryable once connected.

### [severity: major] Cold-start restores are not fenced against a concurrent dashboard switch
**File:** `Conduit/Services/AppState.swift:2526-2536`, `Conduit/Services/AppState.swift:2581`
**Issue:** The switch-initiated restores carry `switchGeneration` fences, but the cold-start paths call `restoreSavedCredentials(credentials, dashboardID: activeID)` and `restoreSavedConnection(saved)` with no generation. If the user switches dashboards while a cold-start restore is awaiting biometric/network, the stale restore resumes, commits to the old dashboard, and calls `connect(with:)`, which unconditionally runs `adoptDashboard(forNormalizedURL:)` (AppState.swift:2581) and flips `activeDashboardID` back to the outgoing dashboard, installing its client/connection. This contradicts the stated invariant that the newest selection owns the flow.
**Suggestion:** Capture `dashboardSwitchGeneration` (and the target UUID) before the cold-start await and reject the restore after the await if either changed; have `connect` refuse to adopt/install when a newer generation exists. Alternatively route all restores through a single fenced entry point.

### [severity: minor] `.disabled(row.isActive)` may suppress the row's context menu
**File:** `Conduit/Views/SavedDashboardsView.swift:124`
**Issue:** The active row's `Button` is disabled and the `contextMenu` (Sign Out of This Dashboard / Remove Dashboard) is attached to that same button. `.disabled(true)` sets the `isEnabled` environment and can prevent context-menu presentation. `removeDashboard` has no other call site (grep confirms only `SavedDashboardsView.swift:71,212`), so the active dashboard's Remove may be unreachable.
**Suggestion:** Keep the container interactive and gate only the tap: ignore taps when `row.isActive` instead of disabling the view, or move the actions to a visible trailing `Menu`.

### [severity: minor] Non-string `dashboard_id` is treated as an absent legacy identity
**File:** `Conduit/Services/PushNotificationService.swift:754-769`
**Issue:** `(payload["dashboard_id"] as? String)` returns nil for a present-but-non-string JSON value (number, object, stray `null`), setting `hasMalformedDashboardID = false`. With ≤1 saved dashboard this falls through to the legacy `.route` path, bypassing the explicit fail-closed intent for malformed identities.
**Suggestion:** Detect key presence separately (e.g. `payload.keys.contains("dashboard_id")`) and mark any present-but-non-String/unusable value as malformed so it always fails closed.

### [severity: minor] Newly added dashboards leak a per-presentation WebKit store
**File:** `Conduit/Views/LoginView.swift:547`
**Issue:** Adding a dashboard uses a throwaway identifier (`resolveDashboardID(..., registerIfMissing: false) ?? UUID()`) for the sign-in WebView, while the durable mirror is later captured under the dashboard UUID registered by `dashboardIDProvider`. The identified `WKWebsiteDataStore` for the temporary UUID is persistent and never cleaned (and `remove(forIdentifier:)` is deliberately avoided), so each add-dashboard sign-in leaks a store.
**Suggestion:** After the real dashboard UUID is resolved at capture time, clear the temporary store's contents (the same cookie-wipe used by `clearWebKitSession`), or register/resolve the dashboard before constructing the WebView so the real UUID is used throughout.

### [severity: minor] Migration tests read/write `UserDefaults.standard`
**File:** `ConduitTests/SavedDashboardRegistryTests.swift:254`, `:258`, `:311`
**Issue:** `SavedDashboardMigrator.loadRegistry()` defaults to `.standard`; these calls read host state and `execute` writes the relabeled chat-resume identity into the shared standard domain, which teardown (per-suite only) never cleans. This can leak across suites and contradicts the claim that injected defaults pass through every `loadRegistry` call.
**Suggestion:** Pass `defaults: defaults` at all three call sites.

### [severity: nit] Remaining implicitly unwrapped optionals in the new test fixture
**File:** `ConduitTests/DashboardSessionIsolationTests.swift:19-21`, `:30`
**Issue:** `defaultsSuite: String!`, `defaults: UserDefaults!`, `backend: InMemoryKeychainBackend!`, and `UserDefaults(suiteName: defaultsSuite)!` remain; the recent hardening commit converted cookie force unwraps but not these. A fixture-allocation failure crashes the runner instead of reporting an XCTest failure.
**Suggestion:** Make the properties non-optional and initialize via `XCTUnwrap` in a throwing `setUpWithError()`.

## Verdict
REQUEST_CHANGES

The multi-dashboard scoping, migration ordering, and cookie-isolation work look solid; no retain cycles, force unwraps, or MainActor violations were found in the production additions. Findings #1 and #2 should be fixed before merge because they defeat the switch guarantee the PR is built around.

<a href="https://opencode.ai/s/3gAqJvtx">New%20session%20-%202026-09-13T13%3A36%3A42.902Z</a>
[opencode session](https://opencode.ai/s/3gAqJvtx)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34757178093)

kaishi00:
## Summary
Adds a UUID-keyed multi-dashboard registry with dashboard-scoped Keychain records, a one-shot legacy migration, dashboard switching/sign-out/removal, per-dashboard WebKit/native cookie stores, and push/relay dashboard ownership gating.

## Findings

### [severity: major] Cross-dashboard push route cancels the task that performs the switch
**File:** `Conduit/ConduitApp.swift:115-122`, `Conduit/ConduitApp.swift:131-132`, `Conduit/Services/AppState.swift:3147`
**Issue:** `openNotificationTarget` runs inside `.task(id: notificationRouteKey)`, and `notificationRouteKey` embeds `appState.isConnected`. For a push owned by another dashboard, `openNotificationTarget` calls `await switchDashboard(to:)` (AppState.swift:7560), which synchronously calls `retireConnectionRuntimeForDashboardSwitch()` and sets `isConnected = false` (AppState.swift:3147) before its first suspension. That flips the task id, so SwiftUI cancels the in-flight task at the next `await`. The credential/biometric/network restores observe cancellation (and the ticket path is fenced by `transportContinuation`'s `Task.isCancelled` check), so the target never connects; the replacement task then bails on `guard appState.isConnected`. The headline "banner tap switches to A, then opens the decision" path is dropped and `pendingTarget` is left unresolved. The same `isConnected` gate also means a push for A cannot trigger a switch when Conduit is currently disconnected.
**Suggestion:** Decouple routing from connection state. Drop `isConnected` from `notificationRouteKey` and perform the switch in an unstructured `Task` not owned by the modifier, or re-trigger routing via a generation the switch itself does not mutate. Ensure a cancelled/superseded route leaves `pendingTarget` retryable, and allow the switch to start while disconnected.

### [severity: major] Cold-start restores are not fenced against a concurrent dashboard switch
**File:** `Conduit/Services/AppState.swift:2526-2536`, `Conduit/Services/AppState.swift:2581`
**Issue:** Switch-initiated restores carry a `switchGeneration` fence, but the cold-start paths call `restoreSavedCredentials(credentials, dashboardID: activeID)` and `restoreSavedConnection(saved)` with no generation. If the user (or a push) switches dashboards while a cold-start restore is suspended on biometrics or the network, the stale restore resumes, passes `restoreOwnsFlow(nil)` (which always returns true), and calls `connect(with:)`, which unconditionally runs `adoptDashboard(forNormalizedURL:)` (AppState.swift:2581), flipping `activeDashboardID` back to the outgoing dashboard and installing its client/connection. This defeats the stated "newest selection owns the flow" invariant.
**Suggestion:** Capture `dashboardSwitchGeneration` and the target UUID before the cold-start await and reject the restore after the await if either changed (reuse `restoreOwnsFlow`); additionally have `connect`/`adoptDashboard` refuse to adopt or install when a newer switch generation exists, or route all restores through a single fenced entry point.

### [severity: minor] `.disabled(row.isActive)` can suppress the active row's Sign Out / Remove actions
**File:** `Conduit/Views/SavedDashboardsView.swift:124-138`, `Conduit/Views/SavedDashboardsView.swift:210`
**Issue:** The row `Button` is disabled for the active dashboard, and the `contextMenu` holding "Sign Out of This Dashboard" / "Remove Dashboard" is attached to that same disabled button. `.disabled(true)` sets `isEnabled = false` for the subtree, which can prevent context-menu presentation/activation. `removeDashboard` has no other call site (only here and the confirmation-dialog handler), so the active dashboard's Remove may be unreachable without first switching away.
**Suggestion:** Keep the container interactive and gate only the tap action (`if !row.isActive { switchTo(row) }`), or move the actions to a visible trailing `Menu` button.

### [severity: minor] Present-but-non-string `dashboard_id` is treated as legacy/absent
**File:** `Conduit/Services/PushNotificationService.swift:821-836`
**Issue:** `payload["dashboard_id"] as? String` returns nil for a present-but-non-string JSON value (number, object, stray `null`), so `hasMalformedDashboardID` is set false and routing falls through to the legacy `.route` path when ≤1 dashboard is saved, bypassing the explicit fail-closed intent for malformed identities.
**Suggestion:** Detect key presence separately (`payload.keys.contains("dashboard_id")`) and mark any present-but-invalid value as malformed so it always fails closed.

### [severity: minor] New-dashboard sign-in leaks a per-presentation WebKit data store
**File:** `Conduit/Views/LoginView.swift:547`
**Issue:** Adding a dashboard uses a throwaway identifier (`resolveDashboardID(..., registerIfMissing: false) ?? UUID()`) for the sign-in WebView, while the durable mirror is captured later under the dashboard UUID registered by `dashboardIDProvider`. The identified `WKWebsiteDataStore(forIdentifier:)` created for the throwaway UUID is persistent and never cleaned (and `remove(forIdentifier:)` is deliberately avoided), so each add-dashboard sign-in leaves a stale store behind.
**Suggestion:** Register/resolve the dashboard before constructing the WebView (use `registerIfMissing: true` at the sign-in intent), or clear the temporary store's contents once the real UUID is resolved.

### [severity: minor] Migration tests still read/write `UserDefaults.standard`
**File:** `ConduitTests/SavedDashboardRegistryTests.swift:254`, `:258`, `:311`
**Issue:** `SavedDashboardMigrator.loadRegistry()` defaults to `.standard`, so these calls read `conduit.dashboardURL` / the chat-resume identity from, and (inside `execute`) write the relabeled identity into, the shared standard domain that the per-suite teardown never cleans. This can leak into other suites and contradicts the claim that injected defaults pass through every `loadRegistry` call.
**Suggestion:** Pass `defaults: defaults` at all three call sites.

### [severity: nit] Implicitly unwrapped optionals in new test fixtures
**File:** `ConduitTests/DashboardSessionIsolationTests.swift:19-21`, `:30`; `ConduitTests/SavedDashboardRegistryTests.swift:22-24`
**Issue:** `defaultsSuite: String!`, `defaults: UserDefaults!`, `backend: InMemoryKeychainBackend!`, and `UserDefaults(suiteName: defaultsSuite)!` remain; an allocation failure crashes the runner instead of reporting an XCTest failure.
**Suggestion:** Make the properties non-optional and initialize them via `XCTUnwrap` in a throwing `setUpWithError()`.

### [severity: nit] Relay gateway discriminator map grows unbounded
**File:** `Conduit/Services/PushNotificationService.swift:377`, `:747-761`
**Issue:** `relayGatewayIDsByRequestID` is written for every decision push and never pruned when a decision is answered, cleared, or expires, so it grows for the app's lifetime with no bound; it also keys only by request id, so two gateways reusing a request id would overwrite each other's discriminator.
**Suggestion:** Remove the entry when the target is cleared/answered, and/or bound the map's size.

### [severity: nit] Accessibility identifiers are identical for every dashboard row
**File:** `Conduit/Views/SavedDashboardsView.swift:125`
**Issue:** Every row shares `.accessibilityIdentifier("settings.dashboard.row")` and the `ProgressView` has no accessibility label, making UI tests and VoiceOver navigation unable to distinguish rows/state.
**Suggestion:** Include the dashboard label/UUID in the identifier and add an accessibility label to the progress indicator.

## Verdict
REQUEST_CHANGES

The scoping, migration ordering, and cookie-isolation work are sound, and the prior locking/fencing fixes are present. The two major findings should be resolved before merge, since both defeat the multi-dashboard switch guarantee the PR is built around.

<a href="https://opencode.ai/s/MjfiUrRX">New%20session%20-%202026-09-13T14%3A40%3A58.428Z</a>
[opencode session](https://opencode.ai/s/MjfiUrRX)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34763070440)

PR #167, opened 2026-09-13
Fix session, tool activity, and approval recovery after app reopen

mpenelange:
Fix session, tool activity, and approval recovery after app reopen

  After iOS suspends or terminates Conduit, reopening the app can lose the current conversation view, omit in-flight
  tool activity, or leave pending approval prompts unavailable.

  ## Changes

  - Preserve the selected conversation while the session catalog catches up.
  - Handle deleted saved sessions without treating transient failures as deletion.
  - Restore distinct tool calls and match lifecycle events by call identity.
  - Recover pending and queued approvals, with responses tied to the correct request.
  - Prevent delayed recovery responses from updating a different conversation or connection.

  ## Validation

  - Added and ran regression tests for session restoration, tool lifecycle handling, and approval recovery.
  - Built and exercised the recovery changes on a physical iPhone 17 Pro Max.
  - Checked approval behavior against public upstream Hermes source.

  Happy to split this further or adjust the approach to fit the project.

## Summary by CodeRabbit

- **New Features**
  - Added support for tracking multiple approval requests independently, including restored pending approvals.
  - Added clearer approval expiration handling, with expired requests visibly marked.
  - Improved tool activity tracking so concurrent and out-of-order tool calls remain correctly associated.

- **Bug Fixes**
  - Improved session resume and foreground recovery, preserving the visible conversation and composer state.
  - Improved recovery when sessions are temporarily unavailable during compression.
  - Improved restoration of pending tool activity and conversation presentation after reconnecting or resuming.


kaishi00:
## Summary
PR hardens app-reopen recovery: it preserves the selected conversation when the cold catalog lags, correlates tool lifecycle events by gateway call identity, restores queued approvals by `request_id`, distinguishes deleted sessions from transient resume failures, and fences delayed recovery responses by client/profile/session/generation.

## Findings

### [minor] Cold-catalog-lag resume bypasses viewport restoration and leaves stale restoration state
**File:** Conduit/Services/AppState.swift:3236 (and branch at 3305)
**Issue:** When `missingSavedSessionID` is non-nil, the code skips `selectChatResumeTarget(...)` and resumes the identity directly. `selectChatResumeTarget` is the only caller that invokes `chatResumeCoordinator.selectTarget`, which for `.automaticReturn` sets `pendingSessionKey`, freezes the viewport, clears the previous `chatResumeRestorationRequest`, and later lets `settleReconciliationAndPublish` publish the saved scroll snapshot. In this new path none of that happens, so a `.continueWhereLeftOff` cold launch resumes the correct conversation but never restores/republishes its saved scroll position, and any prior `chatResumeRestorationRequest` is not cleared.
**Suggestion:** After a successful direct resume in this branch, seed the coordinator's pending key/restoration explicitly (or route through a `selectChatResumeTarget`-equivalent target so the normal freeze/publish machinery runs). At minimum clear `chatResumeRestorationRequest` for `.automaticReturn`.

### [minor] Synchronous full-store persistence on every `tool.start` bypasses the debounce and runs on the main actor
**File:** Conduit/Services/SessionPresentationCache.swift:412 (`recordPendingToolStart`), called from Conduit/Services/AppState.swift:12675
**Issue:** `recordPendingToolStart` does `load()` (JSON-decode the entire multi-session store) and `persist()` (JSON-encode + `UserDefaults.set`) synchronously for every tool-start event. `schedulePresentationCacheFlush` was deliberately built to avoid serializing the whole store on every stream frame (see comment at AppState.swift:1245). With `maxSessions = 32` and `maxMessagesPerSession = 320`, this can be a multi-hundred-KB encode/decode on the main thread per tool call, and it defeats the coalescing it otherwise relies on.
**Suggestion:** Keep the synchronous write only as a small, bounded "pending tool" side-record (separate key), or coalesce the synchronous path (e.g., write only the incremental record and rely on the debounced flush for the full snapshot). If synchronous full-store writes are required, consider moving the encode off the main actor.

### [minor] New reasoning-stream tests write to the shared presentation cache (standard `UserDefaults`)
**File:** ConduitTests/AppStateReasoningStreamTests.swift:483 and :528 (helper at :22)
**Issue:** `makeAppState()` passes only an isolated `UserDefaults` for `AppState` but does not pass a matching `sessionPresentationCache`, so `AppState` defaults to `SessionPresentationCache.shared` (`UserDefaults.standard`). Previously the tool-start cache write went through the 2s-debounced flush and generally never landed during a test; now `recordPendingToolStart` writes synchronously, so these tests persist running-tool records for fixed ids like `default|stored-a` into global storage and never clean them up. This violates the documented rule that presentation-cache tests use isolated suites and risks cross-test contamination/flakiness.
**Suggestion:** Pass `sessionPresentationCache: SessionPresentationCache(defaults: defaults)` from `makeAppState()` (as `AppStateChatResumeTests.makeHarness` does) so tool-start records stay in the per-test suite.

### [nit] Legacy no-ID gateways can render a duplicate running tool card
**File:** Conduit/Services/SessionPresentationCache.swift:382–399 (`containsResolvedTool`) and merge at 279–304
**Issue:** For a cached running tool and a same-name completed gateway row where neither side carries a stable tool id, `containsResolvedTool` cannot correlate them (returns `false`) and `bestMatch` deliberately refuses to consume a running candidate, so the stale running card is re-appended alongside the completed row. This is intentional per the new tests, but on older gateways (no `tool_id`) it will visibly duplicate tools whose completion event was missed.
**Suggestion:** If acceptable, document this as a known legacy limitation; otherwise fall back to best-effort name+input matching when neither side has an id (guarded to a single trailing candidate) to suppress the duplicate.

## Verdict
APPROVE

The recovery logic is well-fenced (client identity, profile, reconciliation token, per-request decision keys, and `resolved == 0` semantics all check out against upstream `_approval_request_payload`/`resolve_gateway_approval`), no force unwraps or retain cycles were introduced, and the new tests cover the primary failure modes. The findings above are non-blocking.

<a href="https://opencode.ai/s/fXsSmNhG">New%20session%20-%202026-09-13T06%3A46%3A33.476Z</a>
[opencode session](https://opencode.ai/s/fXsSmNhG)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34743531160)

mpenelange:
ran into a real major bug. will fix and resubmit.

mpenelange:
Re-opening PR pending Hermes patch https://github.com/NousResearch/hermes-agent/pull/109841.

kaishi00:
## Summary
PR #167 hardens app-reopen recovery by preserving the selected conversation through cold-catalog lag, correlating tool lifecycle events by gateway call identity, restoring/refreshing queued approvals by `request_id`, and fencing delayed recovery responses by client/profile/session/generation.

## Findings

### [major] Incomplete automatic sync leaves `currentPurpose` sticky, defeating the new `.preserveCurrent` foreground recovery
**File:** Conduit/Services/AppState.swift:3109 (intent at :5942-5948; failure paths at :3455-3481, :4204-4233)
**Issue:** `performSyncSession` resolves the purpose via `beginChatResumeRecovery` → `recoverySequence.register`. `register` only ever upgrades to `.automaticReturn` and never downgrades (`ChatResumeRecoverySequence` at :355-360), so once an automatic-return sync fails, `currentPurpose` stays `.automaticReturn`. The failure paths (`reconcile` catch and `performSyncSession`'s gateway-session catch) call `settleReconciliation`/`scheduleReconnect` but never reset `recoverySequence`, and `complete()` only runs on success (:3651). A later foreground recovery passes `.preserveCurrent` (the whole point of `foregroundRecoveryPurpose`), but `register(.preserveCurrent)` returns the sticky `.automaticReturn`, so `ChatResumeSessionResolver` can still select the newest conversation instead of preserving the visible one — exactly the class of bug this PR targets. The new test passes only because it uses a fresh `recoverySequence`.
**Suggestion:** On the automatic-sync failure paths, demote/reset the sequence before scheduling the reconnect (e.g. `recoverySequence.preserveTransportAfterAutomaticIntentCancellation()` or `complete()` alongside `cancelScheduledReconnect()`), or allow an explicit `.preserveCurrent` request to override a stale automatic-return purpose in `register`.

### [minor] Authoritative approval replay removes a request-id-less card that is mid-submission
**File:** Conduit/Services/AppState.swift:12817-12833
**Issue:** `applyApprovalActivity(authoritative: true)` removes every request-id-less card whose status satisfies `isPendingDecision`, which includes `.submitting` (SessionPresentationCache.swift:26-30). `recordNotificationDecision` always creates cards with `requestId == nil` (:7606-7616). If the user taps approve, `respondToApproval` sets `.submitting` and awaits the RPC; a concurrent authoritative resume then deletes the card, so the response's message-id guard (:11539/:11554) silently discards the result, and the identified snapshot is appended as a fresh `.pending` card. Net effect: a lost submission result and a re-armed/duplicate card.
**Suggestion:** Exclude `.submitting` from the `removeAll` predicate and defer the identified snapshot upsert while a matching legacy submission is in flight (as suggested in the review thread), or reconcile the in-flight card instead of replacing it.

### [minor] Synchronous full-store persistence on every `tool.start` bypasses the debounce and runs on the main actor
**File:** Conduit/Services/SessionPresentationCache.swift:412-441 (call site Conduit/Services/AppState.swift:12675)
**Issue:** `recordPendingToolStart` performs `load()` (JSON-decode the entire multi-session store) and `persist()` (JSON-encode + `UserDefaults.set`) synchronously for every tool-start event on the `@MainActor`. This is exactly what `schedulePresentationCacheFlush` was built to avoid (AppState.swift:1245-1249). With `maxSessions = 32` and `maxMessagesPerSession = 320`, this is a potentially multi-hundred-KB encode/decode per tool call on the main thread, and it affects normal streaming, not just recovery.
**Suggestion:** Keep only a small bounded "pending tool" side record under a separate key, or coalesce this path and rely on the debounced full snapshot. If a synchronous full snapshot is required, move the encode off the main actor.

### [minor] New reasoning-stream tests persist running-tool records into the shared `UserDefaults.standard` cache
**File:** ConduitTests/AppStateReasoningStreamTests.swift:22-35
**Issue:** `makeAppState()` passes an isolated `UserDefaults` for `AppState` but no `sessionPresentationCache`, so `AppState` falls back to `SessionPresentationCache.shared` (`UserDefaults.standard`, AppState.swift:1408). The newly added tests drive `.toolStart` with fixed ids like `default|stored-a`, and because `recordPendingToolStart` now writes synchronously instead of via the 2s debounce, these records land in global storage and are never cleaned up. This violates the documented rule (AGENTS.md, Repository Notes) that presentation-cache tests use isolated suites, and risks cross-test contamination/flakiness.
**Suggestion:** Pass `sessionPresentationCache: SessionPresentationCache(defaults: defaults)` from `makeAppState()`, as `AppStateChatResumeTests.makeHarness` does.

### [minor] Cold-catalog-lag direct resume skips viewport restoration and leaves a stale restoration request
**File:** Conduit/Services/AppState.swift:3241-3250, 3305-3317
**Issue:** When `missingSavedSessionID` is non-nil, the code bypasses `selectChatResumeTarget`, which is the only caller of `chatResumeCoordinator.selectTarget` — the method that sets `pendingSessionKey`, freezes the viewport, and clears any prior `chatResumeRestorationRequest` for `.automaticReturn` (ChatResumeCoordinator.swift:91-111; AppState.swift:3572-3574). Consequently a `.continueWhereLeftOff` cold launch resumes the right conversation but never republishes its saved scroll snapshot (`publishChatResumeRestorationIfReady` sees no pending key), and a prior restoration request is not cleared.
**Suggestion:** After a successful direct resume in this branch, seed the coordinator's pending key/restoration for the resumed identity (or route through a `selectChatResumeTarget`-equivalent), and clear `chatResumeRestorationRequest` for `.automaticReturn`.

### [minor] Negative fence tests use `Task.yield()` and can assert before the refresh applied the fence
**File:** ConduitTests/AppStateDecisionFenceTests.swift:243-244, 263-264
**Issue:** `deliverResult` resumes the parked RPC continuation, but the unretained `@MainActor` refresh task applies its generation/client fence only after that await. Two `Task.yield()` calls do not prove completion, so these negative tests can assert before the stale result is fenced and a regression could pass.
**Suggestion:** Retain and `await` the refresh task (or await a deterministic completion signal from it) before asserting the stale-fence behavior.

### [nit] Legacy no-ID gateways can render a duplicate running tool card
**File:** Conduit/Services/SessionPresentationCache.swift:279-304, 382-406
**Issue:** For a cached running tool and a same-name completed gateway row where neither side carries a stable tool id, `containsResolvedTool` cannot correlate them and returns `false`, while `bestMatch` deliberately refuses to consume a running candidate. The stale running card is then re-appended next to the completed row. This is intentional per the new tests, but on older gateways (no `tool_id`) whose completion event was missed it visibly duplicates tools.
**Suggestion:** Document this as a known legacy limitation, or fall back to a guarded single-trailing-candidate name+input match to suppress the duplicate when neither side has an id.

## Verdict
REQUEST_CHANGES

The recovery fencing (client identity, profile, viewport generation, reconciliation token, request-id keys, and `resolved == 0` handling) is well constructed, and no force unwraps, implicitly unwrapped optionals, or retain cycles were introduced (the new approvals refresh Task uses `[weak self]`). The sticky `recoverySequence` purpose, however, can defeat the PR's central "preserve the visible conversation" guarantee, and the approval-removal race can drop an in-flight submission result; both should be addressed before merge.

<a href="https://opencode.ai/s/D8Z0311L">New%20session%20-%202026-09-13T15%3A58%3A44.229Z</a>
[opencode session](https://opencode.ai/s/D8Z0311L)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34767139135)

mpenelange:
Pushed fixes for the functional review findings, including recovery intent, scroll restoration, approval submission races, tool-start cache persistence, and test isolation/synchronization.

The separate Hermes WAL issue was fixed upstream in NousResearch/hermes-agent#109841, and reopening works really well on my patched server.

kaishi00:
## Summary
Hardens Conduit's app-reopen recovery by preserving the selected conversation through cold-catalog lag, correlating tool lifecycle events by gateway call identity, restoring/refreshing queued approvals by `request_id`, and fencing delayed recovery responses. Adds an `expired` approval state and a battery of regression tests.

## Findings

### [minor] Identified approval arriving during an in-flight legacy submission can be dropped permanently after a failed response
**File:** Conduit/Services/AppState.swift:12845-12860 (and Conduit/Services/AppState.swift:11578-11592)
**Issue:** When any request-id-less legacy card for the session is `.submitting`, `applyApprovalActivity` returns early for *every* activity that carries a `request_id`, including a newly queued identified approval. That is the intended guard against re-arming the in-flight card, but on the response failure path (`respondToApproval` catch) no `schedulePendingApprovalsRefresh` is issued (only the success path schedules one at :11575). So if the gateway queued a second approval while the first legacy submission was pending and the submission then fails (network/4002), the identified approval is never re-surfaced until the next resume/foreground cycle.
**Suggestion:** Also schedule `schedulePendingApprovalsRefresh(sessionId:using:)` in the `catch` branch (the refresh preserves non-`.pending` local statuses, so it won't re-arm the failed card), or defer only the conflicting snapshot rather than dropping all identified activities.

### [minor] Sticky `.automaticReturn` is reset only in `scheduleReconnectAfterFailedSync`; the raw transport-connect failure path still bypasses it
**File:** Conduit/Services/AppState.swift:2672 (see also Conduit/Services/AppState.swift:355, 5325-5335, 4481)
**Issue:** `ChatResumeRecoverySequence.register` remains upgrade-only (`AppState.swift:355`), so `foregroundRecoveryPurpose == .preserveCurrent` cannot override a stale `.automaticReturn`. The new helper `scheduleReconnectAfterFailedSync()` correctly calls `complete()` before re-arming, and the reconcile/sync failure paths use it. But the transport-connect failure catch still calls plain `scheduleReconnect(purpose: continuation.purpose)`, and `transportContinuation` can hand back `.preserveCurrent` with `handedOffAutomaticIntent == true` (:1883) while `recoverySequence.currentPurpose` is still `.automaticReturn`. `register(.preserveCurrent)` then returns `.automaticReturn` and queues an automatic-return retry — the exact "replace the visible conversation" behavior this PR targets.
**Suggestion:** Route the connect-failure catch (and the create-session catch at :4481, for consistency) through `scheduleReconnectAfterFailedSync()`, or let an explicit `.preserveCurrent` request demote a stale automatic-return purpose inside `register`.

### [nit] Legacy no-ID gateways can still render a duplicate running tool card
**File:** Conduit/Services/SessionPresentationCache.swift:386-406
**Issue:** For a cached running tool and a same-name completed gateway row where neither side carries a stable tool id, `containsResolvedTool` cannot correlate them and `bestMatch` deliberately refuses to consume a running candidate, so the stale running card is re-appended beside the completed row. The comment documents this trade-off (matching by name/input would be unsafe with overlapping calls), but on older gateways whose completion event was missed it visibly duplicates tools.
**Suggestion:** Keep as-is if acceptable, but document it as a known legacy limitation; a guarded single-trailing-candidate name match could suppress the duplicate without affecting ID-bearing gateways.

### [nit] Test-only force unwrap in the new client test helper
**File:** ConduitTests/HermesClientTests.swift:192
**Issue:** `try JSONSerialization.data(withJSONObject: ...)!` is force-unwrapped inside `capturedApprovalRespond`. It is test code and practically cannot throw, but the file otherwise uses `XCTUnwrap`.
**Suggestion:** Use `try XCTUnwrap(JSONSerialization.data(...))`.

## Verdict
COMMENT

The earlier blockers (in-flight legacy approval replay, sticky recovery purpose on the reconcile/sync failure paths, task-await synchronization in the fence tests, and isolated presentation-cache suites) are addressed, and the new tests cover the main failure modes. The two minor items above are narrow but worth a follow-up; no force unwraps, retain cycles, or MainActor violations were introduced in production code (the new refresh task uses `[weak self]` and is generation-fenced).

<a href="https://opencode.ai/s/mGC55JY6">New%20session%20-%202026-09-13T18%3A13%3A01.029Z</a>
[opencode session](https://opencode.ai/s/mGC55JY6)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34773879090)

kaishi00:
## Summary
PR #167 hardens Conduit's app-reopen recovery: it keeps the selected conversation through cold-catalog lag, correlates tool lifecycle events by gateway call identity, restores/refreshes queued approvals by `request_id`, fences delayed recovery by client/profile/session/generation, and adds an `expired` approval state with regression coverage.

## Findings

### [severity: major] `.preserveCurrent` reconnect still inherits a sticky `.automaticReturn` recovery purpose
**File:** Conduit/Services/AppState.swift:2775-2779 (root cause at 374-378; consumed at 3531)
**Issue:** `ChatResumeRecoverySequence.register` is upgrade-only: once `currentPurpose` becomes `.automaticReturn`, `register(.preserveCurrent)` returns `.automaticReturn` and never demotes. The latest fix only resets the sequence in `scheduleReconnectAfterFailedSync()` (5787). The connect-failure catch now branches on `continuation.purpose`, but when it is `.preserveCurrent` it still calls `scheduleReconnect(purpose: .preserveCurrent)` (2776), which goes through `planReconnect` → `register` and can queue an `.automaticReturn` retry. `performSyncSession` then does `let purpose = beginChatResumeRecovery(purpose:)` (3531) and treats that sticky purpose as authoritative, so the reconnect can select the newest catalog row instead of preserving the visible conversation — the exact failure this PR targets. This is reachable when a stale automatic attempt is still registered while a foreground/preserve-current sync or handoff runs (`transportContinuation` returns `.preserveCurrent` with `handedOffAutomaticIntent == true`, AppState.swift:1931-1932). The new regression test only exercises the `.automaticReturn` branch with a fresh sequence.
**Suggestion:** In the `.preserveCurrent` branch, call `recoverySequence.complete()` (or `preserveTransportAfterAutomaticIntentCancellation()`) before `scheduleReconnect(...)`, or make `register` honor an explicit `.preserveCurrent` request over a stale automatic-return purpose. For consistency, route the create-session failure at 4927 through `scheduleReconnectAfterFailedSync()` as well.

### [severity: minor] `resolvePendingTool`/`removePendingTools` still decode (and sometimes re-encode) the full presentation store on the main actor per event
**File:** Conduit/Services/SessionPresentationCache.swift:483 (and 502), 512
**Issue:** The tool-start path was correctly moved to a bounded side record (`recordPendingToolStart`, 419-448), but the completion path calls `load()` unconditionally at 483, which JSON-decodes the entire `conduit.sessionPresentation.v1` store (`maxSessions = 32`, `maxMessagesPerSession = 320`) on every `tool.complete`; it then `persist()`s the whole store when it removes a committed running tool. `removePendingTools` (512) does the same on an idle resume. This reintroduces the main-thread full-store JSON work the debounced `schedulePresentationCacheFlush` was built to avoid, just on the completion side.
**Suggestion:** Only touch the pending-tools side record in the common path, and let the debounced flush fold removal into the full snapshot; or do the full-store read/write off the main actor.

### [severity: nit] Test-only force unwraps in the new approval tests
**File:** ConduitTests/HermesClientTests.swift:718, ConduitTests/HermesClientTests.swift:780
**Issue:** `String(data: try JSONSerialization.data(withJSONObject: ...), encoding: .utf8)!` is force-unwrapped. It practically cannot fail, but the surrounding helper favors `XCTUnwrap`.
**Suggestion:** Wrap with `try XCTUnwrap(String(data: ...))` to keep the test failure path explicit.

### [severity: nit] Legacy no-ID gateways can render a duplicate running tool card
**File:** Conduit/Services/SessionPresentationCache.swift:386-407 (merge append at 283-304)
**Issue:** For a cached running tool and a same-name completed gateway row where neither side carries a stable id, `containsResolvedTool` returns false and `bestMatch` refuses to consume a running candidate, so the stale running card is appended beside the completed row. The comment documents the trade-off, but on older gateways whose completion event was missed this visibly duplicates tools.
**Suggestion:** Leave as-is if acceptable, but document it as a known legacy limitation (or add a guarded single-trailing-candidate name+input match when neither side has an id).

## Verdict
REQUEST_CHANGES

The approval-fencing, identity-gating, and tool-correlation work is thorough, and the prior rounds' blockers (in-flight legacy replay, test isolation, deterministic fence synchronization, viewport restoration) are addressed. The remaining sticky-purpose gap, however, can still let a reconnect replace the visible conversation and should be closed before merge.

<a href="https://opencode.ai/s/yIRNkswm">New%20session%20-%202026-09-13T19%3A05%3A39.259Z</a>
[opencode session](https://opencode.ai/s/yIRNkswm)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34776566428)

kaishi00:
## Summary
This PR hardens Conduit's app-reopen recovery: it preserves the selected conversation through cold-catalog lag, correlates tool lifecycle events by gateway call identity, restores/refreshes queued approvals by `request_id`, fences delayed recovery responses by client/profile/session/generation, and adds an `expired` approval state with regression coverage.

## Findings

### [severity: major] Sticky `.automaticReturn` purpose still defeats `.preserveCurrent` recovery on the connect-failure and create-session paths
**File:** Conduit/Services/AppState.swift:374-378 (`register`), :2775-2779 (connect catch), :3531 (`performSyncSession`), :4926-4928 (create catch)
**Issue:** `ChatResumeRecoverySequence.register` is upgrade-only: once `currentPurpose` is `.automaticReturn`, `register(.preserveCurrent)` returns `.automaticReturn` and never demotes. The fix only resets the sequence in `scheduleReconnectAfterFailedSync()` (5787). The transport-connect failure branch calls `scheduleReconnect(purpose: continuation.purpose)` when `continuation.purpose == .preserveCurrent`, and `scheduleReconnect` → `planReconnect` → `register` therefore queues an `.automaticReturn` retry. `performSyncSession` also calls `beginChatResumeRecovery` (3531), which will silently upgrade a `.preserveCurrent` request to the stale `.automaticReturn`. The state is reachable: `transportContinuation` returns `(.preserveCurrent, nil, true)` whenever the automatic token is stale (1931-1932), and `ChatResumeCoordinator.removeSessions` bumps `automaticCancellationEpoch` (ChatResumeCoordinator.swift:308) without resetting `recoverySequence`. The retry can then select the newest catalog row / direct-resume the saved id instead of preserving the visible conversation — the exact failure this PR targets. The create-session failure branch (4926-4928) has the same gap.
**Suggestion:** In the `.preserveCurrent` connect-failure branch call `recoverySequence.complete()` (or `preserveTransportAfterAutomaticIntentCancellation()`) before `scheduleReconnect`, route the create-session failure through `scheduleReconnectAfterFailedSync()`, and/or make `register` honor an explicit `.preserveCurrent` request over a stale automatic-return purpose.

### [severity: major] `AnyCodable.intValue` can trap on a malformed gateway `resolved` value
**File:** Conduit/Services/HermesClient.swift:126-129 (call site :1293)
**Issue:** `intValue` does `Int(n)` on a `Double`, which traps for finite out-of-range values (e.g. JSON `1e100`) and NaN. `respondToApproval` now feeds gateway-controlled `resolved` through it, so a malformed/oversized value can crash the app instead of surfacing an error. This matches the merge-risk note that malformed gateway data can terminate the app.
**Suggestion:** Validate before conversion, e.g. `guard let value = rawResolved.doubleValue, let resolved = Int(exactly: value) else { throw HermesError.invalidResponse }`, while keeping the `true` fallback when `resolved` is omitted. A non-numeric `resolved` currently also silently defaults to `true`; consider rejecting that too.

### [severity: minor] Identified queued approvals are dropped when the legacy card cannot be located
**File:** Conduit/Services/AppState.swift:4741-4758
**Issue:** In `schedulePendingApprovalsRefresh(replacingLegacyErrorMessageID:)`, the guard requires the exact legacy message by id with `requestId == nil` and `status == .error` before applying `identifiedApprovals`. A concurrent authoritative `applyApprovalActivity(authoritative: true)` can sweep/alter that legacy card while this async task is suspended (it removes any request-id-less pending-decision card, and `.error` is a pending decision per SessionPresentationCache.swift:26-30). The guard then fails and the function returns, discarding the entire fetched `identifiedApprovals` array — including a genuinely new unrelated approval that would otherwise be surfaced. `applyApprovalActivity` already upserts by request id, so the legacy lookup is not needed for safety.
**Suggestion:** Apply `identifiedApprovals` unconditionally (only `guard !identifiedApprovals.isEmpty`), or defer only the specific conflicting snapshot rather than dropping the whole batch.

### [severity: minor] `resolvePendingTool`/`removePendingTools` decode/persist the full store on the main actor per completion
**File:** Conduit/Services/SessionPresentationCache.swift:483, 502, 512
**Issue:** Tool starts were correctly moved to a bounded side record, but the completion path still calls `load()` (JSON-decode the entire `conduit.sessionPresentation.v1` store; `maxSessions = 32`, `maxMessagesPerSession = 320`) synchronously on every `tool.complete`, and `persist()`s the whole store when it removes a committed running record. `removePendingTools` does the same on an idle resume. This reintroduces the multi-hundred-KB main-thread encode/decode that `schedulePresentationCacheFlush` was built to avoid.
**Suggestion:** Only touch the pending-tools side record in the common path and let the debounced flush fold removal into the full snapshot, or perform the full-store read/write off the main actor.

### [severity: minor] Legacy id-less replayed `tool.start` appends a duplicate that completion cannot fully resolve
**File:** Conduit/Services/AppState.swift:13540-13550
**Issue:** The duplicate-start guard only runs `if let stableToolID`. For a gateway that sends no per-tool id, a replayed `tool.start` (e.g. reconciliation buffer replay after resume) appends a second running card with a different timestamp-based id. The id-less `toolComplete` handler matches only `messages.lastIndex(where: ... status == .running)` (13592-13595), so it completes one card and leaves the earlier duplicate stuck in `.running` forever.
**Suggestion:** Add a name+input dedup fallback when `stableToolID == nil`, guarded against overlapping same-name calls (e.g. match only an existing running card with identical name and input).

### [severity: nit] Legacy no-ID gateways can render a duplicate running tool on restore
**File:** Conduit/Services/SessionPresentationCache.swift:386-407 (`containsResolvedTool`), merge append at :283-307
**Issue:** For a cached running tool and a same-name completed gateway row where neither side carries a stable id, `containsResolvedTool` returns `false` and `bestMatch` deliberately refuses to consume a running candidate, so the stale running card is appended beside the completed row. The comment documents the trade-off, but on older gateways the user sees a duplicate/stuck tool.
**Suggestion:** If acceptable, document this as a known legacy limitation; otherwise add a guarded single-trailing-candidate name+input match when neither side has an id.

### [severity: nit] Test-only force unwraps in the new approval tests
**File:** ConduitTests/HermesClientTests.swift:718, :780
**Issue:** `String(data: try JSONSerialization.data(withJSONObject: ...), encoding: .utf8)!` is force-unwrapped in the new helpers; surrounding helpers prefer `XCTUnwrap`.
**Suggestion:** Wrap with `try XCTUnwrap(String(data: ...))`.

## Verdict
REQUEST_CHANGES

The recovery fencing (client identity, profile, viewport generation, reconciliation token, request-id keys, `resolved == 0` handling) is well constructed, and no production force unwraps or retain cycles were introduced (the new approvals refresh Task uses `[weak self]` and is generation-fenced). However, the sticky recovery purpose can still let a reconnect replace the visible conversation, the `resolved` conversion is a remote crash vector, and the identified-approval guard can silently drop a queued request; these should be addressed before merge.

<a href="https://opencode.ai/s/e2LNgxmE">New%20session%20-%202026-09-13T22%3A24%3A23.403Z</a>
[opencode session](https://opencode.ai/s/e2LNgxmE)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34786636448)

kaishi00:
## Summary
This PR hardens Conduit's app-reopen recovery: it preserves the selected conversation through cold-catalog lag, correlates tool lifecycle events by gateway call identity, restores/refreshes queued approvals by `request_id`, and fences delayed recovery responses by client/profile/session/generation.

## Findings

### [severity: major] Malformed gateway `resolved` value can crash the app
**File:** Conduit/Services/HermesClient.swift:1310 (root at :126-128)
**Issue:** `respondToApproval` now feeds the gateway-controlled `resolved` field through `AnyCodable.intValue`, which does `Int(n)` on a `Double`. `Int(_:)` from a `Double` traps for NaN or finite out-of-range values (e.g. JSON `1e100`), so a malformed/oversized gateway response terminates the app at Conduit/Services/HermesClient.swift:127 instead of surfacing `HermesError`. The file already documents the correct pattern for exactly this case at Conduit/Services/HermesClient.swift:2134 (`doubleValue` + `Int(exactly:)`). This same crash surface is now on the approval path, which is reachable from user taps.
**Suggestion:** Validate before converting, preserving the omitted-field fallback:
```swift
guard let rawResolved = result.objectValue?["resolved"] else { return true }
guard let value = rawResolved.doubleValue, let resolved = Int(exactly: value) else {
    throw HermesError.invalidResponse
}
return resolved > 0
```

### [severity: major] Create-session failure leaves `.automaticReturn` purpose sticky, defeating `.preserveCurrent` recovery
**File:** Conduit/Services/AppState.swift:4960 (root cause :374-378)
**Issue:** `ChatResumeRecoverySequence.register` is upgrade-only, and the failed `createAndReconcileSession` path still calls `scheduleReconnect(purpose: resumePurpose)` directly instead of `scheduleReconnectAfterFailedSync()`. When `resumePurpose == .automaticReturn`, `register` sets `currentPurpose = .automaticReturn` and it is never demoted. A later visible-conversation recovery requests `.preserveCurrent` (via `foregroundRecoveryPurpose`), but `beginChatResumeRecovery`/`register` returns the sticky `.automaticReturn`, so `ChatResumeSessionResolver.target` can select the newest catalog row instead of preserving the on-screen conversation — the exact regression this PR targets. The connect-failure path (AppState.swift:2807) has the same latent upgrade because `scheduleReconnect(purpose: .preserveCurrent)` runs through `planReconnect` → `register`.
**Suggestion:** Route the automatic-return branch at :4960 through `scheduleReconnectAfterFailedSync()` (which calls `recoverySequence.complete()` first), and/or make `register` allow an explicit `.preserveCurrent` request to demote a stale `.automaticReturn`. Also consider resetting in the `.preserveCurrent` connect-failure branch at :2807.

### [severity: minor] Identified queued approvals are dropped if the legacy error card is gone
**File:** Conduit/Services/AppState.swift:4774-4785
**Issue:** `schedulePendingApprovalsRefresh(replacingLegacyErrorMessageID:)` gates the entire fetched `identifiedApprovals` batch on the legacy card still existing at that id with `requestId == nil` and `status == .error`. A concurrent authoritative `applyApprovalActivity` sweeps request-id-less pending-decision cards (and `.error` counts as pending, SessionPresentationCache.swift:26-30), so the lookup can fail during the await. The function then returns, silently discarding unrelated newly queued identified approvals in the same response.
**Suggestion:** Apply `identifiedApprovals` whenever non-empty; `applyApprovalActivity` already upserts by request id, so the legacy-card lookup is not needed for safety. At minimum, defer only the conflicting card rather than dropping the whole batch.

### [severity: minor] Full presentation store is decoded on the main actor per tool completion
**File:** Conduit/Services/SessionPresentationCache.swift:483 (and :512)
**Issue:** `recordPendingToolStart` was correctly moved to a bounded side record, but `resolvePendingTool` still calls `load()` unconditionally on every `tool.complete` (id-less matches `continue` before using it but the decode already ran), and `removePendingTools` does the same on idle resume. With `maxSessions = 32` / `maxMessagesPerSession = 320`, this reintroduces the multi-hundred-KB JSON decode the debounced `schedulePresentationCacheFlush` was built to avoid, on the main thread, on every tool completion.
**Suggestion:** Only touch the pending-tools side record in the common path and let the debounced flush fold removal into the full snapshot, or move the full-store read/write off the main actor.

### [severity: nit] Legacy no-ID gateways can show a duplicate/stuck running tool
**File:** Conduit/Services/AppState.swift:13988 (see also SessionPresentationCache.swift:386-407)
**Issue:** The replayed-`tool.start` dedup guard only runs `if let stableToolID`. For gateways without per-tool ids, a replayed start appends a second running card, and the id-less `toolComplete` handler completes only the newest matching card, leaving the earlier one stuck `.running`; the cache merge exhibits the same duplicate on restore. The new tests document this as intentional, but it is a visible degradation on older gateways.
**Suggestion:** Add a guarded name+input fallback for `stableToolID == nil` (matching a single trailing running candidate), or explicitly document it as a known legacy limitation.

### [severity: nit] Test-only force unwraps in new approval helpers
**File:** ConduitTests/HermesClientTests.swift:718, :780
**Issue:** `String(data: try JSONSerialization.data(withJSONObject: ...), encoding: .utf8)!` is force-unwrapped. It cannot realistically fail, but the surrounding helpers and peers use `XCTUnwrap`.
**Suggestion:** Wrap with `try XCTUnwrap(String(data: ...))`.

## Verdict
REQUEST_CHANGES

The recovery fencing (client identity, profile, viewport generation, reconciliation token, request-id keys, `resolved == 0` handling) is well constructed, and no production force unwraps, retain cycles, or MainActor violations were introduced (the new approvals refresh `Task` uses `[weak self]` and is generation-fenced). The remote-crash vector in `intValue` and the sticky `.automaticReturn` purpose on the create-session failure path should be closed before merge.

<a href="https://opencode.ai/s/6P3zdhJh">New%20session%20-%202026-09-15T06%3A48%3A05.599Z</a>
[opencode session](https://opencode.ai/s/6P3zdhJh)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34938492205)

kaishi00:
Thank you so much for this excellent recovery series, @mpenelange!

We have taken over this work in a clean replacement PR: #190.

All 11 of your original commits have been preserved and cherry-picked directly onto current \main\ with full git author credit (\michael <michaeljlange@icloud.com>\). We've layered on a hardening commit resolving recovery purpose transitions during reconnect retries, approval replay handling against in-flight submissions, and integer parsing safety, all verified across 449 unit tests on iOS 18 simulator.

Closing this PR in favor of #190. Thank you again for your valuable contributions to Conduit!

PR #168, opened 2026-09-13
fix(chat): run /compress through the dedicated session.compress RPC

kaishi00:
## Summary

`/compress` (and `/compact`) no longer ride the generic `slash.exec` RPC. They now use the gateway's dedicated `session.compress` RPC with a 660-second budget, matching upstream Hermes Desktop (`SESSION_COMPRESS_TIMEOUT_MS = 660_000`, `apps/desktop/src/app/session/hooks/use-prompt-actions/slash.ts`).

**The bug:** manual compression is LLM-bound and routinely runs 45–70s+ on real sessions, but Conduit sent `/compress` through `slash.exec` with the generic 30s `requestTimeout`. The client reported `Request timed out: slash.exec` while the gateway kept compressing; when the real response finally arrived it was discarded as an unmatched RPC id, so the compressed transcript was never adopted. Worse, Conduit's error-then-fallback retried through `command.dispatch` — the exact cascade upstream fixed as #44456 ("not a quick/plugin/skill command: compress").

### `HermesClient`
- `sessionCompressTimeout = 660s` — deliberately above the gateway's compute-host compression wait cap (`_COMPUTE_HOST_COMPRESS_WAIT_CAP_SECS = 630`, `tui_gateway/compute_host_bridge.py`); below that cap the gateway answers `status: "pending"` and a smaller budget reports a false client timeout while the host is still compressing (upstream #97948). The generic 30s `requestTimeout` is untouched.
- `compressSession(sessionId:focusTopic:timeout:)` → `session.compress` with `{session_id, focus_topic?}`; `focus_topic` is omitted when empty/whitespace (Desktop parity).
- Typed `SessionCompressResult` covering the verified upstream response shapes:
  - in-process (`_compress_live`): `status: "compressed"|"aborted"`, `removed`, `summary{headline, token_line, note, aborted}`, `messages` (same shape `session.resume` returns);
  - compute-host (`_compress_via_compute_host`): host result forwarded verbatim, or `status: "pending"` + `message` when the bounded wait expired;
  - concurrent-compression lock: `lock_held: true` + `message`, no `status`;
  - `host_ack.output` as the record-text fallback (Desktop order: summary → host output → removed-count).
- `HermesClient.isMissingRPCMethod(_:)` classifies the legacy-gateway signal (`-32601` / "method not found" / "unknown method" / "no such method", mirroring Desktop's `isMissingRpcMethod`). Timeouts, 4009 busy rejections, compression errors, and transport loss are deliberately NOT in that class.

### `AppState`
- `executeSlashCommand` routes `compress`/`compact` to the dedicated path before the generic slash machinery. No busy guard: the gateway's own 4009 "session busy" answer is the accurate one (compute-host sessions legitimately compress mid-turn).
- Result semantics (Desktop parity):
  - **success** (`messages` present) → transcript adopted immediately via `adoptCompressedTranscript`, which mirrors `applyChatResume`'s invariants (viewport replacement bookkeeping, presentation-cache enrichment, review-card reattachment, authoritative clears of turn/ordering debt) without touching session identity; summary lines render as the slash-output record. **Mid-turn guard:** if a turn started while the RPC was in flight (`turnState == .running` or a locally-owned in-flight turn exists), adoption is deferred — the live turn's rows survive, `transcriptFreshnessIsStale = true` converges via the existing reconciliation machinery.
  - **`status: "pending"`** → informational record ("compression continues in the background"); never an error, never retried. The gateway pushes `session.info` + a `compacted` status edge on completion.
  - **`lock_held`** → the gateway's message is surfaced; no second compression launched.
  - **aborted** (status or `summary.aborted`) → transcript still adopts the gateway's authoritative (unchanged) history — Desktop adopts whenever `messages` is present — but the record is a ⚠️ note instead of a success summary.
  - An empty `messages` array is adopted as authoritative (pinned deliberately by test).
- Per-session coalescing: `compressingSessionIDs: Set<String>` — a second `/compress` for the same conversation is refused; different conversations compress concurrently (Desktop's `compressInFlightRef` model). `ComposerBar` shows a small "Compressing…" spinner while the *active* conversation is compressing; no progress bar (upstream exposes no incremental compression progress).
- Legacy fallback: only a missing-method failure falls back to today's `slash.exec`/`command.dispatch` path, preserving older-gateway compatibility (Desktop does the same via `runExec`). A compression timeout is never retried as a legacy command — that would start a second server-side compression.

### Not changed
Generic slash behavior, `prompt.submit`, session resume/admission machinery, transcript ownership, and the generic 30s `requestTimeout` are untouched (diff audited hunk-by-hunk). Errors render as the existing `⚠️` local `.system` rows — never submitted to Hermes.

## Test plan

- New `AppStateSessionCompressTests` (18 cases): routing of `/compress` + `/compact` to the dedicated RPC (and *not* `slash.exec`), `focus_topic` passthrough, missing-method → legacy fallback, timeout/busy → **no** fallback, pending/lock-held informational records, aborted-with-messages adoption + ⚠️, `summary.aborted` flag, empty-payload adoption, `host_ack.output` fallback + summary precedence, mid-turn defer, duplicate-compression coalescing, cross-session concurrency, and stale-completion isolation after a session switch.
- `HermesClientTests` additions: wire format of `session.compress` (params, omitted empty `focus_topic`), response parsing per upstream shape, timeout surfaces as `HermesError.timeout("session.compress")`, the 660s budget pinned (> `requestTimeout`), and `isMissingRPCMethod` classification table.
- Mac verification at head `ec410e7`: focused suites 166/166, **full unit suite 2400/2400**; `scripts/check-l10n-coverage.py` green (1688 sites, both new keys carry zh-Hans).

## Compatibility

`session.compress` has existed since the TUI gateway's introduction (upstream 2026-04), but nothing assumes it: gateways that predate the RPC are detected at runtime via the `-32601` family and keep today's `slash.exec` behavior. No version gates. Known boundary (deliberate): Desktop's stale-runtime `withSessionNotFoundResume` recovery is not replicated — a stale session id after sleep/wake surfaces the gateway's error instead of silently re-resuming.

## Follow-ups (out of scope here)

- Generic transient error surface for slash-command failures (today they render as `.system` transcript rows).
- Stale-id recovery for `session.compress` after sleep/wake (Desktop's `withSessionNotFoundResume`).
- Surfacing the gateway's `compacted` status edge as a live UI affordance.

## Summary by CodeRabbit

* **New Features**
  * Added manual conversation compression through the `/compress` and `/compact` commands.
  * Added optional focus-topic support for compression.
  * Added progress feedback while compression is underway.
  * Added support for pending, locked, aborted, and completed compression results.
  * Added Simplified Chinese translations for compression status messages.
* **Bug Fixes**
  * Improved transcript updates after compression while preserving active conversations and ongoing turns.
  * Added compatibility fallback for older servers that do not support compression.
  * Prevented duplicate compression requests for the same conversation.


kaishi00:
## Summary
Routes `/compress` and `/compact` from the generic `slash.exec` path to the dedicated `session.compress` RPC with a 660s budget, typed response handling, transcript adoption, per-session coalescing, a "Compressing…" affordance, and a missing-method-only legacy fallback.

## Findings

### [minor] New user-facing fallback strings bypass localization
**File:** Conduit/Services/AppState.swift:9636, 9647, 9683, 9694
**Issue:** Several new user-visible strings are raw literals, not `AppLocalization.string(...)`, and are absent from `Localizable.xcstrings`:
- `"Compression is already in progress."`
- `"Compression continues in the background; the transcript will refresh when it finishes."`
- `"⚠️ Context compression was aborted; the transcript is unchanged."`
- `"Compressed \(removed) messages."` / `"Nothing to compress."`

They render as `.system` transcript rows, so zh-Hans users see English. The `check-l10n-coverage.py` gate does not catch them (it only scans `AppLocalization.string`/SwiftUI literal/modifier sites plus literal `errorMessage`/`help`/`purposeText` assignments), so a green run does not actually cover them.
**Suggestion:** Wrap each in `AppLocalization.string(...)` and add the corresponding entries (with matching placeholder type, e.g. `"Compressed %lld messages."`) to `Localizable.xcstrings`.

### [minor] Spinner uses raw-id equality while coalescing uses alias-aware equivalence
**File:** Conduit/Services/AppState.swift:520-523, 9553
**Issue:** `compressingSessionIDs` stores the captured routing id (`insert(sessionID)`), and `isCompressingActiveSession` compares it for exact equality against `activeSessionId`. The duplicate guard immediately above uses `composerSessionIDsAreEquivalent`, which resolves aliases. If the active conversation's routing id rotates to an alias mid-compression, the "Compressing…" spinner disappears while the second `/compress` is still refused with "already in progress," which is a confusing inconsistency. The inline comment acknowledges the equality choice but the UX outcome is a real mismatch.
**Suggestion:** Either canonicalize the stored id or have `isCompressingActiveSession` use `composerSessionIDsAreEquivalent` (filtering the set) rather than raw `Set.contains`.

### [minor] Empty `messages` payload destructively clears the transcript
**File:** Conduit/Services/AppState.swift:9703, 9728; Conduit/Services/HermesClient.swift:352
**Issue:** `hasMessagesPayload` is true for `messages: []`, so a present-but-empty payload replaces the local transcript with an empty one (pinned by `testEmptyMessagesPayloadAdoptsEmptyTranscript`). This is deliberate Desktop parity, but it means any malformed/truncated gateway response that yields an empty array silently wipes the conversation view; there is no sanity check against a locally non-empty transcript.
**Suggestion:** At minimum keep the deliberate behavior but log the case; consider refusing to adopt an empty payload when the local transcript is non-empty (or only when `status`/`summary` confirms a genuine compaction), to bound the blast radius of an empty/truncated response.

### [nit] `isMissingRPCMethod` substring classification can false-positive
**File:** Conduit/Services/HermesClient.swift:1379-1390
**Issue:** Matching `contains("-32601")`, `"method not found"`, `"unknown method"`, `"no such method"` against `error.localizedDescription`/`description` can classify an unrelated provider/compression error as a missing RPC method, triggering a second server-side compression attempt through the legacy path — exactly what the comment says must never happen.
**Suggestion:** Prefer the structured `RpcError.code == -32601` signal, and only fall back to message matching when no `RpcError` is present and the message also references the expected method.

## Verdict
COMMENT

No critical or major defects found: no retain cycles, no new force-unwraps in production code, MainActor confinement is consistent (`AppState`/`HermesClient` are `@MainActor`), the stale-completion and mid-turn guards are tested, and the legacy fallback is correctly gated to missing-method failures only. The findings above are localization/consistency hardening.

<a href="https://opencode.ai/s/SpyuqZfQ">New%20session%20-%202026-09-13T13%3A01%3A23.744Z</a>
[opencode session](https://opencode.ai/s/SpyuqZfQ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34758608017)

kaishi00:
Review-round follow-up (commit 7c7c5b2 → a68966e), addressing the two findings from the gate:

**1. Persisted-history pagination/ordering invalidation.** Compression rewrites the gateway's persisted history, so a pre-compression `persistedTranscriptWindow.nextOffset`, `durablePersistedRowIDs`, and `persistedOrderingFrontier` describe a row universe that no longer exists. `adoptCompressedTranscript` now invalidates all three, and `reestablishPersistedHistoryAfterCompression` performs the same validated `order=latest offset=0` hydration the resume path uses (`persistedTranscriptOutcome`), identity-gated via `transcriptMatchesSession` and fenced by `isCurrentComposerSubmission` before and after. The window/frontier/row provenance are rebuilt **only** from that response; a legacy one-shot response or any failure leaves the window nil and backfill disabled until the next authoritative sync. A reconciliation-token rotation after adoption fences an in-flight reconcile that fetched pre-compression rows from overwriting the rebuilt invariants. Regression tests: `testCompressionRehydratesWindowFromFreshOffsetZeroResponse` (a pre-compression nextOffset cannot survive a second compression) and `testFailedRehydrationLeavesBackfillDisabled`.

**2. Non-trapping `removed` conversion.** `Int(Double)` traps outside the Int64 range and `removed` is gateway-authored; `SessionCompressResult` now uses an exact conversion (finite, within [-2^63, 2^63), whole values only) degrading bad payloads to nil. Covered by `testSessionCompressResultRejectsOutOfRangeAndInexactRemoved` (1e300, 2^63, 2.5, Int.min, and the 2^63−1024 edge → Int.max−1023).

Verification at head a68966e: focused suites 169/169, full unit suite 2403/2400+3 on the Mac; `check-l10n-coverage.py` green. One incidental cleanup in this delta: `isMissingRPCMethod`'s `(error as? CustomStringConvertible)` middle clause was dropped — behavior-neutral (`String(describing:)` already dispatches to `description`), it just silenced an always-succeeds-cast compiler warning.

kaishi00:
## Summary
Routes `/compress` and `/compact` through the dedicated `session.compress` RPC with a 660s budget, typed response handling, transcript adoption with persisted-history rehydration, per-session coalescing, a spinner, and a missing-method-only legacy fallback.

## Findings

### [major] Deferred mid-turn compression leaves persisted-history pagination stale
**File:** Conduit/Services/AppState.swift:9692-9696
**Issue:** When the RPC succeeds but a turn started in flight, the branch sets `transcriptFreshnessIsStale = true` and returns `.deferred` without invalidating `persistedTranscriptWindow`, `durablePersistedRowIDs`, or `persistedOrderingFrontier`. The server-side compression did rewrite the persisted row universe, so `nextOffset`/`canLoadEarlier` still describe the pre-compression history. `canLoadEarlierMessagesForActiveConversation` (AppState.swift:11023) and `loadEarlierMessages()` (10774) do not gate on `isBusy`, so a user tap during the live turn can fetch `offset=nextOffset` against the new (shortened) history and skip/garble older rows — the exact failure class the adopted path was just hardened against. `transcriptFreshnessIsStale` only converges lazily on the next send/foreground reconcile, so the window is stale in the interim.
**Suggestion:** In the deferred branch, invalidate the same invariants as `adoptCompressedTranscript` (at minimum `persistedTranscriptWindow = nil`, `durablePersistedRowIDs = []`, `persistedOrderingFrontier = PersistedOrderingFrontier()`) so backfill is disabled until the next authoritative sync, or schedule the rehydration once the turn settles.

### [minor] New user-facing fallback strings bypass localization
**File:** Conduit/Services/AppState.swift:9595, 9635, 9661, 9672, 9709, 9720
**Issue:** The fallback/informational literals rendered as `.system` transcript rows are raw strings, not `AppLocalization.string(...)`, and are absent from `Localizable.xcstrings`: `"Compression is already in progress."`, `"Compression continues in the background; …"`, `"⚠️ Context compression was aborted; the transcript is unchanged."`, `"Compressed \(removed) messages."`, `"Nothing to compress."`, and both `"⚠️ Compression failed: …"` rows. The zh-Hans catalog only carries `"Compressing…"` and `"Compression is already in progress for this conversation."`, so the other paths stay English. `check-l10n-coverage.py` does not scan `appendSlashOutput` arguments, so a green gate does not cover them.
**Suggestion:** Wrap each in `AppLocalization.string(...)` and add catalog entries (using a placeholder key such as `"Compressed %lld messages."` for the count), then update the tests that pin the English literals.

### [minor] `isMissingRPCMethod` substring classification can false-positive
**File:** Conduit/Services/HermesClient.swift:1385-1404
**Issue:** Only a `-32601` code short-circuits. Every other `RpcError` and every non-`RpcError` falls through to lowercased substring matching over `description`/`errorDescription`. A provider/compression error whose text happens to contain e.g. `"unknown method"` is classified as a missing method and triggers `runLegacyCompressionFallback`, starting a second server-side compression — the outcome the comment says must never happen. The check also does not require the message to reference `session.compress`.
**Suggestion:** Treat `RpcError.code == -32601` as the sole structured signal and, for message-based fallback, also require the text to name the method (and prefer no message matching when an `RpcError` with a code is present).

### [minor] Empty `messages: []` payload destructively clears a non-empty transcript
**File:** Conduit/Services/AppState.swift:9726-9730, Conduit/Services/HermesClient.swift:340
**Issue:** `hasMessagesPayload` is true for `messages: []`, so a present-but-empty payload replaces any local transcript with empty (pinned by `testEmptyMessagesPayloadAdoptsEmptyTranscript`). A malformed/truncated gateway response yielding `[]` silently wipes the conversation view with no sanity check or log. The parity rationale is understood, but the blast radius is unbounded.
**Suggestion:** Keep the deliberate behavior, but log the case and consider refusing to adopt an empty payload when the local transcript is non-empty unless `status`/`summary` confirms a genuine compaction.

### [nit] Spinner uses raw-id equality while the duplicate guard is alias-aware
**File:** Conduit/Services/AppState.swift:515-523, 9549
**Issue:** `compressingSessionIDs` stores the captured routing id and `isCompressingActiveSession` does exact `Set.contains(activeSessionId)`, while the duplicate guard two functions up uses `composerSessionIDsAreEquivalent`. If the active conversation's routing id rotates to an alias mid-compression, the "Compressing…" spinner disappears while a second `/compress` is still refused as "already in progress."
**Suggestion:** Canonicalize the stored id or make `isCompressingActiveSession` filter with `composerSessionIDsAreEquivalent`.

## Verdict
REQUEST_CHANGES

The core routing, timeout budget, response parsing (non-trapping `Int` conversion), token fencing, and stale-completion guards are solid, with good test coverage. Blocking on the deferred-path stale window since it can silently skip persisted history through the same mechanism this PR set out to fix; the rest are hardening.

<a href="https://opencode.ai/s/N21DD47o">New%20session%20-%202026-09-13T14%3A52%3A03.221Z</a>
[opencode session](https://opencode.ai/s/N21DD47o)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34763850297)

kaishi00:
CodeRabbit follow-up (commit a633f03):

- **Major**: a deferred compression adoption (live turn owns the transcript) now also invalidates `persistedTranscriptWindow`, `durablePersistedRowIDs`, and `persistedOrderingFrontier` — compression rewrote the persisted history server-side even though the transcript was not replaced, so a pre-compression Load Earlier window must not remain usable. `transcriptFreshnessIsStale` stays true, and the branch deliberately does NOT rehydrate: the live turn is why adoption was deferred, and the authoritative reconciliation path re-establishes pagination/provenance when it settles. Regression test `testDeferredCompressionInvalidatesBackfillWindow` pins the invalidated window, disabled backfill, surviving live-turn rows, and that no rehydration runs in the deferred branch (`hydrationCalls == 1`).
- **Minor**: the `persistedTranscript` seam now receives the production-built query string (third argument), so tests can observe the actual request. `testCompressionRehydratesWindowFromFreshOffsetZeroResponse` asserts both post-adoption rehydrations requested the validated latest tail — one expectation hardcoded to `?limit=120&offset=0&order=latest&include_compacted=true` so a query-builder regression cannot pass tautologically. All ~110 existing seam closures were updated mechanically.

Multi-model gate: 3× APPROVE. Verification at head a633f03: focused suites 170/170, full unit suite 2404/2404 on the Mac; l10n checker green.

kaishi00:
## Summary
Routes `/compress` and `/compact` through the dedicated `session.compress` RPC with a 660s budget, typed response handling, transcript adoption + persisted-history rehydration, per-session coalescing, a spinner, and a missing-method-only legacy fallback. The core routing, timeout budget, non-trapping `removed` conversion, and stale-completion guards are solid.

## Findings

### [minor] New user-facing strings bypass localization
**File:** Conduit/Services/AppState.swift:9603, 9643, 9669, 9680, 9729, 9740
**Issue:** The `.system` transcript rows rendered via `appendSlashOutput` are raw English literals and absent from `Localizable.xcstrings`:
- `"Compression is already in progress."` (fallback when `result.message == nil`)
- `"Compression continues in the background; the transcript will refresh when it finishes."`
- `"⚠️ Context compression was aborted; the transcript is unchanged."`
- `"Compressed \(removed) messages."` / `"Nothing to compress."`
- both `"⚠️ Compression failed: \(error.localizedDescription)"` rows

`scripts/check-l10n-coverage.py` only scans `AppLocalization.string`/SwiftUI initializers and literal `errorMessage`/`help`/`purposeText` assignments (see scripts/check-l10n-coverage.py:76), so a green gate does not cover `appendSlashOutput` arguments. zh-Hans users see English. Only `"Compressing…"` and `"Compression is already in progress for this conversation."` were added to the catalog.
**Suggestion:** Wrap each in `AppLocalization.string(...)`, add catalog entries with the correct placeholder family (e.g. `"Compressed %lld messages."`), and update the tests that pin the English literals.

### [minor] `isMissingRPCMethod` message-substring matching can false-positive
**File:** Conduit/Services/HermesClient.swift:1397
**Issue:** Only `RpcError.code == -32601` short-circuits. Every other `RpcError` (and any non-`RpcError`) falls through to substring matching over the message. A provider/compression error whose text happens to contain `"unknown method"`, `"no such method"`, etc. is classified as a missing method and triggers `runLegacyCompressionFallback`, starting a second server-side compression — exactly the outcome the comment says must never happen. The check also never requires the message to name `session.compress`.
**Suggestion:** Treat `RpcError.code == -32601` as the only structured signal; for message-based matching, require both a missing-method phrase and the method name, and skip message matching when an `RpcError` with a non-nil code is present.

### [minor] Present-but-empty `messages: []` destructively clears a non-empty transcript
**File:** Conduit/Services/AppState.swift:9686, Conduit/Services/HermesClient.swift:361
**Issue:** `hasMessagesPayload` is true for `messages: []`, so a malformed/truncated gateway response that yields an empty array replaces the local transcript with empty (pinned by `testEmptyMessagesPayloadAdoptsEmptyTranscript`). There is no sanity check against a locally non-empty transcript and no log, so the blast radius of a bad payload is unbounded.
**Suggestion:** Keep the deliberate Desktop-parity behavior but log the case, and consider refusing to adopt an empty payload when the local transcript is non-empty unless `status`/`summary` confirms a genuine compaction.

### [minor] Spinner uses raw-id equality while the duplicate guard is alias-aware
**File:** Conduit/Services/AppState.swift:528, Conduit/Services/AppState.swift:9557
**Issue:** `compressingSessionIDs` stores the captured routing id, and `isCompressingActiveSession` does exact `Set.contains(activeSessionId)`, while the duplicate guard immediately above uses `composerSessionIDsAreEquivalent`. If the active conversation's routing id rotates to an alias mid-compression, the `"Compressing…"` notice vanishes while a second `/compress` is still refused as "already in progress."
**Suggestion:** Canonicalize the stored id, or make `isCompressingActiveSession` filter with `composerSessionIDsAreEquivalent` instead of `Set.contains`.

### [minor] Deferred path invalidates the window but does not fence an in-flight reconciliation
**File:** Conduit/Services/AppState.swift:9702
**Issue:** The `.adopted` path rotates `reconciliationToken` (line 9585) specifically so a reconcile that fetched pre-compression history cannot overwrite the rebuilt invariants. The `.deferred` branch performs the same invalidation (`persistedTranscriptWindow = nil`, `durablePersistedRowIDs = []`, `persistedOrderingFrontier = ...`) but does not rotate the token. A reconcile already in flight whose pre-compression response commits after the deferred branch can write back the pre-compression window/frontier and clear `transcriptFreshnessIsStale`, defeating the invalidation this branch just performed.
**Suggestion:** Rotate `reconciliationToken` in the deferred branch too (mirroring the adopted path), or document why an in-flight reconcile cannot commit while `turnState == .running` / a locally owned turn exists.

### [nit] Compression spinner's `ProgressView` has no accessibility label
**File:** Conduit/Views/Components/ComposerBar.swift:469
**Issue:** The spinner is decorative but unlabeled; VoiceOver announces only the adjacent `Text`, and the two elements are not combined. Minor since the label conveys the state.
**Suggestion:** Add `.accessibilityLabel(AppLocalization.string("Compressing…"))` (or `.accessibilityHidden(true)` on the `ProgressView`) and consider `.accessibilityElement(children: .combine)`.

## Verdict
COMMENT

No critical/major defects: no retain cycles or new force-unwraps in production code, `AppState`/`HermesClient` remain `@MainActor`-confined, the timeout/cancellation path cleans up `pending`, and the deferred/empty-payload/stale-completion semantics are deliberate and tested. The findings are localization and hardening items.

<a href="https://opencode.ai/s/2kqLtyMZ">New%20session%20-%202026-09-13T16%3A25%3A52.380Z</a>
[opencode session](https://opencode.ai/s/2kqLtyMZ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34768491258)

PR #169, opened 2026-09-13
feat(chat): compression lifecycle parity — stale-runtime recovery, compaction edges, alias-aware spinner

kaishi00:
## Summary

Follow-up to #168 closing the remaining `/compress` lifecycle-parity gaps with upstream Hermes Desktop. Scope is deliberately narrow: stale-runtime recovery, `status.update` compaction edges, compression-indicator identity correctness, and compression-only localization. The generic slash-command error presentation, `AnyCodable` numeric hardening beyond the compress path, and progress UI are explicitly out of scope.

## 1. Stale runtime recovery for `session.compress`

Upstream Desktop wraps `session.compress` in `withSessionNotFoundResume`. Conduit now does the equivalent with its own identity machinery (`compressSessionWithRecovery`):

- **Trigger**: only a genuine session-not-found response — the gateway answers session-scoped RPCs for a reaped/detached runtime with `_err(4001, "session not found")` (`tui_gateway/server.py::_sess_nowait`); `HermesClient.isSessionNotFoundError` matches code 4001 or the `/session not found/i` message family. Timeouts, 4009 busy rejections, connection loss, compression failures, and missing-method are **never** recovery signals (timeout-based recovery is deliberately not opted into — compression is legitimately LLM-bound, matching Desktop's opt-in design).
- **Recovery**: resume the durable stored session once via the existing `openChatResumeSession` path (compact resume, `omit_messages`), using the stored id the catalog anchors the conversation to (catalog-row containment check — not the never-failing `canonicalSessionID` fallback).
- **Drift fence**: `isCurrentComposerSubmission` is re-evaluated AFTER the resume await; if the user navigated away or identity changed, the original compression error is rethrown — no rebind, no retry.
- **Rebind**: `setActiveSessionState(id: recoveredSessionID)` (the canonical state helper) updates the active runtime and re-anchors the persisted identity; the submission context is re-based onto the recovered runtime so every post-retry fence validates the new identity.
- **Retry**: exactly one, against the recovered runtime. A second failure is a real error surfaced against the recovered identity (`CompressionRecoveryContextError` carries the rebased context so the reporter never fences against the dead runtime). A resume failure rethrows the ORIGINAL compression error — the more meaningful one.

Deliberately deferred (Desktop's `registerRecoveredRuntime` cache): recording the recovered runtime as a catalog alias — exact-match claim/springer state covers the recovery window; noted in the code and in follow-ups.

## 2. `status.update` parsing

`StreamEvent.statusUpdate(sessionId:kind:text:)` + `StatusUpdateKind` (`.compacting` | `.compacted` | `.other(String)`), verified against `tui_gateway/server.py::_status_update`: `compacting` is active server-side compaction (auto-compaction progress re-tagged from the generic `lifecycle` kind), `compacted` is the terminal completion edge (the compute-host late ack). The in-process manual path emits unrelated kinds (`compressing`, then a bare `status`) which land in `.other` — they can never masquerade as compaction edges.

## 3. Compaction lifecycle in AppState

`handleStreamEvent` routes `.statusUpdate` before the reconcile buffer — these are session-scoped edges that must work for background conversations too, and are never transcript projections.

- **`compacting`**: records the conversation in `serverCompactingSessionIDs` (deduped across equivalent ids).
- **`pending` RPC results**: also record server-side compaction — the RPC ending is not the compression finishing, so the "Compressing…" affordance stays truthful while the host continues.
- **`compacted`** (terminal): clears the server-side claim; for the affected **active** conversation, invalidates `persistedTranscriptWindow` / `durablePersistedRowIDs` / `persistedOrderingFrontier` (shared `invalidatePersistedHistoryInvariants()` helper — the same invariants as PR #168) and, when no live turn owns the transcript, drives the existing `syncSession()` authoritative reconcile so the compressed transcript appears without leaving the session (Desktop's rationale: a manual `/compress` may return `pending`, and with no turn-end event nothing else refreshes it). During a live turn: no transcript replacement — provenance invalidated, `transcriptFreshnessIsStale = true`, reconcile token rotated so a pre-compaction reconcile cannot commit underneath the turn. For background conversations: only session-scoped state clears; the next open/resume hydrates authoritatively.
- **Trust rule**: `handleDisconnect` clears server-side claims — gateway state is unverifiable across socket loss, and a `compacted` edge may never arrive.

## 4. Alias-aware compression indicator

`isCompressingActiveSession` now unions local RPC claims and server-side compaction state with `composerSessionIDsAreEquivalent` membership (previously raw equality), so the "Compressing…" affordance survives a runtime rebind mid-compression and stays visible for pending server-side compressions. Duplicate detection was already equivalence-based — unchanged.

## 5. Localization

The five compression fallback strings (lock-held, pending, aborted, removed-count, nothing-to-compress) route through `AppLocalization.string` with zh-Hans `Localizable.xcstrings` entries (`%lld` for the count). `scripts/check-l10n-coverage.py` green (1693 sites).

## Test plan

- `StreamEventParserTests`: `compacting`/`compacted` parse with session id + text; unrelated kinds (`compressing`, `status`, `ready`, `lifecycle`) stay `.other`.
- `AppStateSessionCompressTests` (+8): stale recovery (resume → retry once → rebind → adoption → window rehydration), drift prevents retry (no rebind, no retry), resume failure surfaces the original error, timeout never recovers, double-4001 resumes only once, retry failure surfaces against the recovered runtime, pending keeps the server-side claim until the `compacted` edge (and invalidates provenance then), compacted during a live turn defers replacement, inactive-session compacted touches nothing but its own state, spinner survives a runtime alias rebind, duplicate compression coalesces across equivalent runtime ids.
- `HermesClientTests`: `isSessionNotFoundError` classification table (4001/message vs busy/timeout/connection/compression failures).

## Verification (head 61f1532)

- Focused suites: **384/384** (SessionCompress, StreamEventParser, HermesClient, AppStateChatResume, CompactResume, ForegroundLifecycle).
- Full unit suite: **2420/2420** on an erased simulator (an AppLanguage local-runner flake passed 20/20 in isolation — local defaults contamination, not a product regression).
- `scripts/check-l10n-coverage.py`: green.
- Multi-model gate: round 1 = 1× APPROVE + 2× REQUEST_CHANGES (all findings fixed: error-swallow on the post-rebind retry, alias-invisible rebind, exact-match set removal, dead stored-id guard); round 2 = **3× APPROVE**.

## Follow-ups (out of scope)

- Record the recovered runtime as a catalog alias after stale-runtime recovery (current exact-match state covers the recovery window; cosmetic divergence only).
- Recovery runtime cache (Desktop's single-flight `registerRecoveredRuntime`) to avoid re-minting runtimes across surfaces.
- Generic slash-command error presentation; broader `AnyCodable` numeric hardening; unrelated `status.update` kinds surfacing.

## Summary by CodeRabbit

- **New Features**
  - Added Simplified Chinese messages for context-compression progress, completion, interruption, background processing, failure, and empty states.
  - Added clearer compression status updates for active conversations, including processing and completion states.

- **Bug Fixes**
  - Improved recovery when an active session becomes unavailable during compression, with bounded retry behavior.
  - Improved session identity and status accuracy across reconnects and multiple conversations.
  - Prevented stale or expired compression operations from affecting current session state.
  - Ensured completed compression results consistently update conversation history.


kaishi00:
## Summary
Adds compression-lifecycle parity for `/compress`: stale-runtime recovery (resume stored session, rebind, retry once), `status.update` parsing for `compacting`/`compacted` edges, an alias-aware "Compressing…" indicator, and localization of the compression fallbacks.

## Findings

### [major] Server-side compaction claim can leave the spinner stuck for the connection lifetime
**File:** Conduit/Services/AppState.swift:531, 12702-12723, 5032
**Issue:** `serverCompactingSessionIDs` is only cleared by a `status.update` `.compacted` edge or `handleDisconnect()`. If the terminal edge is lost on a live socket (or arrives under an alias the catalog doesn't yet know — precisely the deferred `registerRecoveredRuntime` follow-up), the "Compressing…" affordance stays true indefinitely with no timeout and nothing else to clear it. The `.compacting` insert path (12706-12712) also means a background conversation's claim leaks until disconnect.
**Suggestion:** Add an in-band clear/refresh: clear an entry when a `session.info` snapshot for that session no longer reports compaction, and/or expire claims after a bounded interval (mirroring the local RPC timeout rationale), so a dropped edge cannot wedge the composer.

### [minor] Drift fence runs after the resume side effect
**File:** Conduit/Services/AppState.swift:9693-9705
**Issue:** `openChatResumeSession(storedSessionID, ...)` is awaited before `isCurrentComposerSubmission(context)` is evaluated. If the user navigated away during the long LLM-bound compression RPC, recovery still issues a stateful `session.resume` (minting/re-attaching a runtime) for an abandoned submission; the fence only prevents the rebind/retry, not the resume.
**Suggestion:** Re-check ownership before resuming (in addition to after), and bail with `originalError` without calling `openChatResumeSession`.

### [minor] Recovered runtime id is not validated before rebinding
**File:** Conduit/Services/AppState.swift:9701-9708
**Issue:** `setActiveSessionState(id: recoveredSessionID)` is called with whatever `resumed.sessionId` is, with no guard that it is non-empty or actually belongs to `storedSessionID`. Compare the sibling `createAndReconcileSession`, which guards `!runtimeSessionID.isEmpty`. A malformed/empty resume response would rebind `activeSessionId` and then retry compression against a bogus id.
**Suggestion:** Guard `!recoveredSessionID.isEmpty` (and prefer `resumed.sessionId` only when it is equivalent to the stored/requested identity); otherwise rethrow `originalError`.

### [minor] Empty `session_id` status updates insert `""` into the claim set
**File:** Conduit/Services/StreamEventParser.swift:64-72, Conduit/Services/AppState.swift:12706-12712
**Issue:** The parser defaults a missing `session_id` to `""` and unconditionally emits `.statusUpdate`. `composerSessionIDsAreEquivalent` rejects empty operands, so the `.compacting` dedup guard fails and `""` is inserted into `serverCompactingSessionIDs`, where it accumulates and can never be cleared.
**Suggestion:** Return `nil` from the parser (or guard `!sessionId.isEmpty` in `applyStatusUpdate`) when `status.update` carries no session id.

### [nit] `isCompressingActiveSession` is now O(catalog) per evaluation
**File:** Conduit/Services/AppState.swift:538-542
**Issue:** Each `contains(where:)` calls `composerSessionIDsAreEquivalent`, which scans `(sessions + cronSessions)` inside the closure. This computed property is read from `ComposerBar.body`, so with a large catalog it runs a catalog scan per element on every render.
**Suggestion:** Resolve the active conversation's equivalent-id set once (or keep a canonical key on the claim entries) rather than re-scanning the catalog inside the membership test.

### [nit] Unused `text` parameter
**File:** Conduit/Services/AppState.swift:12702
**Issue:** `applyStatusUpdate(sessionId:kind:text:)` never uses `text`; the comments explain it is display-only upstream, so the parameter is dead weight that invites accidental use.
**Suggestion:** Drop the parameter from the handler (the enum already carries the kind), or explicitly surface `text` in the compression notice if intended.

## Verdict
COMMENT

The core recovery/drift/rebind logic and the edge-routing look correct and are well covered by the new tests. The stuck-claim path is the one worth addressing before merge; the rest are robustness/nit cleanups.

<a href="https://opencode.ai/s/u4uzPzab">New%20session%20-%202026-09-13T21%3A07%3A58.519Z</a>
[opencode session](https://opencode.ai/s/u4uzPzab)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34782827063)

kaishi00:
## Summary
Adds stale-runtime recovery for `session.compress`, parses gateway `status.update` compaction edges into AppState lifecycle state, makes the "Compressing…" affordance alias-aware, and localizes the compression fallbacks.

## Findings

### [major] Server-side compaction claim has no clear path except the terminal edge or disconnect
**File:** Conduit/Services/AppState.swift:531, 9813-9817, 12706-12719, 5032
**Issue:** `serverCompactingSessionIDs` is inserted on `status.update(.compacting)` and on a `pending` compress result, but is only removed by a `status.update(.compacted)` edge or `handleDisconnect()`. There is no timeout and no in-band clear from `session.info`/`syncSession` (nothing in `SessionRuntimeSnapshot` reports compaction). A dropped terminal edge — or an edge addressed under an alias the catalog hasn't recorded yet, which is exactly the deferred `registerRecoveredRuntime` follow-up — wedges the "Compressing…" notice for the rest of the connection. A subsequent successful `compressed` result also does not clear it. Background claims likewise leak until disconnect.
**Suggestion:** Add a bounded expiry or clear an entry when an authoritative `session.info`/reconcile snapshot for that session no longer reports compaction; additionally clear the matching claim whenever a terminal compress result (non-pending, non-lock-held) lands.

### [major] Recovery rebinds the active conversation without the resume identity admission gate
**File:** Conduit/Services/AppState.swift:9701-9708
**Issue:** `setActiveSessionState(id: resumed.sessionId)` adopts the gateway-returned runtime directly. Every other resume path routes the response through `ConversationIdentityGate.admit` / `adoptChatResumeResult` before touching conversation-owned state (AppState.swift:3711-3739, 4301), and `createAndReconcileSession` guards `!runtimeSessionID.isEmpty` (AppState.swift:4180). Here there is no admission against `storedSessionID` and no empty check: a malformed/empty or contradictory resume response can set `activeSessionId` to `""` or to another conversation's runtime and then retry compression against it, adopting that transcript.
**Suggestion:** Guard `!recoveredSessionID.isEmpty` and validate it belongs to `storedSessionID` (e.g., reuse `ConversationIdentityGate.admit` with the catalog row's identity); on failure rethrow `originalError` without rebinding.

### [minor] Drift fence runs after the stateful resume side effect
**File:** Conduit/Services/AppState.swift:9693-9705
**Issue:** `openChatResumeSession(storedSessionID, ...)` is awaited before `isCurrentComposerSubmission(context)` is evaluated. If the user navigated away during the long, LLM-bound compression RPC, recovery still issues a stateful `session.resume` (minting/re-attaching a runtime) for an abandoned submission; the fence only prevents the rebind/retry.
**Suggestion:** Re-check ownership before resuming (in addition to after) and bail with `originalError` without calling `openChatResumeSession`.

### [minor] Empty `session_id` status updates insert `""` into the claim set
**File:** Conduit/Services/StreamEventParser.swift:64-72, Conduit/Services/AppState.swift:12709-12712
**Issue:** The parser defaults a missing `session_id` to `""` and unconditionally emits `.statusUpdate`. `composerSessionIDsAreEquivalent` rejects empty operands, so the `.compacting` dedup guard fails and `""` is inserted into `serverCompactingSessionIDs`, where it accumulates and can never be cleared by any `compacted` edge.
**Suggestion:** Return `nil` from the parser (or guard `!sessionId.isEmpty` in `applyStatusUpdate`) when `status.update` carries no session id.

### [minor] `applyCompactionCompleted` can spawn a redundant reconcile during an in-flight sync
**File:** Conduit/Services/AppState.swift:12740-12751
**Issue:** The deferred branch only checks `turnState == .running` / `locallyOwnedInFlightTurn`. During `.synchronizing` a `compacted` edge falls through to `Task { syncSession() }`, starting a second reconcile and rotating the reconciliation token underneath the in-flight one.
**Suggestion:** Treat `.synchronizing` like the live-turn case (invalidate + mark stale) or no-op until the active sync settles.

### [nit] `applyStatusUpdate`'s `text` parameter is unused
**File:** Conduit/Services/AppState.swift:12702
**Issue:** `text` is never read; the comment explains it is display-only upstream, so it is dead weight that invites accidental use.
**Suggestion:** Drop the parameter from the handler (the enum carries the kind) or surface it in the compression notice if intended.

### [nit] `isCompressingActiveSession` rescans the catalog per element
**File:** Conduit/Services/AppState.swift:538-542
**Issue:** Each `contains(where:)` calls `composerSessionIDsAreEquivalent`, which builds `(sessions + cronSessions)` and a `Set` per candidate row. Read from `ComposerBar.body` (Conduit/Views/Components/ComposerBar.swift:316), this is O(catalog) per evaluation for each claim.
**Suggestion:** Resolve the active conversation's equivalent-id set once, or key claims by a canonical conversation id.

### [nit] `"⚠️ Compression failed:"` remains unlocalized
**File:** Conduit/Services/AppState.swift:9640, 9774
**Issue:** The five fallback strings in this PR route through `AppLocalization.string`, but the two compression-failure prefixes still interpolate hardcoded English, leaving a mixed-language surface in zh-Hans.
**Suggestion:** Route the format string through `AppLocalization.string("⚠️ Compression failed: %@", errorForReporting.localizedDescription)` (with a catalog entry).

## Verdict
REQUEST_CHANGES

The recovery/drift/rebind logic and edge routing are well-covered by tests, but the unbounded server-side claim can permanently wedge the composer, and rebinding the active conversation without the identity admission gate violates the invariant every other resume path enforces.

<a href="https://opencode.ai/s/YdvR3ctF">New%20session%20-%202026-09-13T21%3A18%3A16.923Z</a>
[opencode session](https://opencode.ai/s/YdvR3ctF)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34783351768)

kaishi00:
## Summary
Adds stale-runtime recovery to `/compress` (resume stored id, rebind, single retry), parses gateway `status.update` `compacting`/`compacted` edges into new AppState lifecycle state, makes the compression affordance alias-aware, and localizes the compression fallbacks.

## Findings

### [major] `serverCompactingSessionIDs` has no bounded clear path (stuck spinner / unbounded set)
**File:** Conduit/Services/AppState.swift:531, 9817-9821, 12706-12723, 5032
**Issue:** The claim is inserted on `status.update(.compacting)` and on a `pending` compress result, but only removed by a `.compacted` edge or `handleDisconnect()`. There is no timeout and no in-band clear from any authoritative `session.info`/reconcile snapshot (nothing reports compaction). Consequences:
- A dropped/never-sent terminal edge wedges "Compressing…" for the remainder of the connection; `isCompressingActiveSession` stays true indefinitely.
- A later terminal compress result (adopted, non-pending) never clears a pre-existing server claim for that session.
- Race: if `.compacted` is processed *before* the `pending` RPC continuation runs, `applyStatusUpdate(.compacted)` clears nothing, then `applySessionCompressionResult` inserts the claim at 9820 — permanently stuck.
- Background conversations' claims accumulate until disconnect.
**Suggestion:** Add a bounded expiry (mirroring the local RPC timeout rationale) and/or clear the matching claim whenever a terminal (non-pending, non-lock-held) compress result lands, and ideally on an authoritative `session.info` that no longer reports compaction.

### [major] Recovery rebinds the active conversation without the identity admission gate
**File:** Conduit/Services/AppState.swift:9703-9715
**Issue:** `setActiveSessionState(id: recoveredSessionID)` adopts the gateway-returned runtime directly. Every other resume path validates the response through `ConversationIdentityGate.admit`/`applyChatResume` (AppState.swift:3711-3742, 4301) and records the runtime→durable mapping via `conversationIdentityIndex.recordAuthoritative`. This path does neither, and there is no empty check (unlike `createAndReconcileSession` at 4179-4186). A malformed or contradictory resume response can set `activeSessionId` to `""` or to a foreign conversation's runtime, after which compression is retried against it and `applySessionCompressionResult` can adopt that foreign transcript into the current conversation. It also means `canonicalID` (ChatScrollState.swift:405-414) falls back to the ephemeral runtime, so `chatResumeCoordinator.rememberSessionID` persists a non-durable id and the durable fence for the retry is weakened.
**Suggestion:** Validate `!recoveredSessionID.isEmpty` and admit it against `storedSessionID`/the catalog row (reuse `ConversationIdentityGate`), and record the authoritative identity mapping before retrying; on rejection rethrow `originalError` without rebinding.

### [minor] Drift fence runs after the stateful resume side effect
**File:** Conduit/Services/AppState.swift:9695-9707
**Issue:** `openChatResumeSession(storedSessionID, ...)` is awaited before `isCurrentComposerSubmission(context)` is checked. If the user navigated away during the LLM-bound compression call, recovery still issues a stateful `session.resume` (minting/re-attaching a runtime for an abandoned submission); the fence only prevents the rebind/retry, leaking the minted runtime.
**Suggestion:** Re-check ownership before the resume as well, and bail with `originalError` without calling `openChatResumeSession`.

### [minor] Empty `session_id` is inserted into the claim set
**File:** Conduit/Services/StreamEventParser.swift:64-72, Conduit/Services/AppState.swift:12706-12716
**Issue:** The parser defaults a missing `session_id` to `""` and unconditionally emits `.statusUpdate`. `composerSessionIDsAreEquivalent` rejects empty operands, so the dedup guard fails and `""` is inserted, where no `.compacted` edge can ever clear it (equivalence fails on empty).
**Suggestion:** Return `nil` from the parser when `status.update` carries no session id, or guard `!sessionId.isEmpty` in `applyStatusUpdate`.

### [minor] `applyCompactionCompleted` starts a second reconcile during `.synchronizing`
**File:** Conduit/Services/AppState.swift:12744-12755
**Issue:** The deferral branch only checks `turnState == .running || locallyOwnedInFlightTurn != nil`. During `.synchronizing`, a `.compacted` edge falls through to `Task { syncSession() }`, which calls `beginReconciliation()` (AppState.swift:3393-3415) and rotates `reconciliationToken` underneath an in-flight reconcile (which will then fail its commit-time ownership check).
**Suggestion:** Treat `.synchronizing` (and arguably `.reconnecting`) like the live-turn case — invalidate and mark stale, or no-op until the active sync settles.

### [nit] `isCompressingActiveSession` rescans the catalog per element
**File:** Conduit/Services/AppState.swift:538-542
**Issue:** Each `contains(where:)` calls `composerSessionIDsAreEquivalent`, which builds `(sessions + cronSessions)` and a `Set` per candidate. This property is read from `ComposerBar.body` (Conduit/Views/Components/ComposerBar.swift:316), so it is O(catalog) per claim per render.
**Suggestion:** Resolve the active conversation's equivalent-id set once, or key claims by a canonical conversation id.

### [nit] `applyStatusUpdate`'s `text` parameter is unused
**File:** Conduit/Services/AppState.swift:12706
**Issue:** `text` is never read; the comment explains it is display-only upstream, so the parameter invites accidental use.
**Suggestion:** Drop it, or surface the gateway copy in the compression notice if intended.

No force unwraps, retain cycles, or MainActor/Sendable violations were introduced (the new closure is non-escaping and runs on `@MainActor`).

## Verdict
REQUEST_CHANGES

<a href="https://opencode.ai/s/5ek2y9aa">New%20session%20-%202026-09-13T22%3A48%3A41.484Z</a>
[opencode session](https://opencode.ai/s/5ek2y9aa)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34787825183)

kaishi00:
## Summary
This PR adds `/compress` lifecycle parity with upstream Hermes Desktop: stale-runtime recovery (resume durable id, identity-admit, rebind, retry once), `status.update` `compacting`/`compacted` parsing into bounded/generation-fenced AppState claims, an alias-aware compression spinner, persisted-history invalidation on terminal compaction, and zh-Hans localization of the compression fallbacks.

## Findings

### [minor] `compacted` during `.synchronizing` starts a second reconcile and abandons the in-flight one
**File:** Conduit/Services/AppState.swift:12955
**Issue:** `applyCompactionCompleted` only treats `turnState == .running || locallyOwnedInFlightTurn != nil` as "transcript owned". During `.synchronizing` (an active reconcile/resume), it falls through to `Task { syncSession() }`, which calls `beginReconciliation()` and rotates `reconciliationToken` while a reconcile is mid-hydration. The in-flight reconcile then fails its commit-time token check and its `settleReconciliation` no-ops, so the sync is silently superseded and re-run. This is extra LLM/socket work and can cancel a user-initiated restoration (`syncSession()` calls `cancelChatResumeRestoration()`), though state does eventually converge because buffered events are carried over in `beginReconciliation`.
**Suggestion:** Treat `.synchronizing` (and arguably `.reconnecting`) like the live-turn case — invalidate provenance, set `transcriptFreshnessIsStale`, rotate the token, and let the in-flight sync settle rather than launching a competing one.

### [minor] Refreshing a claim under an equivalent alias leaks expiry tasks/claims until they expire
**File:** Conduit/Services/AppState.swift:12902
**Issue:** The comment on `establishServerCompactionClaim` says "Replace any prior expiry task for this conversation so repeated compactions never accumulate sleeping tasks", but only `serverCompactionClaimExpiryTasks[sessionId]` is cancelled for the exact alias. A `.compacting` edge addressed to a different but equivalent runtime id writes a second `serverCompactionClaims`/`serverCompactionClaimExpiryTasks` entry and leaves the first one sleeping until it fires. Behavior stays correct (the expiry path keeps the spinner while an equivalent claim is live and `clearServerCompactionClaim` is equivalence-aware), but the stated invariant is violated and entries accumulate per alias over a long connection.
**Suggestion:** Cancel and remove all equivalence-matching claim/task entries before inserting the refresh (reuse the same `composerSessionIDsAreEquivalent` sweep `clearServerCompactionClaim` uses).

### [minor] Deferred `syncSession()` Task does not re-validate the active conversation
**File:** Conduit/Services/AppState.swift:12964
**Issue:** `applyCompactionCompleted` verifies `composerSessionIDsAreEquivalent(sessionId, activeSessionId)` synchronously, then defers the reconcile in a `Task`. If the user navigates to another conversation before the task runs, `syncSession()` reconciles whichever session is now active, not the compacted one (harmless but spurious, and it can cancel an unrelated in-flight restoration).
**Suggestion:** Capture the identity and re-check equivalence inside the Task before calling `syncSession()`.

### [minor] Disconnect does not bump terminal generations, so a late `pending` continuation can resurrect a cleared claim
**File:** Conduit/Services/AppState.swift:5052
**Issue:** `handleDisconnect` clears `serverCompactingSessionIDs`, `serverCompactionClaims`, and cancels tasks, but leaves `serverCompactionTerminalGenerations` unchanged. The `pending` fence in `applySessionCompressionResult` compares the generation, so a `session.compress` that resolves `pending` around a socket drop can re-establish a server claim that disconnect just deliberately discarded ("gateway state is unverifiable across socket loss"). In practice the same socket loss usually fails the RPC, so this is a narrow race.
**Suggestion:** Bump (or clear) the generation for the disconnected client's sessions on disconnect so no pre-loss continuation can re-establish a claim.

### [nit] `text` parameter is dead
**File:** Conduit/Services/AppState.swift:12820
**Issue:** `applyStatusUpdate(sessionId:kind:text:)` never reads `text`; the doc comment explains it is display-only upstream. It is unused weight that invites accidental use.
**Suggestion:** Drop the parameter (the enum carries the kind) or surface the gateway copy in the compression notice.

### [nit] `isCompressingActiveSession` rescans the catalog per claim element
**File:** Conduit/Services/AppState.swift:558
**Issue:** Each `contains(where:)` invokes `composerSessionIDsAreEquivalent`, which can build `(sessions + cronSessions)` and a `Set` per candidate. This property is read from `ComposerBar.body`, so with a large catalog and multiple claims it is O(catalog × claims) per render. Cheap in the common empty-set case, but unbounded in the alias-heavy case.
**Suggestion:** Resolve the active conversation's equivalent-id set once, or key claims by a canonical conversation id.

### [nit] Stray double blank line
**File:** Conduit/Services/AppState.swift:12939
**Issue:** Two consecutive blank lines before the `applyCompactionCompleted` doc comment.
**Suggestion:** Remove one.

## Verdict
COMMENT

No force unwraps, retain cycles, or MainActor/Sendable violations were introduced (the in-flight claim closure is non-escaping and MainActor-bound; the expiry task uses `[weak self]`). The prior blocking issues (unbounded claim, missing identity admission, pre-resume drift fence, empty session id) are addressed in HEAD with strong test coverage. The remaining items are robustness/nit cleanups, not merge blockers.

<a href="https://opencode.ai/s/5bLpmJ5F">New%20session%20-%202026-09-14T01%3A43%3A57.074Z</a>
[opencode session](https://opencode.ai/s/5bLpmJ5F)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34796803487)

kaishi00:
## Summary

Adds `/compress` lifecycle parity with upstream Hermes Desktop: stale-runtime recovery (resume durable id, identity admission, single retry), typed `status.update` `compacting`/`compacted` parsing with bounded/generation-fenced server-compaction claims, alias-aware compression spinner, persisted-history invalidation on terminal compaction, and zh-Hans localization of the compression fallbacks.

## Findings

### [minor] `compacted` during `.synchronizing` starts a competing reconcile and abandons the in-flight one
**File:** Conduit/Services/AppState.swift:12965
**Issue:** `applyCompactionCompleted` treats only `turnState == .running || locallyOwnedInFlightTurn != nil` as "transcript owned". During `.synchronizing` (an active reconcile/resume) it falls through to `Task { syncSession() }`, which calls `beginReconciliation()` and rotates `reconciliationToken` while a reconcile is mid-hydration. Because `handleStreamEvent` deliberately routes `.statusUpdate` *before* `bufferIfReconciling`, this collision is guaranteed whenever compaction completes during a sync, not rare. The superseded reconcile then fails its commit-time token check; the new `syncSession()` also calls `cancelChatResumeRestoration()`, so a user-initiated restoration can be cancelled (state eventually converges).
**Suggestion:** Treat `.synchronizing` (and arguably `.reconnecting`) like the live-turn case — invalidate provenance, set `transcriptFreshnessIsStale`, rotate the token, and let the in-flight sync settle rather than launching a competing one.

### [minor] Deferred `syncSession()` Task does not re-validate the active conversation
**File:** Conduit/Services/AppState.swift:12974
**Issue:** `applyCompactionCompleted` verifies `composerSessionIDsAreEquivalent(sessionId, activeSessionId)` synchronously, then defers the reconcile in a `Task`. If the user navigates to another conversation before the task runs, `syncSession()` reconciles whatever is active then, cancelling that conversation's restoration/causing a spurious reconcile.
**Suggestion:** Capture the identity and re-check equivalence inside the Task before calling `syncSession()`.

### [minor] Refreshing a claim under an equivalent alias leaks the prior claim and expiry task
**File:** Conduit/Services/AppState.swift:12903-12920
**Issue:** The comment states repeated compactions never accumulate sleeping tasks, but only `serverCompactionClaimExpiryTasks[sessionId]` is cancelled for the exact alias. A `.compacting` edge addressed to a different but equivalent runtime id writes a second `serverCompactionClaims`/`serverCompactionClaimExpiryTasks` entry and leaves the first sleeping until it fires. Behavior stays correct (both are equivalence-aware) but entries accumulate per alias over a long connection.
**Suggestion:** Cancel and remove all equivalence-matching claim/task entries before inserting the refresh (reuse the same `composerSessionIDsAreEquivalent` sweep `clearServerCompactionClaim` uses).

### [minor] Disconnect does not fence pending continuations; generation map is never pruned
**File:** Conduit/Services/AppState.swift:5052-5057, 9938
**Issue:** `invalidateServerCompactionState()` clears the claim set and expiry tasks but leaves `serverCompactionTerminalGenerations` unchanged. A `session.compress` resolving `pending` around a socket drop then passes the `terminalServerCompactionGeneration(...) == invocationGeneration` fence and re-establishes a server claim that disconnect deliberately discarded ("gateway state is unverifiable across socket loss"). It is now bounded by the 11-minute expiry, but the deliberate clear is undone. Separately, the generation dictionary is never pruned anywhere, so it grows for the app/account lifetime and survives sign-out/server replacement.
**Suggestion:** Bump (or clear) the generation for the disconnected client's sessions on disconnect so no pre-loss continuation can re-establish a claim; prune/clear the generation map at teardown boundaries as well.

### [nit] `text` parameter is dead
**File:** Conduit/Services/AppState.swift:12830
**Issue:** `applyStatusUpdate(sessionId:kind:text:)` never reads `text`; the doc comment explains it is display-only upstream. Unused weight that invites accidental use.
**Suggestion:** Drop the parameter (the enum already carries the kind) or surface the gateway copy in the compression notice if intended.

### [nit] `isCompressingActiveSession` rescans the catalog per claim element
**File:** Conduit/Services/AppState.swift:538-542, 7714-7730
**Issue:** Each `contains(where:)` calls `composerSessionIDsAreEquivalent`, which scans `(sessions + cronSessions)` and builds a `Set` per candidate row. This property is read from `ComposerBar.body`, so it is O(catalog × claims) per render. Cheap in the common empty/one-element case, but unbounded in alias-heavy cases.
**Suggestion:** Resolve the active conversation's equivalent-id set once, or key claims by a canonical conversation id.

### [nit] Stray double blank line
**File:** Conduit/Services/AppState.swift:12949-12950
**Issue:** Two consecutive blank lines before the `applyCompactionCompleted` doc comment.
**Suggestion:** Remove one.

No force unwraps, retain cycles, or MainActor/Sendable violations were introduced (the recovery closure is non-escaping and MainActor-bound; expiry tasks use `[weak self]`). The previous blocking issues — unbounded claim, missing identity admission, pre-resume drift fence, empty session id, terminal-result claim clearing — are addressed at HEAD with strong test coverage.

## Verdict
COMMENT

<a href="https://opencode.ai/s/01qD8OJL">New%20session%20-%202026-09-14T02%3A42%3A44.267Z</a>
[opencode session](https://opencode.ai/s/01qD8OJL)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34800057534)

PR #173, opened 2026-09-14
feat(auth): add native OAuth for Google OIDC dashboards

### LINKED ISSUE #172: Support Hermes native OAuth flow for Google OIDC dashboards
# Support Hermes native OAuth flow for Google OIDC dashboards

## Summary

Conduit cannot currently sign in to a Hermes dashboard whose only interactive authentication provider is Google OIDC, even though the same dashboard works with the official Hermes Desktop app.

This should not require any change to the dashboard's existing OIDC provider or Google client configuration. Current Hermes already exposes a native-app OAuth broker used by Hermes Desktop.

## Current behaviour

Conduit discovers dashboard providers through `GET /api/auth/providers`, but OAuth-only providers are handed to `AuthWebView`, which loads the dashboard `/login` page in a `WKWebView`.

Google rejects OAuth authorization in embedded user agents such as `WKWebView`, usually with `disallowed_useragent`. Google requires an external/system browser flow for native apps.

There is also an earlier UI obstacle: `LoginView.hasConnectableInput()` and `LoginView.connect()` require non-empty username and password fields before provider discovery runs. An OAuth-only dashboard should require only its URL before the provider is known.

Relevant Conduit code:

- `Conduit/Views/LoginView.swift`
  - `hasConnectableInput()` requires URL, username and password
  - OAuth-only discovery routes to `AuthWebView`
  - `AuthWebView` hosts the login flow in `WKWebView`
- `Conduit/Services/DashboardTicketBridge.swift`
  - authenticated REST and WebSocket-ticket minting currently depend on WebKit session cookies

## Expected behaviour

Given a Hermes dashboard already configured with Google OIDC:

1. Enter the dashboard URL in Conduit without dummy username/password values.
2. Conduit discovers the available provider(s).
3. Google authentication opens in a system-browser surface accepted by Google.
4. Conduit receives Hermes native tokens, stores them securely per dashboard, and connects.
5. Reconnect, refresh and logout work without changing the server's existing Google OIDC configuration.

## Suggested implementation

Port the native OAuth flow already used by the official Hermes Desktop app:

1. Generate a PKCE verifier/challenge and random `state`.
2. Start a short-lived callback listener bound to `127.0.0.1` on an ephemeral port.
3. Open the system-browser authentication URL at `/auth/native/authorize` with the PKCE challenge, callback URI and state.
4. Receive the one-time code at the loopback callback and verify `state`.
5. Exchange it at `POST /auth/native/token`.
6. Store the returned Hermes access token, refresh token, provider and expiry in Keychain, scoped to the saved dashboard UUID.
7. Use `Authorization: Bearer <access_token>` for dashboard REST calls and `POST /api/auth/ws-ticket`.
8. Refresh through `POST /auth/native/refresh`, with one forced refresh after a structured 401 only where replay is safe.
9. Clear both native tokens and legacy cookie state on logout/removal.

The existing `DashboardTicketBridge.requestJSON()` / `mintTicket()` interface could gain a bearer-backed mode, avoiding changes across its many callers while retaining cookie/password compatibility.

Hermes Desktop references:

- `apps/desktop/electron/native-oauth-login.ts`
- `apps/desktop/electron/native-oauth.ts`
- `apps/desktop/electron/native-access-token.ts`
- `apps/desktop/electron/oauth-rest-request.ts`

Hermes server endpoints already available:

- `GET /auth/native/authorize`
- `POST /auth/native/token`
- `POST /auth/native/refresh`
- bearer-authenticated `POST /api/auth/ws-ticket`

## Compatibility and security requirements

- Do not ask users to create a second Google client or alter working Hermes OIDC configuration.
- Do not spoof a browser user agent or attempt Google login inside `WKWebView`.
- Keep PKCE verifier and bearer tokens out of logs.
- Store tokens in Keychain and scope them to a dashboard identity.
- Preserve the existing password/cookie flow.
- Preserve path-prefixed dashboard URLs and Cloudflare Access headers.
- Keep token refresh single-flight and guard against stale results after dashboard switching or logout.

## Validation

At minimum:

- OAuth-only provider accepts URL-only setup.
- Password-only and mixed-provider dashboards continue to work.
- PKCE/state verification, cancellation, timeout and replay rejection are covered.
- Google account selection, MFA/passkey and consent complete on a physical iPhone.
- Access-token expiry refreshes without reauthentication when a valid refresh token exists.
- Confirmed refresh rejection returns to sign-in.
- Cold launch, logout and multiple saved dashboards retain correct token isolation.
- WebSocket ticket minting and ordinary dashboard REST calls both use the same authenticated identity.

Google policy reference: https://developers.googleblog.com/upcoming-security-changes-to-googles-oauth-20-authorization-endpoint-in-embedded-webviews/

lucvan:
## Summary

Adds native Hermes OAuth/PKCE sign-in for iOS dashboards that use providers such as Google OIDC.

Conduit currently opens interactive dashboard authentication in `WKWebView`. Google rejects embedded user agents, while Hermes Desktop uses Hermes' native broker flow in the system browser. This change ports that brokered flow to Conduit without requiring changes to Hermes or the configured Google/OIDC provider.

Closes #172.

## What changed

- Allows URL-only provider discovery, while preserving credential requirements for password login.
- Detects Hermes' advertised `native_pkce` capability.
- Opens `/auth/native/authorize` with a loopback redirect, random state, and PKCE S256 challenge.
- Exchanges and refreshes Hermes-issued bearer tokens through `/auth/native/token` and `/auth/native/refresh`.
- Stores tokens per dashboard in Keychain and clears them with the saved connection.
- Uses bearer authentication for dashboard REST calls and `/api/auth/ws-ticket` while preserving the existing cookie/WebKit path.
- Preserves base-URL path prefixes and dashboard-scoped Cloudflare Access credentials.
- Adds unit coverage for PKCE, authorization URL construction, callback validation, token decoding/storage isolation, and provider selection.

## macOS and physical-device validation

**Physical-device retesting completed for commit `2fd0a9e97f827b2c270ee7bb5a59bae8e306ca08` on 2026-09-15.** The tester performed the interactions on a physical iPhone 16 Pro running iOS 27.0; Xcode supplied build/install status and sanitized stage evidence.

Environment: macOS 26.1, Xcode 26.3 (17C529), iOS SDK 26.2. XcodeGen generated the local project. The latest OAuth code was freshly built, signed, installed, and launched before these tests.

| Physical test | Result |
| --- | --- |
| Tailscale connected and existing dashboard reachable in device Safari | Passed; tester reconfirmed |
| Force-quit and reopen from Home Screen with a non-default profile selected | Passed; same profile and same chat restored |
| Sign Out of This Dashboard | Passed; returned to sign-in and subsequent Connect opened Google authorization |
| Cancel Google authorization | Passed; returned cleanly without connection or error |
| Retry after cancellation with empty username/password | Passed; Google login completed and dashboard/chat worked |
| Real temporary 127.0.0.1 callback and state validation | Passed; confirmed by stage diagnostics |
| Token exchange, WebSocket-ticket minting, and WebSocket activation | Passed; confirmed by stage diagnostics and tester observation |

Sanitized cancellation evidence:

```text
OAuth stage: listener ready
OAuth stage: Safari presented
OAuth stage: user cancelled browser
OAuth stage: cancelled
OAuth stage: sign-in view removed
```

Sanitized successful retry evidence:

```text
OAuth stage: listener ready
OAuth stage: Safari presented
OAuth stage: callback validated
OAuth stage: token exchange succeeded
OAuth stage: ticket minted
WebSocket handshake completed
OAuth stage: sign-in view removed
```

The latest full simulator unit suite recorded **2529 total: 2528 passed, 0 failed, 1 skipped** on iPhone 17 Pro / iOS 26.3.1, arm64. The skipped test requires the newer CarPlay SDK. Portable validation also passed: `git diff --check`, localization coverage (1739 call sites), test-plan validation (134 unit + 6 UI classes), and Python tooling (160 total: 158 passed, 2 skipped).

Local build adaptations: Personal Team automatic Debug signing, a local bundle identifier, empty Debug entitlements, and an external CarPlay compatibility source copy for this older SDK. These adaptations are excluded from the commits; repository CarPlay changes were reverted in `87b1b71`. The tracked tree remained clean. Release signing, dashboard configuration, and Google OIDC configuration were unchanged.

**Privacy:** Only manually curated outcomes and fixed stage labels are published here. No raw Xcode/device console logs, chat messages, chat titles, session identifiers, private dashboard addresses, cookies, OAuth codes, tokens, or credentials are attached or included.

## Remaining validation

The primary physical OAuth flow and the requested profile-restoration, logout, cancellation, and retry scenarios now have evidence for the latest OAuth revision. Upstream required CI and human approval remain separate gates; local results do not substitute for them.

Real token-expiry/rotation scenarios, multiple dashboards, path-prefixed deployments, separate password-login regression, and Cloudflare Access combined with dashboard OIDC have not been physically exercised in this run. The documented limitation on attaching Cloudflare service-token headers to Safari's browser authorize request remains.

## Security notes

- Google authorization is not performed in `WKWebView`.
- Uses high-entropy state and PKCE S256.
- Accepts only Hermes' loopback callback path and validates callback state.
- Tokens are dashboard-scoped and stored in Keychain using `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`.
- Bearer tokens are attached only to URLs validated against the configured dashboard origin.
- A 401-triggered refresh automatically replays only GET, HEAD, and the single-use WebSocket-ticket POST, not arbitrary mutations.

## Summary by CodeRabbit

- **New Features**
  - Added native OAuth sign-in for supported dashboards through a secure external browser flow.
  - Added OAuth token persistence, refresh, and automatic session restoration.
  - Added support for OAuth-only dashboards without username or password credentials.
  - Preserved browser-based sign-in for dashboards without native OAuth support.
  - Added Simplified Chinese translations for sign-in, dashboard, and Bot Mode messages.

- **Documentation**
  - Documented native OAuth limitations for Cloudflare authorization policies.


kaishi00:
## Summary

Adds a native OAuth/PKCE broker flow for OAuth-only Hermes dashboards: provider discovery gains URL-only gating, a loopback `NWListener` receives the OIDC callback, tokens are exchanged/refreshed/stored per-dashboard in Keychain, and `DashboardTicketBridge` transparently routes REST/ws-ticket calls through bearer auth while preserving the cookie/WebKit path. Test coverage is added for PKCE, URL construction, callback validation, token storage, and provider classification.

## Findings

### [severity: major] Loopback server reads a single arbitrary chunk and never accumulates
**File:** Conduit/Services/NativeOAuth.swift:438
**Issue:** `connection.receive(minimumIncompleteLength: 1, maximumLength: 16_384)` completes as soon as **one** byte arrives, and there is no receive loop. The request line/query can therefore be parsed while partially received (e.g. `"GET "` or a target split across TCP segments), `target.contains("code=")` fails, the handler returns without registering another receive, and the login hangs until the 5-minute timeout. It also cannot handle requests larger than 16 KB.
**Suggestion:** Loop `receive` accumulating `Data` until `\r\n\r\n` is seen (bounded), then parse the request line; respond only afterwards.

### [severity: major] `SFSafariViewController` is embedded, not presented
**File:** Conduit/Services/NativeOAuth.swift:560 (used at :531)
**Issue:** `SafariAuthenticationView` embeds `SFSafariViewController` as a SwiftUI child via `UIViewControllerRepresentable`. `SFSafariViewController` is documented for modal presentation; hosting it as a child view controller is unsupported and can break its presentation behavior and the `safariViewControllerDidFinish` delegate (the only cancel path besides swipe-dismiss). This is the sole browser surface of the feature and the PR states it is untested on device.
**Suggestion:** Present the controller modally (e.g. from a hosting controller via a representable that calls `present(_:animated:)`, or wrap it in a `.fullScreenCover` presentation), and verify the Done/cancel delegate on device.

### [severity: minor] Callback validates `error` before `state`
**File:** Conduit/Services/NativeOAuth.swift:137-138
**Issue:** `callbackCode` throws `.callbackRejected` on any non-empty `error` before checking `state`. Since the loopback listener accepts connections from any local process on `127.0.0.1`, a local app can send `/callback?error=x` (no valid state needed) and abort the user's pending sign-in. The server also unconditionally returns a "Signed in" success page even when the callback is rejected.
**Suggestion:** Validate `state` first, then treat `error` as rejection; render success/failure HTML based on the validated result.

### [severity: minor] Cancellation during `start()` is reported as a listener failure
**File:** Conduit/Services/NativeOAuth.swift:473-476
**Issue:** `finish` always resumes `startContinuation` with `NativeOAuthError.listenerFailed`. If `stop()` is called (sheet dismissed) or the timeout fires before the listener is ready, `run` throws `listenerFailed`, which is not a `CancellationError`, so the `.task` catch routes to `onError` and surfaces a spurious "Could not start the secure sign-in callback." (and `.timedOut` is never surfaced).
**Suggestion:** Resume `startContinuation` with the actual `result` failure (or `CancellationError` when stopped/cancelled).

### [severity: minor] Error copy is unlocalized
**File:** Conduit/Services/NativeOAuth.swift:52-64
**Issue:** `NativeOAuthError.errorDescription` returns raw English literals, while `AuthClientError` (the analogous type) routes user-facing text through `AppLocalization.string`. These strings are shown directly via `LoginView`'s `onError`. The l10n checker only scans wrapped call sites, so it passes while zh-Hans users see English detail text.
**Suggestion:** Wrap the descriptions in `AppLocalization.string(...)` and add catalog entries.

### [severity: minor] `URLSession` and delegate are never invalidated
**File:** Conduit/Services/NativeOAuth.swift:164 (also NativeAuthClient.swift:194)
**Issue:** A `URLSession` retains its delegate until invalidated; neither `NativeOAuthAPIClient` nor `NativeAuthClient` invalidates it. Each login attempt creates a fresh session/delegate, leaking them.
**Suggestion:** Give the API clients a `deinit` (or explicit teardown) that calls `session.invalidateAndCancel()`, and release `nativeOAuthSession` in `DashboardTicketBridge.invalidate()`.

### [severity: minor] Provider classification treats missing flags as OAuth-capable
**File:** Conduit/Services/NativeAuthClient.swift:142-158
**Issue:** `supports_session != false && supports_password != true` classifies a provider that declares **neither** flag as OAuth-capable (`nil != false` and `nil != true` both hold). A password provider that simply omitted `supports_session` can be routed into the native flow (and even pinned as `provider=` when it's the only entry).
**Suggestion:** Require `supports_session == true` explicitly (or at least require `supports_password != true` plus an explicit session signal).

### [severity: minor] Repeated 403 clears tokens and forces re-login
**File:** Conduit/Services/NativeOAuth.swift:208-216, 322-325
**Issue:** Any 401 **or 403** becomes `.signInRequired`; after refresh, a second 403 clears the Keychain tokens and forces re-authentication. A legitimate authorization (permission) 403 would log the user out. (The existing cookie path also maps 403 to `signInRequired`, but the native path additionally destroys the stored session.)
**Suggestion:** Distinguish 401 from 403 for token clearing, or only clear tokens when refresh itself fails with 401.

### [severity: minor] Missing tests for the refresh/replay and rejection paths
**File:** ConduitTests/NativeOAuthTests.swift
**Issue:** The most behavior-sensitive logic — 401→refresh→replay safety gating, refresh failure clearing tokens, and `NativeOAuthSession` per-dashboard isolation — has no coverage.
**Suggestion:** Add tests injecting a fake API client/session to exercise refresh success, refresh failure, and the non-replayable-method path.

### [severity: nit] Stale comments / duplicated Cloudflare write
**File:** Conduit/Views/LoginView.swift:482-483, 170-188; ConduitUITests/LoginKeyboardUITests.swift:128
**Issue:** The `connectInputsArePresent` doc and the UI-test helper comment still claim username/password are required for Connect. The `.sheet` onSuccess path also persists Cloudflare access a second time (already saved in `connect()`), and does not reset a prior `failure`.
**Suggestion:** Update the comments; drop the redundant `saveCloudflareAccess` and clear `failure` on entry.

## Verdict

REQUEST_CHANGES

The core design (PKCE S256, high-entropy state, origin-validated bearer requests, non-replayable mutating calls, dashboard-scoped Keychain) is sound, but the loopback request handling and browser presentation need to be fixed before device validation, and the draft's own admission that the full flow is untested on a physical device supports holding the change.

<a href="https://opencode.ai/s/28YK016x">New%20session%20-%202026-09-14T20%3A32%3A42.427Z</a>
[opencode session](https://opencode.ai/s/28YK016x)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34893438854)

lucvan:
Addressed the review findings in `9ef8dd3`:

- loopback HTTP handling now accumulates bounded chunks through the complete header terminator before parsing;
- Safari is presented modally from a dedicated presenter controller and retains the Done/cancel delegate path;
- callback state is validated before provider errors, malformed/wrong-state local probes no longer terminate the pending flow, and callback pages reflect success/failure;
- cancellation/timeout during listener startup preserves the actual error;
- native OAuth errors are localized with zh-Hans catalog entries;
- native URL sessions are invalidated and the bridge releases its native session on teardown;
- 403 permission failures no longer refresh or destroy stored tokens; only authentication 401s do;
- added tests for 401 refresh + safe replay, non-idempotent replay refusal, rejected-refresh token clearing/isolation, 403 preservation, and state-before-error validation;
- updated stale URL-only gating comments and removed the duplicate Cloudflare write.

One suggested change was intentionally not applied as written: requiring `supports_session == true` would disable native OAuth against current Hermes. `/api/auth/providers` is already generated from `list_session_providers()` and currently emits only `name`, `display_name`, and `supports_password`; endpoint membership is therefore the explicit session-capability signal. The code and a regression test now document that server contract.

Portable gates pass again: localization/placeholder coverage, test-plan validation, catalog JSON validation, `git diff --check`, and 160 script tests (2 skipped). Swift/XCTest and physical-device callback validation remain pending macOS CI/device access.

kaishi00:
## Summary
Adds a native OAuth/PKCE sign-in path for OAuth-only Hermes dashboards: URL-only provider discovery, an `NWListener` loopback callback, `/auth/native/*` token exchange/refresh stored per-dashboard in Keychain, and a bearer-backed transport in `DashboardTicketBridge`. The second commit addresses most of the prior review round (loopback accumulation, Safari modal presentation, state-before-error, 401-only token clearing, URLSession invalidation, localized copy).

## Findings

### [severity: minor] Error copy key uses `%@` but the runtime interpolation is an integer
**File:** Conduit/Services/NativeOAuth.swift:63 (catalog: Conduit/Localizable.xcstrings:14898)
**Issue:** `case .requestFailed(let status): return AppLocalization.string("Dashboard sign-in failed (HTTP \(status)).")` interpolates an `Int`, which `String.LocalizationValue` renders as `%lld`. The catalog entry added is `"Dashboard sign-in failed (HTTP %@)."`. The checker's heuristic (`is_int_interpolation` only recognizes `Int(...)`, count/index/total identifiers) classifies `status` as `%@`, so the coverage check passes, but the runtime key is `%lld` and the zh-Hans translation will never be found (English fallback). The rest of the codebase wraps such values (`String(status)`) specifically to force `%@`.
**Suggestion:** Use `AppLocalization.string("Dashboard sign-in failed (HTTP \(String(status))).")` (or catalog the `%lld` key).

### [severity: minor] Bearer transport is selected by token presence, but password/browser sign-in never clears native tokens
**File:** Conduit/Services/DashboardTicketBridge.swift:387-389; Conduit/Views/LoginView.swift:206-219, 605-627
**Issue:** `NativeOAuthSession` is constructed whenever any `native-oauth-tokens.v1` record exists for the dashboard, so the bridge uses bearer auth for all REST/ws-ticket traffic regardless of how the current connection was authenticated. The password path (`LoginView.connect`) and the WebView `onTicket` path clear credentials/cookies but never call `clearNativeOAuthTokens`. If a stale token set survives a non-OAuth sign-in, the fresh password/browser session is silently shadowed (or the user is signed out when the stale refresh is rejected). Reachability is narrow today (explicit sign-out and 400/401 refresh clear tokens), but the transport mode is keyed off durable storage rather than the connection's actual auth mode.
**Suggestion:** Clear native OAuth tokens in the password and WebView success paths (mirroring `clearCredentials`), or thread the resolved auth mode through `HermesConnection` so the bridge chooses transport explicitly.

### [severity: minor] Provider classification treats an absent `supports_password` as OAuth-capable
**File:** Conduit/Services/NativeAuthClient.swift:142-157
**Issue:** `hasNativeOAuthProvider` / `nativeOAuthProvider` use `supports_password != true`, so a provider entry that omits the key (or carries `supports_session: false` with no password flag) is treated as an interactive OAuth provider. The removed `supports_session != false` guard actually rejected `supports_session: false`; the new code cannot. Misclassification is gated by `supportsNativeOAuth()`, but a legacy password provider without the flag can still be routed into the native flow and pinned as `provider=`.
**Suggestion:** Require positive evidence, e.g. treat `supports_password == false` (explicitly) as OAuth, or require an explicit session signal in addition to `supports_password != true`.

### [severity: minor] Cloudflare Access service tokens cannot authenticate the browser authorize step
**File:** Conduit/Services/NativeOAuth.swift:570 (authorize URL), 99-122
**Issue:** The native flow attaches dashboard-scoped Cloudflare Access headers only inside `NativeOAuthAPIClient` (token exchange/refresh/REST). The `/auth/native/authorize` URL is opened in `SFSafariViewController`, which cannot receive those headers, so a dashboard whose authorize path is protected by a Cloudflare *service-token* policy will be challenged (or redirected to interactive Cloudflare login) and the brokered flow cannot complete. The PR lists "Cloudflare Access combined with dashboard OIDC" as untested.
**Suggestion:** Document this limitation and validate on device; if service-token-protected authorize paths must work, consider whether Hermes exposes an authorize endpoint reachable without the Access challenge, or detect the CF redirect and fall back to the WebView path.

### [severity: minor] `NativeAuthClient`'s `URLSession` is never invalidated
**File:** Conduit/Services/NativeAuthClient.swift:193
**Issue:** The struct builds `URLSession(configuration:delegate:)` per login and never invalidates it. `URLSession` retains its delegate and itself until invited; `NativeOAuthAPIClient` was given a `deinit`/`invalidate`, but the password client was not. Each `LoginView.connect()`/discovery attempt leaks a session and delegate.
**Suggestion:** Add an explicit teardown (or make the client a final class with `deinit { session.invalidateAndCancel() }`) and call it from the login/discovery call sites.

### [severity: nit] Recursive `receive` can grow the stack for a byte-at-a-time local client
**File:** Conduit/Services/NativeOAuth.swift:463-480
**Issue:** `receiveRequest` recurses from each receive callback until `\r\n\r\n` or the 64 KiB cap. A local process feeding one byte per segment can drive recursion depth proportional to the request size.
**Suggestion:** Drive accumulation with a loop (`while` + a single in-flight receive) or a small state machine rather than direct recursion.

### [severity: nit] New loopback request handling and timeout/cancel paths are untested
**File:** ConduitTests/NativeOAuthTests.swift
**Issue:** The prior major defect lived in the loopback accumulation/parsing and `finish`/continuation logic, yet the new tests cover only PKCE, URL construction, pure callback validation, token decoding/storage, classification, and refresh/replay. No coverage exercises `receiveRequest` accumulation, the timeout resume, or cancellation-during-`start()`.
**Suggestion:** Add focused tests for the parsing/accumulation logic (e.g. split TCP chunks) and for `NativeOAuthLoopbackServer` continuation behavior where feasible.

### [severity: nit] Force-unwraps in tests
**File:** ConduitTests/NativeOAuthTests.swift:74-78
**Issue:** `values["code_challenge"]!`, etc., force-unwrap dictionary lookups; a malformed URL turns a failed assertion into a crash.
**Suggestion:** Use `XCTUnwrap`.

## Verdict
COMMENT

The core security design (PKCE S256, high-entropy state, origin-validated bearer requests, non-replayable mutations, dashboard-scoped Keychain) is sound and the previous blocking defects appear resolved. The remaining items are correctness/robustness gaps (localization key mismatch, stale-token transport selection, provider heuristic) rather than blockers, but they are worth fixing before/around device validation.

<a href="https://opencode.ai/s/azW94Mxf">New%20session%20-%202026-09-14T21%3A54%3A28.940Z</a>
[opencode session](https://opencode.ai/s/azW94Mxf)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34901246707)

lucvan:
Addressed the second review round in `5a541d3`:

- forced the localized HTTP status interpolation to `String(status)`, matching the catalog's `%@` runtime key;
- password-login and WebView-success paths now clear any prior dashboard-scoped native tokens so bearer transport cannot shadow the newly validated cookie session;
- provider classification now requires `supports_password == false` explicitly, while retaining compatibility with Hermes' current session-provider payload;
- converted `NativeAuthClient` to a final class and invalidated its `URLSession` in `deinit`;
- replaced callback-style recursive accumulation with an async `while` read loop and bounded accumulator;
- added accumulator split-chunk/bound tests and deterministic cancellation-before-listener-readiness coverage;
- replaced authorization-URL test force-unwraps with `XCTUnwrap`;
- documented the Cloudflare Access limitation in the README: `SFSafariViewController` cannot attach service-token headers to `/auth/native/authorize`. Native token exchange/refresh/API requests still carry them, but a service-token-only browser challenge cannot be bypassed or safely downgraded to embedded Google OAuth.

Portable gates pass again: localization and placeholder coverage, test-plan validation, String Catalog JSON validation, `git diff --check`, and 160 script tests (2 skipped). The PR remains draft pending macOS Swift/XCTest CI and physical-device loopback validation.

kaishi00:
## Summary

Ports Hermes Desktop's brokered OAuth/PKCE flow to iOS: URL-only provider discovery, an `NWListener` loopback callback, `/auth/native/{authorize,token,refresh}` with bearer-only REST/ws-ticket transport, dashboard-scoped Keychain tokens, plus tests. The latest commit closes most of the prior review round (bounded request accumulation, modal Safari, state-before-error, 401-only clearing, localized copy, session invalidation).

## Findings

### [severity: major] Bearer transport can still be selected from stale tokens after a non-OAuth sign-in
**File:** Conduit/Services/AppState.swift:2963, :3364, :5619 (effect at Conduit/Services/DashboardTicketBridge.swift:387-389)
**Issue:** `DashboardTicketBridge` chooses bearer auth whenever a `native-oauth-tokens.v1` record exists for the dashboard, independent of how the current connection actually authenticated. The last commit clears those tokens in `LoginView`'s password path (LoginView.swift:612) and WebView path (:216), but the equivalent AppState success paths do not: saved-password restore (`:3364`), silent-renewal re-auth (`:5619`), and the `.browserSignIn` repair path (`:2963`, which only clears password credentials). On those paths a leftover native token makes the fresh cookie session silently shadowed; if the stale refresh is rejected the user is immediately signed out despite having just authenticated.
**Suggestion:** Clear `KeychainHelper.clearNativeOAuthTokens(dashboardID:)` on those success paths too, or thread the resolved auth mode through `HermesConnection`/the bridge so transport is chosen explicitly rather than inferred from durable storage.

### [severity: minor] Connection Setup probe rejects native-OAuth-only dashboards
**File:** Conduit/Services/ConnectionSetupTest.swift:408
**Issue:** When discovery returns providers with no password support, the probe reports `.unexpectedServerResponse`. A dashboard whose only session provider is Google/OIDC — exactly the target of this PR — therefore fails the wizard's "Test Connection" even though `LoginView` can now sign into it.
**Suggestion:** When `HermesProviderCheck.hasNativeOAuthProvider(providers)` holds (and `supportsNativeOAuth()` verifies), route to `.requiresInteractiveSignIn`/a terminal native outcome instead of a dashboard failure.

### [severity: minor] Provider classification ignores explicit `supports_session: false`
**File:** Conduit/Services/NativeAuthClient.swift:142-156
**Issue:** `hasNativeOAuthProvider`/`nativeOAuthProvider` key only on `supports_password == false`, so an entry like `{"supports_password": false, "supports_session": false}` is treated as an interactive OAuth provider. Current Hermes omits `supports_session`, but a future/alternate payload that emits it as `false` would be misrouted and pinned as `provider=`.
**Suggestion:** Require positive evidence, e.g. `supports_password == false && supports_session != false`, in both helpers.

### [severity: minor] Loopback read task/connection is not cancelled on timeout or finish
**File:** Conduit/Services/NativeOAuth.swift:484-522
**Issue:** `finish` cancels the `NWListener` but never the in-flight `NWConnection`s or their `readRequest` Tasks. A local client that connects and withholds `\r\n\r\n` leaves the Task awaiting `receive` indefinitely after the 5-minute timeout fires; repeated attempts accumulate live connections and Tasks for the process lifetime.
**Suggestion:** Track active connections (or the pending read task) and cancel/close them from `finish`/`stop`.

### [severity: minor] `SFSafariViewController` shares the Safari cookie jar for OAuth
**File:** Conduit/Services/NativeOAuth.swift:672-714
**Issue:** `SFSafariViewController` is not session-isolated — it reads/writes Safari's cookies — and `ASWebAuthenticationSession` is the platform-recommended OAuth surface, with cleaner dismissal/lifecycle semantics. For a Google OIDC flow this is worth confirming on device (Google acceptance and cookie side effects), especially since the PR's premise is avoiding disallowed user agents.
**Suggestion:** Consider `ASWebAuthenticationSession` (ephemeral session) for the authorize step, or explicitly validate and document the `SFSafariViewController` behavior on device.

### [severity: nit] `hasConnectableInput` retains now-unused credential parameters
**File:** Conduit/Views/LoginView.swift:490-492
**Issue:** `username`/`password` are ignored, which is misleading for the shared Connect-gate helper and can mask a future regression.
**Suggestion:** Drop the parameters (updating call site and `LoginFieldNavigationTests`) or rename the helper to reflect URL-only gating.

### [severity: nit] Force-unwraps remain in the URLProtocol test stub
**File:** ConduitTests/NativeOAuthTests.swift:16-20
**Issue:** `request.url!` and `HTTPURLResponse(...)!` turn a malformed stub request into a crash rather than an assertion failure.
**Suggestion:** Guard/`XCTUnwrap` and fail the load with an error instead.

## Verdict

COMMENT

The core security design (PKCE S256, high-entropy state, origin-validated bearer requests, non-replayable mutations, dashboard-scoped Keychain, state-before-error callback handling) is sound, and the previously blocking loopback/presentation defects are resolved. The remaining items are correctness/robustness gaps — chiefly the incomplete stale-token cleanup in AppState's non-OAuth sign-in paths — that should be fixed before physical-device validation.

<a href="https://opencode.ai/s/OHbfi1y4">New%20session%20-%202026-09-14T22%3A03%3A32.861Z</a>
[opencode session](https://opencode.ai/s/OHbfi1y4)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34902042435)

lucvan:
I handled this as a consolidated whole-PR review rather than another comment-by-comment patch cycle. The resulting revision is `3e11d26`.

This revision addresses the confirmed integration and lifecycle issues together:

- routes OAuth-only and mixed-provider setup/repair through Hermes native OAuth when `native_pkce` is advertised
- keeps transient capability-probe failures from silently downgrading to the embedded browser path
- makes native-token, password-cookie, and browser-cookie transitions dashboard-scoped and transactional
- restores previous authentication state if repaired or replacement connection activation fails
- persists a successfully issued Hermes grant before the first ticket mint, allowing transient ticket failures and cold starts to retry without repeating Google sign-in
- fences logout/session replacement against late refresh completion and deduplicates refresh after concurrent 401s
- bounds bearer and token response reads while streaming
- closes accepted loopback connections on success, timeout, or cancellation
- adds regression coverage for setup/repair routing, token rollback/commit, cold-start native recovery, session invalidation, response limits, and loopback cancellation

Portable validation on Linux/WSL passes:

- `git diff --check`
- localization catalog JSON and coverage
- test-plan validation: 133 unit and 6 UI test classes discovered
- repository Python CI-tooling suite: 160 passed, 2 skipped

I am deliberately keeping the PR in draft. The remaining acceptance gates cannot be established from WSL or by further static review alone:

1. upstream macOS/Xcode build and XCTest execution, once the first-time-contributor workflow is approved
2. physical iPhone/iPad validation of the complete Google OAuth flow, especially Safari/system-browser lifecycle and the `127.0.0.1` callback

I will treat any further findings as triage input against those explicit gates rather than automatically creating another incremental revision.

kaishi00:
## Summary

Ports Hermes Desktop's brokered OAuth/PKCE flow to iOS: URL-only provider discovery, an `NWListener` loopback callback, `/auth/native/{authorize,token,refresh}` exchange with bearer-backed REST/ws-ticket transport, dashboard-scoped Keychain tokens, and a native-OAuth repair path. The fourth commit (`3e11d26`) consolidates the cookie↔bearer lifecycle into `NativeAuthConnection.commitCookies` and makes bridge reuse transport-aware.

## Findings

### [severity: minor] Concurrent refresh waiters bypass the rejected-refresh error taxonomy
**File:** Conduit/Services/NativeOAuth.swift:428-433 (vs. 442-454)
**Issue:** `refresh()` maps a rejected grant to `DashboardTicketBridgeError.signInRequired` and cancellation to `CancellationError` only on the path that *created* `refreshTask`. A second concurrent request that reaches `if let refreshTask { let refreshed = try await refreshTask.value ... }` rethrows the raw `NativeOAuthError.requestFailed(400/401)` (or a raw `URLError.cancelled`) instead. `AppState` keys re-auth off `error as? DashboardTicketBridgeError, case .signInRequired` (AppState.swift:5716), so that second caller will not run sign-in recovery and may surface an unexpected error.
**Suggestion:** Funnel the awaited `refreshTask.value` result through the same `do/catch` mapping (400/401 → `.signInRequired` + clear; invalidated cancellation → `CancellationError`) as the creator path, or extract the shared error-mapping helper.

### [severity: minor] Cold-start native restore failure leaves the live bridge in bearer mode with cleared tokens
**File:** Conduit/Services/AppState.swift:2622-2638
**Issue:** `restoreNativeOAuthConnection` calls `mintChatResumeTicket`, which lazily creates `dashboardTicketBridge` with a `NativeOAuthSession` (`usesNativeOAuth == true`, `isReady == true`). When a rejected refresh clears the Keychain grant, the catch only flips `showLogin`/`isConnected`; it never invalidates or rebuilds the bridge. Until the next `prepareDashboardBridge`, a request routed through it uses a bearer session whose durable tokens were just cleared (and whose in-memory `tokens` were not reset). The repair/login paths already handle this via `restoreNativeOAuthTokens`.
**Suggestion:** In the failure branch, call `restoreNativeOAuthTokens(KeychainHelper.loadNativeOAuthTokens(dashboardID:), dashboardID:baseURL:)` (or invalidate and nil the bridge) so the stale bearer session is torn down on the same boundary.

### [severity: minor] Bearer/refresh tokens use a weaker Keychain accessibility class than password credentials
**File:** Conduit/Services/AppState.swift:15585
**Issue:** Native OAuth tokens are saved with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, while saved passwords use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` (AppState.swift:15615). Bearer/refresh tokens are readable while the device is locked after first unlock; unless background refresh genuinely requires it, this weakens the token store relative to the password it shadows.
**Suggestion:** Use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` unless a background-refresh requirement is documented.

### [severity: minor] OAuth authorize step uses the shared Safari cookie jar
**File:** Conduit/Services/NativeOAuth.swift:826-830
**Issue:** The flow presents `SFSafariViewController`, which reads/writes Safari's persistent cookies and cannot be session-isolated. `ASWebAuthenticationSession` (with `prefersEphemeralWebBrowserSession`) is the platform-recommended OAuth surface, gives cleaner dismissal/lifecycle semantics, and avoids reusing an existing Google session or leaving OAuth cookies in Safari. This was raised previously; it remains a design/device-validation risk rather than a resolved item.
**Suggestion:** Either move the authorize step to `ASWebAuthenticationSession` or document/validate the `SFSafariViewController` cookie-sharing behavior on device.

### [severity: nit] Test URLProtocol stub has thread-unsafe shared state
**File:** ConduitTests/NativeOAuthTests.swift:7, 351-382, 384-411
**Issue:** `NativeOAuthURLProtocolStub.handler` is a static mutable var set on the test thread and invoked from URL-loading threads, while tests capture and mutate `apiCalls`/`refreshCalls` from the handler and read them on the test thread. This is a data race that makes the assertions nondeterministic under concurrency checking (and would fail strict Swift 6 test compilation).
**Suggestion:** Confine the handler/counters behind a lock or actor, or capture state with an injected per-test recorder.

### [severity: nit] Loopback result pages lack basic document metadata
**File:** Conduit/Services/NativeOAuth.swift:643
**Issue:** The success/failure HTML is a bare fragment with no `lang` or viewport meta; it is rendered by the system browser, but as the only user-visible completion surface it is also the accessibility fallback.
**Suggestion:** Add `<html lang="…">` and a viewport meta (and ideally reuse the localized strings already added) if the page is intended for users.

## Verdict
COMMENT

The core security design is sound (PKCE S256, high-entropy state, state-before-error validation, origin-validated bearer requests, non-replayable mutations, per-dashboard Keychain isolation), and the earlier blocking loopback/presentation/token-shadowing defects are resolved. The remaining items are narrow correctness and hardening gaps, plus a device-validation dependency (loopback reachability, Cloudflare-protected authorize route) that the PR itself acknowledges; none are blockers, but the refresh-error taxonomy and cold-start bridge teardown are worth fixing. No retain cycles or force unwraps were found in the new production code.

<a href="https://opencode.ai/s/thlFrHFu">New%20session%20-%202026-09-14T22%3A58%3A19.811Z</a>
[opencode session](https://opencode.ai/s/thlFrHFu)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34906571766)

lucvan:
Pushed `bf2df018bd0d2e29099ce436a24450ce73fb6a95` with the fixes found during local Xcode and physical-iPhone testing.

The initial phone attempt opened Google sign-in briefly, then dismissed it when Safari/password UI covered the sign-in sheet. The OAuth operation now belongs to a stable presentation coordinator; covering the sheet no longer cancels it. Actual teardown or browser Done still cancels the operation. Three regression tests cover operation lifetime and cancellation.

Validation:

- Xcode 26.3 (17C529), iOS SDK 26.2: simulator build succeeded; focused XCTest suite **146 passed, 0 failed, 0 skipped** on iPhone 17 Pro / iOS 26.3.1.
- Before the final presentation-lifecycle change, the complete unit suite recorded **2523 total: 2522 passed, 0 failed, 1 skipped** (newer CarPlay SDK required). That full run is not a full-suite result for the final lifecycle change.
- Built, installed, and ran on a physical iPhone 16 Pro / iOS 27.0. The tester confirmed Tailscale connectivity and dashboard reachability in device Safari, then connected using the dashboard URL with empty username/password.
- After the fix, sanitized device logs confirmed `listener ready` → `Safari presented` → `callback validated` → `token exchange succeeded` → `ticket minted`. The tester subsequently reported that the app was working well.
- Localization coverage, test-plan validation, and `git diff --check` passed.

The commit also includes the Xcode 26.3 compilation fixes and corrections to the capability fixtures and asynchronous credential-cleanup test. Local Personal Team signing, bundle-ID, entitlements, and generated-project settings were excluded. No dashboard or Google OIDC configuration was changed.

This provides local Xcode/XCTest and primary physical Google OAuth/127.0.0.1 callback evidence. The PR remains draft: upstream macOS CI and dedicated physical cancellation/retry, cold-start, logout/removal, refresh, and deployment/regression scenarios remain to be validated. No credentials, OAuth codes, or tokens are included in this update.

kaishi00:
## Summary
This PR adds a native OAuth/PKCE flow for OAuth-only Hermes dashboards on iOS: URL-only provider discovery, an `NWListener` loopback callback, `/auth/native/{authorize,token,refresh}` with dashboard-scoped Keychain tokens, a bearer-backed transport in `DashboardTicketBridge`, and transactional cookie↔bearer transitions. The core security design is sound and earlier blocking defects (single-chunk loopback parse, embedded Safari, stale-token shadowing) are addressed.

## Findings

### [severity: minor] Concurrent refresh callers still bypass the rejected-refresh error taxonomy
**File:** Conduit/Services/NativeOAuth.swift:429
**Issue:** The creator of `refreshTask` maps a rejected grant (`400/401`) to `DashboardTicketBridgeError.signInRequired` and clears the grant (NativeOAuth.swift:443-455), but a second request that reaches `if let refreshTask { let refreshed = try await refreshTask.value }` rethrows the raw `NativeOAuthError.requestFailed(400/401)` (or a raw `URLError.cancelled`). `AppState` keys re-auth recovery off `error as? DashboardTicketBridgeError, case .signInRequired` (AppState.swift:5716), so the concurrent caller will not run sign-in recovery and will surface an unexpected error. This was flagged in an earlier round and remains unfixed.
**Suggestion:** Funnel the awaited `refreshTask.value` result through the same `do/catch` mapping as the creator path, or extract a shared mapping helper that both paths use.

### [severity: minor] Cold-start native restore failure leaves the live bridge in bearer mode
**File:** Conduit/Services/AppState.swift:2632
**Issue:** `restoreNativeOAuthConnection` calls `mintChatResumeTicket`, which lazily creates `dashboardTicketBridge` with a `NativeOAuthSession` (`usesNativeOAuth == true`). When a rejected refresh clears the Keychain grant, the `catch` only flips `showLogin`/`isConnected`; it neither invalidates nor rebuilds the bridge, and the session's in-memory `tokens` are not reset. Until the next `prepareDashboardBridge`, any request routed through that bridge uses a bearer session whose durable tokens were just cleared.
**Suggestion:** In the failure branch, call `restoreNativeOAuthTokens(KeychainHelper.loadNativeOAuthTokens(dashboardID:), dashboardID:baseURL:)` (or invalidate/nil the bridge) so the stale bearer session is torn down on the same boundary.

### [severity: minor] Bearer/refresh tokens use a weaker Keychain accessibility class than the password they shadow
**File:** Conduit/Services/AppState.swift:15585
**Issue:** Native OAuth tokens are stored with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, while saved passwords use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` (AppState.swift:15615). Bearer/refresh tokens are therefore readable while the device is locked after first unlock. Unless a background-refresh/relay requirement genuinely needs it, this weakens the token store relative to the credential it replaces.
**Suggestion:** Use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` unless a background-authentication requirement is documented.

### [severity: minor] Connection Setup probe still rejects OAuth-only dashboards that LoginView can sign into
**File:** Conduit/Services/ConnectionSetupTest.swift:410
**Issue:** When discovery returns an OAuth provider, the probe only accepts it if `supportsNativeOAuth()` is true; otherwise it falls through to `guard supportsPassword else { .unexpectedServerResponse }`. `LoginView.connect` instead falls back to the browser path for an OAuth-only dashboard when `native_pkce` is absent (`requiresBrowserSignIn = !supportsPassword`). The wizard's "Test Connection" therefore fails for a dashboard the normal login flow can still reach.
**Suggestion:** Route the non-native OAuth-only provider case to `.requiresInteractiveSignIn` (or a browser outcome) to match `LoginView`, or document why the probe intentionally diverges.

### [severity: minor] `supportsNativeOAuth` treats an interactive redirect as a fatal error
**File:** Conduit/Services/NativeAuthClient.swift:278
**Issue:** `guard (200...299).contains(http.statusCode)` throws `providerDiscoveryFailed` for any 3xx. `authProviderDiscovery` treats a 3xx as the interactive sign-in signal, so a dashboard where `/api/auth/providers` is reachable but `/api/status` is redirected to an auth challenge now aborts login with a classified failure instead of falling back to interactive sign-in.
**Suggestion:** Handle a redirect response here consistently with `authProviderDiscovery` (return `false`/interactive) rather than throwing, or explicitly document that `/api/status` is assumed unprotected.

### [severity: minor] OAuth authorize step uses Safari's shared cookie jar
**File:** Conduit/Services/NativeOAuth.swift:908
**Issue:** `SFSafariViewController` is not session-isolated: it reads/writes Safari's persistent cookies, so an existing Google/Safari session is reused and OAuth cookies remain in Safari. `ASWebAuthenticationSession` (with `prefersEphemeralWebBrowserSession`) is the platform-recommended OAuth surface with cleaner dismissal/lifecycle semantics. This is a design/device-validation risk previously raised and not resolved.
**Suggestion:** Move the authorize step to `ASWebAuthenticationSession`, or explicitly validate and document the cookie-sharing behavior on device.

### [severity: nit] Test URLProtocol stub has thread-unsafe shared state
**File:** ConduitTests/NativeOAuthTests.swift:7
**Issue:** `NativeOAuthURLProtocolStub.handler` is a static mutable var set on the test thread and read from URL-loading threads, and handlers mutate captured `apiCalls`/`refreshCalls` (NativeOAuthTests.swift:354-411) that are read back on the test thread. This is a data race and would fail strict Swift 6 test compilation.
**Suggestion:** Confine the handler/counters behind a lock or actor, or inject a per-test recorder.

### [severity: nit] Stale localization entry and unlabeled/under-specified callback UI
**File:** Conduit/Localizable.xcstrings:24; Conduit/Services/NativeOAuth.swift:644, :885
**Issue:** `"Preparing secure sign-in…"` is now orphaned — the final commit removed its only call site when replacing the `ProgressView` with the presentation coordinator (the l10n checker only verifies call sites→catalog, not the reverse). The callback HTML is a bare fragment with no `lang`/viewport and hardcoded English (including `✓ Signed in…`), and `UIActivityIndicatorView` in `SafariPresenterViewController` has no accessibility label.
**Suggestion:** Remove the unused catalog entry; add `lang`/viewport metadata and reuse the localized strings for the completion page; give the spinner an accessibility label.

## Verdict
COMMENT

The core security posture (PKCE S256, high-entropy state + state-before-error validation, origin-validated bearer requests, non-replayable mutation guard, per-dashboard Keychain isolation, dashboard-scoped cookie↔bearer transitions) is sound, and the previously blocking loopback/presentation/token-shadowing issues are resolved. The remaining items are narrow correctness and hardening gaps — chiefly the concurrent refresh error mapping and cold-start bridge teardown — plus the still-pending physical-device/CI validation the PR itself acknowledges.

<a href="https://opencode.ai/s/aO8ZhzgS">New%20session%20-%202026-09-15T00%3A37%3A43.167Z</a>
[opencode session](https://opencode.ai/s/aO8ZhzgS)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34913864868)

kaishi00:
## Summary
Adds a native Hermes OAuth/PKCE sign-in path for iOS (external `SFSafariViewController`, `127.0.0.1` loopback callback, `/auth/native/*` exchange/refresh, per-dashboard Keychain bearer tokens, and a bearer transport in `DashboardTicketBridge`), plus OAuth-only provider discovery and transactional cookie↔bearer transitions.

## Findings

### [severity: major] Native OAuth cold-start restore resets the active profile to `default`
**File:** Conduit/Services/AppState.swift:2622-2625
**Issue:** `restoreNativeOAuthConnection` calls `connect(with: HermesConnection(...))`, which defaults `profile` to `"default"`. `connect` then calls `setActiveProfile(profile)` and `defaults.set(profile, forKey: activeProfileKey)` (AppState.swift:2722-2725), so on every cold launch of a native-OAuth dashboard whose persisted `activeProfile` is not `"default"`, the workspace silently switches to `default` and that switch is persisted. The sibling restore paths already do the right thing: `restoreSavedConnection` (AppState.swift:3429) and `restoreSavedCredentials` (AppState.swift:3476) both pass `profile: activeProfile`. `activeProfile` is restored at AppState.swift:1509 before `loadSavedConnection()` runs, so the correct value is available.
**Suggestion:** Pass `profile: activeProfile`: `await connect(with: HermesConnection(baseUrl: baseURL, ticket: ticket), profile: activeProfile)`.

### [severity: major] Bridge-reuse guard ignores rotated bearer tokens, so a stale session can clear fresh grants
**File:** Conduit/Services/AppState.swift:3517-3531 (effect at Conduit/Services/DashboardTicketBridge.swift:327-331, 391-393)
**Issue:** `prepareDashboardBridge` keeps the existing bridge whenever `usesNativeOAuth == shouldUseNativeOAuth` — a boolean. `NativeOAuthSession` captures its `tokens` once at construction (`NativeOAuthSession.init?(baseURL:dashboardID:cloudflareAccess:)` at NativeOAuth.swift:336-341) and only updates them through its own `refresh()`. When a repair (`performConnectionRepair` `.nativeOAuth`, AppState.swift:2932-2953) or a re-auth saves a newly rotated token set for a dashboard that is already in native mode, the guard sees `true == true` and reuses the bridge with the old in-memory session. If the fresh grant rotated/invalidated the old refresh token, the stale session's next refresh fails and `NativeOAuthSession.refresh` unconditionally clears the Keychain entry (NativeOAuth.swift:448-451) — which now holds the *new* tokens — signing the user out immediately after a successful sign-in. `restoreNativeOAuthTokens` (AppState.swift:2986) rebuilds the bridge for rollback, but the success path relies solely on this insufficient guard.
**Suggestion:** Compare the stored token value, not just presence. Expose a comparable snapshot of the `NativeOAuthTokenSet` captured at session construction (e.g. `nativeOAuthTokenSnapshot`) and rebuild when `shouldUseNativeOAuth && bridge.nativeOAuthTokenSnapshot != storedNativeOAuthTokens`.

### [severity: minor] Concurrent refresh callers bypass the rejected-refresh error mapping
**File:** Conduit/Services/NativeOAuth.swift:429-434 (vs 443-455)
**Issue:** Only the task *creator* maps a rejected grant to `DashboardTicketBridgeError.signInRequired` (and clears the grant). A second caller that reaches `if let refreshTask { let refreshed = try await refreshTask.value }` rethrows the raw `NativeOAuthError.requestFailed(400/401)` (or a raw `URLError.cancelled`). `AppState` keys sign-in recovery off `error as? DashboardTicketBridgeError, case .signInRequired` (AppState.swift:5716), so the second caller skips recovery and can surface an unexpected error.
**Suggestion:** Funnel the awaited `refreshTask.value` through the same `do/catch` mapping (or extract a shared mapping helper) so both paths produce `signInRequired`/`CancellationError` consistently.

### [severity: minor] Cold-start restore failure leaves the live bridge in bearer mode with cleared tokens
**File:** Conduit/Services/AppState.swift:2632-2637
**Issue:** `mintChatResumeTicket` lazily builds `dashboardTicketBridge` with a `NativeOAuthSession` (AppState.swift:5840-5846). When a rejected refresh clears the grant, the `catch` only sets `showLogin`/`isConnected`; it never invalidates/rebuilds the bridge, and the session's in-memory `tokens` are not reset. Any request routed through that bridge until the next `prepareDashboardBridge` uses a bearer session whose durable grant was just cleared. The repair paths already call `restoreNativeOAuthTokens`; this path does not.
**Suggestion:** In the failure branch, call `restoreNativeOAuthTokens(KeychainHelper.loadNativeOAuthTokens(dashboardID:), dashboardID:baseURL:)` (or invalidate/nil the bridge) on the same boundary.

### [severity: minor] `supportsNativeOAuth()` treats an auth redirect as fatal, unlike provider discovery
**File:** Conduit/Services/NativeAuthClient.swift:269-283 (compare 214-239)
**Issue:** `authProviderDiscovery` classifies 3xx-with-Location as `.interactiveSignInRequired`, but `supportsNativeOAuth` throws `providerDiscoveryFailed` for any non-2xx. For a dashboard where `/api/auth/providers` is reachable but `/api/status` is redirected to an interactive auth challenge, `LoginView.connect` (LoginView.swift:558-559) and `ConnectionSetupProbe` (ConnectionSetupTest.swift:412-420) abort with a classified failure rather than falling back to the browser path they support. This is an intentional "don't silently downgrade" choice, but it diverges from discovery and can make an otherwise reachable dashboard unreachable.
**Suggestion:** Handle the redirect consistently with `authProviderDiscovery` (return `false`/interactive) or explicitly document that `/api/status` is required to be unprotected.

### [severity: minor] Native bearer/refresh tokens use a weaker Keychain class than the password they shadow
**File:** Conduit/Services/AppState.swift:15580-15586
**Issue:** `saveNativeOAuthTokens` uses `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, while saved passwords use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` (AppState.swift:15615). The bearer/refresh token — which grants full API access — is therefore readable while the device is locked after first unlock.
**Suggestion:** Use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` unless a documented background-refresh requirement needs earlier access; if so, add that rationale near the call.

### [severity: minor] Authorize step uses Safari's shared cookie jar
**File:** Conduit/Services/NativeOAuth.swift:867-918
**Issue:** `SFSafariViewController` is not session-isolated — it reads/writes Safari's persistent cookies, so an existing Google/Safari session is reused and OAuth cookies persist in Safari. `ASWebAuthenticationSession` (with `prefersEphemeralWebBrowserSession`) is the platform-recommended OAuth surface and gives cleaner dismissal/lifecycle semantics. This is a design/device-validation risk (the PR premise is avoiding a disallowed embedded user agent).
**Suggestion:** Move the authorize step to `ASWebAuthenticationSession`, or explicitly validate and document the cookie-sharing behavior on device.

### [severity: nit] Orphaned localization entry
**File:** Conduit/Localizable.xcstrings:24
**Issue:** `"Preparing secure sign-in…"` was added in this PR but its only call site was removed when the `ProgressView` was replaced by the presentation coordinator. The l10n checker only validates call-site→catalog, so this dead entry passes.
**Suggestion:** Remove the unused catalog entry.

### [severity: nit] Callback page and spinner lack accessibility metadata
**File:** Conduit/Services/NativeOAuth.swift:644, 885
**Issue:** The loopback success/failure HTML is a bare fragment with no `lang`/viewport (and hardcoded English), and the `UIActivityIndicatorView` in `SafariPresenterViewController` has no accessibility label. The completion page is the only user-visible surface after the browser callback.
**Suggestion:** Add `<html lang=…>`/viewport meta (reusing the localized strings) and give the spinner an accessibility label.

### [severity: nit] URLProtocol test stub uses thread-unsafe shared state
**File:** ConduitTests/NativeOAuthTests.swift:7, 351-411
**Issue:** `NativeOAuthURLProtocolStub.handler` is a static mutable `var` set on the test thread and invoked from URL-loading threads, and handlers mutate captured `apiCalls`/`refreshCalls` that are read back on the test thread. This is a data race that can make assertions nondeterministic and would fail strict Swift 6 test compilation.
**Suggestion:** Confine the handler/counters behind a lock or actor, or inject a per-test recorder.

## Verdict
REQUEST_CHANGES

The security core (PKCE S256, high-entropy state + state-before-error validation, origin-validated bearer requests, non-replayable mutation guard, per-dashboard Keychain isolation) is sound, and no force unwraps or retain cycles were found in the new production code. However, the cold-start profile reset and the stale-token bridge-reuse path are user-visible correctness bugs that should be fixed before merge.

<a href="https://opencode.ai/s/xnMUYRk6">New%20session%20-%202026-09-15T08%3A59%3A55.482Z</a>
[opencode session](https://opencode.ai/s/xnMUYRk6)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34949940402)

lucvan:
@coderabbitai The fixes for both review findings are now in commit `2fd0a9e97f827b2c270ee7bb5a59bae8e306ca08`. GitHub is returning server errors for in-thread replies, so I am linking each finding here.

### [Preserve the active profile on cold start](https://github.com/kaishi00/hermes-conduit/pull/173#discussion_r4010949559)

Fixed in [2fd0a9e](https://github.com/lucvan/hermes-conduit/commit/2fd0a9e97f827b2c270ee7bb5a59bae8e306ca08). `restoreNativeOAuthConnection` now passes `profile: activeProfile` to `connect(with:)`. The cold-start regression starts with a non-default profile and verifies it survives successful native OAuth restoration.

Validation on Xcode 26.3 / iPhone 17 Pro simulator (iOS 26.3.1): full unit suite **2528 passed, 0 failed, 1 skipped**. The skip is the newer CarPlay API; this local run used external SDK-compatibility source copies, which are not included in the commit. Physical-device retesting of these latest changes is still pending.

### [Replace the bridge when the native grant changes](https://github.com/kaishi00/hermes-conduit/pull/173#discussion_r4010949569)

Fixed in [2fd0a9e](https://github.com/lucvan/hermes-conduit/commit/2fd0a9e97f827b2c270ee7bb5a59bae8e306ca08). Bridge reuse now compares the stored grant with the session’s **current live grant**, using boolean comparison helpers rather than exposing tokens. A same-mode replacement invalidates and rebuilds the bridge; an unchanged grant still reuses it. Internal refresh updates the compared live grant too, so a successful refresh does not leave a stale construction-time snapshot.

Regression coverage verifies unchanged-grant reuse, same-mode replacement/reconstruction, and live-grant matching after refresh. The same commit also normalizes refresh failures inside the shared refresh task for all concurrent callers and discards the stale bridge after a failed cold-start restoration.

Validation on Xcode 26.3 / iPhone 17 Pro simulator (iOS 26.3.1): full unit suite **2528 passed, 0 failed, 1 skipped**. The skip is the newer CarPlay API; this local run used external SDK-compatibility source copies, which are not included in the commit. Physical-device retesting of these latest changes is still pending.

kaishi00:
## Summary

Adds a native Hermes OAuth/PKCE broker flow for iOS (external `SFSafariViewController`, `127.0.0.1` loopback callback, `/auth/native/*` exchange/refresh, per-dashboard Keychain bearer tokens, bearer transport in `DashboardTicketBridge`) plus URL-only provider discovery and transactional cookie↔bearer transitions. The two previously blocking integration bugs (cold-start profile reset, stale-grant bridge reuse) are fixed in `2fd0a9e`.

## Findings

### [severity: minor] Bearer/refresh tokens stored with a weaker Keychain class than the password they replace
**File:** Conduit/Services/AppState.swift:15589
**Issue:** `saveNativeOAuthTokens` uses `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, while saved passwords use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` (AppState.swift:15623). The bearer + refresh token grants full API access and is now readable while the device is locked after first unlock. This was raised in earlier rounds and remains; no background-refresh requirement is documented.
**Suggestion:** Use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` unless a concrete background-auth need exists, in which case document the rationale next to the call.

### [severity: minor] `supportsNativeOAuth()` treats any non-2xx as fatal, unlike provider discovery
**File:** Conduit/Services/NativeAuthClient.swift:278
**Issue:** `authProviderDiscovery` classifies a 3xx-with-Location as `.interactiveSignInRequired`, but `supportsNativeOAuth` throws `providerDiscoveryFailed` for any non-2xx. On a dashboard whose `/api/auth/providers` is reachable but `/api/status` is redirected to an auth challenge, `LoginView.connect` (LoginView.swift:559) aborts with a classified failure instead of the supported browser fallback. The "don't silently downgrade" intent is defensible, but it diverges from discovery and can make an otherwise reachable dashboard unreachable.
**Suggestion:** Treat a 3xx consistently with discovery (return `false`/interactive), or explicitly document that `/api/status` is assumed unprotected.

### [severity: minor] Connection Setup probe rejects an OAuth-only dashboard that `LoginView` can still reach
**File:** Conduit/Services/ConnectionSetupTest.swift:413-436
**Issue:** When discovery returns an OAuth provider but `supportsNativeOAuth()` is false, the probe falls through to `guard supportsPassword` and reports `.unexpectedServerResponse`. `LoginView.connect` instead sets `requiresBrowserSignIn = true` (LoginView.swift:563) for the same shape, so "Test Connection" fails for a dashboard the normal login path can still sign into. `ConnectionSetupView.beginRepairSignIn` also falls back to `.browser` here, so the probe is stricter than both callers.
**Suggestion:** Route this case to `.requiresInteractiveSignIn` (matching `LoginView`/repair), or document why the probe intentionally requires a password provider.

### [severity: minor] Authorize step uses Safari's shared cookie jar
**File:** Conduit/Services/NativeOAuth.swift:920
**Issue:** `SFSafariViewController` is not session-isolated — it reads/writes Safari's persistent cookies, so an existing Google/Safari session is reused and OAuth cookies persist in Safari. `ASWebAuthenticationSession` with `prefersEphemeralWebBrowserSession` is the platform-recommended OAuth surface and gives cleaner dismissal/lifecycle semantics. Since the PR's premise is avoiding a disallowed embedded user agent, this is worth resolving or explicitly validating on-device.
**Suggestion:** Move the authorize step to `ASWebAuthenticationSession`, or document/validate the cookie-sharing behavior.

### [severity: nit] Orphaned localization entry
**File:** Conduit/Localizable.xcstrings:24
**Issue:** `"Preparing secure sign-in…"` has no remaining call site (it was dropped when the `ProgressView` was replaced by the presentation coordinator). The l10n checker validates call-site→catalog, not the reverse, so it passes.
**Suggestion:** Remove the unused catalog entry.

### [severity: nit] Callback page and spinner lack accessibility metadata
**File:** Conduit/Services/NativeOAuth.swift:653, Conduit/Services/NativeOAuth.swift:894
**Issue:** The success/failure HTML is a bare fragment with no `lang`/viewport and hardcoded English, and the `UIActivityIndicatorView` has no accessibility label. The completion page is the only user-visible surface after the browser callback.
**Suggestion:** Add `<html lang="…">` and a viewport meta (reusing the localized strings) and set an accessibility label on the spinner.

### [severity: nit] Test URLProtocol stub uses unsynchronized shared state
**File:** ConduitTests/NativeOAuthTests.swift:7
**Issue:** `NativeOAuthURLProtocolStub.handler` is a `static var` set on the test thread and invoked from URL-loading threads, and handlers mutate captured `apiCalls`/`refreshCalls` read back on the test thread (e.g. lines 354-381, 432-453). This is a data race and would fail strict Swift 6 test compilation.
**Suggestion:** Confine the handler/counters behind a lock or actor, or inject a per-test recorder.

## Verdict

COMMENT

No force unwraps or retain cycles were found in the new production code, and the core security design (PKCE S256, high-entropy state with state-before-error validation, origin-validated bearer requests, non-replayable mutation guard, per-dashboard Keychain isolation, transactional cookie↔bearer transitions) is sound. Remaining items are hardening/consistency nits plus the still-pending physical-device and CI validation the PR itself acknowledges.

<a href="https://opencode.ai/s/L902QvNR">New%20session%20-%202026-09-15T09%3A29%3A17.567Z</a>
[opencode session](https://opencode.ai/s/L902QvNR)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34952485658)

lucvan:
@coderabbitai Physical-device retesting has now taken place for `2fd0a9e97f827b2c270ee7bb5a59bae8e306ca08`, following a fresh build/install on an iPhone 16 Pro running iOS 27.0.

The tester confirmed:

- Tailscale connectivity and dashboard reachability in device Safari.
- A non-default profile and the same chat survived force-quit and Home Screen relaunch.
- Sign Out of This Dashboard returned to sign-in and required authorization again.
- Cancelling Google login returned cleanly without connecting or showing an error.
- Retrying with empty username/password completed Google sign-in and restored a working dashboard/chat.

The successful retry produced these sanitized stage labels:

```text
OAuth stage: listener ready
OAuth stage: Safari presented
OAuth stage: callback validated
OAuth stage: token exchange succeeded
OAuth stage: ticket minted
WebSocket handshake completed
OAuth stage: sign-in view removed
```

The cancelled attempt separately logged `user cancelled browser`, `cancelled`, and `sign-in view removed`, without token exchange or ticket minting. This establishes the real loopback callback and the requested physical lifecycle checks for the latest OAuth changes. The latest full simulator suite also passed: **2528 passed, 0 failed, 1 skipped**.

Local Personal Team Debug signing and an external CarPlay source compatibility copy were needed for Xcode 26.3; neither is included in the PR. These results do not claim a byte-identical release build or replace upstream CI/human approval. Real token-expiry/rotation and broader deployment regressions remain outside this physical run.

The PR description has been updated. To protect confidential chat data, this update contains only curated test outcomes and fixed stage labels. No raw device/Xcode logs, chat messages/titles, session identifiers, private dashboard addresses, cookies, codes, tokens, or credentials are being shared.

lucvan:
Any chance a maintainer could review this please?

kaishi00:
Yea, sorry I'm swamped. I didn't forget about this.

kaishi00:
## Summary
Adds a native Hermes OAuth/PKCE sign-in path for iOS OAuth-only dashboards: URL-only provider discovery, an `NWListener` loopback callback, brokered token exchange/refresh with per-dashboard Keychain storage, and bearer-backed REST/WebSocket-ticket transport that coexists with the existing cookie/WebKit path.

## Findings

### [severity: minor] Bearer/refresh tokens use a weaker Keychain accessibility class than the password they shadow
**File:** Conduit/Services/AppState.swift:17063
**Issue:** `saveNativeOAuthTokens` stores the grant with `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`, while saved passwords use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` (AppState.swift:15615). The bearer + refresh token grants full dashboard API access and is therefore readable while the device is locked after the first unlock. No background-refresh requirement for these tokens is documented in the PR.
**Suggestion:** Use `kSecAttrAccessibleWhenUnlockedThisDeviceOnly` unless a concrete background-auth need exists; if one does, document the rationale next to the call.

### [severity: minor] `supportsNativeOAuth()` treats any non-2xx as fatal, diverging from provider discovery
**File:** Conduit/Services/NativeAuthClient.swift:278
**Issue:** `authProviderDiscovery` classifies a 3xx-with-Location as `.interactiveSignInRequired` (NativeAuthClient.swift:243-249), but `supportsNativeOAuth` throws `providerDiscoveryFailed` for any non-2xx. On a dashboard whose `/api/auth/providers` is reachable but `/api/status` is redirected to an auth challenge, `LoginView.connect` aborts with a classified failure instead of using the browser fallback it supports. The "don't silently downgrade into embedded OAuth" intent is reasonable, but the inconsistent handling makes an otherwise reachable dashboard unreachable.
**Suggestion:** Treat a 3xx consistently with discovery (return `false`/interactive), or explicitly document that `/api/status` must be unprotected for password/browser login.

### [severity: minor] Connection Setup probe rejects OAuth-only dashboards that `LoginView` can still reach
**File:** Conduit/Services/ConnectionSetupTest.swift:413-436
**Issue:** When discovery returns an OAuth provider but `supportsNativeOAuth()` is false, the probe falls through to `guard supportsPassword` and reports `.unexpectedServerResponse`. `LoginView.connect` instead sets `requiresBrowserSignIn = true` for the same shape (LoginView.swift:562-564), and `ConnectionSetupView.beginRepairSignIn` falls back to `.browser` (ConnectionSetupView.swift:294-296). "Test Connection" therefore fails for a legacy OAuth-only dashboard the normal login flow can still reach.
**Suggestion:** Route the non-native OAuth-only provider case to `.requiresInteractiveSignIn` (or a browser outcome) to match the login and repair paths, or document why the probe intentionally diverges.

### [severity: minor] Loopback completion page and spinner lack accessibility/basic document metadata
**File:** Conduit/Services/NativeOAuth.swift:653, Conduit/Services/NativeOAuth.swift:894
**Issue:** The success/failure HTML is a bare fragment with no `lang` or viewport meta and hardcoded English, even though it is the only user-visible confirmation surface after the browser callback. The `UIActivityIndicatorView` in `SafariPresenterViewController` also has no accessibility label.
**Suggestion:** Add `<html lang="…">` and a viewport meta (reusing `AppLocalization` strings for the visible message) and set an accessibility label on the spinner.

### [severity: nit] Orphaned localization entry
**File:** Conduit/Localizable.xcstrings:24
**Issue:** `"Preparing secure sign-in…"` was added by this PR but its only call site was dropped when the `ProgressView` was replaced by the presentation coordinator. The l10n checker validates call-site→catalog only, so the dead entry passes.
**Suggestion:** Remove the unused catalog entry.

### [severity: nit] Test URLProtocol stub uses unsynchronized shared state
**File:** ConduitTests/NativeOAuthTests.swift:7
**Issue:** `NativeOAuthURLProtocolStub.handler` is a static mutable `var` written on the test thread and read from URL-loading threads, and handlers mutate captured `apiCalls`/`refreshCalls` that are read back on the test thread (e.g. lines 354-381, 432-453). This is a data race that can make assertions nondeterministic and would fail strict Swift 6 concurrency checking.
**Suggestion:** Confine the handler and counters behind a lock/actor, or inject a per-test recorder instead of static state.

### [severity: nit] Loopback accumulation path is only unit-tested at the accumulator level
**File:** ConduitTests/NativeOAuthTests.swift:208
**Issue:** The bounded split-header accumulator is covered directly, but no test drives `readRequest`/`receiveChunk` with a real request split across multiple TCP segments, which was the source of the earlier blocking parse defect. The partial-request test (line 237) only exercises cancellation of a withheld terminator.
**Suggestion:** Add a loopback test that sends a valid callback in two segments to assert accumulation across `receiveChunk` boundaries.

### [severity: nit] Native cold-start failure leaves `turnState` at `.synchronizing`
**File:** Conduit/Services/AppState.swift:2775-2790
**Issue:** `restoreNativeOAuthConnection` sets `turnState = .synchronizing` before the task and, on failure, resets `showLogin`/`isConnecting`/`isConnected` but never resets `turnState`, unlike the no-reusable-auth path that sets `.idle` (AppState.swift:3454).
**Suggestion:** Reset `turnState = .idle` (or restore the prior value) in the catch before presenting the login surface.

## Verdict
COMMENT

No force unwraps, implicitly unwrapped optionals, or retain cycles were found in the new production code; the PKCE S256, high-entropy state with state-before-error validation, origin-validated bearer requests, non-replayable mutation guard, dashboard-scoped Keychain isolation, and transactional cookie↔bearer transitions are sound. The remaining items are hardening/consistency/accessibility gaps plus the physical-device and CI validation the PR itself still lists as open.

<a href="https://opencode.ai/s/ZBCrmEOo">New%20session%20-%202026-09-19T12%3A33%3A38.863Z</a>
[opencode session](https://opencode.ai/s/ZBCrmEOo)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35443151311)

PR #171, opened 2026-09-14
ci: classify and recover the hosted CoreAudio wedge (host-level, every unit lane)

kaishi00:
# CI: classify and recover the hosted CoreAudio wedge (host-level, every unit lane)

> **REDESIGN (second round).** The first version of this PR gated wedge
> recovery on a two-class "audio-sensitive inventory" and reserved those
> classes into a dedicated `unit-audio` lane, under the assumption that the
> wedge was class-local. That assumption was wrong and this PR's own CI
> disproved it: run 34885607864 saw `Unit - unit-audio` pass on its own
> healthy runner while `Unit - unit-2` stalled on a poisoned one (178
> AURemoteIO -10851 lines, 39 HALC overload skips, watchdog timeout, zero
> failure records, 16 minutes of pointless isolation). The wedge is a
> RUNNER failure: every matrix lane gets its own Mac, so a class-based
> gate cannot contain it. This round removes the inventory and the lane
> reservation and puts the host-health classifier in EVERY unit lane,
> including the timeout path.

## Root cause / problem

A broken GitHub-hosted macOS audio host does not fail loudly. In run
34822043959 (post-#169 main, retried by hand **8 times over ~5.5 hours**), every
attempt showed the same signature — **184–186 `AURemoteIO.cpp:1135 failed:
-10851` lines plus `[AMCP] HALC_ProxyIOContext::IOWorkLoop: skipping cycle due
to overload` bursts (48 vs ≤7 healthy)** — while the *failing tests moved
between attempts* (`AppStateVoiceSuspensionTests` assertion failures at four
different tests, a `CarPlayVoiceCoordinatorTests` isolation hang, partial
transcripts). The wedge starves async state machines host-wide; it does not
produce deterministic product failures. Because the affected invocations exit 65
with ordinary assertion records, the lane runner classified every attempt as a
product test failure and a human re-ran the lane by hand.

Calibration data (full-lane invocation logs):

| lane state                        | AURemoteIO -10851 | HALC overload skips |
|-----------------------------------|------------------:|--------------------:|
| healthy unit-1 (green)            | ~8                | ~0–2                |
| healthy unit-2 (green, 2 runs)    | 107–116           | 4–7                 |
| slow-timeout unit-1 (red, NOT audio) | 48             | 4                   |
| wedged unit-2 (8 attempts)        | 184–186           | 48                  |

## The host-health classifier (fail closed)

`scripts/classify-coreaudio-wedge.py` is a pure host-health classifier: the
[...]

PR #176, opened 2026-09-15
ci(unit): recover a watchdog-stalled lane once on a fresh hosted runner

kaishi00:
## Goal

Smallest possible CI mechanism for the residual GitHub-hosted macOS runner stalls demonstrated by PR #175: a unit lane's **primary** invocation watchdogs, the same-runner isolation passes classes individually until the budget exhausts, and the lane finishes red with **zero identified test failures**. Same-runner retries cannot fix this — every invocation inside one job stays on the same hosted Mac — so the retry is a **second GitHub Actions job** (`runs-on: macos-26`), which GitHub allocates on a **new** runner.

Evidence anchoring the predicate (downloaded from #175's failed runs): `lane-result.json` with `status=timeout`, `attempts=[{mode:lane,n:1,status:timeout},{mode:isolation,status:incomplete}]`, `hung_class=null`, `failures=null` (the watchdog killed xcodebuild mid-write; extraction failed safely), isolation classes passing individually. No audio/hardware classifier, no test allowlists.

## Architecture

```
plan → build → unit matrix
                 ↓ (always)
             unit-recovery-plan (ubuntu): classify each lane from lane-result.json
             vs the authoritative plan; emit recovery matrix from the PLAN's
             original lane inputs (empty = healthy)
                 ↓ (only affected lanes)
             unit-recovery (macos-26 matrix, NEW runner per stalled lane):
             rerun the exact lane once, marked --fresh-runner-recovery
                 ↓ (always)
             CI Gate: per-lane adjudication   →   report / timing history
```

## Recovery-eligibility predicate

All must hold (else fail closed):
1. unit lane; 2. the **primary** full-lane invocation hit its watchdog (attempt chain `mode:lane → timeout`); 3. zero failing test methods known anywhere in the lane metadata (failure records, `test-failures` attempt status, isolation class failures — a watchdog-killed extraction with no readable result counts as *nothing identified*, which is exactly the demonstrated shape); 4. the lane result is complete and self-consistent (embedded classification == re-derived classification; document agrees with the plan on kind/membership/target/watchdog/schema).

Never eligible: ordinary test failures; unclassifiable failures; infra errors that already burned their same-runner attempt-2 (or a retry-timeout whose primary never watchdoged); any fresh-runner retry (anti-recursion).

## CI Gate (branch protection context unchanged: **CI Gate**)

`needs.unit.result` is no longer the unit truth. The gate adjudicates every planned lane from `lane-unit-*` + `lane-recovery-*` artifacts against `plan.json`:

| Original lane | Recovery | Final |
|---|---|---|
| PASS | — | accepted |
| recoverable stall | exactly one fresh-runner PASS | accepted — *recovered infrastructure PASS* |
| recoverable stall | missing / fail / timeout | **FAIL** |
| test failure | (never requested) | **FAIL** |
| unclassifiable / malformed / plan disagreement | — | **FAIL** (fail closed) |
| fresh-runner result for unplanned lane / disagreeing with plan / without embedded classification | — | **FAIL** |

`unit` may legitimately report job-level `failure` while the run is green (recovered stall); `cancelled`/`skipped` still fail. `unit-recovery-plan` must be `success` (it runs `if: always()`); `unit-recovery` may be `skipped` (empty matrix). Every planned lane gets exactly one disposition; ambiguous duplicates (same `finished_at`, differing docs) fail closed. Re-running a **failed** recovery leg follows the repo's usual re-run policy: the newest attempt's artifact supersedes (lane artifacts carry `-attempt-N` so this is deterministic).

## Anti-recursion

The recovery job passes `--fresh-runner-recovery` to `ci-test-lane.sh` (exactly once in the workflow, pinned by a contract test). The lane result is marked `fresh_runner_recovery: true` and its classification can never come out recovery-eligible again — a retry that stalls or fails is final. There is no third attempt.

## Timing history

Main-only update now runs behind the same per-lane adjudication (`ci_lane_recovery.py adjudicate` preflight): a recovered main run still updates history; a genuinely red run never does. Per lane, only **clean primary passes** merge — the stalled invocation's partial timings and the recovery run's timings are excluded in this first version.

## Report

The CI Test Report gains a Final column and a dedicated *Fresh-runner recovery* section rendering each affected lane as `original / fresh runner / final`, e.g. (live from this PR's runs):

```
- unit-2
  - original: recoverable-timeout-zero-failures, zero XCTest failures identified (...)
  - fresh runner: timeout (fresh-runner retry stalled again with zero identified test failures - no further attempt is permitted)
  - final: FAIL
```

## Verification

- `python3 -m unittest discover -s scripts/tests`: **236 tests OK** (was 160; +76 covering the required scenarios: all-lanes-pass→no matrix, test-failure→not eligible, watchdog-zero-failures→matrix contains exactly that lane, recovery pass/fail/timeout/missing/unplanned/ambiguous-duplicate, malformed/missing/tampered metadata, multiple recoverable lanes, non-recursive marking, UI/build/plan/self-test semantics unchanged, recovered-run-not-rejected)
- `bash scripts/tests/test_lane_runner.sh` on the Mac build box: **148 passed / 0 failed** (incl. 3 new cases: primary stall classifies recoverable, fresh-runner retry marked non-recursive, green-lane classification)
- `bash scripts/tests/test_simulator_destination.sh`: **36/36**
- `python3 scripts/plan-tests.py validate`: OK
- Multi-model review gate: 2× APPROVE + 1× REQUEST CHANGES (majority approve), no blockers; all findings fixed and re-verified. CodeRabbit: 2 inline findings (shared plan-agreement fence, non-dict plan-entry guard) — both fixed.

## LIVE EXERCISE on this PR (2026-09-15) — the mechanism worked at every decision point

The very first CI run of this branch hit the real failure domain and every rule fired as designed, repeatedly:

1. **Run 1:** unit-2's primary stalled with the exact #175 shape → classified `recoverable-timeout-zero-failures` → a real **fresh-runner recovery job spawned** on another hosted Mac. That recovery ALSO stalled (fleet-wide window); isolation then surfaced a REAL assertion failure (`AppStateVoiceSuspensionTests.testSpokenGoodbyeBeforeBackgroundPreventsRestoration` — a pre-existing main-branch test, not #175's) → correctly classified `test-failure` → **gate RED (correct: recovery has XCTest failures).**
2. **Re-run 1:** unit-2 stalled again → recoverable → recovery stalled again → `nonrecoverable-infrastructure` (anti-recursion held, no third attempt) → **gate RED (correct).**
3. **Final head run + Re-run 2:** identical pair of stall→recovery-stall cycles → **gate RED (correct).**

Across ~12 hours, unit-2's full-lane invocation stalled on **10+ different hosted runners** while unit-1 (identical predicted load, 372.7s vs 372.6s, same 932s watchdog) and all UI lanes passed every time. unit-2 carries 19 voice/CarPlay/audio classes vs unit-1's 12 — the strongest remaining signal for the underlying hosted-fleet problem (consistent with PR #171's audio-wedge domain; this PR deliberately does NOT classify by audio signature). The persistent red gate is the mechanism honestly refusing to mask a real, still-broken environment: per spec, *"fresh-runner retry watchdogs → RED, no third attempt."* Once the fleet (or the lane-composition issue it exposes) recovers, the same pipeline demonstrates the green `recovered infrastructure PASS` path without any further changes — that path is covered by the regression suites and the live gate/report adjudication.

## Scope

CI tooling only — no production Swift, no test assertions touched, nothing skipped, no `continue-on-error` masking, no AURemoteIO/HALC thresholds, no general flaky retries. PRs #171 and #175 untouched. **Do not merge.**


kaishi00:
## Summary
This PR adds a second, bounded, fresh-runner GitHub Actions job that re-runs a unit lane once when its primary invocation watchdogged with zero identified test failures, and moves the unit verdict from the raw `needs.unit.result` to per-lane artifact adjudication (`ci_lane_recovery.py`) consumed by `ci-gate.py`, the report, and timing history. No production Swift or test assertions change, so the Swift-specific lenses (retain cycles, MainActor, SwiftUI lifecycle, accessibility) are not applicable; the review below covers the CI tooling.

## Findings

### [minor] Unguarded `.get()` crashes adjudication on a malformed plan
**File:** scripts/ci_lane_recovery.py:391
**Issue:** The first plan pass explicitly tolerates a non-dict `unit_lanes` entry (`lane_entry.get("lane") if isinstance(lane_entry, dict) else None` and records "plan contains a unit lane without a name"), but the second pass at line 391 calls `lane_entry.get("lane")` unconditionally. A plan whose `unit_lanes` list contains a scalar raises `AttributeError`, which is not caught by `ci-gate.py` (catches only `OSError/ValueError/JSONDecodeError`) or by `_cmd_adjudicate`, so the job dies with a traceback instead of the precise fail-closed message the module documents.
**Suggestion:** Change the second loop to `lane = lane_entry.get("lane") if isinstance(lane_entry, dict) else None` and `continue` (or reuse the pre-validated `planned_lanes` list).

### [minor] Dynamic recovery matrix can break workflow evaluation on a fail-closed plan job
**File:** .github/workflows/ci.yml:373 (with `if` at 364-368)
**Issue:** `unit-recovery` is gated with `always() && ...` and takes `matrix: ${{ fromJSON(needs.unit-recovery-plan.outputs.matrix) }}`. If `ci-recovery-plan.py` fails closed, `steps.recovery.outputs.matrix` is never written, so the job output is the empty string. Because `always()` forces the job to be considered, GitHub may evaluate the `fromJSON('')` matrix expression even though the `if` result check is false, producing a workflow-level expression error rather than the intended clean fail-closed gate red. The primary `unit` job avoids this only because it has no `always()` (so a failed `plan` skips it before matrix expansion).
**Suggestion:** Give the output a safe default, e.g. `matrix: ${{ steps.recovery.outputs.matrix || '{\"include\":[]}' }}`, or use `fromJSON(needs.unit-recovery-plan.outputs.matrix || '{\"include\":[]}')`, and/or write an empty `{"include":[]}` matrix file in a `trap`/always step so the output is never empty.

### [minor] Recovery matrix builder can raise on non-numeric estimates instead of failing with a message
**File:** scripts/ci-recovery-plan.py:105-106
**Issue:** `missing` only checks that each class is a key in `plan["estimates"]`; it doesn't validate the value. `"{0}={1:.1f}".format(c, estimates[c])` raises `ValueError`/`TypeError` for a non-numeric estimate, and `entry["classes"]`/`entry["lane"]` are accessed without guarding. This produces an uncaught traceback, contradicting the module's stated "fails closed with a precise message instead of a traceback" contract (the `try/except ValueError` in `main` does not cover `TypeError` from `.format`).
**Suggestion:** Validate/coerce estimate values (wrap the `.format` in the existing `try` and convert with `float()`, reporting the offending class), and use `.get("classes", [])`.

### [minor] A positively identified hang is still eligible for fresh-runner recovery
**File:** scripts/ci_lane_recovery.py:181-196
**Issue:** The `status == "timeout"` branch returns `CLASS_RECOVERABLE_TIMEOUT_ZERO_FAILURES` whenever the primary watchdogged and no test failures are known, without checking `hung_class`. An isolation run that positively pinned a hung class (via the `run_isolation` break path) therefore still schedules a recovery retry. This cannot produce a false green (a fresh-runner re-hang is classified `nonrecoverable-infrastructure`), but it burns a hosted Mac for a case that is not the demonstrated "zero identified failures" stall shape.
**Suggestion:** If intentional, document it; otherwise treat a non-null `hung_class` on the primary as nonrecoverable, matching the "zero identified failures" predicate.

### [nit] Recovery dedupe orders by `finished_at` and ignores the attempt suffix
**File:** scripts/ci_lane_recovery.py:274-324
**Issue:** Records are sorted by `(artifact_name, path)` but the winner is chosen with `max(key=_finished_at_key)`; ties fall back to the first element of the ascending artifact-name sort, i.e. the *older* attempt. `now_iso()` has only second granularity. The equal-stamp/differing-doc check converts genuine ties into a fail-closed conflict, so this isn't a wrong-winner bug, but a re-run whose two attempts land in the same second would fail CI even though the artifact name encodes the ordering.
**Suggestion:** Prefer the parsed `attempt` number as the primary sort key, with `finished_at` as tie-break, and only call a tie ambiguous when both attempt and timestamp agree but docs differ.

### [nit] Timing preflight scans the same directory twice
**File:** .github/workflows/ci.yml:679-680
**Issue:** `ci_lane_recovery.py adjudicate` is invoked with `--unit-lanes-dir ci-update/lanes --recovery-lanes-dir ci-update/lanes`. `_cmd_adjudicate` concatenates both scans, so every record is processed twice; the partition-then-dedupe logic happens to collapse the duplicates (identical doc + identical stamp is not a conflict), so behavior is correct but the duplication is surprising and makes the "recovery lane must never collapse into its primary" invariant rely on the `fresh_runner_recovery` partition rather than on separate roots.
**Suggestion:** Download `lane-recovery-*` into a separate `ci-update/recovery` directory and pass it explicitly, mirroring the gate.

### [nit] Report can mislabel a corroborating recovery as a recovered stall
**File:** scripts/extract-test-timings.py:524-545 (called at 614-617)
**Issue:** When a lane's latest primary doc is a clean `pass` but a recovery artifact also exists (the documented re-run race where a recovery was planned, then the primary re-ran and passed), `_lane_final_label` returns `None` yet `aggregate` still emits recovery rows because `recovery_doc is not None`. `_lane_recovery_rows` then prints `original: pass` / `fresh runner: pass` / `final: recovered infrastructure PASS`, implying the lane was recovered when it actually passed on its own.
**Suggestion:** Only render the recovery section when the primary classification is recovery-eligible (or when the primary is non-pass), and/or have `_lane_recovery_rows` label a pass-primary plus pass-recovery as "superseded by primary pass".

## Verdict
COMMENT

The recovery-eligibility predicate, anti-recursion fence, plan-vs-artifact validation, and gate/timing-history gating are sound and fail closed in the paths I traced; I found no path to a false green. The findings above are robustness/reporting issues (two real uncaught-exception paths and one matrix-evaluation fragility) worth tightening before merge, but none blocks the core mechanism.

<a href="https://opencode.ai/s/nunweNbq">New%20session%20-%202026-09-15T10%3A28%3A37.257Z</a>
[opencode session](https://opencode.ai/s/nunweNbq)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/34958017548)

kaishi00:
## Summary

Adds a bounded second GitHub Actions job (`unit-recovery`) that re-runs a unit lane once on a fresh `macos-26` runner when the primary invocation watchdogged with zero identified failures, and moves the unit verdict from `needs.unit.result` to per-lane artifact adjudication shared by the gate, report, and timing-history gating. Swift-specific lenses (retain cycles, force unwraps, `MainActor`, SwiftUI lifecycle, accessibility) are not applicable — this PR touches only CI tooling.

## Findings

### [major] Recovery matrix can fail workflow evaluation when the planner fails closed
**File:** `.github/workflows/ci.yml:373` (output declared at 314, planner step at 335-345)
**Issue:** `unit-recovery` uses `matrix: ${{ fromJSON(needs.unit-recovery-plan.outputs.matrix) }}`. The `matrix` output is written only by the `id: recovery` step, which runs under `set -euo pipefail` and returns non-zero on a fail-closed plan (`ci-recovery-plan.py` exits 1 before writing/echoing the output). In that case `steps.recovery.outputs.matrix` is the empty string. Because `unit-recovery` has `always()` in its `if`, the matrix expression can still be evaluated (matrix/`strategy` expansion is not reliably short-circuited by a false job-level `if`), producing a workflow-level "Unexpected end of JSON input" error instead of the intended clean gate red. The primary `unit` job only avoids this because it has no `always()`.
**Suggestion:** Give the output a safe default: `matrix: ${{ fromJSON(needs.unit-recovery-plan.outputs.matrix || '{"include":[]}') }}` (and/or write an empty `{"include":[]}` in an `always()`/`trap` step so the output is never empty).

### [minor] `_classes_match` raises `TypeError` on a `classes: null` plan entry, escaping the fail-closed handlers
**File:** `scripts/ci_lane_recovery.py:350-353` (caller `:387`, gate catch `scripts/ci-gate.py:155`, planner `scripts/ci-recovery-plan.py:128`)
**Issue:** `plan_disagreements` passes `lane_entry.get("classes", [])` to `_classes_match`. `.get(default)` returns `None` when the key exists with a null value, and `_classes_match` then runs `sorted(map(str, planned))` on `planned=None` → `TypeError`. `ci-gate.py` catches only `OSError/ValueError/JSONDecodeError`, so this becomes an uncaught traceback instead of the module's documented precise fail-closed message. `ci-recovery-plan.py` calls `plan_disagreements` outside its `try`, so it crashes the same way. (Still a red run, but silently not the intended reason.)
**Suggestion:** Guard in `_classes_match`: `if not isinstance(planned, list) or not isinstance(documented, list): return False`, and/or normalize with `lane_entry.get("classes") or []`.

### [minor] `recovery_matrix_entry` can still raise a non-`ValueError` on malformed estimates/classes
**File:** `scripts/ci-recovery-plan.py:64-77`
**Issue:** The `missing` check only verifies each class is a key in `estimates`; it does not validate the value or that `entry["classes"]` is a list. `"{0}={1:.1f}".format(c, estimates[c])` raises `TypeError` for a `None`/non-numeric estimate, and `",".join(entry["classes"])` raises `TypeError` for a null/non-list. The caller's `except ValueError` (line 151) does not cover these, contradicting the docstring's "precise message instead of a traceback".
**Suggestion:** Coerce/validate with `float()`, report the offending class, use `.get("classes") or []`, and broaden the `except` to `(TypeError, ValueError)`.

### [minor] Report can mislabel a corroborating recovery as a recovered stall
**File:** `scripts/extract-test-timings.py:524-545` (selection at `:614-617`)
**Issue:** When a lane's latest primary doc is a clean `pass` but a recovery artifact also exists (documented re-run race), `_lane_final_label` returns `None`, yet `aggregate` still emits recovery rows because `recovery_doc is not None`. `_lane_recovery_rows` then prints `original: pass` / `fresh runner: pass` / `final: recovered infrastructure PASS`, implying the lane was recovered when it passed on its own.
**Suggestion:** Render the recovery section only when the primary classification is recovery-eligible/non-pass, or have `_lane_recovery_rows` label pass-primary + pass-recovery as "superseded by primary pass".

### [nit] Dedupe ignores the attempt suffix and treats same-second re-runs as ambiguous
**File:** `scripts/ci_lane_recovery.py:309-323`
**Issue:** Winners are chosen by `max(finished_at)` with only second granularity (`now_iso`), and the ambiguity check fails closed when two attempts share a timestamp but differ. A sanctioned re-run whose two attempts land in the same second would fail CI even though the artifact name already encodes attempt ordering deterministically.
**Suggestion:** Prefer the parsed `attempt` number as the primary ordering key with `finished_at` as tie-break; only call a tie ambiguous when attempt and timestamp both agree but docs differ.

### [nit] Timing preflight scans the same directory as both unit and recovery roots
**File:** `.github/workflows/ci.yml:679-680`
**Issue:** `--unit-lanes-dir ci-update/lanes` and `--recovery-lanes-dir ci-update/lanes` are identical, so `_cmd_adjudicate` concatenates the same records twice. Correctness currently rests on content-based dedupe plus the `fresh_runner_recovery` partition rather than on separate roots, which is surprising and fragile.
**Suggestion:** Download `lane-recovery-*` into a distinct directory (e.g. `ci-update/recovery`) as the gate does, and pass it explicitly.

### [nit] Empty `--lane-results` silently skips every timing merge
**File:** `scripts/update-timing-history.py:43-63`
**Issue:** `_lane_result_gate([])` returns `{}` (not `None`), so `observations_allowed` rejects every observation with "no paired lane result". If observations exist but lane-result files are missing/undownloadable, history silently stops evolving. It is fail-closed, but indistinguishable from a real red run in the logs.
**Suggestion:** Emit an explicit warning/error when `--lane-results` is passed but yields no pairings, so the skip is diagnosable.

## Verdict
COMMENT

The recovery predicate, anti-recursion fence, plan/artifact adjudication, and gate/timing gating are sound and fail closed on every path I traced; I found no route to a false green. The matrix-evaluation fragility (major) and the two uncaught-exception paths (minor) are worth tightening before merge, but none blocks the core mechanism.

<a href="https://opencode.ai/s/2yPeXBMZ">New%20session%20-%202026-09-15T20%3A13%3A50.095Z</a>
[opencode session](https://opencode.ai/s/2yPeXBMZ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35018228458)

kaishi00:
Closing as superseded by #184/#185.

The fresh-host recovery experiment was useful in proving that moving the same large lane to another hosted Mac did not address the underlying stall reliably.

The final solution works at a much narrower boundary: small fresh-xcodebuild batches inside the existing hosted job, with at most one retry of the affected batch rather than repeating an entire lane on another Mac.

PR #179, opened 2026-09-16
Messaging: local send ack + admit-time overlay bind

jcmcneal:
## Summary

Implements the two coupled fixes from the messaging latency diagnosis (see `docs/messaging-latency-diagnosis.md`).

### 1) Local send ack (UX probe 1)
- `send()` commits the optimistic user bubble and returns immediately; HTTP submit runs in a background task.
- Send cool-down is a **100ms local** gate (`sendCooldownActive` + `pendingDelivery`), not `hasPending` / HTTP completion.
- Composer still clears on send; failed 4xx keeps the user bubble with a visible failed state (no silent removal); uncertain outcomes surface **Check delivery** / **Retry delivery**.

### 2) Admit-time bind (UX probes 2 & 4)
- `MessagingSendReceipt` accepts optional `runs[]` (backward compatible when absent).
- `accepted()` calls `bindAdmittedRun()` to bind `conversation_id`, `run_id`, and `session_id` on the live overlay before history poll.
- `routeKey` also matches `join.runID` for mid-stream reconnect into the **same** assistant bubble — no unkeyed alias-to-only-turn fallback.

## UX probe mapping

| Probe | Expected improvement |
|---|---|
| **1** Optimistic send | Bubble + local return <100ms; FAILED keeps bubble with error chrome |
| **2** Shell before tokens | Early WS deltas route into shell once receipt/`messaging.run.start` binds session |
| **4** Run-join / reconnect | `run_id` join key + admitted session bind resume same bubble |

## Test plan

- [x] `MessagingTests` — local ack timing, failed delivery keeps pending, awaiting-reply latch
- [x] `MessagingLiveTurnTests` / `MessagingStreamRoutingTests` — receipt bind, run-id routing, no cross-talk regressions
- [ ] Manual: kill network after send tap → bubble <100ms, Send re-enables after brief cool-down, failure is honest
- [ ] Manual: delay first token 2s → shell dots visible quickly; tokens append into same bubble after admit bind

## Residual risks

- **Server contract:** admit bind only helps when bot-coms returns `runs` with `session_id` on the send receipt (or WS `messaging.run.start` with join keys). Older gateways remain poll-bound for first token.
- **In-flight delivery:** while `pendingDelivery == .inFlight`, composer blocks duplicate sends; uncertain state requires Check/Retry before another send.
- **Out of scope:** TOOL_APPROVAL sheet, nav highlight, poll throttle, soft reasoning row (per diagnosis).

Made with [Cursor](https://cursor.com)


kaishi00:
## Summary
Implements optimistic local send acknowledgement (bubble commits and returns before HTTP completes, with failed/uncertain delivery states) and admit-time binding of run/session/conversation keys from the send receipt so early WS tokens route into the live overlay.

## Findings

### [major] Automatic `checkDelivery()` can run concurrently and re-submit hard-rejected messages
**File:** Conduit/Views/Messaging/MessagingConversationView.swift:127 and Conduit/Services/MessagingConversationStore.swift:151
**Issue:** The poll task invokes `checkDelivery()` whenever `pending != nil`. Previously a 4xx cleared `pending`, so this path was inert after a hard rejection. Now `pending` is retained on failure, so `checkDelivery` runs automatically. Because `markLiveTurnFailed` / `clearAwaitingReply()` change the `.task(id:)` key (`prefersUrgentPolling`, `liveTurn.phase.isActive`), SwiftUI cancels and restarts the task, producing overlapping `checkDelivery` calls. `checkDelivery` no longer guards `!sending`, and there is no in-progress latch, so two invocations can both `reconcile` and then both `await submit(pending)` — retrying a message the server just rejected (400/403/409/422) without user action. The `sending` flag only disables the button, not the automated path.
**Suggestion:** Re-add an in-progress guard (e.g. `guard !sending else { return }`) and a separate "check in flight" latch, serialize via the existing `submitTask`, and only auto-reconcile `.uncertain` — leave `.failed` to the explicit "Retry delivery" button. Also check `Task.isCancelled` after `await submitTask.value` before mutating.

### [minor] `accept(receipt:)` replaces the whole run list with `receipt.runs`
**File:** Conduit/Services/MessagingHistoryCache.swift:133
**Issue:** `runs: receipt.runs.isEmpty ? (previous?.runs ?? []) : receipt.runs` fully replaces known runs whenever the receipt carries any run. If the ack returns only the newly admitted run(s) (likely for group sends), concurrently active runs for other profiles are dropped from the cache and from `syncLiveTurn` until the next poll, causing run-presence avatars to flicker/disappear. (`merge` also selects `incoming.runs` since `keepCurrentMetadata` is false.)
**Suggestion:** Merge runs by `id` (dedupe, prefer incoming status) rather than replacing, consistent with how messages are merged at line 215.

### [minor] Validation failure in `accepted` pins delivery at `.inFlight` with no recovery
**File:** Conduit/Services/MessagingConversationStore.swift:168
**Issue:** When `receipt.conversation` fails the destination validation, `accepted` returns after setting `error` but leaves `pendingDelivery == .inFlight`. `composerSendBlocked` then stays true (no Retry/Check button is shown, since those require `.uncertain`/`.failed`), so the composer is permanently blocked for that conversation with no user recourse.
**Suggestion:** On the invalid-response paths reset `pendingDelivery = .uncertain` (or clear `pending`) and call `owner.markLiveTurnFailed(...)` so the Check/Retry affordance appears.

### [minor] `bindAdmittedRun` last-run-wins when creating a new overlay
**File:** Conduit/Services/MessagingStore.swift:466
**Issue:** In the "no existing turn" branch the loop assigns `turn.runID = run.id` unconditionally, so for multiple admitted runs the last one wins nondeterministically relative to bind order, while the existing-turn branch correctly uses `if turn.runID == nil`. This makes the runID routing key inconsistent between the two paths.
**Suggestion:** Use `if turn.runID == nil { turn.runID = run.id }` in both branches.

### [minor] Synchronous disk decode in `AgentAvatar.body`
**File:** Conduit/Views/Components/AgentAvatar.swift:34 (cache at :509)
**Issue:** `photoURL.flatMap { AgentAvatarImageCache.shared.image(at: $0) }` runs inside `body` and `image(at:)` performs `UIImage(contentsOfFile:)` synchronously on the main thread on a cache miss (body is re-evaluated frequently during streaming). The cache is also unbounded (no cost/count limit), so it grows for the process lifetime.
**Suggestion:** Load/decode off the main actor (or prefetch on URL change), and add a count/byte eviction policy to the cache.

### [nit] `submitTask` not cancelled on deinit
**File:** Conduit/Services/MessagingConversationStore.swift:58
**Issue:** `deinit` cancels the polling/cooldown tasks but not `submitTask`, and `Task { await submit(value) }` (line 127) strongly captures `self`. The store cannot deallocate while a submit is in flight, so this is not a leak, but cleanup is inconsistent.
**Suggestion:** Add `submitTask?.cancel()` in `deinit` (or capture `[weak self]` and bail if gone).

### [nit] `routeKey` id lookups are nondeterministic on collisions
**File:** Conduit/Services/MessagingStore.swift:552
**Issue:** `liveTurns.first(where: { $0.value.runID == runID })` iterates a dictionary; if two live turns ever share a `runID`/`conversationID` (e.g. a DM destination and a conversation destination for the same conversation), routing picks an arbitrary turn.
**Suggestion:** If duplicates are possible, prefer an active-phase match or deterministic tie-break; otherwise document the invariant.

## Verdict
REQUEST_CHANGES

<a href="https://opencode.ai/s/u7FqXofk">New%20session%20-%202026-09-16T01%3A49%3A38.022Z</a>
[opencode session](https://opencode.ai/s/u7FqXofk)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35045521372)

kaishi00:
## Summary
Implements optimistic local send acknowledgement (bubble commits before HTTP completes, with failed/uncertain delivery states) and admit-time binding of run/session/conversation keys from the send receipt so early WS tokens route into the live overlay.

## Findings

### [major] Hard-rejected messages are automatically re-submitted by the poll task
**File:** Conduit/Views/Messaging/MessagingConversationView.swift:127 and Conduit/Services/MessagingConversationStore.swift:152
**Issue:** The poll task calls `checkDelivery()` whenever `model.pending != nil`. On a hard reject `submit()` sets `pendingDelivery = .failed`, and `markLiveTurnFailed` + `clearAwaitingReply` flip both `liveTurn?.phase.isActive` (true→false) and `prefersUrgentPolling` (true→false), which changes the `.task(id:)` key at line 127. SwiftUI cancels/restarts the task, `checkDelivery()` runs automatically for the retained `.failed` pending, `reconcile` 404s (the message was rejected, so no server record exists), and the `status == 404` branch calls `await submit(pending)` — re-POSTing a message the server just answered 400/403/409/422 without any user action. `sending` only disables the button; it never gates this path.
**Suggestion:** Only auto-reconcile `.uncertain`. Do not auto-invoke `checkDelivery` for `.failed` (leave that to the explicit "Retry delivery" button), and/or make the automated path reconcile-only, never resubmit.

### [major] `accepted()` can clobber a newer pending send (no pending-identity check)
**File:** Conduit/Services/MessagingConversationStore.swift:169 and :178
**Issue:** `accepted(_:pending:)` never verifies the `value` it is completing is still the current `pending`; it unconditionally does `pending = nil`, `pendingDelivery = nil`, and removes the persisted pending. In the 404 path, `checkDelivery()` calls `submit(pending)` directly (not through `submitTask`), while `pendingDelivery` is still `.failed`, so `composerSendBlocked` is false and the user can send a new message B. When the older in-flight submit for A then completes, `accepted(A)` clears `pending` (B) and its persisted state, dropping the newer optimistic bubble.
**Suggestion:** Guard on identity, e.g. `guard pending?.id == value.id else { return }` before mutating, and route every resubmit through a serialized `submitTask`.

### [major] Invalid receipt leaves `pendingDelivery == .inFlight`, permanently blocking the composer
**File:** Conduit/Services/MessagingConversationStore.swift:170
**Issue:** When the receipt fails conversation-destination validation, `accepted` returns after setting `error` but does not reset `pendingDelivery`. It stays `.inFlight`, so `composerSendBlocked` is true. Because the Check/Retry button is only shown for `.uncertain`/`.failed` (MessagingConversationView.swift:755), there is no user recourse — the composer is locked for that conversation until the view/store is recreated.
**Suggestion:** On the invalid-response paths reset `pendingDelivery = .uncertain` (or clear `pending`) and call `owner.markLiveTurnFailed(...)` so the Check/Retry affordance appears.

### [major] `checkDelivery()` has no in-progress latch, allowing overlapping reconcile/resubmit
**File:** Conduit/Services/MessagingConversationStore.swift:152
**Issue:** `sending = true` is set but never used as a guard, and the previous `guard !sending` was removed. Two overlapping `checkDelivery` invocations (plausible given the `.task(id:)` restart behavior above) can both pass the `pendingDelivery` guard, both `await service.reconcile`, and both fall into the 404 branch, issuing duplicate `submit` calls. The `await submitTask.value` at line 153 only serializes against the original send, not between `checkDelivery` callers.
**Suggestion:** Add an in-progress latch (e.g. a `checkDeliveryTask`/bool) and check `Task.isCancelled` after awaits before mutating.

### [minor] `accept(receipt:)` replaces the entire run list
**File:** Conduit/Services/MessagingHistoryCache.swift:136
**Issue:** `runs: receipt.runs.isEmpty ? (previous?.runs ?? []) : receipt.runs` (and `merge` with `keepCurrentMetadata: false` selecting `incoming.runs`) discards all previously known runs whenever the ack carries any run. If the receipt contains only newly admitted runs (likely for group sends), concurrently active runs for other profiles vanish until the next poll, causing run-presence avatars to flicker/disappear.
**Suggestion:** Merge runs by `id` (dedupe, prefer incoming status) instead of replacing, as is done for messages at line 215.

### [minor] `bindAdmittedRun` uses last-run-wins when creating a new overlay
**File:** Conduit/Services/MessagingStore.swift:468
**Issue:** In the "no existing turn" branch the loop assigns `turn.runID = run.id` unconditionally, so with multiple admitted runs the last one wins nondeterministically, while the existing-turn branch (line 454) correctly uses `if turn.runID == nil`. This makes the runID routing key inconsistent between the two paths.
**Suggestion:** Use `if turn.runID == nil { turn.runID = run.id }` in both branches.

### [minor] Synchronous disk decode in `AgentAvatar.body`
**File:** Conduit/Views/Components/AgentAvatar.swift:34 (cache at :509)
**Issue:** `photoURL.flatMap { AgentAvatarImageCache.shared.image(at: $0) }` runs during `body`; on a cache miss `image(at:)` performs `UIImage(contentsOfFile:)` synchronously on the main thread, and `body` is re-evaluated frequently during streaming. The cache is also unbounded (no cost/count limit), growing for the process lifetime.
**Suggestion:** Decode off the main actor (or prefetch on URL change) and add an eviction policy to the cache.

### [nit] `submitTask` not cancelled on `deinit`
**File:** Conduit/Services/MessagingConversationStore.swift:58
**Issue:** `deinit` cancels the polling/cooldown/awaiting tasks but not `submitTask`, and `Task { await submit(value) }` (line 127) strongly captures `self`, creating a temporary retain cycle while the request is in flight. Not a permanent leak given the 12s request timeout, but cleanup is inconsistent.
**Suggestion:** Add `submitTask?.cancel()` in `deinit` (or capture `[weak self]`).

### [nit] `routeKey` run/conversation lookups depend on dictionary iteration order
**File:** Conduit/Services/MessagingStore.swift:544 and :552
**Issue:** `liveTurns.first(where: { $0.value.runID == runID })` (and the conversationID lookup at :544) iterates a dictionary; if two live turns ever share a `runID`/`conversationID`, routing picks an arbitrary turn.
**Suggestion:** Prefer an active-phase match or add a deterministic tie-break, or document the uniqueness invariant.

## Verdict
REQUEST_CHANGES

<a href="https://opencode.ai/s/HreXF8Sg">New%20session%20-%202026-09-16T02%3A41%3A59.634Z</a>
[opencode session](https://opencode.ai/s/HreXF8Sg)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35048910388)

PR #178, opened 2026-09-16
ci(unit): split unit-2 audio workload for stall diagnosis

kaishi00:
## OUTCOME (2026-09-16, first run + one rerun)

**Hypothesis B supported for unit-2; Hypothesis A (audio-specific) refuted; the stall also appeared in unit-3, pointing at large-lane invocations generally.**

| Lane | First run | Rerun (failed jobs only) | Elapsed |
|---|---|---|---|
| `unit-1` (36 classes, unchanged) | PASS | PASS | ~12m / ~11m |
| **`unit-2-audio`** (14 audio classes) | **PASS** | **PASS** | **~5m / ~5m** |
| **`unit-2-rest`** (34 classes) | **PASS** | **PASS** | **~6.5m / ~10m** |
| `unit-3` (48 classes, unchanged) | **STALL** (watchdog; zero XCTest failures; isolation passed 20/20 classes individually, budget exhausted) | **STALL** (identical signature: `lane:timeout → isolation:incomplete`, zero failures, 8/8 isolation classes passed, no named hang) | ~16m / ~13m |

Deterministic XCTest failures: **none, anywhere.** unit-3's first attempt pinned `MarkdownLargeDocumentTests` as a class-budget timeout in isolation once; the same class passed in isolation inside the #175 artifacts (92s) — consistent with slow-host variance, not a product hang.

Reading:

1. **A is dead.** All 14 Voice/CarPlay/audio classes of the old stalling unit-2 passed together, twice, in ~5 minutes. The audio subset does not stall lanes.
2. **B confirmed for unit-2.** The identical 48-class set that stalled 10+ consecutive times as one lane passed 4/4 attempts when split 14+34. Class set unchanged; only batching changed.
3. **The stall is not unit-2-specific — it tracks large single invocations.** unit-3 (audio-free, markdown/connection-heavy, 48 classes) stalled twice in the same window with the exact historical unit-2 signature. Earlier unit-3 passes show the lane is not inherently bad; the wedge attaches to whichever large lane runs long enough during a degraded-host window.

Conclusion: the residual stall is a **long-invocation × degraded-hosted-runner** phenomenon. The audio subset is exonerated; shrinking unit lanes (or the PR #176 fresh-runner recovery as a safety net) is the durable direction. `MarkdownLargeDocumentTests` is worth a look purely as the slowest CPU-bound class, but it produced no assertion failure.

---

## What this is

[...]

PR #180, opened 2026-09-16
ci(unit): cap dynamic unit-lane size at 30 classes to reduce hosted-runner stalls

kaishi00:
## OUTCOME (first run + one rerun, 2026-09-16)

The cap produced a 6-lane unit matrix (memberships 1/26/25/26/27/27 - all <=30). Results:

| Lane | Classes | Predicted | Result | Actual | Notes |
|---|---|---|---|---|---|
| unit-1 | 1 | 258.2s | PASS x2 | ~4m | the 810s-baseline outlier class alone (timing model's choice) |
| unit-2 | 26 | 97.4s | PASS | ~4m | |
| unit-3 | 25 | 97.4s | PASS | ~5m | |
| unit-4 | 26 | 97.4s | PASS | ~5m | |
| unit-5 | 27 | 97.4s | **STALL x2** | 1362s / 1282s vs 600s watchdog | run 1: `lane:timeout -> isolation:class-failed` (SettledMessageIsolationTests (assertion failure at 106s in isolation)); rerun: `lane:timeout -> isolation:incomplete`, zero failures, 13/13 reached classes passed |
| unit-6 | 27 | 97.4s | XCTest failure -> PASS on rerun | ~5m | `AppStateForegroundLifecycleTests.testAcceptedUnsettledPreservesStalenessFromNewerBoundary()` failed 3 native retries, then passed clean on a fresh runner - a flake, reported separately |

**Interpretation: the 30-class cap alone did NOT eliminate the watchdog stall.** A 27-class lane stalled twice with the exact historical zero-XCTest-failure signature while four sibling <=30-class lanes passed in the same window. Lane size is a contributing factor (PR #178: 14/34-class splits of the cursed 48-class lane passed repeatedly) but not a sufficient mitigation. The remaining signal points at (a) specific test content from the recently-merged session-compression work - `SettledMessageIsolationTests`, `TranscriptPerformanceFixtureTests`/`TranscriptPerfLedgerContractTests`, `AppStateForegroundLifecycleTests` all landed with #168/#169 - and/or (b) per-invocation hosted-runner degradation windows: unit-5 was predicted at 97.4s and ran 1282-1362s (~13x slowdown).

Per the experiment protocol the cap stays at 30 (no knee-jerk lowering; the signature persisted across one rerun). The next diagnostic should bisect unit-5's membership (e.g. isolate the session-compression/transcript-perf classes into their own tiny lane) rather than shrink lanes further. The two flaky failures live in recently-merged (#168/#169) test code and warrant separate attention from CI infrastructure.

---

## What this is

A small CI-only planner change: **unit lanes are capped at 30 test classes**, as the next diagnostic/mitigation step after PR #178's experiment. This is general planner policy - it does not depend on #178 and carries none of #178's frozen experiment machinery (`--unit-lanes-override` is intentionally NOT carried forward).

## Evidence (from #178 - strongly supports, does not prove, root cause)

- 48-class unit lanes repeatedly watchdog-stalled on hosted macos runners with **zero identified XCTest failures**.
- Splitting one 48-class lane into **14 + 34** classes made both halves pass repeatedly.
- An unrelated **unchanged 48-class lane** then stalled with the same signature.

This PR changed exactly one independent variable: the maximum number of unit classes per lane.

## The change

In `scripts/plan-tests.py`:

```
lane count = max(existing timing-based selection, ceil(unit classes / 30))
then: nudge upward while the deterministic LPT assignment still packs any
      lane over 30 classes (LPT balances time, not count)
```

- The timing model **keeps owning class assignment** - this is a floor on lane count, not a replacement planner. Timing-driven counts above the floor are retained.
- `validate_plan` enforces the invariant **fail-closed**: every unit lane <= 30 classes (cap read from the plan's embedded config), full inventory covered exactly once, and the lane-count bound stretches legitimately when an inventory outgrows `max_lanes x 30` (loud planning failure if ever unsatisfiable).
- New flag `--max-unit-classes-per-lane` (default 30, positive-int enforced).
- `docs/CI.md` documents the real selection algorithm.

Explicitly unchanged: watchdog durations, retry policy, runner type, macOS/Xcode/Simulator versions, UI lanes, report/gate semantics, timing-history semantics, no audio classification, no Swift/XCTest changes.

## Local verification

- CI-tooling suite: **170 tests OK** (+10 planner tests: floor lifts a 132-class inventory to >=5 lanes; timing-chosen-larger counts retained; <=30 per lane; union == inventory & disjoint; small inventories unchanged; knob overridable; nudge-loop regression (1 huge + 45 tiny classes -> 3 lanes [1/22/23]); 300-class inventory validates at 10 lanes; invalid cap values fail closed; UI planning bit-identical under the cap).
- Multi-model review gate consolidated (2x REQUEST CHANGES + 1x APPROVE) - blockers fixed: cap<=0 crash (positive-int argparse + ValueError guard), false LPT-monotonicity comment corrected (linear re-check documented, never binary-search), cap-aware lane-count bound, nudge-loop and stale-table tests, docs formula.

## Summary by CodeRabbit

- **New Features**
  - Unit test planning now limits each execution lane to 30 test classes by default.
  - The planner automatically adds lanes when needed to meet the class limit while continuing to balance estimated execution time.
  - The maximum classes per lane can be configured through the planning command.
  - Configuration validation now rejects invalid limits and plans that contain oversized lanes.
  - UI test lane planning remains unchanged.


kaishi00:
## Summary
Caps unit-test lane size at 30 classes by adding a class-count floor (plus an LPT-count nudge) to the existing timing-based lane planner, with a new `--max-unit-classes-per-lane` flag, fail-closed validation, docs, and 10 tests.

## Findings

### [severity: major] Nudge loop's upper bound can leave an oversized lane even when a valid lane count exists
**File:** scripts/plan-tests.py:422-426
**Issue:** The loop upper bound is `n_bound = min(max(cfg["max_lanes"], size_floor), len(items))`, but the comment at lines 419-420 justifies it with "at n == len(items) every lane holds one class" — which is false whenever `max_lanes >= size_floor`, because the clamp makes the reachable bound `max(max_lanes, size_floor)`, not `len(items)`. LPT balances *time*, so a single dominant slow class plus many tiny classes can pack `ceil((N-1)/(size_floor-1)) > cap` into one lane. Concrete failure with defaults (cap=30, max_lanes=8): 212 unit classes where one is an extreme outlier and 211 are tiny → `size_floor = 8`, `n_bound = 8`; LPT at 8 lanes puts `1 + 211` as `31` tiny classes in one lane. The loop refuses the `n=9` increment that would fix it, `build_plan` returns a plan with a 31-class lane, and `validate_plan` then hard-fails the plan. This also affects inventories beyond `max_lanes × 30` (e.g. 300 classes + outlier: `size_floor = 10`, still 34 classes in the light lane), contradicting docs/CI.md:86-96 ("the floor only guarantees no unit lane carries more than 30 classes"). Current inventory (~72 classes) is far below the trigger and the failure is loud (validation error, no plan written), so it is not a silent misplan — but the planner cannot satisfy its own new invariant as inventory grows with realistic skew.
**Suggestion:** Bound the loop by the true reachable maximum, `n_bound = len(items)` (at `n == len(items)` LPT yields one class per lane and the cap is trivially satisfied). The loop stops at the first `n` that satisfies the cap, so this cannot over-create lanes. Equivalently: `n_lanes = max(lane_count_for(items, cfg), size_floor); while any(len(l) > cap for l in lanes): n_lanes += 1; lanes = LPT(items, n_lanes)`. Add a skewed large-inventory test (e.g. one 10000s class + 260 tiny classes) proving the planner returns a valid, cap-respecting plan.

### [severity: minor] Plan summary prints stale lane bounds when the floor lifts lane count above `max_lanes`
**File:** scripts/plan-tests.py:609-610
**Issue:** The summary emits `lanes N (bounds {min_lanes}-{max_lanes})`, but the class-count floor (and the `validate_plan` stretch at lines 562-565) legitimately produces `lane_count > max_lanes`. For a large inventory the CI summary will read e.g. `lanes 10 (bounds 1-8)`, which looks like a violation.
**Suggestion:** Report the effective upper bound (`max(max_lanes, ceil(unit classes / cap))`), or drop the bounds parenthetical and state the cap.

### [severity: nit] Fallback `cfg.get` for the cap is inconsistent with the required `cfg[k]` in the plan config
**File:** scripts/plan-tests.py:399 vs 477-483
**Issue:** `cap = cfg.get("max_unit_classes_per_lane", MAX_UNIT_CLASSES_PER_LANE)` implies the key is optional, but the plan's embedded `config` comprehension does `cfg[k]` for the same key, so a caller relying on the fallback gets a `KeyError` while building the plan. All current callers supply it, so this is latent.
**Suggestion:** Use `cap = cfg[k]` (and require the key) or make both sites consistent.

### [severity: nit] `SCHEMA_VERSION` not bumped after adding a plan-schema field
**File:** scripts/plan-tests.py:46, 478-479
**Issue:** `plan["config"]["max_unit_classes_per_lane"]` is a new field in the serialized schema while `SCHEMA_VERSION` stays `1`. The addition is backward-compatible for readers that ignore unknown keys, but any strict consumer/versioning discipline should reflect the change.
**Suggestion:** Either bump `SCHEMA_VERSION` or document that config additions are non-breaking for this schema.

### [severity: nit] Docs/comment overstate the guarantee
**File:** docs/CI.md:86-96, scripts/plan-tests.py:415-420
**Issue:** Both assert that the nudge guarantees "no unit lane carries more than 30 classes" / "The bound is reachable: at n == len(items) every lane holds one class", which is not true for the clamped loop bound described above.
**Suggestion:** Correct the prose to describe the loop bound and the fail-closed validator, or fix the bound as in the major finding.

No Swift/iOS changes are present in this PR, so retain-cycle, force-unwrap, concurrency, SwiftUI-lifecycle, accessibility, and memory concerns are not applicable.

## Verdict
REQUEST_CHANGES

<a href="https://opencode.ai/s/8zG0kTci">New%20session%20-%202026-09-16T02%3A13%3A39.856Z</a>
[opencode session](https://opencode.ai/s/8zG0kTci)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35047089034)

kaishi00:
## Summary
This PR caps unit-test lane size by adding a `ceil(classes / 30)` lane-count floor plus an LPT-count nudge to the timing-based planner, with a new `--max-unit-classes-per-lane` flag, cap-aware fail-closed validation, docs, and tests. The follow-up commit correctly replaces the buggy `min(max_lanes, …)` loop bound with `len(items)` and makes `validate_plan` cap-aware, so the previous blocking issue (oversized lane even though a valid count exists) is resolved. Remaining items are minor/latent.

## Findings

### [minor] Plan summary still prints timing bounds as if they bounded the plan
**File:** scripts/plan-tests.py:625-629
**Issue:** `plan_summary_md` renders `lanes N (bounds {min_lanes}-{max_lanes})`. The class-count floor and nudge legitimately produce `lane_count > max_lanes` (e.g. the 300-class test yields 10 lanes with `max_lanes=8`), so the CI step summary would read `lanes 10 (bounds 1-8)`, which looks like a self-violation. The earlier review flagged this as minor and commit 9ecf466 did not address it. Not reachable with today's ~72-class inventory, but it is a real latent inconsistency.
**Suggestion:** Report the cap-adjusted upper bound (`max(max_lanes, ceil(unit classes / cap))`), or drop the parenthetical and state the cap.

### [minor] Cap-active validation silently drops the `min_lanes` lower bound (and lane-count consistency)
**File:** scripts/plan-tests.py:568-582
**Issue:** When the cap is present (always, for plans from `build_plan`), the branch at 571 only checks `lane_count <= len(unit_names)` and `lane_count >= ceil(n/cap)`. It no longer checks `min_lanes`, and it still doesn't check `lane_count == len(plan["unit_lanes"])`. A planner regression that under-splits (`lane_count` below `--min-lanes`, e.g. `--min-lanes 4` with a large cap) or that reports a `lane_count` inconsistent with the emitted `unit_lanes` (which `ci.yml` reads from `plan.json` while the matrix is built from `unit_lanes`) now passes the fail-closed guard.
**Suggestion:** Keep the cap as the effective upper-bound policy but retain the lower-bound check: `if plan["lane_count"] < max(floor_n, lo)` and add `if plan["lane_count"] != len(plan["unit_lanes"])`.

### [nit] `n_bound` reachability claim assumes strictly positive estimates
**File:** scripts/plan-tests.py:419-420, 426-430
**Issue:** The comment asserts "at n == len(items) every lane holds one class." That holds only for strictly positive weights. `longest_processing_time_first` breaks heap ties by lane index, so with zero-weight items every item is repeatedly assigned to lane 0; even at `n == len(items)` lane 0 holds all classes, the loop exits at `n_bound`, and `build_plan` returns a plan that `validate_plan` then rejects. `load_estimates` filters `secs <= 0`, so production inputs are safe, but `build_plan` is callable directly (as the tests do) with arbitrary estimates/negatives.
**Suggestion:** Either document/validate the positive-weight precondition in `build_plan`, or make LPT's tie-break prefer the least-loaded-by-count lane so the "one class per lane at n == len(items)" invariant holds unconditionally.

### [nit] `cfg.get` fallback is inconsistent with the required `cfg[k]` in the embedded plan config
**File:** scripts/plan-tests.py:399 vs 481-483
**Issue:** Line 399 reads `cfg.get("max_unit_classes_per_lane", MAX_UNIT_CLASSES_PER_LANE)`, implying the key is optional, but the plan's config comprehension uses `cfg[k]`, so a caller relying on the fallback gets a `KeyError` while building the plan. All current callers supply it, so this is latent.
**Suggestion:** Use `cap = cfg[k]` (require the key) or make both sites use the fallback.

### [nit] `SCHEMA_VERSION` unchanged after adding a serialized plan field
**File:** scripts/plan-tests.py:46, 481-483
**Issue:** `plan["config"]["max_unit_classes_per_lane"]` is a new serialized schema field while `SCHEMA_VERSION` remains `1`. The addition is backward-compatible for readers that ignore unknown keys, but versioning discipline should reflect it.
**Suggestion:** Bump `SCHEMA_VERSION`, or explicitly document that config additions are non-breaking for this schema.

### [nit] Docs credit the floor for a guarantee the nudge provides
**File:** docs/CI.md:95-96
**Issue:** "the floor only guarantees no unit lane carries more than 30 classes" conflates the floor (`ceil(N/cap)`) with the nudge. The floor alone does not guarantee the cap because LPT balances time, not count; the nudge loop is what enforces it (as the skewed-HugeTests case demonstrates).
**Suggestion:** Attribute the guarantee to "floor + nudge" and note the positive-weight caveat if finding above is not fixed.

No Swift/iOS surface is present in this PR (Python planner, docs, tests only), so retain cycles, force unwraps, concurrency/MainActor, SwiftUI lifecycle, accessibility, and memory concerns are not applicable.

## Verdict
COMMENT

The previously blocking `n_bound` defect is fixed and covered by the new regression tests; the remaining items are non-blocking but worth a follow-up, with the stale summary bounds and the weakened `min_lanes` validation being the most substantive.

<a href="https://opencode.ai/s/HehhlIYJ">New%20session%20-%202026-09-16T03%3A31%3A09.702Z</a>
[opencode session](https://opencode.ai/s/HehhlIYJ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35052019275)

kaishi00:
Closing as diagnostic/superseded.

The <=30-class lane experiment showed that reducing lane size improved behavior but was not sufficient by itself: a 27-class lane still reproduced the watchdog stall repeatedly.

The final architecture landed through #184/#185 instead: keep timing-aware lanes, but recycle xcodebuild/testhost between <=7-class batches and cap the number of sequential batches per hosted job.

PR #182, opened 2026-09-16
ci(unit): bisect reproducing unit-5a stall

kaishi00:
## OUTCOME (first run + one rerun, 2026-09-16)

**The watchdog stall did NOT reproduce at 7/7 granularity.** Both 7-class halves completed under their watchdogs across all executions; CI Gate went green after the one allowed rerun.

| Lane | Classes | Predicted | Result | Actual |
|---|---|---|---|---|
| **unit-5a1** | 7 | 84.4s | **XCTest failure → PASS on rerun** | 488s (fail) / 283s (pass) |
| **unit-5a2** | 7 | 10.0s | **PASS** | 297s |
| unit-1 / 2 / 3 / 4 / 5b / 6 (pinned) | 1/26/25/26/13/27 | — | PASS | |

Details:

- **unit-5a1 attempt 1:** a real assertion failure — `TranscriptPerformanceFixtureTests.testStreamingTicksLeaveSettledMarkdownDormant_PlainTextTranscript()` — survived 3 native retries (actual 488s < 600s watchdog; **not a stall**). Reported separately per protocol; not fixed here. This test is from the transcript-performance fixture family (#124/#126 lineage) — the same content family the stall lanes are weighted toward.
- **unit-5a1 rerun:** passed clean (283s). So that failure was a flake, and no stall appeared in either attempt of either half.

**Interpretation.** Splitting the reproducing 14-class membership into 7/7 eliminated the stall (0 occurrences across 3 executions), while the same 14 as one invocation stalled 2/2 in PR #181. Per the experiment's decision table this is the "both pass" branch: **strong evidence the stall requires interaction across the two 7-class halves — accumulated test-host/Simulator state over a larger single invocation — rather than one independently toxic class.** The surfaced `TranscriptPerformanceFixtureTests` flake is consistent with the same resource-degradation dynamic showing through at reduced scale (an assertion failure instead of a wedge); one run cannot distinguish that from an unrelated transient, so it is flagged, not claimed.

Cumulative picture: 48 classes → stalls; 27 → stalls; 14 (this membership) → stalled 2/2; 7/7 → no stall. Lane size of THIS membership is the dominant variable, with content determining sensitivity (an audio-only 14-class lane passed in #178).

---

## What this is

[...]

PR #181, opened 2026-09-16
ci(unit): bisect unit-5 hosted-runner stall

kaishi00:
## OUTCOME (first run + one rerun, 2026-09-16)

The mechanical bisect is decisive: **the stall lives in `unit-5a`.**

| Lane | Classes | Predicted | Result | Actual |
|---|---|---|---|---|
| **unit-5a** | 14 | 94.4s | **STALL** | **1118s** vs 600s watchdog; zero XCTest failures; isolation hung `DashboardSessionIsolationTests` (1 pass, 12 not_diagnosed) |
| unit-5a rerun | 14 | 94.4s | **STALL** | **1315s**; the pure signature: `lane:timeout -> isolation:incomplete`, zero failures, 13/13 reached classes PASSED, 1 not_diagnosed |
| **unit-5b** | 13 | 3.0s | **PASS** | 241s |
| unit-1 / 2 / 3 / 4 / 6 (pinned) | 1/26/25/26/27 | — | PASS (unit-2 failed once, see below) | |

Reading:

1. **The stall is now localized to the 14-class `unit-5a` membership** - the next step (separate PR) is to bisect those 14 classes 7/7 with the same frozen-table approach. Note the wedge manifests only in the whole-lane invocation: in both stalled attempts, every isolation class that actually ran individually PASSED (`DashboardSessionIsolationTests` hung once in isolation, then passed there in the rerun) - accumulated test-host/Simulator state during the batched invocation remains the leading mechanical explanation, now for a specific 14-class population.
2. This population is dominated by recently-merged session/dashboard-isolation and transcript-scaling content (#118/#122/#168/#169 lineage: `SettledMessageIsolationTests`, `DashboardSessionIsolationTests`, `TranscriptPerformanceFixtureTests`, `ChatViewFollowCorrectionTests`), while #178's 14-class audio lane passed cleanly - content-dependent, not size-dependent and not audio.
3. **Real XCTest failures, reported separately from the watchdog problem:** `AVAudioCaptureServiceGenerationTests.testStaleInterruptionFromTornDownGenerationCannotStopLiveCapture()` (unit-2) failed 3 native retries, then **passed on rerun - a flake**.

Per protocol: one rerun only; the branch stays red on unit-5a (that is the experiment working).

---

## What this is

[...]

PR #183, opened 2026-09-16
ci(unit): test fresh xcodebuild boundary on one hosted Mac

kaishi00:
## OUTCOME (run 35123747377, 2026-09-16) — GREEN END-TO-END

| Invocation | Classes | Predicted | Result | Actual |
|---|---|---|---|---|
| #1 — unit-5a1 | 7 | (plan repricing) | **PASS** | **488s** |
| #2 — unit-5a2 | 7 | (plan repricing) | **PASS** | **211s** |

**Both 7-class invocations passed SEQUENTIALLY on ONE hosted macos-26 runner with the same Simulator session and NO erase between them.** `seq-result.json`: `same_runner: true, simulator_erased_between_invocations: false, completed: true`. The whole run went green — every pinned control lane, both sequential invocations, UI lanes, CI Gate.

## Answer to the experiment's question

**Terminating xcodebuild/XCTest/testhost between small batches is SUFFICIENT. A fresh hosted runner is NOT required.**

That resolves PR #182's ambiguity (the 7/7 matrix split changed fresh-xcodebuild AND fresh-runner simultaneously): the boundary that matters is the process boundary, not the machine boundary. Cumulative chain:

| Configuration | Stalls |
|---|---|
| 14 classes, one invocation (PR #181) | 2/2 |
| same 14 as two 7-class SEPARATE RUNNERS (PR #182) | 0/2 (one unrelated flake) |
| **same 14 as two 7-class SEQUENTIAL invocations, ONE runner, no erase (this PR)** | **0/2** |

## Recommended durable CI architecture

One hosted unit job per sized runner, with small batches and a fresh xcodebuild invocation per batch:

```
unit job (macos-26)
  batch (<= ~7 classes, sized to stay well inside the watchdog) -> xcodebuild exits
  batch -> fresh xcodebuild
  batch -> fresh xcodebuild ...
```

Sizing guidance from the chain: a 14-class invocation of this membership ran 1118-1315s against a 600s watchdog when the host degraded; 7-class invocations of the same content completed in 211-488s. Batch size must be chosen so a degraded-host multiple (observed ~13x) still fits the batch watchdog — i.e. keep predicted-per-batch small and/or scale watchdogs with observed (not optimistic-history) durations. This PR deliberately does not implement that architecture; it is the evidence for it.

(First truncated attempt on `78daa99`: invocation 1 stalled at 606s — the historical signature — but a dropped counter initialization aborted the script before invocation 2 (`SEQ_FAILED: unbound variable`), so no invocation-2 data existed; fixed in `ec0dbf4` and re-run. The truncated run's partial result was still captured by the signal traps, as designed.)

---

## What this is

A tiny CI-only **diagnostic experiment**: the same mechanically-defined 7 + 7 split from PR #182, run **sequentially inside one macos-26 GitHub job** — one hosted Mac, one Simulator session, two sequential xcodebuild invocations with no erase/reset between them. Isolates whether terminating the xcodebuild/XCTest/testhost process between the halves is sufficient, versus needing a fresh hosted runner (PR #182's 7/7 matrix split changed both at once).

## Membership (byte-proven)

Identical to PR #182's table — cross-checked byte-for-byte against both #182's frozen table (`99ab945`) and run `35085756610`'s plan artifact:

- **unit-5a1 (7):** `ChatViewFollowCorrectionTests`, `DashboardSessionIsolationTests`, `TranscriptPerformanceFixtureTests`, `ChatTypographyTests`, `CrossProfilePresentationCacheTests`, `SettledMessageIsolationTests`, `AppStateDecisionFenceTests`
- **unit-5a2 (7):** `NativeAuthClientIntegrationTests`, `MessageReadAloudControllerTests`, `SessionYoloPersistenceTests`, `KanbanV3ATests`, `AppStateVoiceSpokenPhraseTests`, `ChatTextSelectionTests`, `AuthWebViewNavigationPolicyTests`

## How

- The unit matrix **filters** `unit-5a1`/`unit-5a2` out (jq filter in the plan step — `strategy.exclude` is a no-op for include-only matrices); `scripts/ci-sequential-xcodebuild-diagnostic.sh` runs the two groups sequentially in the new `unit-seq-diagnostic` job: per-invocation planned watchdog, the SAME native retry flags as a normal lane, per-invocation logs/result bundles, `seq-result.json` via jsonl records + EXIT/TERM traps (evidence survives even a job-kill truncation), one bounded `simctl shutdown all` before invocation 1 for startup parity — deliberately NOT repeated between the groups.
- The six sibling lanes run byte-identical to #181's table as the normal matrix. The seq job is instrumentation, not part of the CI Gate. The table must never merge while frozen (fail-closed on inventory drift).
- No Swift, no test assertions, no watchdog/retry-policy changes; no erase/reset/reboot between the halves.

## Verification

- CI-tooling suite: **170 tests OK** (machinery + partition pins carried from #181/#182).
- jq matrix filter verified on a real bash; script parse of the two lane groups verified locally; provenance byte-proven vs #182's table and run artifact.
- Multi-model review gate: blockers fixed before push — CSV class extraction (list-repr would have produced garbage `-only-testing` filters), jq filter mechanism (exclude→filter), retry-flag parity, fatal+trapped result writer, empty-class guards, artifact naming outside the `lane-*` glob.

Do not merge.


kaishi00:
## Summary
This PR adds a one-shot diagnostic experiment that runs the frozen `unit-5a1` and `unit-5a2` class groups as two sequential `xcodebuild` invocations on a single hosted Mac (no Simulator reset between them), plus the planner `--unit-lanes-override` machinery, a frozen lane table, and regression tests. The goal is to isolate the fresh xcodebuild/testhost boundary from the fresh-runner variable.

## Findings

### [severity: critical] `SEQ_FAILED` is never initialized, breaking the both-pass outcome and possibly invocation 2
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:138` and `scripts/ci-sequential-xcodebuild-diagnostic.sh:212`
**Issue:** The script runs under `set -u` (line 27) but `SEQ_FAILED` is never initialized. On the primary hypothesized outcome (both invocations pass), `record_invocation` never executes the failure branch, so line 212 `[ "$SEQ_FAILED" -gt 0 ]` expands an unset variable and the shell aborts with `SEQ_FAILED: unbound variable`, exiting non-zero. I reproduced this exact pattern on bash 5.2: the script prints the error and exits 1 despite both records passing — the job goes red on the success case. Additionally, on any bash that honors `nounset` inside arithmetic (bash 4.4+), the first failing invocation dies at line 138 (`SEQ_FAILED=$(( SEQ_FAILED + 1 ))`), so invocation 2 never runs — directly violating the documented contract "Invocation 2 ALWAYS runs, even if invocation 1 fails or stalls." (On macOS `/bin/bash` 3.2 the arithmetic tolerates it, but the both-pass bug above still hits.)
**Suggestion:** Initialize `SEQ_FAILED=0` alongside the other globals (e.g. at line 32 or immediately before the first invocation) and keep the `set -u`-safe read at line 212.

### [severity: major] Diagnostic lane override is wired into the normal plan/validate path with no guard
**File:** `.github/workflows/ci.yml:96` and `.github/workflows/ci.yml:116`
**Issue:** `--unit-lanes-override scripts/diagnostic-unit-split.json` is now passed unconditionally in the `plan` job's validate and plan steps, and the matrix is hard-filtered by lane name in jq (line 125). If this branch is ever merged (or if the diagnostic files are forgotten), lane formation is permanently frozen: `plan-tests.py` fails closed on any newly added/renamed class, so ordinary test additions break CI, and the jq filter silently references experiment-only lanes. The PR body says "Do not merge", but nothing in the workflow prevents it.
**Suggestion:** Gate the override behind an explicit branch/workflow input or event check, or remove the override + jq filter before merging. At minimum, add a comment/lint that makes the experiment-only nature fail loudly on merge.

### [severity: minor] New bash script has no regression coverage
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:1`
**Issue:** The repo invests heavily in bash state-machine tests for `ci-test-lane.sh`/`ci-lib.sh`, and the new diagnostic script is the centerpiece of this experiment, but no test exercises it. The `SEQ_FAILED` bug and the trap behavior would have been caught by even a lightweight stub harness mirroring the existing CI-tooling suites.
**Suggestion:** Add a test that stubs `xcodebuild`/`run_with_deadline` (or drives the exit paths with fixtures) and asserts: both-pass exits 0, one-fail still runs invocation 2, and `seq-result.json` is valid JSON on both paths.

### [severity: minor] TERM/INT traps handle the signal but do not terminate
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:115-116`
**Issue:** `trap write_seq_result TERM` / `INT` write the result and return, after which bash resumes the interrupted command (e.g. the `sleep` inside `run_with_deadline`). The process is not actually terminated, so an outer cancellation may leave the script running until GitHub's SIGKILL grace expires.
**Suggestion:** Have the handlers write the result and then `exit` (e.g. `trap 'write_seq_result; exit 143' TERM`), or call `exit` explicitly at the end of the handler.

### [severity: nit] `seq-result.json` is written twice on the normal path
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:210` and `scripts/ci-sequential-xcodebuild-diagnostic.sh:114`
**Issue:** `write_seq_result` is called explicitly at line 210 and again by the `EXIT` trap on `exit 0`/`exit 1`, producing a redundant second write (and a duplicated "result written" log line).
**Suggestion:** Drop the explicit call at line 210 and rely on the `EXIT` trap, or set a guard so the writer is idempotent.

## Verdict
REQUEST_CHANGES

<a href="https://opencode.ai/s/qaG1qhFK">New%20session%20-%202026-09-16T16%3A18%3A56.480Z</a>
[opencode session](https://opencode.ai/s/qaG1qhFK)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35120972369)

kaishi00:
## Summary

Adds a one-shot diagnostic experiment: a `--unit-lanes-override` frozen-lane path in `plan-tests.py`, a sequential two-invocation `xcodebuild` runner, and a non-gating `unit-seq-diagnostic` job that holds one hosted Mac/Simulator constant while varying only the fresh `xcodebuild`/testhost boundary.

## Findings

### [severity: major] Diagnostic override and matrix filter are wired into the normal plan path with no guard
**File:** `.github/workflows/ci.yml:96`, `.github/workflows/ci.yml:116`, `.github/workflows/ci.yml:125`
**Issue:** `--unit-lanes-override scripts/diagnostic-unit-split.json` is now passed unconditionally to both `validate` and `plan`, and the unit matrix is permanently filtered by the experiment-only lane names `unit-5a1`/`unit-5a2`. If this branch is merged, two things break: (1) `load_lane_override` + `validate_plan` fail closed on any newly added/renamed/removed test class, so ordinary test additions turn the `plan` job red; (2) the 14 classes in `unit-5a1`/`unit-5a2` are removed from the `unit` matrix and only run in `unit-seq-diagnostic`, which is deliberately **not** in `ci-gate.needs`, so a green CI Gate would no longer execute those 14 tests. The jq filter also hardcodes the lane names. The PR body says "Do not merge", but nothing in the workflow enforces that.
**Suggestion:** Gate the override and the jq filter behind an explicit trigger (e.g. `github.event_name == 'pull_request' && github.head_ref == '...'`, a `workflow_dispatch` input, or a branch check), or use a `strategy.matrix` `exclude`/`include` construction that disappears with the experiment. At minimum, remove the override and filter before any merge to `main`.

### [severity: minor] TERM/INT trap handlers record evidence but do not terminate
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:120`
**Issue:** `trap write_seq_result TERM` / `INT` write the result and return, after which bash resumes the interrupted command (e.g. the `sleep` inside `run_with_deadline`). The trapped signal no longer terminates the script, so an outer cancellation can leave it running until GitHub's untrappable SIGKILL, during which no further evidence is captured and the runner keeps being consumed.
**Suggestion:** Exit from the handler, e.g. `trap 'write_seq_result; exit 143' TERM` / `trap 'write_seq_result; exit 130' INT` (the EXIT trap double-write is harmless/idempotent).

### [severity: minor] Hardcoded 75-minute ceiling is not derived from the planned watchdogs
**File:** `.github/workflows/ci.yml:303`
**Issue:** `timeout-minutes: 75` is fixed, while the two invocation budgets come from `plan.json` `timeout_s` values, and `timeout_for` has only a floor (`LANE_TIMEOUT_MIN_S = 600`) — no upper cap. Timing history can grow both watchdogs so that `2 × (budget + finalize_grace) + setup` exceeds 4500s, letting GitHub SIGKILL the job before the second invocation or final write.
**Suggestion:** Emit a sequential ceiling from the plan (sum of both lanes' watchdogs plus finalize/setup/download slack, including the `XCODEBUILD_FINALIZE_GRACE_S` behavior) and consume it for this job's `timeout-minutes`, as already done for `matrix.job_timeout_min` on the normal lanes.

### [severity: minor] The new diagnostic script has no regression coverage
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:1`
**Issue:** The repo invests heavily in bash state-machine tests for `ci-test-lane.sh`/`ci-lib.sh`, but the experiment's centerpiece has none. The earlier `SEQ_FAILED` abort (and the trap/exit semantics) would have been caught by a stub harness driving the pass/fail/timeout paths.
**Suggestion:** Add a test mirroring the existing CI-tooling suites that stubs `xcodebuild`/`run_with_deadline` and asserts: both-pass exits 0, one-fail still runs invocation 2, and `seq-result.json` is valid JSON on both paths.

### [severity: nit] `seq-result.json` is written twice on the normal path
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:215`
**Issue:** `write_seq_result` is called explicitly at line 215 and again by the `EXIT` trap installed at line 119 when the script exits at lines 219/222, producing a redundant second write and duplicated "result written" log line.
**Suggestion:** Drop the explicit call and rely on the `EXIT` trap, or make the writer idempotent with a guard flag.

### [severity: nit] Dead state variables
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:64`
**Issue:** `SEQ_N` and `SEQ_RECORDS` (line 66) are initialized but never read; `SEQ_RECORDS` appears to be a leftover from the pre-jsonl string-concatenation design.
**Suggestion:** Remove them, or use them.

### [severity: nit] `lane-count` output no longer matches the unit matrix
**File:** `.github/workflows/ci.yml:126`
**Issue:** `lane-count` is read from `plan.json` and reports 8, while the filtered matrix now contains 6 entries. It is currently unused by any downstream job, but the two values are semantically inconsistent.
**Suggestion:** Derive `lane-count` from the filtered matrix value, or drop the output.

### [severity: nit] Empty classes list serializes as `[""]`
**File:** `scripts/ci-sequential-xcodebuild-diagnostic.sh:136`
**Issue:** `sys.argv[6].split(',')` turns an empty CSV into `[""]` rather than `[]`. This is currently unreachable (line 151 exits on empty lane resolution) but is brittle if the guard is ever relaxed.
**Suggestion:** Guard with `sys.argv[6].split(',') if sys.argv[6] else []`.

## Verdict

COMMENT

Note: no Swift/SwiftUI files changed, so retain-cycle, force-unwrap, concurrency, and accessibility concerns are not applicable. The previously reported critical `SEQ_FAILED`-under-`set -u` bug is fixed at `scripts/ci-sequential-xcodebuild-diagnostic.sh:65`/`143`. The remaining items are non-blocking for an explicitly do-not-merge experiment branch, but the override wiring (major) and the trap behavior must be resolved before any merge.

<a href="https://opencode.ai/s/OfJV1Rdd">New%20session%20-%202026-09-16T16%3A44%3A59.982Z</a>
[opencode session](https://opencode.ai/s/OfJV1Rdd)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35123743694)

kaishi00:
Closing as completed diagnostic work.

This experiment established the key architectural result: the two small groups can run sequentially on the same hosted Mac and Simulator when a fresh xcodebuild/testhost process is used between them. A fresh GitHub runner per batch is not required.

#184 implements that execution model generally, and #185 adds the job-level parallelism needed to keep CI wall time reasonable.

PR #184, opened 2026-09-16
ci(unit): recycle xcodebuild between small test batches

kaishi00:
## Summary

Unit lanes keep their GitHub-job identity, but each lane now executes its classes internally as **multiple small sequential `xcodebuild test-without-building` batches** (at most 7 classes per invocation), each batch a **fresh xcodebuild/XCTest/testhost process** on the **same hosted runner and same Simulator session** — no erase, no fresh runner.

This turns the diagnostic conclusion of PRs #178/#180/#181/#182/#183 into normal CI behavior:

- large single unit invocations repeatedly watchdog-stalled on hosted macos-26;
- audio-specific hypotheses did not hold (#178);
- a 30-class lane cap did not help (#180);
- the reproducing set narrowed to 14 classes (#181/#182);
- the same 14 completed as two 7-class invocations run back-to-back inside ONE job on ONE runner with ONE Simulator session (#183) — and one earlier 7-class invocation did stall once, so batching reduces exposure without being infallible (hence the narrow one-retry safety net).

Evidence strongly supports recycling xcodebuild/XCTest/testhost state between small batches as a materially more reliable execution model on GitHub-hosted macOS runners. The root cause inside XCTest/CoreSimulator is **not** proven and is not claimed here.

## Architecture

**Planner (`plan-tests.py`) — single authority, unchanged lane structure**

- The existing timing-aware LPT lane balance is untouched; lanes stay `unit-1…unit-N` in the GitHub matrix.
- Each lane's class list is chunked **mechanically, in stored order**: `classes[0:7]`, `classes[7:14]`, … — no regrouping, no subsystem/audio special-cases (`MAX_UNIT_CLASSES_PER_XCODEBUILD_BATCH = 7`, an empirically conservative starting point, not a proven threshold).
- Every batch gets its own watchdog: `max(600s, ceil((240s invocation overhead + predicted execution) × 2.5))` from the same timing data that balanced the lane.
- `validate_plan` enforces the invariants: every class exactly once, concatenation reproduces the lane order exactly, batch size ≤ 7, positive integer budgets.
- The batch layout + budgets travel in the unit matrix (`batches` compact-JSON string → env var → `--batches-json`); the runner refuses to start unless the batches exactly reproduce the lane's class list.

**Runner (`ci-test-lane.sh`) — batch execution + narrow recovery**

Per batch: fresh `xcodebuild test-without-building`, the existing destination/environment/native-retry semantics (`-retry-tests-on-failure -test-iterations 3`), a per-batch watchdog, full process-group kill on expiry, result recorded, next batch.

Recovery is batch-level, exactly one retry per batch, never a lane rerun, never a fresh runner:

| attempt-1 outcome | action |
|---|---|
| pass (all classes recorded) | continue; no reset/erase |
| real test failures (surviving native retry) | **lane fails** — no retry masquerade |
| unclassifiable result | **lane fails** — never retried into a green lane |
| watchdog stall, zero known failures | bounded simulator shutdown (NO erase) → retry THIS batch once |
| infra wedge (nonzero exit, zero known failures) | erase (historical semantics) → retry THIS batch once |
| retry stalls/wedges again | **lane fails**, `hung_batch` names the culprit |

A stall whose killed xcresult still records surviving failures is classified as a test failure (never retried). An exit-0 batch whose xcresult lacks a class's record fails the lane (fail-closed). A green verdict additionally requires a passing record for **every planned batch index** in the lane bookkeeping.

**Retired:** the whole-lane single invocation, the full-lane infra retry, and lane-level class-granular isolation (units). UI lanes are untouched. The CI Gate is untouched (lane pass/fail remains the input).

## Reporting & timing history

- `lane-result.json` gains a `batches` document (per batch: classes, planned budget, attempt chain with elapsed/failure counts, final status, `not_run` for batches a stopped lane never reached).
- The CI Test Report renders a **"Unit lane batches"** section — `batch 3/5 watchdog -> retry PASS` — so a reviewer never reads raw Actions logs to find a stall; a double stall renders `HUNG: unit batch N stalled twice (classes: …)`.
- Timing history is unchanged in spirit and machinery: batch extractions fold through the existing `merge-parts` (last full-class attempt wins; a retried batch's passing rerun owns its classes' samples), so only successful observations reach the EWMA — stalled wall-clock never poisons the history.

## Tests

- Planner: partition scenarios (0 / 1–7 / 8 / 14 / 15 classes, concatenation identity, duplicates/omissions, cap, budget math, ceiling ≥ worst path, matrix carries batches, validate rejects invariant violations).
- Runner state machine (bash, 173/173 on macOS): all-pass, real failure (no retry, later batches `not_run`), unclassified, watchdog→same-batch retry→green (no erase), double stall (exactly 2 invocations, `hung_batch`), infra wedge→erase retry→green, persistent infra, malformed/missing `--batches-json` refuse to start, stall-with-failures never retried, exit-0-with-missing-class fails closed, finalize-grace regressions.
- Python tooling suite 179 passing locally; bash lane-runner suite 173/173 and destination suite 36/36 on the Mac build host.

## Live verification (run [35158637061](https://github.com/kaishi00/hermes-conduit/actions/runs/35158637061), all green, CI Gate success)

| lane | classes | batches (sizes) | per-batch outcomes |
|---|---|---|---|
| unit-1 | 63 | 9 (7×9) | 9/9 PASS on attempt 1, no retries; wall ≈ 29 min |
| unit-2 | 69 | 10 (7×9 + 6) | batch 1/10 **watchdog stall at 1203s → same-batch retry (fresh xcodebuild, no erase) PASSED in 326s**; batches 2–10 PASS on attempt 1; wall ≈ 48 min |

The batch-1 stall-recovery in unit-2 is the new recovery path exercised live: previously this stall shape killed the whole lane (attempt 2 = full-lane rerun / isolation); here the lane recovered in ~5.5 extra minutes and continued green. Every unit lane's log shows the fresh-xcodebuild invocation boundaries (`batch N/M attempt 1` groups), and the lane artifact's `batches` document records the exact attempt chain (`(1, 'timeout', 1208), (2, 'passed', 326)`).

The PR is **not** to be merged as part of this change.

No production Swift, XCTest assertions, runner/Xcode/Simulator versions, watchdog policy for UI, or gate semantics are modified.


kaishi00:
## Summary
This PR changes unit CI lanes from one whole-lane `xcodebuild` invocation into sequential per-lane batches (≤7 classes each, fresh `xcodebuild` per batch, same runner/Simulator), with planned per-batch watchdogs, bounded batch-level retry, fail-closed bookkeeping, and new batch reporting in the CI Test Report. UI lanes and the gate are untouched.

## Findings

### [minor] Green unit-lane retry-bundle pruning deletes evidence when more than one batch was retried
**File:** scripts/ci-test-lane.sh:590-596
**Issue:** `keep=$(awk -F'|' '$2 == 2 {print $1}' ...)` yields a newline-separated list (e.g. `2\n3`), but the matcher `case " $keep " in *" $bn "*)` assumes space-separated tokens. With `keep="2\n3"`, the subject is `" 2\n3 "` and no `" 2 "`/`" 3 "` substring exists, so **every** `batch-*.xcresult` is pruned. This silently drops the attempt-1/attempt-2 bundles for recovered watchdog/infra batches, contradicting the code comment and docs/CI.md ("on a green lane preserve both attempt bundles of any batch/class that needed its retry"). Existing tests only exercise a single retried batch, so the multi-retry case is uncovered.
**Suggestion:** Normalize `keep` before matching, e.g. `keep=$(awk ... | tr '\n' ' ')`, or compare with `case " $(echo $keep) "`; add a green-lane test with two retried batches.

### [minor] `--unit-batch-max-classes 0` crashes the planner before validation can report it
**File:** scripts/plan-tests.py:378-380 (and main order at scripts/plan-tests.py:985-986)
**Issue:** `build_plan` runs before `validate_plan`. `unit_batches_for` calls `range(0, len(classes), size)`; `size == 0` raises `ValueError: range() arg 3 must not be zero`, producing a traceback instead of the intended clean `unit_batch_max_classes must be >= 1` error. Negative values happen to produce empty batches and are caught later, so behavior is inconsistent.
**Suggestion:** Validate `unit_batch_max_classes >= 1` (and other numeric config) before `build_plan`, or guard `size` inside `unit_batches_for`.

### [minor] `bounded_run ... > diagfile` writes an empty diagnostic file
**File:** scripts/ci-test-lane.sh:789 (also 951, 1161)
**Issue:** `bounded_run` captures the command's stdout/stderr into an internal file and returns it via `BOUNDED_OUTPUT`; it never prints it. Redirecting `bounded_run` to a file therefore produces an empty `simctl-devices-after-batch-N-timeout.txt`, so the on-timeout device dump is useless. This pattern is carried over from the old timeout paths, but the new batch path introduces another instance.
**Suggestion:** Drop the redirect and persist the captured output, e.g. `bounded_run 45 xcrun simctl list devices || true; printf '%s\n' "$BOUNDED_OUTPUT" > "$LOG_DIR/..."`.

### [minor] Claim that only successful observations reach the EWMA is not enforced
**File:** scripts/ci-test-lane.sh:530-541 / scripts/update-timing-history.py:104-127
**Issue:** `merge-parts`/`observations.json` are written and uploaded (`if: always()`) for failed lanes too, and `update-timing-history.py` consumes every lane's `observations.json` regardless of `status`. A batch that stalls twice can therefore contribute its partial class timings to the EWMA. The PR body/docs state "only successful observations reach the EWMA; stalled wall-clock never poisons the history," which overstates the guarantee (the non-zero outlier clamp only bounds the damage).
**Suggestion:** Either filter to lane results with `status == "pass"` when building the `--observations` list for `update-timing-history`, or soften the docs/PR claim.

### [nit] Job ceiling is not clamped to GitHub's hosted-runner maximum
**File:** scripts/plan-tests.py:424-441
**Issue:** `unit_job_timeout_min` can exceed 360 minutes for large batches/estimates (e.g. two ~5600s budgets), and the value flows straight into `timeout-minutes`. If it exceeds GitHub's hosted cap the job cannot start; there is no clamp or warning. (The same structural risk pre-exists for `ui_job_timeout_min`.)
**Suggestion:** Clamp to 360 and surface a warning, or assert the ceiling stays under the cap at plan time.

### [nit] `isolation` lane-result field and report section are now dead
**File:** scripts/extract-test-timings.py:244-246, 733-741
**Issue:** The runner no longer passes `--isolation-json` for any lane, so `isolation` is always `None` and the aggregate's "isolation per-class results" rendering can never execute. `SCHEMA_VERSION` was also bumped to 2 while `--isolation-json` remains in the CLI.
**Suggestion:** Remove the dead field/rendering/arg, or keep it explicitly documented as legacy.

### [nit] Batch summary example/status wording is inconsistent
**File:** scripts/extract-test-timings.py:492-495; docs/CI.md (Watchdogs/artifacts sections)
**Issue:** The docstring/docs promise `batch 3/5 watchdog -> retry PASS`, but status values are `timeout`/`passed`, so the actual rendering is `batch 3/5 timeout -> retry PASSED` (as the tests assert). Also `docs/CI.md` still shows the old example string.
**Suggestion:** Align the docstring/docs example with the real output.

### [nit] Misleading log line for UI lanes
**File:** scripts/ci-test-lane.sh:216
**Issue:** The unconditional log prints `... classes in 0 sequential xcodebuild batches` for UI lanes (`BATCH_CLASSES_ARR` is empty for `--kind ui`), which is confusing.
**Suggestion:** Emit batch-count wording only when `KIND=unit`.

## Verdict
REQUEST_CHANGES

The core batch state machine and planner invariants look sound (real-failure/terminal ordering, fail-closed green check, exactly-one batch retry, not_run bookkeeping, and planner/runner cross-validation are well covered). The blocking item is the new green-lane bundle-retention logic in `finish_lane`, which is wrong for any lane with more than one retried batch and contradicts both its comment and the documented evidence policy.

<a href="https://opencode.ai/s/XzPoYEDw">New%20session%20-%202026-09-16T22%3A39%3A28.513Z</a>
[opencode session](https://opencode.ai/s/XzPoYEDw)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35158637005)

PR #185, opened 2026-09-17
ci(unit): parallelize sequential xcodebuild batches

kaishi00:
## Summary

#184's small-batch execution model proved reliable (live run green, including a real watchdog-stalled batch recovered by the same-batch retry), but its first run consolidated the suite into only **2 unit jobs of 9–10 sequential batches each → ~29–48 minutes of unit wall time**. This PR keeps that execution model untouched and adds one planner policy: **a unit lane may chain at most `MAX_UNIT_BATCHES_PER_JOB = 4` sequential xcodebuild batches**, so batches run concurrently across hosted Macs instead of serially on one.

The cap is on **batches per job**, not on an arbitrary "safe class count" — with the 7-class batch cap this bounds a normal lane at roughly 28 classes.

## Planner policy

- The existing timing-aware lane-count selection stays authoritative. After it runs, the **actual LPT assignment is verified** against the policy: `ceil(lane class count / 7) ≤ 4` for every lane. If any lane violates it, the lane count is **incremented (never lowered)** and the deterministic assignment re-run until every lane fits.
- Verifying the real assignment is load-bearing: LPT balances *load*, not class counts, so `ceil(N / 28)` alone does not guarantee it (a skewed inventory can pour dozens of tiny classes onto one lane; pinned by a dedicated test where the naive floor needs 5 increments).
- **Fail-closed**: if the configured lane bounds (`MAX_LANES = 8`) make the policy unsatisfiable, `validate_plan` reports a per-lane policy error and planning exits nonzero without writing artifacts. An invalid policy value (`unit_max_batches_per_job = 0` — the natural "disable the cap" attempt) is a structured `::error::`, never a crash. A `::warning::` tripwire announces when the lane count saturates `max_lanes` (the policy bounds the suite at 8 × 4 × 7 = 224 classes with defaults).
- One variable per change: watchdog floors/multipliers/timeouts are deliberately **not** touched.

## Not changed (all #184 behavior preserved)

Sequential batch execution, one fresh xcodebuild per batch, same runner/Simulator, same-batch one-retry for watchdog+zero-failure stalls and infra wedges, failure classification, native XCTest retries, simulator-reset semantics, report/gate behavior, timing ingestion, batch artifact format, UI planning (byte-identical under the new policy — pinned by test).

Retries remain an emergency safety net; healthy CI should normally finish all batches on their first attempt and complete unit CI in roughly the **10–20 minute** range rather than 30–50 minutes.

## Tests

12 new `UnitBatchParallelismTests` + 2 robustness tests: small inventories stay one lane; 28 classes fit one 4-batch job; 29 need ≥2 jobs; 132-class inventory → every lane ≤4 batches; timing logic can still exceed the floor (8 lanes); skewed estimates verified-incremented to a satisfying assignment (with mutation-verified load-bearing-ness); coverage exactly once; no duplicates; UI planning identical under cap 1 vs 4; matrix batch metadata valid; `validate_plan` fails closed on a policy-violating plan; invalid policy value fails closed without crashing; empty/single-class inventories. Full tooling suite: **193 passing** locally and on the macOS build host; bash lane-runner suite 173/173 (runner untouched).

## Live verification (run [35166035967](https://github.com/kaishi00/hermes-conduit/actions/runs/35166035967), all green, CI Gate success)

Generated plan (warm timing history): **6 unit lanes, every lane ≤ 4 batches, every batch ≤ 7 classes**:

| job | classes | batches (sizes) | job wall (incl. setup) | batch retries |
|---|---|---|---|---|
| unit-1 | 1 (810s outlier alone) | 1 (1) | 6.6 min | 0 |
| unit-2 | 26 | 4 (7/7/7/5) | 17.2 min | 0 |
| unit-3 | 25 | 4 (7/7/7/4) | 14.3 min | 0 |
| unit-4 | 26 | 4 (7/7/7/5) | 12.6 min | 0 |
| unit-5 | 27 | 4 (7/7/7/6) | 13.1 min | 0 |
| unit-6 | 27 | 4 (7/7/7/6) | 15.3 min | 0 |

**All 25 batches passed on attempt 1** (2–7 min per batch). Unit-phase wall: ~29 min end-to-end, but the slowest single job was **17.2 min** (vs 48 min in #184) and three jobs queued ~12 min behind GitHub's hosted-macOS concurrency limit — i.e. the phase is now bounded by runner availability, not by the planner. Without queueing the critical path is ~17 min, inside the 10–20 minute target. Compared with #184's ~29–48 minute per-job walls this is the intended step change; the remaining lever is Actions macOS job concurrency, not lane sizing.

The PR is **not** to be merged as part of this change. Stacked on #184 (base branch `main` will show both PRs' commits until #184 merges).

## Summary by CodeRabbit

- **New Features**
  - Unit-test execution is now split into sequential batches of up to seven classes.
  - Jobs support a maximum of four unit-test batches, with automatic lane adjustments to meet the limit.
  - Test plans, reports, and matrix output now include batch details and counts.
  - Planning and validation support new controls for batch size, timeouts, recovery overhead, and per-job batch limits.

- **Documentation**
  - CI documentation now describes batching limits, lane planning, and the maximum supported suite size.


kaishi00:
## Summary
Adds a planner policy (`MAX_UNIT_BATCHES_PER_JOB = 4`) that raises the unit lane count until every LPT lane fits at most four sequential 7-class xcodebuild batches, so batches fan out across more hosted Macs; the execution model, runner, and watchdog budgets are unchanged. The policy logic is sound and well-tested, but the invalid-policy fail-closed path it advertises is incomplete.

## Findings

### [severity: major] `--unit-batch-max-classes 0` still crashes planning before `validate_plan` can report it
**File:** scripts/plan-tests.py:469-473 (claim), scripts/plan-tests.py:387 (crash)
**Issue:** The new early-return comment states the invalid value is handled "exactly like `unit_batch_max_classes < 1`". It is not. When `unit_batch_max_classes < 1`, `enforce_batches_per_job` correctly skips the floor, but `build_plan` then unconditionally calls `unit_batches_for`, which does `range(0, len(classes), size)` (line 389) with `size = cfg["unit_batch_max_classes"]`. `range(..., 0)` raises `ValueError: range() arg 3 must not be zero`, so `build_plan` aborts with a traceback. `validate_plan` (line 621) is only reached *after* `build_plan` returns, so its `unit_batch_max_classes must be >= 1` check is unreachable for exactly the input it exists to catch. `0` is the natural "disable batching" value an operator would try (the same class of input the 95c502c fix handled for the sibling flag), so this is a real user-facing crash.
**Suggestion:** Guard before batching, e.g. in `_cfg_from_args`/early in `main` return a structured error for `a.unit_batch_max_classes < 1` (and `a.unit_max_batches_per_job < 1`), or make `unit_batches_for` use `cfg.get(..., MAX_UNIT_CLASSES_PER_XCODEBUILD_BATCH)` and clamp/return a single batch when `size < 1`. Fix the comment to reflect reality.

### [severity: minor] Invalid `unit_max_batches_per_job` is not caught when there are no unit lanes
**File:** scripts/plan-tests.py:702-705
**Issue:** The `< 1` check for `unit_max_batches_per_job` lives inside the `for lane in plan["unit_lanes"]` loop. With an empty unit inventory (empty `unit_lanes`), the loop body never runs, so `--unit-max-batches-per-job 0` produces no error and planning succeeds and writes artifacts containing the invalid config value. This contradicts the fail-closed guarantee and is asymmetric with `unit_batch_max_classes`, whose check is at top level (line 621).
**Suggestion:** Hoist the `unit_max_batches_per_job < 1` check to the top of `validate_plan`, next to the `unit_batch_max_classes` check, and drop (or keep as a no-op) the in-loop duplicate.

### [severity: minor] Saturation warning claims a class bound that LPT skew can invalidate
**File:** scripts/plan-tests.py:520-528, docs/CI.md:130
**Issue:** The warning fires whenever `n_lanes >= max_lanes` and announces the suite is "bounded at `max_lanes * unit_max_batches_per_job * 7` = 224 classes". Two problems: (a) that product is only an upper bound — because LPT balances load, not counts, a skewed inventory can fail the policy closed far below 224 (the skew test itself demonstrates this), so stating it as the bound can mislead capacity planning; (b) it fires even when the policy is comfortably satisfiable and the timing model merely preferred `max_lanes` (e.g. 40 heavy classes pin `lane_count == 8`), so a routine plan emits a "growth will fail planning closed" warning that isn't near true. The docs say the planner warns "as the lane count approaches the bound", but code warns only at saturation.
**Suggestion:** Reword to an explicit upper bound ("at most ..., and less if LPT skews") and/or only warn when the policy is actually near-binding (e.g. any lane at `max_batches`). Align docs wording.

### [severity: nit] Inconsistent config access in the new warning
**File:** scripts/plan-tests.py:526-528
**Issue:** `enforce_batches_per_job` defensively uses `cfg.get(...)` with module defaults, but the adjacent warning dereferences `cfg["unit_max_batches_per_job"]` / `cfg["unit_batch_max_classes"]` directly. Any caller constructing a config without those keys (as the helper explicitly tolerates) would `KeyError` in the warning instead of degrading.
**Suggestion:** Use `.get(..., MODULE_DEFAULT)` consistently, matching the helper.

## Verdict
REQUEST_CHANGES

Note: this PR is CI tooling (Python/bash), so the Swift/iOS focus areas (retain cycles, force unwraps, MainActor, SwiftUI lifecycle, a11y) are not applicable — no Swift sources changed.

<a href="https://opencode.ai/s/g0IA3GoF">New%20session%20-%202026-09-17T00%3A19%3A57.740Z</a>
[opencode session](https://opencode.ai/s/g0IA3GoF)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35166035971)

kaishi00:
## Summary
Adds a planner policy (`MAX_UNIT_BATCHES_PER_JOB = 4`) that raises the unit lane count until every LPT lane holds at most four sequential 7-class xcodebuild batches, leaving the #184 execution model, runner, and watchdog budgets untouched. The policy logic and test coverage are generally solid, but one advertised fail-closed path (invalid `unit_batch_max_classes`) still crashes planning.

## Findings

### [severity: major] `--unit-batch-max-classes 0` still crashes before `validate_plan` can report it
**File:** scripts/plan-tests.py:469-473 (early return) and scripts/plan-tests.py:387-389 (crash)
**Issue:** The early-return comment in `enforce_batches_per_job` says an invalid value is handled "exactly like `unit_batch_max_classes < 1`", but that is not true. When `unit_batch_max_classes < 1`, the helper now correctly skips the floor, but `build_plan` still calls `unit_batches_for(classes, unit_est, cfg)` unconditionally (line 538), which evaluates `range(0, len(classes), size)` with `size = cfg["unit_batch_max_classes"] == 0` and raises `ValueError: range() arg 3 must not be zero`. `validate_plan`'s `unit_batch_max_classes must be >= 1` check (line 621) runs only after `build_plan` returns, so it is unreachable for exactly the input it exists to catch, and `main` has no guard/try-except. The PR body's "invalid policy value is a structured `::error::`, never a crash" guarantee therefore holds only for `unit_max_batches_per_job`, not its sibling. There is also no test exercising `unit_batch_max_classes=0`.
**Suggestion:** Reject `a.unit_batch_max_classes < 1` with a structured error before calling `build_plan` (e.g. in `_cfg_from_args`/start of `main`), or make `unit_batches_for` return a single batch / clamp when `size < 1` so `validate_plan` can fail closed. Fix the misleading comment and add a regression test mirroring `test_invalid_policy_value_fails_closed_without_crashing`.

### [severity: minor] Invalid `unit_max_batches_per_job` is not caught when the unit inventory is empty
**File:** scripts/plan-tests.py:702-705
**Issue:** The `unit_max_batches_per_job < 1` check lives inside `for lane in plan["unit_lanes"]`. With an empty unit inventory (`enforce_batches_per_job` returns `0, []`), the loop body never runs, so `--unit-max-batches-per-job 0` validates clean and writes artifacts containing the invalid policy value. This contradicts the fail-closed guarantee and is asymmetric with `unit_batch_max_classes`, whose check is at top level (line 621). When multiple lanes do exist the error is also appended once per lane (duplicated).
**Suggestion:** Hoist the check to the top of `validate_plan`, next to the `unit_batch_max_classes` check, and keep only the per-lane count comparison in the loop.

### [severity: minor] Saturation warning overstates the guaranteed class bound and fires on routine plans
**File:** scripts/plan-tests.py:520-528; docs/CI.md:129-133
**Issue:** The warning fires whenever `n_lanes >= max_lanes` and announces the suite is "bounded at `max_lanes * unit_max_batches_per_job * 7` = 224 classes". That product is only a naive upper bound: because LPT balances load, not counts, a skewed inventory can fail the policy closed far below 224 (the new `test_skewed_estimates_are_rebalanced_until_assignment_fits` demonstrates exactly this). It also fires when the timing model — not the policy — happens to pick `max_lanes` (e.g. 40 heavy classes pin the count at 8), emitting a "growth will fail closed" warning that is not near true. The docs claim the planner "warns as the lane count approaches the bound", but the code only warns at exact saturation.
**Suggestion:** Reword to an explicit upper bound ("at most …, and less if LPT skews") and/or warn only when the policy is actually near-binding (e.g. some lane already at `max_batches`), then align the docs wording.

### [severity: nit] Inconsistent config access in the new warning
**File:** scripts/plan-tests.py:526-528
**Issue:** `enforce_batches_per_job` defensively uses `cfg.get("unit_max_batches_per_job", MAX_UNIT_BATCHES_PER_JOB)` / `cfg.get("unit_batch_max_classes", ...)`, but the adjacent warning dereferences `cfg["unit_max_batches_per_job"]` and `cfg["unit_batch_max_classes"]` directly. A caller constructing a config without those keys (which the helper explicitly tolerates) would `KeyError` in the warning instead of degrading.
**Suggestion:** Use `.get(..., MODULE_DEFAULT)` consistently in the warning.

### [severity: nit] Stale report example contradicts the new cap
**File:** docs/CI.md:384 (pre-existing from #184, not touched by this PR)
**Issue:** The "Unit lane batches" example reads `batch 3/5 watchdog -> retry PASS`, which can no longer occur now that `MAX_UNIT_BATCHES_PER_JOB = 4`. The new docs block (120-133) documents the 4-batch cap, so the example is internally inconsistent.
**Suggestion:** Change the example to `batch 3/4` (or fewer). Similarly, the `unit_batch_timeout` formula example at docs/CI.md:307 (`invocation_overhead 240s + predicted x 2.5`) does not match `unit_batch_timeout_for` (`(invocation_overhead + predicted) x 2.5`); both are pre-existing from #184 and outside this PR's diff, but worth fixing while the file is being edited.

## Verdict
REQUEST_CHANGES

No Swift/iOS sources changed, so retain cycles, force unwraps, MainActor/Swift 6 concurrency, SwiftUI lifecycle, accessibility, and memory-leak focus areas are not applicable.

<a href="https://opencode.ai/s/Q57t6wW9">New%20session%20-%202026-09-17T01%3A10%3A35.675Z</a>
[opencode session](https://opencode.ai/s/Q57t6wW9)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35169445133)

PR #189, opened 2026-09-17
feat(rooms): ANVIL Room read/replay client seam (H1 wire contract)

naytewilson:
## Summary
Reconciles the historical Room seam (`2ad81c83`) onto current `main` (`edd146b`):

- `RoomModels` — H1 wire-contract models (`ProjectedRoom`, `RoomParticipant`, `RoomEvent`, `RoomEventPage`)
- `RoomProjectionClient` — `/api/v1/rooms`, `/{roomId}`, `/{roomId}/events` with classified failures (401/403/no-grant/not-found/projection-unavailable/transport), response bounds, transport seam
- `RoomReplayCoordinator` — snapshot + cursor-paged replay, `(room_id, room_seq)` dedupe, durable-before-ACK ordering, stale-on-failure semantics
- `RoomReplayStore` — per-(dashboard, room) persisted cursor/snapshot/tail in UserDefaults, bounded windows
- `KeychainHelper.Backend` seam — production system keychain; tests install `InMemoryKeychainBackend` (simulator host lacks keychain entitlements, `-34018`)
- Dashboard-scoped Room Hub credentials (`roomScopedAccount`); CF access bound to hub origin
- `AppLanguage.swift` ported for `AppLocalization` (seam dependency; self-contained)

#### Adaptations from the upstream seam
- Dropped unrelated `Localizable.xcstrings` changes (fork deleted that file; seam uses no localized strings)
- `loadCloudflareAccess(for:)` — origin-bound variant; fork has no dashboard-scoped CF store
- No `scopedAccount`/dashboard-registry import — local `roomScopedAccount` helper instead

#### Test plan
- [x] `RoomHubCredentialTests`, `RoomProjectionClientTests`, `RoomReplayCoordinatorTests` — pass on Neo simulator (signing disabled)
- [x] Full `ConduitTests`: 131 tests, 0 failures
- [x] Live replay vs Hub H1 (`bc4679c`, tailnet :3100): campaign Room `3955c547` — 14 committed events, cursor=15, seq=13 gap preserved, restart restore, credential isolation (driver test, uncommitted)
- [ ] Device/signing validation (provisioning profile unavailable on build host)

Generated with [Devin](https://devin.ai)

## Summary by CodeRabbit

* **New Features**
  * Added support for viewing ANVIL Hub rooms, participants, snapshots, and event history.
  * Added reliable replay synchronization with resume support, duplicate prevention, and stale-state indicators.
  * Added in-app language selection for system, English, and Simplified Chinese.
  * Added dashboard-specific secure storage for Room Hub credentials.
  * Added clear handling for authentication, permissions, unavailable rooms, and connection errors.
* **Tests**
  * Added comprehensive coverage for room loading, replay, localization-related storage, credential isolation, and error scenarios.


naytewilson:
Closing: retargeted to the fork's own main — this seam is campaign work that lands in naytewilson/hermes-conduit, not upstream.

kaishi00:
## Summary
Adds a dashboard-scoped, read-only client seam for the Hub H1 ANVIL Room projection: wire models, a classified HTTP client with an injectable transport, a durable per-(dashboard, room) replay store, a MainActor replay coordinator, a Keychain backend test seam, and credential isolation tests. No UI wiring or production call sites are added.

## Findings

### [severity: major] Unbounded replay loop can spin forever on a misbehaving/regressing server
**File:** Conduit/Services/RoomReplayCoordinator.swift:134
**Issue:** `while true` is bounded only by `page.hasMore`. If the Hub returns `hasMore == true` with a `nextCursor` that does not advance (or regresses below the committed cursor), `after = page.nextCursor` re-issues the same request indefinitely. There is no page cap, no progress assertion, and `after` can move *backwards* since the committed cursor is `max(current, page.nextCursor)` while the request cursor is unconditionally `page.nextCursor`.
**Suggestion:** Add a maximum page count (e.g. derived from `latestSeq`) and an explicit progress guard that throws/breaks when `page.nextCursor <= after`, plus never set `after` below the committed cursor (`after = max(page.nextCursor, committedCursor)`). Server-side `events.count == 0 && hasMore` should also terminate.

### [severity: major] Durable event tail truncation silently drops history while reporting `.live`
**File:** Conduit/Services/RoomReplayStore.swift:119 / Conduit/Services/RoomReplayCoordinator.swift:218
**Issue:** `commitPage` persists only `appliedEvents.suffix(500)`, and the coordinator's in-memory `events` array is never trimmed. After a process restart, `projection.events` contains only the last 500 events, but the cursor is at the high-water and `freshness` is reported `.live`. Because replay only ever fetches `room_seq > cursor`, the earlier events can never be re-fetched — history beyond 500 is permanently lost from the projection with no indication.
**Suggestion:** Either keep the restore window large enough to be authoritative and surface a "history truncated" state, or stop persisting events entirely and re-fetch from 0 (or a snapshot endpoint) on cold start. Do not mark `.live` when the rendered projection is known to be incomplete.

### [severity: minor] Closed enums make the whole response fail forward-compatibility
**File:** Conduit/Models/RoomModels.swift:20-33
**Issue:** `RoomStatus` and `RoomEventKind` are fixed `String` enums. A single unknown value from a newer Hub causes `JSONDecoder` to throw, which the client maps to `.undecodable` for the *entire* page/list — one new event kind loses all events in the response. The seam is explicitly meant to pass authority projections through verbatim.
**Suggestion:** Add an `unknown`/`other(String)` case via a custom `init(from:)` (or decode the raw kind as `String` and expose a typed accessor), so unknown authority values degrade instead of failing the whole payload.

### [severity: minor] `markSyncFailed` overwrites `lastSyncAt` with the failure time
**File:** Conduit/Services/RoomReplayStore.swift:129-134
**Issue:** On failure, `record.lastSyncAt = at` stamps the *failed attempt* time, not the last successful sync. After restore, `Projection.lastSyncAt` therefore reports a time when no data was actually synced, and it diverges from the in-memory projection (which keeps the prior successful timestamp).
**Suggestion:** Keep `lastSyncAt` for successful commits only; track the failure time separately (e.g. `lastSyncAttemptAt`) or derive staleness from `lastSyncError` alone.

### [severity: minor] `restore` clobbers the session-exact in-memory dedupe set
**File:** Conduit/Services/RoomReplayCoordinator.swift:101-118
**Issue:** `restore` unconditionally reassigns `appliedKeys[roomID]` from the durable bounded window. The type's own comment states the in-memory set is intentionally unbounded so a long session cannot regress. Any caller invoking the public `restore` mid-session shrinks that set; events applied earlier but outside the durable window can then be re-applied.
**Suggestion:** Make `restore` union into `appliedKeys[roomID]` (or restrict it to cold start), rather than replacing it.

### [severity: minor] `nonisolated(unsafe)` mutable global backend
**File:** Conduit/Services/AppState.swift:8069-8075
**Issue:** `backend` is a mutable global protocol existential accessed from arbitrary threads by `save`/`load`/`delete`, with the `unsafe` annotation suppressing the Sendable check. The safety argument rests entirely on XCTest running serially; any parallel test execution or a stray production call to `useBackendForTesting` is a data race.
**Suggestion:** Gate the mutator with `#if DEBUG`/test-only visibility, and prefer `@TaskLocal`, a lock, or an actor-isolated injectable store over a `nonisolated(unsafe)` global.

### [severity: minor] Response size bound is applied after the whole body is buffered
**File:** Conduit/Services/RoomProjectionClient.swift:233-235
**Issue:** `result.0.count <= maxResponseBytes` runs only after `URLSession` has already buffered the full response, so it does not bound peak memory (24 MB default, and the coordinator loops over many pages). It is a post-hoc rejection, not a guard.
**Suggestion:** Use `session.bytes(for:)`/streaming and abort once the byte count exceeds the limit, or document that the bound is validation-only.

### [severity: nit] Force unwraps in failure classification
**File:** Conduit/Services/RoomProjectionClient.swift:251,265,267,269
**Issue:** `problem!` is force-unwrapped inside `case` branches whose pattern implies `problem != nil`, so it currently cannot trap — but it couples correctness to the switch pattern and will crash if the classification is refactored.
**Suggestion:** Bind the value: `case "capability_denied" where problem != nil:` or `if case .some(let p) = problem { ... }`.

### [severity: nit] `outOfOrderDeliveries` undercounts
**File:** Conduit/Services/RoomReplayCoordinator.swift:206
**Issue:** `zip(...).contains(where:)` increments the counter once per page regardless of how many inversions occurred; a single descending pair and a fully reversed page report the same value. This is a diagnostic surface.
**Suggestion:** Count the inversions (`reduce(0) { ... }` / explicit loop) or rename the field to `pagesWithOutOfOrderDelivery`.

### [severity: nit] Test-only keychain hook lives in the production target
**File:** Conduit/Services/AppState.swift:8071-8075
**Issue:** `useBackendForTesting` is compiled into the shipping app and can swap credential storage at runtime.
**Suggestion:** Wrap in `#if DEBUG` or move the seam behind an internal initializer that tests can reach without a global mutator.

## Verdict
REQUEST_CHANGES

The core contract work (dedupe by `(room_id, room_seq)`, durable-before-advance cursor commit, per-dashboard credential scoping, classified failures, no force-unwrap/retain-cycle hazards) is sound. The unbounded pagination loop and the silent post-restart history truncation reported as `.live` should be resolved before this seam is relied upon.

<a href="https://opencode.ai/s/qGQkd1cU">New%20session%20-%202026-09-17T22%3A22%3A56.442Z</a>
[opencode session](https://opencode.ai/s/qGQkd1cU)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35281628873)

PR #198, opened 2026-09-20
fix(ui): build-146 regressions — Voice "On this iPhone", card tap targets, Bots list styling

kaishi00:
Build-146 UI/settings regressions. Three logically separate commits; no session
lifecycle/navigation architecture changes (Bot Mode session restore/resume is a
separate worker's branch).

## 1. Voice: "On this iPhone" could not be selected

**Root cause — pinned to the commit that introduced it.** PR #165 (CarPlay
Voice V1, `fefcae8`, 2026-09-12) narrowed `AppState.hasActiveVoiceSurface` from
`isSceneActive || isCarPlayVoiceSurfaceActive` to
`(isSceneActive && showVoiceSheet) || isCarPlayVoiceSurfaceActive`. That
property was already the single "foreground" flag the voice controller
consulted, and three consumers had **app-foreground** intent, not
voice-surface intent:

| consumer | why it broke |
|---|---|
| `requestOnDeviceTranscriptionPermissions()` | reached from Settings → Voice when selecting "On this iPhone"; the Voice sheet is closed there, so the gate was false and the request was refused |
| `runTranscriptionTest()` (Record ASR) | same screen, same closed sheet → refused with "Voice tests only run while Conduit is in the foreground." |
| `runSpeechTest()` | same |
| `isCurrent(_:)` (in-flight runtime fence) | discarded the provider tests' own progress mid-run |

The user-visible symptom of the permission path is exactly "the option cannot be
selected": `setVoiceTranscriptionMode` returned `false`, the view only surfaces a
message for the three *availability* outcomes (not for a refused foreground
check), so the picker simply never changed — no error, no state change.
Build 146 is the first TestFlight build containing #165 (build 145 tag is
2026-09-07), which matches it being reported as a new regression. Build 145 and
earlier passed `isSceneActive` into the same guard and worked.

**Fix.** Split the one flag into the two facts it was carrying:

- `isForegroundActive` remains the **capture** gate (a Voice surface presents).
  `startListening` / `resumeMicrophone` stay gated on it, so #165's contract is
  preserved: no capture while no surface presents.
- `isApplicationForegroundActive` is the **app-foreground** fact used by the
  permission request, both provider tests, and `isCurrent`.
- `AppState.publishVoiceRuntimeGates()` publishes both together from the same
  surface bookkeeping at every lifecycle boundary (`.active`, `.background`,
  CarPlay surface edges, `reassertVoiceSurfaceGate`, and the attach path), so no
  writer can open one gate without the other.

Removing the surface term from `isCurrent` does not weaken suspension: teardown
is already fenced by the operation generation (`stop()` and
`suspendRuntimeForLifecycle()` both bump it via `releaseRuntimeResources()`), so
superseded/torn-down work is still discarded.

## 2. Connection/setup cards and tap targets

**What I verified first, because it changes the story.** I built a UI test that
opens the setup flow and selects every step by tapping the card's own edges
(`card.frame.minX + 8`, `card.frame.maxX - 8`) rather than its glyphs, then ran
it **with and without** the content-shape changes. It passes both ways on
iOS 26.5: the shipped answer/method/login cards already resolve taps across
their visible area, so the *literal* Connection Setup wizard cards are not
reproducible as a dead region here, and the two files' history shows no
hit-target-related change between build 145 and 146.

**What the audit did find** is the one card-shaped selection surface whose
visible card is drawn **outside** its button: `ProfilePickerRow` (the
profile/dashboard chooser). Its select button wraps only the text column while
`.padding(12)` + the glass card sit on the containing `HStack`, so the card's
padding, the trailing accessory column, and the space beside the name were all
dead — tap the text or nothing, which is the reported symptom verbatim. The card
now selects the profile itself; the nested controls (avatar, photo removal,
reorder arrows) keep their own taps, and a single `canSelect` rule keeps the two
selection paths in agreement.

The explicit card-wide `contentShape` is also declared on the setup answer/method
cards, the login screen's setup card, the app-icon tiles, and the notification
command card, so the contract ("the entire visible card is the target") is
stated rather than inherited from how SwiftUI happens to derive the label's hit
region. Those are contract declarations, not a reproduced fix.

## 3. Bots list styling

The Bots tab rendered as one sharp-edged slab: a plain `List` with the system row
chrome (background, separators, insets) still applied, directly under the
sidebar's rounded glass tab bar. Rows now carry their own rounded card using the
treatment `SessionRow` and `CronJobRow` give the sibling tabs (fill
`Color.primary.opacity(0.045)`, `RoundedRectangle(cornerRadius: 18, style:
.continuous)`, `Color.white.opacity(0.08)` border) with
`listRowBackground(.clear)` / `listRowSeparator(.hidden)` / `listRowInsets`
matching, and the roster's notice row gets the matching rounded treatment
instead of a bare row. Row taps use the existing `contentShape`; accessibility
label/hint unchanged.

I chose individually rounded rows over one rounded section because both sibling
tabs in the same sidebar (Sessions, Cron) do it that way and share the same row
geometry.

## Files changed

- `Conduit/Voice/VoiceConversationController.swift` — gate split; permission/
  provider-test/`isCurrent` consumers moved to the app-foreground gate.
- `Conduit/Services/AppState.swift` — `hasForegroundApplicationSurface`,
  `publishVoiceRuntimeGates()`, gate publication at every boundary + attach.
- `Conduit/Views/ConnectionSetupView.swift`, `LoginView.swift`,
  `AuxiliaryViews.swift` — explicit card-wide hit shapes.
- `Conduit/Views/ProfilePickerSheet.swift` — card-wide selection + `canSelect`.
- `Conduit/Views/BotRosterView.swift` — rounded roster rows and notice row.
- Tests: `VoiceConversationControllerTests`, `AppStateVoiceCapabilityTests`,
  `VoiceBackgroundSuspensionTests` (harness helper now models a foreground return
  for both gates), `ConnectionSetupUITests`.

## Shared root cause?

Not a single shared component. Two distinct causes: (a) one flag carrying two
facts, consumed by three settings-time actions plus the runtime fence — a shared
*cause* with multiple symptoms, fixed at the gate; (b) the "visible card drawn
outside the button" shape, which only `ProfilePickerRow` has among selectable
cards. The other call sites audited (sidebar session/project rows, source-filter
chips, saved-dashboard rows, Kanban card, model/profile pickers) apply their
padding and background inside the label and are unaffected.

## Tests / build

- Full unit suite on ios-mac (iPhone 17, iOS 26.5): **2709 tests, 0 failures**.
- `ConduitUITests/ConnectionSetupUITests` (5 tests incl. the new one): **all pass**.
- New unit tests go RED against the pre-fix code (mutation-checked on the Mac):
  reverting the permission guard, reverting `isCurrent`, and removing the
  `.active` gate publication each fail exactly the test meant to pin them.
- UI A/B for #2 as described above.

## Follow-ups (not in this PR)

- The `Reorder` pill in the profile picker has the same "padding outside the
  button" shape (small dead region); left alone because fixing it means touching
  the localized `Button(String)` initializer.
- Kanban rows without a declared hit shape (audit list in the commit notes);
  unrelated to these regressions.
- Pre-iOS-26 devices use the `.ultraThinMaterial` fallback instead of
  `glassEffect`; the explicit content shapes now cover both rendering paths, but
  no pre-26 simulator was available to A/B the fallback here.

## Summary by CodeRabbit

- **New Features**
  - Voice permissions and provider tests now work whenever the app is foregrounded, without requiring an active Voice screen.
  - Profile rows can be selected by tapping anywhere on the card body.
  - Setup options, login cards, app tiles, and notification cards respond across their full surfaces.

- **UI Improvements**
  - Updated bot roster styling with rounded cards and refined spacing.

- **Bug Fixes**
  - Improved Voice behavior across foreground, background, and CarPlay transitions.


kaishi00:
## Summary
Splits the single Voice "foreground" gate into a capture gate (`isForegroundActive`) and an app-foreground gate (`isApplicationForegroundActive`) so Voice settings can request permissions/run provider tests, and adds card-wide hit shapes plus rounded Bots rows. The gate split is well-reasoned and fenced by generation bumps; no force unwraps, retain cycles, or actor violations were introduced.

## Findings

### [severity: minor] `setCarPlayVoiceSurfaceActive` publishes only one of the two gates
**File:** Conduit/Services/AppState.swift:1040-1046
**Issue:** `isCarPlayVoiceSurfaceActive` feeds *both* `hasActiveVoiceSurface` (capture gate) and `hasForegroundApplicationSurface` (app gate), but this setter only calls `setApplicationForegroundActive(...)`. It updates the capture gate nowhere. The doc comment on `publishVoiceRuntimeGates` (lines 1076-1079) says a CarPlay edge "publishes by the same rule," and the PR body claims "no writer can open one gate without the other," which this setter contradicts. Today the inconsistency is masked because `CarPlayVoiceCoordinator` always follows with `handleCarPlayVoiceSurfaceActivated()` / `handleCarPlayVoiceSurfaceRemoved()`, so the capture gate converges in the same runloop — but the setter is now a latent trap for any future caller.
**Suggestion:** Call `publishVoiceRuntimeGates()` (or at minimum `voiceConversationController.setForegroundActive(hasActiveVoiceSurface)`) in this setter, or correct the comment to state that the CarPlay edges intentionally publish only the app gate and rely on the adjacent handler for the capture gate.

### [severity: minor] No regression test for the one card that actually had the dead region
**File:** ConduitUITests/ConnectionSetupUITests.swift:277-341
**Issue:** The reproduced defect was `ProfilePickerRow` (padding/accessory column outside the select button), but the new UI test exercises only the Connection Setup/login cards that the author states are *not* reproducible as dead regions. The fixed surface therefore has no direct coverage, while the tests added assert a contract that already held. `ProfilePickerRow` is `private` and not otherwise unit-tested.
**Suggestion:** Add a UI test that taps the profile card's padding/accessory area and asserts the profile switches, so the actual fix is pinned; the setup-card test can remain as a contract declaration.

### [severity: minor] Card-wide tap is not exposed to accessibility; behavior differs from the sibling Kanban pattern
**File:** Conduit/Views/ProfilePickerSheet.swift:161-171
**Issue:** The new `.onTapGesture` makes the whole card selectable for sighted users, but unlike the equivalent card in `KanbanViews.swift:1935-1942` it adds no `.accessibilityElement(children: .combine)` / `.accessibilityAddTraits(.isButton)`. VoiceOver users still only get the (disabled-when-current) text button, so the newly-tappable region is invisible to assistive tech. Also, because the nested avatar/trash/reorder controls are `Button`s rather than explicit gesture consumers, the claim that they "never reach this gesture" depends on SwiftUI button gesture precedence (which the Kanban card relies on too, so it is likely fine — but it is implicit, not declared).
**Suggestion:** Mirror the Kanban card's accessibility treatment (combine children + `.isButton` + an accessibility action) or document why the inner button suffices.

### [severity: nit] Notice row gets double vertical padding
**File:** Conduit/Views/BotRosterView.swift:49-55 and 291
**Issue:** `BotModeNoticeRow` already applies `.padding(.vertical, 2)` (line 291) and the call site adds `.padding(.vertical, 10)`, giving 12pt vertical while `BotRosterRow` uses 10pt. Likely unintentional.
**Suggestion:** Drop the internal `.padding(.vertical, 2)` on `BotModeNoticeRow` or reduce the call-site padding so both rows share geometry.

### [severity: nit] `hasForegroundApplicationSurface` can diverge from the published app gate
**File:** Conduit/Services/AppState.swift:6652-6714
**Issue:** `.inactive` sets `isSceneActive = false` but deliberately does not publish the app-foreground gate (correct, since the permission alert itself drives `.inactive`). The consequence is that the computed `hasForegroundApplicationSurface` can read `false` while `isApplicationForegroundActive` remains `true`. The next writer that recomputes from the computed property (e.g. an intervening `setCarPlayVoiceSurfaceActive`) can therefore close the gate while the app is still foreground-but-inactive.
**Suggestion:** Add a brief comment at the `.inactive` branch stating the app gate is intentionally left at its last value, so a future maintainer does not "fix" it into publishing and refuse the in-flight permission request.

## Verdict
APPROVE

The core gate split is correct: `isCurrent` no longer depends on the capture gate, and every path that closes the Voice surface (`showVoiceSheet = false`, suspend) already bumps `operationGeneration` via `stop()`/`endVoiceConversation`, so removing the surface term does not un-fence stale work. Remaining findings are minor/nit.

<a href="https://opencode.ai/s/FG2yPen1">New%20session%20-%202026-09-20T12%3A19%3A21.902Z</a>
[opencode session](https://opencode.ai/s/FG2yPen1)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35510228569)

kaishi00:
Review follow-up on the five findings from the automated review (verdict APPROVE). All four actionable items are in `fe090f4`; the fifth is a documented constraint.

1. **`setCarPlayVoiceSurfaceActive` published only one gate** — fixed. It now calls `publishVoiceRuntimeGates()`, so both `hasActiveVoiceSurface` and `hasForegroundApplicationSurface` are applied together and a CarPlay edge is never half-applied. `handleCarPlayVoiceSurfaceActivated/Removed` still own the suspension side-effect and re-state the same capture value.

2. **No regression test for the one card that actually had the dead region** — accepted, and documented in `ProfilePickerRow`'s doc comment rather than left silent. The picker is only reachable with a connected dashboard **and** more than one discovered profile. The UI-test seam `-CONDUIT_UI_TEST_CONNECTED_DASHBOARD` provides a connected snapshot with no transport (`uiTestConnectedStub`), which yields no `profiles.list` result, so the sheet renders exactly one row — the current profile — whose selection is disabled by design. Reaching a selectable row needs a new profiles-seeding seam; adding a production DEBUG seam to test a hit region is a larger change than the fix, so this stays a documented coverage gap. The setup-card UI test remains as the contract declaration for the card idiom.

3. **Card-wide tap not exposed to accessibility** — documented in the same comment: VoiceOver keeps the inner select button (profile name, disabled state from `canSelect`). The Kanban card's `children: .combine` is wrong here specifically because this row carries two further controls (choose photo, remove photo) that combining would fold into the card and hide; the Kanban card has no such controls.

4. **Notice row double vertical padding** — fixed; `BotModeNoticeRow` no longer pads itself, so the notice row and roster rows share the call site's geometry.

5. **`hasForegroundApplicationSurface` can diverge from the published gate at `.inactive`** — documented at the `.inactive` branch: leaving the gate at its last value is intentional, because that branch runs while iOS presents the microphone/Speech permission alert the settings flow awaited, and closing the gate mid-request would refuse that request's own follow-up work (the second permission, then the provider test).

Verification after the follow-up: full unit suite 2709/0 and `ConnectionSetupUITests` 5/5 on ios-mac (iPhone 17, iOS 26.5).

kaishi00:
## Summary
Splits Voice's single "foreground" gate into a capture gate (`isForegroundActive`) and an app-foreground gate (`isApplicationForegroundActive`) so Voice settings permission/test actions work with the sheet closed, and adds card-wide hit shapes plus rounded Bots roster rows.

## Findings

### [minor] App-foreground fact diverges between the computed property and the published gate at `.inactive`
**File:** Conduit/Services/AppState.swift:6716-6755, 1075-1086
**Issue:** `.inactive` sets `isSceneActive = false` but deliberately does not republish, so `hasForegroundApplicationSurface` reads `false` while the controller's `isApplicationForegroundActive` stays `true`. Any writer that recomputes from the property during this window (`setCarPlayVoiceSurfaceActive`, `reassertVoiceSurfaceGate` via `attachToLiveVoiceConversation`) will close the app gate while the app is still foreground-but-inactive — potentially refusing the in-flight permission request's follow-up (second permission, provider test) that the `.inactive` comment is specifically trying to protect. The follow-up comment documents the intent but does not make it robust.
**Suggestion:** Track a dedicated flag that is true for both `.active` and `.inactive` and false only on `.background` (e.g. `isApplicationActive`), and derive both `hasActiveVoiceSurface` and `hasForegroundApplicationSurface` from it. That removes the divergence instead of documenting it.

### [minor] The actual fixed surface (`ProfilePickerRow`) has no automated coverage
**File:** Conduit/Views/ProfilePickerSheet.swift:74-100; ConduitUITests/ConnectionSetupUITests.swift:277-341
**Issue:** The reproduced dead region was the profile card (padding/accessory column outside the select button). The added UI test exercises only the Connection Setup/login cards that the author states already resolved taps, so the real fix is unpinned. The card's `select()` could regress without a red test.
**Suggestion:** Seed a second profile for the UI-test seam (or extract the `canSelect`/hit-region predicate into a testable helper) and add a test that taps the card's padding/accessory area and asserts selection. At minimum, keep the coverage gap explicit in a tracking issue.

### [minor] Card-wide tap depends on implicit button-gesture precedence and is invisible to VoiceOver
**File:** Conduit/Views/ProfilePickerSheet.swift:169-186
**Issue:** `.onTapGesture` on the card relies on the nested `Button`s (avatar, remove photo, reorder arrows) winning the gesture race. If SwiftUI ever routes a tap on a nested control to the card gesture, the user would switch profiles instead of opening the picker/removing a photo. Separately, unlike the equivalent Kanban card (`KanbanViews.swift:1939-1942`), the new region is not exposed as an accessibility button/action, so VoiceOver only sees the text-column button.
**Suggestion:** Make the precedence explicit (e.g. attach the selection to the inner button's row and use `.contentShape` on the card, or use a `highPriorityGesture`/`simultaneousGesture` strategy) and add an accessibility action/hint for the card region, or state the precedence assumption in the comment.

### [nit] Stale comment on `interruptAssistantPlayback`'s fence
**File:** Conduit/Voice/VoiceConversationController.swift:405
**Issue:** The comment says `isCurrent` "also covers foreground loss and session teardown," but `isCurrent` now uses the *app-foreground* gate, not the capture gate; capture/surface loss is no longer one of its terms.
**Suggestion:** Reword to "covers app-backgrounding and session teardown" to match the new `isCurrent` definition at line 1280.

## Verdict
APPROVE

The gate split is correct and well-fenced: `isCurrent` no longer depends on the capture gate, and every path that closes/torndowns the Voice surface bumps `operationGeneration`. Findings are minor/nit and none block merge.

<a href="https://opencode.ai/s/3qLYSCpI">New%20session%20-%202026-09-20T12%3A37%3A09.309Z</a>
[opencode session](https://opencode.ai/s/3qLYSCpI)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35511050944)

kaishi00:
## Summary
Splits the single Voice "foreground" gate into a capture gate (`isForegroundActive`) and an app-foreground gate (`isApplicationForegroundActive`) so Voice settings permission/provider-test actions work while no Voice sheet presents, and adds card-wide hit shapes plus rounded Bots roster rows.

## Findings

### [minor] App-foreground gate and its computed source diverge during `.inactive`
**File:** Conduit/Services/AppState.swift:1075, 6757-6764
**Issue:** The published app gate is intentionally left at its last value on `.inactive` (good — the permission alert drives `.inactive`), but `hasForegroundApplicationSurface` is computed from `isSceneActive`, which `.inactive` sets to `false`. So during an overlay dip the stored gate says foreground while the computed fact says background. Any writer that republishes during that window — `reassertVoiceSurfaceGate()` via `attachToLiveVoiceConversation()`/`closeVoiceConversation()`/CarPlay edges, or `setCarPlayVoiceSurfaceActive` — recomputes from `hasForegroundApplicationSurface` and can close the gate mid-request, fencing the `isCurrent`-checked continuation the `.inactive` comment is specifically trying to protect. Documenting the intent does not make it robust; the next writer bypasses the documentation.
**Suggestion:** Track a stored fact that is true for `.active` **and** `.inactive` and false only on `.background` (e.g. `isApplicationActive`), and derive both `hasActiveVoiceSurface`/`hasForegroundApplicationSurface` (and the published gate) from it, so the computed property and the gate cannot disagree.

### [minor] The surface that actually had the dead region (`ProfilePickerRow`) is untested, and the new UI test may not actually exercise the declared shape
**File:** Conduit/Views/ProfilePickerSheet.swift:182-191; ConduitUITests/ConnectionSetupUITests.swift:277-341
**Issue:** The reproduced defect was the profile card (padding/accessory column outside the select button), yet the only new UI test exercises the Connection Setup/login cards the author states already resolved taps. Additionally, `tapCardWhitespace` probes at `card.frame.minX + 8` / `maxX - 8`; for a SwiftUI `Button` XCUITest's `frame` frequently reflects the label's accessibility frame, not the extended `contentShape` region, so those points may still land on the label. The test can therefore pass whether or not the card-wide hit shape exists, pinning a contract that already held rather than the fix.
**Suggestion:** Seed a second profile for the `-CONDUIT_UI_TEST_CONNECTED_DASHBOARD` seam (or extract the `canSelect`/hit-region predicate into a testable helper) and assert selection from the card's padding/accessory area. If the setup-card test is kept as a contract declaration, state that explicitly and verify the probe point actually falls outside the accessibility element (assert `!card.frame.contains(point)` for the leading/trailing cases).

### [minor] Card-wide selection depends on implicit SwiftUI gesture precedence and is invisible to VoiceOver
**File:** Conduit/Views/ProfilePickerSheet.swift:166-191
**Issue:** The `.background` button reaches only where nothing interactive sits above it, which relies on the nested avatar/trash/reorder `Button`s winning the hit-test race. This is implicit behavior, not a declared contract; if the layering changes, a tap on "remove photo" could instead switch profiles and dismiss the sheet. Separately, `.accessibilityHidden(true)` leaves the new card-sized region with no accessibility action — VoiceOver only sees the inner text button, so the newly tappable area is undiscoverable to assistive tech and inconsistent with the Kanban card (`KanbanViews.swift:1939-1942`).
**Suggestion:** Make precedence explicit (e.g. attach selection to the row via a `highPriorityGesture`/explicit hit-test path, or `.contentShape` on the card plus an overlay ordering that guarantees nested controls are above) and add an accessibility action/hint for the card region, or state the precedence assumption directly in the comment.

### [nit] Redundant app-gate write in `handleCarPlayVoiceSurfaceRemoved`
**File:** Conduit/Services/AppState.swift:16532-16541
**Issue:** The only reachable path into this handler is `setCarPlayVoiceSurfaceActive(false)` (which already calls `publishVoiceRuntimeGates()`) or `fence()`, which only calls it when `!hasActiveVoiceSurface` — and `hasActiveVoiceSurface` includes `isCarPlayVoiceSurfaceActive`, so the flag is already `false` there. The added `setApplicationForegroundActive(hasForegroundApplicationSurface)` therefore recomputes the value it was just given. Harmless, but the accompanying comment overstates that `fence()` "reaches here directly" with a stale flag.
**Suggestion:** Drop the redundant call or note that it is a defensive re-statement, not a corrective one.

### [nit] Stale comment on `interruptAssistantPlayback`'s fence
**File:** Conduit/Voice/VoiceConversationController.swift:405
**Issue:** The comment says `isCurrent` "also covers foreground loss and session teardown," but `isCurrent` now uses the *app-foreground* gate (line 1285); capture/surface loss is no longer one of its terms (it is covered by the generation bump). "foreground loss" is ambiguous and reads as the surface gate.
**Suggestion:** Reword to "covers app-backgrounding and session teardown," matching the new definition.

## Verdict
APPROVE

The core gate split is correct and well-fenced: the permission/provider-test/`isCurrent` consumers genuinely needed the app-foreground fact, and every realistic path that retires in-flight Voice work (`stop()`, `suspendRuntimeForLifecycle()`, `endVoiceSession()`, sheet dismissal via `closeVoiceConversation`) bumps `operationGeneration`, so removing the capture term from `isCurrent` does not un-fence stale work. No force unwraps, retain cycles, `@MainActor` violations, or leaks were introduced. Remaining findings are minor/nit and none block merge.

<a href="https://opencode.ai/s/JYukqS8c">New%20session%20-%202026-09-20T13%3A17%3A03.672Z</a>
[opencode session](https://opencode.ai/s/JYukqS8c)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35512986564)

kaishi00:
## Multi-model review — consolidated report

Three independent reviewers on the same diff (`origin/main...HEAD`). **Majority verdict: APPROVE (2) / REQUEST CHANGES (1)** — the single dissent raised the one BLOCKER, and it is fixed in `1fa7650`. Duplicate findings are merged; the per-reviewer summary is at the end.

### Consolidated findings, by severity

**[BLOCKER] `ProfilePickerRow` card selection might double-fire on its nested controls** — raised by Reviewer 1 (Reviewers 2 and 3 assert the opposite: that a nested `Button` consumes its own tap before a container gesture). **Resolved by removing the disputed construct rather than adjudicating SwiftUI precedence**: the container `.onTapGesture` is deleted and replaced with the reviewer's own suggested shape — a card-sized selection `Button` *behind* the row. A sibling behind the row cannot receive a tap that lands on a control above it under either reading, and it still covers the card's padding, the trailing accessory column, and the space beside the name. It is `.accessibilityHidden(true)`, so VoiceOver keeps exactly one selection target. If the double-fire claim were true, the shipped shape would have switched profiles (and dismissed the sheet) on every avatar/photo-removal tap; if it were false, the previous shape was fine. Neither risk remains.

**[WARNING] `handleCarPlayVoiceSurfaceRemoved()` depended on a prior `setCarPlayVoiceSurfaceActive(false)`** — Reviewer 1, verified: the CarPlay coordinator's `fence()` path reaches the handler directly. **Fixed**: the handler now states the app-foreground fact itself before suspending (idempotent, removes the implicit ordering coupling).

**[SUGGESTION] `closeVoiceConversation()` left the capture gate stale-true** — Reviewer 1. **Fixed**: it now restates both gates from bookkeeping after the sheet flag clears (a CarPlay-presented conversation keeps capture open by the same rule). Reviewer 1 also asked for an audit of the other direct `showVoiceSheet = false` sites — done, and none needs a restatement: every one of the seven sites is paired with `stop()` / `endVoiceSession()` / `setGateway(nil)` (which performs the authoritative stop), so the generation bump always fences the retired work.

**[SUGGESTION] The `publishVoiceRuntimeGates` invariant comment was overstated** — Reviewer 3, verified: the `.background` branch writes both facts explicitly, and `suspendVoiceConversationForBackground()` writes only the capture gate. **Fixed**: the comment now states the real invariant (both facts stated in one synchronous run, or the same step tears the runtime down).

**[SUGGESTION] `isCurrent` now trusts generation-bump discipline alone** — Reviewer 1. **Fixed** as documentation: `releaseRuntimeResources()` carries the INVARIANT that every retirement path goes through it.

**[NIT] `.inactive` rationale was partly wrong** — Reviewer 3, verified: the permission request checks the gate once at entry, so republishing at `.inactive` would not refuse *that* request; what it would break is `startListening`'s post-permission `isCurrent` continuation during the first-run microphone alert. **Fixed** (comment corrected to the accurate mechanism) **and now pinned** by a new AppState test, `testOverlayDipLeavesThePublishedVoiceGatesUntouched`, which Reviewer 3 noted was enforced only by the absence of a call.

**[NIT] `testProviderTestsRefuseWhileTheApplicationIsBackgrounded` asserted only message copy** — Reviewer 3. **Fixed**: both results now also assert `passed == false`.

**[NIT] `BotModeNoticeRow` vertical padding** — Reviewer 1 calls the removal unrelated; OpenCode's earlier review asked for exactly that removal to equalize row geometry. **Kept deliberately**, per the newer instruction, and noted here.

**[NIT] `tapCardWhitespace` probe offsets are resolution-brittle** — Reviewer 3, no change required: the bounded loop plus hard assertions make it fail loudly rather than flake silently. Left as is; if it ever retries a lane, the inset should be derived from the card width instead of the literal 8 pt.

### Testing gaps raised, and their disposition

- **`ProfilePickerRow` has no automated coverage** (Reviewers 1 and 3): acknowledged, and it is now the documented reason the fix ships the zero-double-fire structure rather than a precedence-dependent one. The picker needs a connected dashboard *and* more than one discovered profile; `-CONDUIT_UI_TEST_CONNECTED_DASHBOARD` supplies neither, so probing it would require a new profiles-seeding seam — larger than the fix. Deferred, documented in the source.
- **No test covered `attachToLiveVoiceConversation()` publishing the app-foreground gate** (Reviewer 3): deferred as a follow-up rather than asserted — the path needs a live voice session plus an attached gateway, so the test would exercise more of the harness than of the gate.
- **First-run permission-alert-over-auto-listen flow** (Reviewer 3): now partially pinned by the new `.inactive` test above.

### Verification

- Reviewer falsification attempts that **failed to falsify** (i.e. author claims held): non-vacuous tests, writer consistency, the Kanban-gesture and `SessionRow` design precedents, and scope containment (no Bot Mode lifecycle/navigation changes, `Conduit.xcodeproj` untouched, no new files).
- Reviewer 3's independent sweep for other consumers of the narrowed gate and other "card outside the button" shapes found none needing this PR.
- Local verification of the earlier commits: full unit suite 2709/0 and `ConnectionSetupUITests` 5/5 on ios-mac, plus mutation checks. **The `1fa7650` commit could not be re-run locally — ios-mac went unreachable mid-cycle — so its verification is CI's: `Build test products` passed on that commit, with the unit and UI lanes in flight.**

### Per-reviewer summary

| Reviewer | Verdict | Key theme |
|---|---|---|
| 1 | REQUEST CHANGES | Card gesture precedence risk (BLOCKER) + CarPlay gate-publishing coupling |
| 2 | APPROVE | Two-gate model correct, all consumers on the right gate, harness change legitimate |
| 3 | APPROVE | Could not falsify any claim; found comment-precision and test-hardening items |

kaishi00:
### CI on `1fa7650` — green

All lanes pass: 7 unit lanes, ui-1, ui-2, ui-3, Build test products, Plan & validate, CI tooling self-test, review, CI Gate.

Two red-then-green events on the way there, both infrastructure and both already documented in this repo's CI notes:

1. **ui-2 first attempt** — `LoginKeyboardUITests.testCloudflareFieldsAndConnectStayReachableWithKeyboardVisible` failed with `Failed to get matching snapshots: Timed out while evaluating UI query` (173.9 s, then 116.7 s on the lane's own targeted retry). This is the second, previously-recorded mechanism for that test (first seen 2026-09-08 on PR #150), where the in-run retry is not enough and `gh run rerun --failed` is the recovery. Attribution evidence, not just a hunch: the previous run on this branch (`35511051930`) executed the **identical** `ui-2/batch-a1` — same two classes, same simulator UDID — with that same test **passing** in 62.6 s, and `1fa7650` touches no login-path code (its changes are `ProfilePickerSheet`, the CarPlay-removed handler, `closeVoiceConversation`, comments and tests).
2. **ui-2 first re-run** — `xcodebuild: error: Unable to find a device matching the provided destination specifier` (runner simulator-pool race, the other recorded hosted transient). The lane failed without executing a single test.

The second re-run passed in 10m58s.

What matters for this PR: the new UI test `testSetupCardsSelectFromWhitespaceOffTheirText` passed in **both** runs where it executed (24.7 s and 19.1 s), and all five `ConnectionSetupUITests` pass in the hosted UI lane.

kaishi00:
## Summary
Splits Voice's single foreground flag into a capture gate (`isForegroundActive`) and an app-foreground gate (`isApplicationForegroundActive`) so Voice settings permission requests/provider tests work with no Voice sheet, and adds card-wide hit shapes plus rounded Bots roster rows. The gate split is sound and generation-fenced; findings are minor/nit.

## Findings

### [minor] Stored app-foreground gate can diverge from its computed source during `.inactive`
**File:** Conduit/Services/AppState.swift:1075-1077, 6780-6787
**Issue:** `.inactive` sets `isSceneActive = false` but deliberately does not republish the gates, so `hasForegroundApplicationSurface` (computed as `isSceneActive || isCarPlayVoiceSurfaceActive`) reads `false` while the published `isApplicationForegroundActive` stays `true`. Any writer that recomputes during that window — `setCarPlayVoiceSurfaceActive(false)` (which now calls `publishVoiceRuntimeGates()`), or `reassertVoiceSurfaceGate()` via an attach/close/CarPlay edge — will close the app gate while the app is still foreground-but-inactive, fencing the very `isCurrent`-checked continuation the `.inactive` comment exists to protect (the first-run mic alert over auto-listen, or a settings permission request's follow-up). The comment documents the intent but does not make the invariant hold.
**Suggestion:** Track a stored fact that is true for `.active` and `.inactive` and false only on `.background` (e.g. `isApplicationActive`), and derive both `hasActiveVoiceSurface`/`hasForegroundApplicationSurface` from it, so the computed property and the published gate cannot disagree.

### [minor] `ProfilePickerRow` doc comment now contradicts the added UI test
**File:** Conduit/Views/ProfilePickerSheet.swift:74-81
**Issue:** The comment states the picker "has no automated coverage" and that the UI-test seams "supply neither" a connected dashboard nor profiles, so "the card's hit region cannot be probed from a test." The head commit adds exactly that: `-CONDUIT_UI_TEST_PROFILES` (`AppState.swift:2765`) and `ConduitUITests/ProfilePickerUITests.swift`, which probes the card body/accessory region. The comment is stale and misleading to the next maintainer.
**Suggestion:** Update the comment to point at `ProfilePickerUITests` and the new seam, and drop the "no automated coverage" claim.

### [nit] Stale comment on `interruptAssistantPlayback`'s fence
**File:** Conduit/Voice/VoiceConversationController.swift:405
**Issue:** The comment says `isCurrent` "also covers foreground loss and session teardown," but `isCurrent` now uses the app-foreground gate (`isApplicationForegroundActive`, line 1285); capture/surface loss is no longer one of its terms (it is covered by the generation bump). "foreground loss" reads as the surface gate.
**Suggestion:** Reword to "covers app-backgrounding and session teardown" to match the new definition.

### [nit] Enlarged card target is invisible to VoiceOver
**File:** Conduit/Views/ProfilePickerSheet.swift:182-191
**Issue:** The new card-sized selection surface is `.accessibilityHidden(true)`, so VoiceOver only sees the inner text-column button; the padding/accessory region that sighted users can now tap has no accessibility counterpart. The same action remains reachable, so this is not a functional gap, but it diverges from the sibling Kanban card and means the card's actual tap area is larger than its accessible area.
**Suggestion:** Acceptable as-is given the nested controls, but consider exposing a selection accessibility action on the row (or state the merge rationale, which the comment already partly does).

## Verdict
APPROVE

The core gate split is correct: `isCurrent` no longer depends on the capture gate, the permission/provider-test consumers genuinely need the app-foreground fact, and every realistic retirement path (`stop()`, `suspendRuntimeForLifecycle()`, `endVoiceSession()`, sheet close) bumps `operationGeneration`. No force unwraps, retain cycles, `@MainActor` violations, or leaks were introduced. Remaining findings are minor/nit and do not block merge.

<a href="https://opencode.ai/s/X1d5bY1f">New%20session%20-%202026-09-20T15%3A30%3A22.345Z</a>
[opencode session](https://opencode.ai/s/X1d5bY1f)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35519722637)

kaishi00:
## Summary
Splits Voice's single foreground flag into a capture gate (`isForegroundActive`) and an app-foreground gate (`isApplicationForegroundActive`) so Voice-settings permission/test actions work with no Voice sheet, and adds card-wide hit regions plus rounded Bots roster rows.

## Findings

### [minor] App-foreground gate can diverge from its computed source during `.inactive`
**File:** Conduit/Services/AppState.swift:1075, 6748-6788
**Issue:** `.inactive` sets `isSceneActive = false` but deliberately does not republish the gates, so `hasForegroundApplicationSurface` (`isSceneActive || isCarPlayVoiceSurfaceActive`) reads `false` while the published app gate stays `true`. Any writer that recomputes during that window — `setCarPlayVoiceSurfaceActive(_:)` (now calling `publishVoiceRuntimeGates()`), `handleCarPlayVoiceSurfaceRemoved()`, or `reassertVoiceSurfaceGate()` via attach/close — closes the app gate while the app is foreground-but-inactive. That fences the `isCurrent`-checked continuation (`startListening` after the first-run microphone alert) the `.inactive` comment exists to protect. The comment documents the intent but does not enforce it.
**Suggestion:** Track a stored fact true for `.active` and `.inactive` and false only on `.background` (e.g. `isApplicationActive`), and derive both `hasActiveVoiceSurface`/`hasForegroundApplicationSurface` (and the published gate) from it, so the computed property and the gate cannot disagree.

### [minor] Enlarged profile-card tap region is invisible to VoiceOver
**File:** Conduit/Views/ProfilePickerSheet.swift:183-192
**Issue:** The card-sized selection surface is `.accessibilityHidden(true)`, so the padding/accessory region sighted users can now tap has no accessibility counterpart; VoiceOver only reaches the (smaller) text-column button. The action remains reachable, so this is not a regression, but it diverges from the sibling Kanban card (`KanbanViews.swift:1939-1942`), which exposes the whole card as an `isButton` with accessibility actions.
**Suggestion:** Add an accessibility action/hint for the card region, or explicitly state in the comment that the inner button is the intended sole VoiceOver target.

### [nit] Stale fence comment in `interruptAssistantPlayback`
**File:** Conduit/Voice/VoiceConversationController.swift:401-405
**Issue:** The comment says `isCurrent` "also covers foreground loss and session teardown," but `isCurrent` (line 1285) now keys on the *app-foreground* gate; capture/surface loss is no longer one of its terms (it is covered only by the generation bump). "foreground loss" reads as the surface gate.
**Suggestion:** Reword to "covers app-backgrounding and session teardown."

### [nit] Redundant app-gate restatement in `handleCarPlayVoiceSurfaceRemoved`
**File:** Conduit/Services/AppState.swift:16555-16563
**Issue:** This handler is only reachable when `!hasActiveVoiceSurface` (which already implies `isCarPlayVoiceSurfaceActive == false`) or from `setCarPlayVoiceSurfaceActive(false)`, which already published the app gate. The added `setApplicationForegroundActive(hasForegroundApplicationSurface)` therefore recomputes the value it was just given. Harmless and documented, but the comment overstates it as a correction.
**Suggestion:** Either drop the call or note that it is a defensive re-statement, not a corrective one.

### [nit] Profile-picker probe assumes window origin `(0,0)` and stable accessory geometry
**File:** ConduitUITests/ProfilePickerUITests.swift:88, 193-204
**Issue:** `tapInWindow` passes an absolute point as the `dx/dy` offset from `window.coordinate(withNormalizedOffset: .zero)`, which is correct only while the window origin is `(0,0)` (true on iPhone, not guaranteed on multi-window/iPad). Also, the probe's `x = accessorySlot.frame.maxX + 6` is derived from the current row's "Current" label; if a prior simulator run left an avatar on the target row, the trailing trash button shifts that column and the point can land on the trash control.
**Suggestion:** Derive the offset from `bounds.origin` (subtract it), and assert the probe point is not inside any interactive descendant before tapping.

## Verdict
APPROVE

The gate split is correct and generation-fenced: the permission/provider-test/`isCurrent` consumers genuinely needed the app-foreground fact, and every path that retires in-flight Voice work (`stop()`, `suspendRuntimeForLifecycle()`, `failForAudioInterruption()`, sheet dismissal via `closeVoiceConversation`) bumps `operationGeneration`, so removing the capture term from `isCurrent` does not un-fence stale work. No force unwraps, retain cycles, `@MainActor` violations, or leaks were introduced; remaining findings are minor/nit.

<a href="https://opencode.ai/s/MJBdqMfL">New%20session%20-%202026-09-20T15%3A47%3A42.505Z</a>
[opencode session](https://opencode.ai/s/MJBdqMfL)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35520631844)

kaishi00:
## Deferred coverage items — closed (2 of 3 fully, 1 justified deferral)

### 1. Profile card interaction coverage — closed

**Test-only seam (DEBUG-only):** `-CONDUIT_UI_TEST_PROFILES "work,staging"` in `AppState.uiTestSeededProfiles()`, applied inside the existing `loadSavedConnection()` stub block and seeding through the same `orderedProfiles` a real `/api/profiles` discovery uses. It was necessary, not convenient: the connection-scoped reset clears the persisted known-profile cache on the way in and a transport-less stub can never repopulate it, so without the seed a profile-scoped UI test only ever sees `default`. It is reusable — `launchWithSeededProfiles` in the new test file documents the pattern — and it writes only in-memory state (never `knownProfilesKey`), so nothing leaks into the next run.

`ConduitUITests/ProfilePickerUITests` seeds a *pair* of profiles so each test acts on whichever is not current, and asserts the two invariants:

- **Tapping the card body invokes selection** — probed just past the row's accessory, the only place the card-sized selection surface can receive a tap (see the review round below), with the picker's title as a negative control that the signal is absent where no selection target exists.
- **Tapping the nested photo control runs its own action and does not switch the profile** — the photo picker appears, and after dismissing it the picker is still presented with the same profile current. This is the invariant that decided the card's structure during the first review round.

One accessibility identifier was added (`sidebar.workspace`) because that button's label embeds the dynamic active-profile name; `accessibilityIdentifier` is invisible to VoiceOver, so there is no user-facing change.

### 2. `attachToLiveVoiceConversation` gate path — closed

Two `AppStateVoiceCapabilityTests`: attaching to a live conversation republishes the app-foreground gate (and returns true with the harness's bridge installed), and attaching with no live conversation is a no-op that leaves the gate closed. Both assert the published contract (`isApplicationForegroundGateOpen` plus the return value), not internals — the observable seam exists precisely because this failure mode leaves no other trace.

### 3. Pre-iOS-26 material path — justified deferral

Verified by direct query: the build box has exactly one simulator runtime (iOS 26.5), and CI's xctestrun is built for the same, so the `#available(iOS 26.0, *)` false-branch cannot execute in any available test environment. Parameterizing a compile-time availability check purely to make a cosmetic branch assertable would test a double rather than the OS-gated compositor. Instead `Conduit/Theme.swift` documents the real invariant for the two branches of both glass helpers, and this stays the one deferred item.

### What the second review round changed

The three-reviewer gate on this work returned 2 APPROVE and 1 REQUEST CHANGES, and the dissenting reviewer falsified two of the author's claims by geometry — both correct, both fixed in `5f2154c`:

- The card probe's x came from the current row's `Current` label midpoint, which on a selectable row lands inside the text column's *button* (the label is far wider than the chevron it sits beside) — which is exactly why the mutation check kept passing with the card surface removed. The probe now sits past the accessory's trailing edge.
- The "which profile is not current" heuristic compared the sidebar label against the seeded names, but that label always contains the literal word "Workspace", so it always matched `Work` and would have failed outright had a run left `staging` active. The target is now derived from row geometry.

Also applied: photo-library permission handling before the picker wait, the negative control moved off the sheet's bottom edge, `ProfilePickerRow`'s now-false "no automated coverage" comment, and a parity note that covers both glass helpers without implying a hit-testing coupling that belongs to callers.

**Verification status.** Unit side verified locally (11/11 `AppStateVoiceCapabilityTests`, including both attach tests). The UI tests were verified locally in earlier runs; the final probe geometry could not be re-verified on this machine — the local simulator service is refusing to launch any test runner ("Busy / Application failed preflight checks"), reproducibly across two devices — so the first execution of the corrected tests is this run's UI lane, and I will treat its tree dump as the record.

kaishi00:
## Summary
Splits the Voice controller's single foreground flag into a capture gate (`isForegroundActive`) and an app-foreground gate (`isApplicationForegroundActive`) so Voice-settings permission/provider-test actions work with no Voice sheet presenting, and makes selectable cards fully tappable plus gives Bots roster rows rounded card styling.

## Findings

### [severity: minor] Computed app-foreground fact can disagree with the published gate during `.inactive`
**File:** Conduit/Services/AppState.swift:1075, 6748-6788
**Issue:** `.inactive` sets `isSceneActive = false` but deliberately does not republish, so `hasForegroundApplicationSurface` (`isSceneActive || isCarPlayVoiceSurfaceActive`) reads `false` while the controller's published `isApplicationForegroundActive` stays `true`. Any writer that recomputes from that property inside the `.inactive` window — `setCarPlayVoiceSurfaceActive(_:)` now calls `publishVoiceRuntimeGates()`, and `handleCarPlayVoiceSurfaceRemoved()`/`reassertVoiceSurfaceGate()` do too — will close the app gate while the app is foreground-but-inactive, fencing the in-flight `isCurrent`-checked permission continuation the `.inactive` comment is trying to protect. The comment documents the intent but does not enforce it.
**Suggestion:** Store a separate `isApplicationActive` fact that is `true` for `.active`/`.inactive` and `false` only on `.background`, and derive both `hasActiveVoiceSurface`/`hasForegroundApplicationSurface` from it, so the computed property and the published gate cannot disagree.

### [severity: minor] Enlarged profile-card hit region is invisible to VoiceOver
**File:** Conduit/Views/ProfilePickerSheet.swift:183-192
**Issue:** The card-sized selection `Button` added behind the row is `.accessibilityHidden(true)`, so the padding/accessory/space-beside-name area sighted users can now tap has no accessible counterpart; VoiceOver only reaches the smaller text-column button. The action remains reachable, so this is not a regression, but it diverges from the sibling Kanban card (`KanbanViews.swift:1938-1942`), which exposes the whole card as an `isButton`.
**Suggestion:** Add a selection accessibility action/hint on the row, or state explicitly in the comment that the inner button is the intended sole VoiceOver target.

### [severity: nit] Stale fence comment in `interruptAssistantPlayback`
**File:** Conduit/Voice/VoiceConversationController.swift:401-405
**Issue:** The comment says `isCurrent` "also covers foreground loss and session teardown," but `isCurrent` (line 1285) now keys on the app-foreground gate; capture/surface loss is no longer one of its terms and is covered only by the generation bump.
**Suggestion:** Reword to "covers app-backgrounding and session teardown."

### [severity: nit] Redundant app-gate restatement in `handleCarPlayVoiceSurfaceRemoved`
**File:** Conduit/Services/AppState.swift:16555-16563
**Issue:** The handler's only reachable paths are `setCarPlayVoiceSurfaceActive(false)` (which already published the app gate) and `fence()` (guarded by `!hasActiveVoiceSurface`, which implies `isCarPlayVoiceSurfaceActive == false`). The added `setApplicationForegroundActive(hasForegroundApplicationSurface)` recomputes the value it was just given; the comment overstates it as a correction.
**Suggestion:** Drop the call or note it as a defensive re-statement, not a corrective one.

### [severity: nit] Test's `selectButton` query can match the sidebar workspace button
**File:** ConduitUITests/ProfilePickerUITests.swift:194-200
**Issue:** `app.buttons.matching(NSPredicate(format: "label BEGINSWITH %@", displayName)).firstMatch` with `displayName == "Work"` also matches the sidebar workspace button, whose accessibility label begins with "Workspace". Which element `firstMatch` returns is hierarchy-order dependent, so the test may assert enablement against the wrong button (it happens to still work because the sidebar button is also disabled during a switch).
**Suggestion:** Disambiguate the predicate (e.g. exclude labels beginning with "Workspace", or match the exact expected select-button label).

### [severity: nit] Absolute-coordinate window tap assumes window origin `(0,0)`
**File:** ConduitUITests/ProfilePickerUITests.swift:204-215
**Issue:** `tapInWindow` treats the absolute point as a `dx/dy` offset from `window.coordinate(withNormalizedOffset: .zero)`, which is only correct while the window origin is `(0,0)` (true on iPhone, not guaranteed on multi-window/iPad).
**Suggestion:** Subtract `bounds.origin` from the point before applying the offset.

## Verdict
APPROVE

The gate split is correct and generation-fenced: the permission/provider-test/`isCurrent` consumers genuinely need the app-foreground fact, and every path that retires in-flight Voice work (`stop()`, `suspendRuntimeForLifecycle()`, `failForAudioInterruption()`, the direct-`showVoiceSheet` sites paired with `stop()`/`endVoiceSession()`/`setGateway(nil)`) bumps `operationGeneration`. No force unwraps, retain cycles, `@MainActor` violations, or leaks were introduced; remaining findings are minor/nit.

<a href="https://opencode.ai/s/9mT7snrt">New%20session%20-%202026-09-20T16%3A26%3A50.849Z</a>
[opencode session](https://opencode.ai/s/9mT7snrt)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35522676840)

kaishi00:
## Verification of the coverage work — final

**Both new UI tests pass in the hosted UI lane** on `c3180ae`: `testTappingTheCardBodyInvokesSelection` (43.6 s) and `testTappingThePhotoControlDoesNotSelectTheProfile` (55.0 s), inside `ui-2` alongside `ConnectionSetupUITests` (5/5) and `LoginKeyboardUITests`.

**The card-body test is no longer vacuous — mutation-proven locally.** Removing the card-sized selection surface from `ProfilePickerRow` makes it fail ("Tapping the card body must invoke selection"); with the surface present it passes. The two earlier versions of this assertion *did* pass under that mutation, which is how the defect was found: first a dismissal-only signal (the picker also disappears for unrelated reasons), then a probe whose x came from the `Current` label's midpoint and therefore landed inside the text column's own button.

**How the signal works now:** a switch attempt is in flight ⇒ `AppState.isProfileSwitching` disables the row's select button. That is immediate and specific (nothing else disables those rows), where the reported failure copy arrives only after the attempt fails — the first CI run's tree dump showed exactly this: the seeded rows' select buttons disabled after the probe, with the failure copy still absent. The control tap asserts the row is *still* enabled, so a signal that fired without a switch would fail loudly.

**`Unit - unit-6` is a pre-existing, load-sensitive failure, not this branch.** The failing tests are `TranscriptPerformanceFixtureTests`' settled-Markdown render counters (`testFirstAppearanceRendersSettledMarkdownOnce`: "lazy mounting never reached a quiet state; … on this runner"; `testStreamingTicksLeaveSettledMarkdownDormant_PlainTextTranscript`: "streaming re-rendered N interior rows", with N varying between runs — 2 then 1 — which is the load sensitivity). Evidence it is unrelated: this branch's earlier head `1fa7650` had **all seven unit lanes green**, and **main itself failed the same family** on 2026-09-20 (`SettledMessageIsolationTests.testIdenticalRowRecreationSkipsSettledMarkdownPresentation`, counter 1 ≠ 0, three times in one run) on a commit that predates this branch. This is the documented load-sensitive settled-Markdown family; `gh run rerun --failed` is the recorded remedy and has been applied twice.

PR #199, opened 2026-09-20
Fix Bot Mode session restoration and resume isolation

kaishi00:
## Why

A TestFlight report: after backgrounding/foregrounding, or on the "continue latest session" path, the app landed in a **Bot Mode** session that was not even the newest conversation.

## Root cause

Conduit's durable resume state stored **only a bare session id** (per dashboard profile) and had no notion of a conversation's *type*. Bot Mode conversations are ordinary gateway sessions living in the **same id space** as the Sessions surface's — the client's only way to tell them apart was ephemeral evidence (the reserved row title, an in-memory bot-chat registry, a roster that loaded only when the Bots tab was opened). After a relaunch a bot chat was therefore indistinguishable from an ordinary conversation:

- `restoreActiveSessionState` (cold launch + profile switch) adopted the stored id blindly;
- the resume resolver's stored-id match had no bot filter at all, and the "saved id absent from the catalog" direct-resume escape hatch **bypassed Bot Chat hygiene entirely**;
- when a bot chat *was* resumed that way it addressed the **dashboard profile's store** instead of the bot's, so identity, label, and surface were all wrong;
- "latest" was list position on a catalog that prepends a retained active-turn row and merges cached rows behind live ones, and `SessionSummary` had **no machine-readable timestamp** (only a formatted label).

Verified against the gateway (`tui_gateway/methods_session.py`): `session.resume` picks its store from the `profile` parameter — "a DEDICATED handle we own until the agent takes it; else the shared launch db" — so a bot conversation resumed under the workspace profile can never reach its session.

## What changed

**New typed `SessionReference`** (`Conduit/Services/SessionReference.swift`) — kind (`.dashboard` / `.bot`), scope profile, bot name, bot label, session id — the durable identity of "the conversation the user was in". It is recorded on every adoption (open, resume, rebind, restore) and resolved from the bot registry across the conversation's **own** ids, so a rotated runtime id cannot read as ordinary and a foreign id cannot classify this conversation.

**Correct Bot Mode profile/scope restoration across process launches.** A `.bot` reference re-registers the bot's scope and label *before* anything resumes, so the resume rides the bot's profile store, the header shows the bot's name, and the dashboard's title cache is never written. Scope resolution is an authority chain: this process's registry → the roster's canonical registries → the durable reference itself (the only authority for a conversation opened in an earlier process). Legacy id-only selections heal to `.bot` on positive evidence, in one direction only.

[...]

PR #197, opened 2026-09-20
Prepare 0.1.11 build 147 for TestFlight / App Store

kaishi00:
## Candidate

- **Marketing version:** 0.1.11
- **Build:** 147
- **Base:** `main` @ `f582a2e50e6e61ebbade02fa4197ba2cef5f3d30` (through PR #200)
- **Head:** `release/0.1.11` — a single release-metadata commit rebased onto that base. This PR remains the audit trail for the 0.1.11 candidate line (build 146 → build 147); no new release PR was opened.

This is the **replacement TestFlight candidate**: the build-146 candidate (head `24e2bec`, based on main through #175) exposed regressions and is superseded by this one. Main has advanced since the old base — including PRs #198, #199, and #200 (settled-Markdown fixture stabilization) — and this candidate carries all of it via the rebase onto `f582a2e`.

## Diff vs `main`

`project.yml` only — `MARKETING_VERSION: "0.1.11"`, `CURRENT_PROJECT_VERSION: "147"`. No code or behavioral changes on the release branch itself; `Conduit.xcodeproj` is generated (xcodegen) and never committed.

## Validation

- **Fresh hosted CI on this exact release head** — this PR's checks are the gate.
- **Local release checks:** `xcodegen generate` from this exact head verified on the Mac build box; the generated project carries `MARKETING_VERSION = 0.1.11` / `CURRENT_PROJECT_VERSION = 147` with no stale values.
- The TestFlight archive/upload must come from **this exact validated release commit** — no other checkout.

## Explicitly out of scope until separately requested

- Merging this release PR into `main`
- Creating the release tag
- App Store submission / phased release

PR #200, opened 2026-09-21
test(perf): stabilize the settled-Markdown dormancy fixtures on hosted runners

kaishi00:
Closes the hosted-CI failures of the settled-Markdown dormancy family that blocked release validation (PR #197 / build 147): `SettledMessageIsolationTests.testIdenticalRowRecreationSkipsSettledMarkdownPresentation` failed 6/6 native-retry iterations across two hosted attempts ("1 vs 0" Markdown evaluations), `TranscriptPerformanceFixtureTests` hit quiet-guard trips and one interior-row dormancy hit, and the same signature appears in recent main post-merge runs (#173-era, #198) — #199's green run only survived via native-retry rescue. The same tests pass repeatedly on a healthy Mac with byte-identical code and identical batch composition.

## Root cause (test harness, not production)

The fixture re-created its row through an `AnyView` `UIHostingController` **root replacement** — a hosting shape no production path exercises. Production re-creates every mounted row as a concrete value through `MessageBubble`'s switch inside an established hierarchy, invalidated by `AppState` publishes, with `SettledAssistantMessageContent.equatable()` holding identical rows at rest.

Hosted-run span diagnostics (added in commit 1) prove the artifact: the stray evaluation lands at **t≈0.001s** in the root swap's first synchronous update pass, stack = pure AttributeGraph body-update (`AG::Graph::UpdateStack::update` → `ViewBodyAccessor.updateBody(of:changed: true)` → `MarkdownText.body`), i.e. the erased-root diff re-evaluated the identical subtree outright. Deliberately checked and rejected alternatives: an identical **concrete** root is pruned wholesale by value diffing before the gate is consulted (verified locally — the first cut passed vacuously until an anti-vacuity guard caught it), so neither root-replacement shape measures the gate.

## Fix

Drive the tick the way production does: publish unrelated live state (`@Published streamingText` — the same vector the transcript fixtures drive), which invalidates the row's `@EnvironmentObject`, re-runs its body chain, and re-creates the `SettledAssistantMessageContent` value with identical inputs; the `.equatable()` gate must hold. Root re-assignment is removed entirely.

- **Anti-vacuity guard:** `AssistantBubble` now notes its settled-bubble body (mirroring `MessageBubble`; compiled out of release builds) and the fixture asserts the body chain re-ran before trusting the stay-at-zero result.
- **Invariant unchanged and strict:** equal inputs must produce exactly 0 settled-Markdown evaluations and 0 `SelectableTextView` updates. No weakening, no retries, no sleeps, no tolerance changes, no CI skips.
- **Diagnostics kept, bounded, DEBUG-only:** evaluation spans (time/source/stack, max 8 per window) surface in failure messages so any future violation names its driver; everything compiles out of release (same pattern as the existing counters). Quiet-guard failsafe cap 15s→45s — it bounds a stuck run; readiness is still decided solely by the sustained-quiet counter condition.

## Validation

- Local Mac (healthy host, exact branch head): `SettledMessageIsolationTests` ×3 isolated, full `TranscriptPerformanceFixtureTests` ×2, and the CI batch composition with CI's own iteration flags — all green; full unit suite **2775/2775, 0 failures** (the local UI bundle's AX bootstrap failure is the documented environmental one; hosted UI lanes are the gate).
- Hosted CI, commit 1 (mutation check): the old AnyView shape reproduced the exact failure 3/3 iterations with the spans shown above — CI run 35549540844, unit-7.
- Hosted CI, commit 2 (this head): the round that must be green.

## Summary by CodeRabbit

* **Tests**
  * Improved coverage for settled transcript behavior during streaming updates and Dynamic Type changes.
  * Added safeguards to prevent performance assertions from passing without exercising the intended rendering path.
  * Extended test wait limits for slower or contended environments.
  * Added diagnostic details to relevant test failures to help identify evaluation timing and call-stack issues.


kaishi00:
## Summary

This commit adds DEBUG-only in-window diagnostics (bounded time-offset/source/call-stack "spans" surfaced into three fixture failure messages) and widens the `settleUntilCountersQuiet` failsafe from 15s to 45s. Per the PR body, it does not fix the flaky fixture yet; the actual fix (concrete-root recreation + anti-vacuity guard) is deferred to a follow-up commit.

## Findings

### [major] The commit does not stabilize the failing tests; the branch as reviewed stays red
**File:** `ConduitTests/SettledMessageIsolationTests.swift:78` (and `PerformanceFixtureWait.swift:57`)
**Issue:** The known-bad `AnyView` root-replacement shape (`mountRow` / `host.rootView = AnyView(...)`) is untouched, the fixture still expects to fail on hosted CI, and the anti-vacuity guard is absent. The PR title says the fixtures are stabilized, but this head only adds diagnostics and a longer failsafe. The test that this PR exists to fix will still fail 6/6.
**Suggestion:** Land the fix commit (concrete root value recreation + anti-vacuity guard) before this is merged, or retitle/mark the PR draft so the title matches the contents. Do not merge just the diagnostics.

### [major] `TranscriptPerf.storage` is read and mutated without synchronization
**File:** `Conduit/Services/TranscriptPerformanceInstrumentation.swift:97-117` (new), `:412` (existing)
**Issue:** `note(_:context:)` is invoked from SwiftUI body/commit paths (`MarkdownText.swift:61`, `ChatView.swift:191`, etc.), and `recordWindowSpan` now appends to the shared `windowEvaluationSpans` array and reads/writes `windowOpenedAt` on the same unguarded static `storage`. Test code reads the same arrays/`Set`s (`allCounters()`, `windowEvaluationSpans`, `recentPreWindowRepeatSources`). If any note site lands off the main thread, concurrent COW mutations of `Set`/`Array` are undefined behavior and a plausible source of the exact non-determinism ("1 vs 0") this PR is chasing. This predates the commit but the new code widens the surface.
**Suggestion:** Serialize all `storage` access (e.g. `os_unfair_lock`/`NSLock` around the mutation block and accessors), or confine instrumentation to the main actor and assert that in debug. Worth ruling out as the actual root cause before attributing the flake solely to the `AnyView` hosting shape.

### [minor] Diagnostic calls `Thread.callStackSymbols` synchronously inside the measured window
**File:** `Conduit/Services/TranscriptPerformanceInstrumentation.swift:105-114`
**Issue:** `Thread.callStackSymbols` is expensive (symbolication/allocation) and is called from within `MarkdownText.body` for the first 8 settled evaluations of every window, including in shipped DEBUG builds (not just tests). Inside the very window being measured, adding main-thread symbolication can perturb commit scheduling and make the flake harder to diagnose. It is bounded to 8 per window, so the impact is limited, but it is avoidable overhead on a hot path.
**Suggestion:** Gate `recordWindowSpan` behind an explicit diagnostic flag (e.g. `CONDUIT_PERF_TRACE == "1"` or a test-only env var) rather than `#if DEBUG` alone, so production DEBUG and normal test runs don't pay for it.

### [minor] Address-stripping logic contradicts its comment
**File:** `Conduit/Services/TranscriptPerformanceInstrumentation.swift:111`
**Issue:** The comment says "Keep the symbol name only; addresses are noise," but `frame.split(separator: " ", maxSplits: 3).last` still retains the frame address (the frame format is `<index> <module> <address> <symbol> + <offset>`, so `maxSplits: 3` leaves `0x… symbol + offset`). The resulting spans are noisier than intended.
**Suggestion:** Use `maxSplits: 4` (or split and drop the address token explicitly) so only the symbol + offset remain.

### [minor] Failure messages can be very large and multi-line
**File:** `ConduitTests/SettledMessageIsolationTests.swift:170-172`, `ConduitTests/TranscriptPerformanceFixtureTests.swift:237,289`
**Issue:** Each of up to 8 spans embeds up to 10 call-stack frames joined by `" <- "`, and `context.prefix(24)` may contain raw newlines from the Markdown source. On failure this produces a huge, hard-to-read message and can bury the actual diagnostic.
**Suggestion:** Compact the stack (fewer frames, shorter separator), and escape/replace newlines in the source snippet (e.g. `context.replacingOccurrences(of: "\n", with: "\\n")`).

### [nit] Typo in doc comment
**File:** `Conduit/Services/TranscriptPerformanceInstrumentation.swift:91`
**Issue:** `call stack.DEBUG-only` is missing a space and the sentence runs together.
**Suggestion:** `call stack. DEBUG-only ...`

### [nit] Inaccurate rationale for the failsafe
**File:** `ConduitTests/PerformanceFixtureWait.swift:48-53`
**Issue:** The comment says tripping the failsafe "fails the lane WITHOUT any assertion signal," but every caller does `guard settled else { XCTFail(...) }`, which is an assertion signal. The real point is that it fails without naming the responsible counter/driver.
**Suggestion:** Reword to say the cap exists only to bound a genuinely stuck run and that tripping it reports no driver classification.

### [nit] Cap widening affects every fixture using the shared helper
**File:** `ConduitTests/PerformanceFixtureWait.swift:57`
**Issue:** Raising the default cap to 45s also triples the worst-case drain for `LongContextScalingFixtureTests` (four call sites) and all three `TranscriptPerformanceFixtureTests` calls. A genuinely stuck run now ties up the lane up to 3x longer per invocation.
**Suggestion:** Acceptable as a failsafe, but consider passing an explicit cap at the sites that need it, or documenting the CI-runtime tradeoff. The comment update is appreciated.

## Verdict

REQUEST_CHANGES

The diagnostics and cap change are reasonable groundwork, but as submitted the PR does not achieve its stated goal (the targeted fixture still fails by design), and the deferral of the actual fix plus the unsynchronized shared `storage` should be resolved/validated before merge.

<a href="https://opencode.ai/s/n9S26MzP">New%20session%20-%202026-09-21T01%3A02%3A46.195Z</a>
[opencode session](https://opencode.ai/s/n9S26MzP)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35549540831)

kaishi00:
## Summary
Stabilizes the settled-Markdown dormancy fixtures by replacing the artificial `AnyView` root-swap tick with an `AppState.streamingText` publish (the production invalidation vector), adds an anti-vacuity guard on `AssistantBubble`, and adds DEBUG-only evaluation-span diagnostics plus a widened quiet-failsafe cap.

## Findings

### [minor] Shared `settledBubbleBody` counter now double-counts assistant rows
**File:** `Conduit/Views/ChatView.swift:1461` (new), `Conduit/Views/ChatView.swift:963`
**Issue:** `MessageBubble` already notes `.settledBubbleBody`; nesting `AssistantBubble` (for `.assistant`/`.partial`) means every assistant-row body pass now increments the shared counter **twice**. The comment says it "mirrors" MessageBubble's counter, but it is an additional increment. Nothing asserts an exact value today (`allCounters()` only checks change/quiet), so there is no immediate break, but the counter's meaning ("settled bubble body evals") silently erodes for future assertions.
**Suggestion:** Either give `AssistantBubble` its own event (e.g. `.assistantBubbleBody`) for the anti-vacuity guard, or update the counter's documented semantics to "bubble body passes" so the double count is intentional and understood.

### [minor] Evaluation spans are unfiltered, so the first 8 may not name the violating row
**File:** `Conduit/Services/TranscriptPerformanceInstrumentation.swift:98-113`
**Issue:** `recordWindowSpan` records the first 8 settled-Markdown evaluations of the window in time order. In `TranscriptPerformanceFixtureTests` the window begins after `reset()` while fresh lazy/edge rows are still mounting, so those first 8 are frequently benign first-render evaluations — exactly the rows the classifier already tolerates — while the interior cascade the assertion cares about can land later and be absent from the diagnostics. The feature is meant to "name WHAT drove the evaluation," but it may not.
**Suggestion:** Filter spans to evaluations that are repeat/pre-window or window-duplicate (i.e. record only when the source is already known or already seen this window), or keep a ring of the last N rather than the first N.

### [minor] `Thread.callStackSymbols` runs on the measured body path in every DEBUG build
**File:** `Conduit/Services/TranscriptPerformanceInstrumentation.swift:106-110`
**Issue:** Symbolication/allocation happens synchronously inside `MarkdownText.body` for the first 8 evaluations after each `reset()`. In production DEBUG there is no reset so it is bounded to 8 per session, but in the fixtures the spans land inside the very window being measured. Counts (not wall time) are asserted, so this is unlikely to flip an assertion, but it is avoidable overhead on a hot path and can perturb commit scheduling in the run being diagnosed.
**Suggestion:** Gate `recordWindowSpan` behind an explicit flag (e.g. `CONDUIT_PERF_TRACE == "1"` or a test-only env var) in addition to `#if DEBUG`.

### [minor] Failure messages can be enormous and contain raw newlines
**File:** `ConduitTests/SettledMessageIsolationTests.swift:238-247`, `ConduitTests/TranscriptPerformanceFixtureTests.swift:232-234`, `ConduitTests/TranscriptPerformanceFixtureTests.swift:287-289`
**Issue:** Each span embeds up to 10 call-stack frames joined with `" <- "`, and up to 8 spans are concatenated; `context.prefix(24)` may also contain embedded newlines from Markdown source. On failure the diagnostics can bury the actual assertion.
**Suggestion:** Compact the stack (fewer frames / shorter separator) and escape newlines (`context.replacingOccurrences(of: "\n", with: "\\n")`).

### [nit] Failsafe comment contradicts its callers
**File:** `ConduitTests/PerformanceFixtureWait.swift:48-53`
**Issue:** The comment says tripping the cap "fails the lane WITHOUT any assertion signal," but every caller does `guard settled else { XCTFail(...) }` — that is an assertion signal. The real point is that tripping it reports no driver classification.
**Suggestion:** Reword to say the cap only bounds a genuinely stuck run and tripping it yields no driver classification.

### [nit] 15s → 45s cap also triples worst-case drain for unrelated suites
**File:** `ConduitTests/PerformanceFixtureWait.swift:57`
**Issue:** The shared default now applies to `LongContextScalingFixtureTests` and all `TranscriptPerformanceFixtureTests` call sites. A genuinely stuck run ties up the lane up to 3× longer per invocation.
**Suggestion:** Acceptable as a failsafe, but consider passing an explicit cap at the sites that need the larger bound, or note the CI-runtime tradeoff.

## Verdict
APPROVE

Test-only, DEBUG-only changes; no force unwraps, retain cycles, or new concurrency introduced, and the anti-vacuity guard closes the vacuous-pass hole. Findings are non-blocking polish.

Note: `TranscriptPerf.storage` is still read/mutated without synchronization (`TranscriptPerformanceInstrumentation.swift:409`), which predates this PR. The new writes only occur on the main-thread `body` path, so I did not treat it as blocking, but asserting main-thread confinement in DEBUG would make that guarantee explicit.

<a href="https://opencode.ai/s/aeZhLHU9">New%20session%20-%202026-09-21T01%3A37%3A41.785Z</a>
[opencode session](https://opencode.ai/s/aeZhLHU9)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35551445211)

kaishi00:
## Summary
This PR stabilizes the settled-Markdown dormancy fixtures by replacing the artificial `AnyView` root swap with the production invalidation vector (`AppState.streamingText` publish), adds an anti-vacuity guard on `AssistantBubble`, and adds DEBUG-only evaluation-span diagnostics plus a widened quiet-failsafe cap. No production behavior changes (release builds compile the instrumentation out).

## Findings

### [minor] Shared `settledBubbleBody` counter now double-counts assistant rows in production
**File:** Conduit/Views/ChatView.swift:1461
**Issue:** `MessageBubble` already notes `.settledBubbleBody` (ChatView.swift:963), and `AssistantBubble` is nested under it for `.assistant`/`.partial`. Every assistant-row body pass now increments the same counter twice. `allCounters()` only uses it for change detection, so nothing breaks today, but the accessor's name/meaning (`settledMessageBubbleBodyEvaluations`) silently erodes. In the new dormancy test the harness skips `MessageBubble`, so the guard sees a single increment — the test's counter semantics differ from production's.
**Suggestion:** Add a distinct event (e.g. `.assistantBubbleBody`) for the anti-vacuity guard, or update the counter's documented semantics to "bubble body passes" so the double count is intentional.

### [minor] `TranscriptPerf.storage` is mutated from the new diagnostics path without synchronization
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:98-113 (write), :300-306 (read), :409 (storage)
**Issue:** `recordWindowSpan` appends to the shared static `windowEvaluationSpans` and reads `windowOpenedAt` with no lock/isolation; tests read the same storage. All current `note` call sites are main-thread SwiftUI/`@MainActor` paths, so this is latent rather than observed, but `TranscriptPerf` is not `@MainActor`-annotated and the project is Swift 5.9 (no strict-concurrency enforcement), so nothing prevents an off-main call from racing COW mutations of `Set`/`Array`. This predates the PR; the new writes widen the surface.
**Suggestion:** Serialize `storage` access with a lock, or annotate the instrumentation `@MainActor` and add a DEBUG `dispatchPrecondition(condition: .onQueue(.main))` at the `note` entry to make the confinement explicit.

### [minor] `Thread.callStackSymbols` runs synchronously on the measured body path
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:106-110
**Issue:** Symbolication/allocation is invoked inside `MarkdownText.body` for the first 8 settled evaluations of every window, in every DEBUG build (not only tests). It is bounded, but it adds main-thread work inside the very window being measured and can perturb commit scheduling in the run being diagnosed.
**Suggestion:** Gate `recordWindowSpan` behind an explicit opt-in (e.g. `CONDUIT_PERF_TRACE == "1"` or a test-only env var) in addition to `#if DEBUG`.

### [minor] Spans capture the first 8 evaluations, which may not include the violating row
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:99
**Issue:** The feature is meant to "name WHAT drove the evaluation," but it records the first 8 evaluations of the window in time order. In `TranscriptPerformanceFixtureTests` the window opens while fresh lazy/edge rows are still mounting, so the first 8 are frequently the benign first-renders the classifier already tolerates, while a later interior cascade can be absent from the diagnostics.
**Suggestion:** Record only repeat/pre-window or window-duplicate evaluations, or keep a ring of the last N rather than the first N.

### [nit] Failsafe comment contradicts its callers
**File:** ConduitTests/PerformanceFixtureWait.swift:48-53
**Issue:** The comment says tripping the failsafe "fails the lane WITHOUT any assertion signal," but every caller does `guard settled else { XCTFail(...) }` — that is an assertion signal. The real point is that it fails without naming the responsible counter/driver.
**Suggestion:** Reword to say the cap only bounds a genuinely stuck run and tripping it yields no driver classification.

### [nit] 45s cap applies to every suite using the shared helper
**File:** ConduitTests/PerformanceFixtureWait.swift:57
**Issue:** The default now also governs `LongContextScalingFixtureTests` and all three `TranscriptPerformanceFixtureTests` call sites; a genuinely stuck run ties up the lane up to 3× longer per invocation.
**Suggestion:** Acceptable as a failsafe, but consider passing an explicit cap only at the sites that need the larger bound (or document the CI-runtime tradeoff).

### [nit] Failure messages can be large and contain raw newlines
**File:** ConduitTests/SettledMessageIsolationTests.swift:239-247, ConduitTests/TranscriptPerformanceFixtureTests.swift:232-234, :287-289
**Issue:** Up to 8 spans each embed up to 10 stack frames joined with `" <- "`, and `context.prefix(24)` can contain newlines from Markdown source, so a failure can bury the actual assertion.
**Suggestion:** Compact the stack (fewer frames/shorter separator) and escape newlines (e.g. `context.replacingOccurrences(of: "\n", with: "\\n")`).

## Verdict
APPROVE

Test-only and DEBUG-only changes; no force unwraps, no retain cycles, no new MainActor violations, and no accessibility impact. The anti-vacuity guard closes the vacuous-pass hole, and the switch to the publish-based tick is a sound fix for the harness artifact. Findings are non-blocking polish.

<a href="https://opencode.ai/s/svOKEntv">New%20session%20-%202026-09-21T02%3A18%3A44.907Z</a>
[opencode session](https://opencode.ai/s/svOKEntv)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35553692492)

PR #201, opened 2026-09-21
test(perf): make settled-Markdown dormancy deterministic under hosted churn

kaishi00:
## Root cause

The build-147 release-head failures (unit-7, run 35597418648, two attempts) were **not gate bypasses**. The bounded window spans from the failure messages showed:

- **Run 1** (`testStreamingTicksLeaveSettledMarkdownDormant_PlainTextTranscript`): one early burst (t=0.076–0.147s) re-evaluating rows at the rendered-window boundary (plain-240/241 — inside the mounted viewport of the bottom-anchored 250-row transcript, but outside the ±8 transcript-index edge margin), never again across 10 streaming ticks.
- **Run 2** (`testIdenticalRowRecreationSkipsSettledMarkdownPresentation`): exactly one MarkdownText evaluation per publish (t≈0.001/0.025/0.052 = ticks 1/2/3), identical stacks, all three hosted iterations — with every gate input provably constant.

## Why #200 was insufficient

#200 fixed the *root-swap* hosting vector and hardened `sizeCategory` on the isolation harness only. Two vectors remained:

1. `TranscriptPerformanceFixtureTests.mountChat` never pinned the environment — the isolation harness's own comment documents the exact failure mode (a late window trait-sync transaction on freshly booted CI simulators re-opens every mounted row's gate in one burst).
2. Both settled gates consume `\.chatTextSize` as an explicit input, but **no harness pinned that environment**. SwiftUI environment changes re-evaluate an `.equatable()`-gated subtree's body even when inputs compare equal, so a per-layout-pass environment push re-runs `MarkdownText` once per publish with every gate input still equal.

## Structural fix

**`DormancyHarnessEnvironment.applying`** is the single sanctioned harness mount; both dormancy suites and the concrete `SettledGateHarnessRow` mount through it, pinning both gate environment inputs below any hosting churn. No production behavior change, no assertion loosening, no new timing gates, no retries.

## Diagnostics (DEBUG-only, Release zero-cost)

- `TranscriptPerf.recentGateReopenReports` — the two settled gates record which Equatable field opened (inside `==`, result unchanged); failures now distinguish input-churn from a true bypass.
- `TranscriptPerf.settledMarkdownFreshMountEvaluations` + process-lifetime ever-seen ledger — classify re-renders as lazy mounts vs at-rest cascades.
- Window spans deepened to 14 frames (still bounded, still failure-message-only).

## Regression test

`testWindowTraitChurnDoesNotReopenSettledGate` delivers the churn vector deterministically (ambient environment mutation above the harness pin), canary-guarded so a no-op churn fails the test instead of passing vacuously.

**Mutation check (executed on a freshly created cold simulator):**
1. fix applied → new regression PASSES ✓
2. sizeCategory pin stripped (one line) → regression FAILS: `evaluations: 1` with span ✓
3. pin restored → all 6 isolation tests PASS ✓

## Local validation (fresh cold simulator, DiagPhone/DiagFix, iOS 26.5)

- Hosted unit-7 batch composition (8 classes incl. both dormancy suites + neighbors): **359/359, 0 gate reopens**
- Same batch at `-test-iterations 3`: **1080 executions, 0 failures**
- Full ConduitTests: pending in this run (will confirm in CI)

## Notably NOT done

- No retries/skips/looser assertions/wider sleeps/failsafe bumps (45s cap untouched)
- No lane isolation changes
- Release #197 and version numbers untouched

Closes the remaining settled-Markdown hosted nondeterminism so release validation for 0.1.11 build 147 can proceed on a genuinely green matrix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability for settled messages when device text-size or accessibility settings change, preventing unnecessary re-evaluation or visual updates.

- **Performance**
  - Improved measurement and monitoring of message rendering behavior, including tracking fresh content mounts and gate reopen events.

- **Tests**
  - Added regression coverage to verify that settled Markdown remains isolated from unrelated environment changes.


kaishi00:
## Summary
Test/diagnostics-only PR that pins the Dynamic Type and chat text-size environment in the settled-Markdown dormancy harnesses, adds DEBUG gate-reopen/mount diagnostics, and adds a window-trait-churn regression test to make hosted settled-Markdown determinism reproducible.

## Findings

### [major] Regression test uses the erased `AnyView` root re-assignment the suite explicitly rejects as non-deterministic
**File:** ConduitTests/SettledMessageIsolationTests.swift:502-527
**Issue:** The new churn test hosts and re-assigns `host.rootView = AnyView(ChurnableRoot(...))`. This is exactly the "erased AnyView root replacement" shape the sibling test `testIdenticalRowRecreationSkipsSettledMarkdownPresentation` (lines 209-215) documents as the hosted artifact the suite chased: "its erased diff remounts the identical row under a loaded scheduler." Under CI load an erased-root remount will fresh-evaluate `MarkdownText` (a mount, not a gate reopen), so the `XCTAssertEqual(evaluations, 0)` assertion can fail even though the pin is working — a self-inflicted flake. The doc comment's claim (lines 465-467) that "the row subtree itself is never re-created through an erased root" is also factually wrong; the root *is* `AnyView`.
**Suggestion:** Keep the root concrete. Move `ChurnableRoot` to file scope and host `UIHostingController<ChurnableRoot>` (or reuse the concrete-root pattern in `mountConcreteRow`) so the churn is delivered by value diffing rather than an erased root swap.

### [minor] Ineffective `chatTextSize` pin for the `mountChat` harness
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:160-171 (see also Conduit/Views/ChatView.swift:210)
**Issue:** `DormancyHarnessEnvironment.applying` writes `\.chatTextSize` *outside* `ChatView`, but `ChatView.body` re-injects `\.chatTextSize` on its inner content, so the harness pin is shadowed for every descendant. The effective value comes from `@AppStorage` reading `UserDefaults.standard`, which is process-global and not reset by `makeAppState()`'s suite. The PR description ("both dormancy suites ... pinning both gate environment inputs") and the comment claiming the pin "mirrors the production environment shape" overstate what is actually guaranteed.
**Suggestion:** Either drop the ineffective chatTextSize write for the `ChatView` harness and document that `ChatView` owns that injection, or make the fixture truly deterministic by seeding/pinning the standard-defaults key the `@AppStorage` reads.

### [minor] Fresh-mount counter inflates once the ledger cap is hit
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:73-77
**Issue:** Insertion stops at 4096 (`if storage.everSeenMarkdownSources.count < 4096`), but the `!contains(context)` test still runs. After the cap, every evaluation of any source not already recorded re-enters the branch and increments `settledMarkdownFreshMountBody` — so "fresh mount" becomes "every mount," silently corrupting the diagnostic.
**Suggestion:** Track an overflow flag and stop classifying when full, or reserve a sentinel entry so a saturated ledger short-circuits.

### [minor] `recentGateReopenReports` is documented as a "bounded ring"/"most recent" but is a first-N log
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:135-146, 150-156
**Issue:** `guard storage.recentGateReopenReports.count < 32 else { return }` keeps the *first* 32 gate reopens and discards all later ones. If a long DEBUG run opens other gates first, the decisive reopen that explains a dormancy failure is never recorded. The comment/name ("Most recent", "bounded ring") is misleading.
**Suggestion:** Implement an actual ring (drop-oldest) or rename to `firstGateReopenReports` and adjust the docs.

### [minor] Side-effecting `Equatable.==` over unsynchronized global mutable state
**File:** Conduit/Views/ChatView.swift:1045-1069 and 1324-1357; Conduit/Services/TranscriptPerformanceInstrumentation.swift:135, 482
**Issue:** `noteGateReopen` mutates `TranscriptPerf.storage` (a plain `static var` with no lock, not actor-isolated) from inside `static func ==`. `Equatable` is supposed to be pure, and if SwiftUI ever diffs off the main thread this is an unsynchronized data race on the array/set/`Storage` value. The project is Swift 5.9 (project.yml:14) so it compiles, but it will not survive strict concurrency and is a latent crash source.
**Suggestion:** Isolate the instrumentation (`@MainActor` on `TranscriptPerf`) or guard `storage` with a lock, and consider moving recording out of `==` into a place that is explicitly main-actor (or at minimum document the main-thread assumption).

### [minor] Comments reference a regression test that does not exist
**File:** ConduitTests/DormancyHarnessEnvironment.swift:18, ConduitTests/SettledMessageIsolationTests.swift:601
**Issue:** Both comments assert the pinning contract "is held by `testDormancyHarnessPinsGateEnvironmentInputs`," but no such test exists anywhere in the repo. The actual probe is `testWindowTraitChurnDoesNotReopenSettledGate`. This makes the mutation-check claim look better substantiated than it is.
**Suggestion:** Rename the references to the real test, or add the missing `testDormancyHarnessPinsGateEnvironmentInputs` probe.

### [nit] Malformed doc-comment line breaks the `///` block
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:157
**Issue:** Line 157 starts with `//` instead of `///`, dropping the rest of the paragraph out of the rendered documentation.
**Suggestion:** Change to `///`.

### [nit] Isolation suite never clears the new process-lifetime mount ledger
**File:** ConduitTests/SettledMessageIsolationTests.swift:22 (vs. Conduit/Services/TranscriptPerformanceInstrumentation.swift:392)
**Issue:** `setUp` calls `TranscriptPerf.reset()`, which deliberately preserves `everSeenMarkdownSources`/`settledMarkdownFreshMountBody`. The `resetRenderLedgerForTesting` doc says it exists so "each test starts with a clean mount classification," but this suite never calls it, so `settledMarkdownFreshMountEvaluations` accumulates across tests there. It's unasserted today, but the documented "clean per-test classification" is not actually delivered in this suite.
**Suggestion:** Call `resetRenderLedgerForTesting()` in this suite's `setUp` too, or scope the doc claim.

## Verdict
REQUEST_CHANGES

The production-side changes are low-risk (DEBUG-gated, Release-equivalent), but the new regression test re-introduces the erased-`AnyView` root re-assignment that this suite's own comments identify as the source of hosted nondeterminism — the exact failure class the PR is trying to eliminate. Fixing that (and the ineffective `chatTextSize` pin in `mountChat`) is needed before this can be trusted as the deterministic fix for build 147.

<a href="https://opencode.ai/s/NQptkBJM">New%20session%20-%202026-09-21T16%3A50%3A05.680Z</a>
[opencode session](https://opencode.ai/s/NQptkBJM)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35627998702)

kaishi00:
## Summary
Test/diagnostics-only change that pins the Dynamic Type and chat text-size environment in the settled-Markdown dormancy harnesses (via `DormancyHarnessEnvironment` and a new `ChatView.chatTextSizeOverride` seam), adds DEBUG gate-reopen/fresh-mount diagnostics, and adds regressions for window-trait and shared-preference churn. The head (8ef6ae6) correctly supersedes the earlier erased-`AnyView`/ineffective-pin/saturation issues, but a few diagnostic-honesty and concurrency-hygiene items remain.

## Findings

### [minor] `recentGateReopenReports` is documented as a "bounded ring"/"most recent" but keeps the first N
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:139, 152-153, 466-468
**Issue:** `guard storage.recentGateReopenReports.count < 32 else { return }` appends and then stops, preserving the *first* 32 reports and dropping all later ones. The accessor doc says "Most recent gate-reopen reports (bounded ring…)". In a long DEBUG run, the decisive in-window reopen can be dropped, and a reader is told they are seeing the most recent events when they are seeing the oldest. The stored-property comment ("Bounded ring of gate-reopen reports") repeats the misnomer.
**Suggestion:** Either implement a true drop-oldest ring (`removeFirst()` before append when full) or rename to `firstGateReopenReports` and correct the doc/storage comments.

### [minor] Gate diagnostics mutate unsynchronized global state from inside `Equatable.==`
**File:** Conduit/Views/ChatView.swift:1069, 1355; Conduit/Services/TranscriptPerformanceInstrumentation.swift:493
**Issue:** `noteGateReopen` writes `TranscriptPerf.storage` (a plain `private static var`, neither `@MainActor` nor lock-guarded) from within `static func ==`. `Equatable` is expected to be side-effect-free, and nothing enforces that SwiftUI only diffs on the main thread. Under the project's current SWIFT_VERSION 5.9 (project.yml:14) this compiles, but it is a latent data race and will not survive Swift 6 strict concurrency.
**Suggestion:** Isolate `TranscriptPerf` to `@MainActor` (or guard `storage` with a lock) and, if feasible, move recording out of `==` into an explicitly main-actor call site; at minimum document the main-thread assumption.

### [minor] Test-only seam added to production `ChatView`
**File:** Conduit/Views/ChatView.swift:185, 196
**Issue:** `var chatTextSizeOverride: ChatTextSize?` is internal, mutable, and compiled into release; production behavior is preserved only by convention (nil). Any in-module code could set it and silently override the user's persisted preference, and it adds surface area to a hot view's value identity.
**Suggestion:** Consider gating the property behind `#if DEBUG`, making it `private(set)` with a dedicated initializer, or documenting at the declaration that production must never set it.

### [minor] Other `ChatView`-hosted fixtures remain unpinned despite the "single sanctioned mount" claim
**File:** ConduitTests/LongContextScalingFixtureTests.swift:122, ConduitTests/ChatViewFollowCorrectionTests.swift:86
**Issue:** `DormancyHarnessEnvironment` is described as "the ONLY sanctioned way to mount a dormancy-fixture hierarchy," but these two hosted performance fixtures still construct `ChatView().environmentObject(...)` via `AnyView` with no `chatTextSizeOverride`, so they read the same shared `@AppStorage` preference the new regression exists to neutralize (and still use `isHidden = false` rather than `makeKeyAndVisible`). If the goal is hosted determinism for build 147, these remain exposed to the same class of churn.
**Suggestion:** Route them through the pinned mount (and the concrete-root pattern), or explicitly scope/justify the claim in `DormancyHarnessEnvironment` to the dormancy suites only.

### [nit] Isolation suite never clears the process-lifetime mount ledger
**File:** ConduitTests/SettledMessageIsolationTests.swift:22 (vs. Conduit/Services/TranscriptPerformanceInstrumentation.swift:403)
**Issue:** `setUp` calls `TranscriptPerf.reset()`, which deliberately preserves `everSeenMarkdownSources`/`settledMarkdownFreshMountBody`. `resetRenderLedgerForTesting()`'s doc promises "each test starts with a clean mount classification," but this suite never calls it, so the fresh-mount counter accumulates across tests there. It is unasserted today, so this is documentation drift rather than a functional bug.
**Suggestion:** Call `resetRenderLedgerForTesting()` in this suite's `setUp`, or scope the doc claim to the suites that use it.

### [nit] Shared-preference regression mutates process-global `UserDefaults.standard`
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:406-426
**Issue:** The test flips `ChatTypography.preferenceKey` in `UserDefaults.standard` (correctly restored via `defer`), but this key is also read by `AuxiliaryViews.swift:905` and any concurrently executed suite. Under parallel test execution this is a cross-suite race; under a hard crash the restore is skipped.
**Suggestion:** Note the serial-execution assumption, or seed/restore through an isolated defaults instance plus the existing override seam for the assertion.

## Verdict
COMMENT

The substantive prior findings (concrete churn roots, inner `chatTextSize` pin, ledger saturation guard, corrected comments) are properly resolved, and no critical/major issue remains. The outstanding items are diagnostic-accuracy and concurrency-hygiene improvements and are non-blocking.

<a href="https://opencode.ai/s/LTGhrsQ5">New%20session%20-%202026-09-21T21%3A22%3A55.532Z</a>
[opencode session](https://opencode.ai/s/LTGhrsQ5)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35656715531)

kaishi00:
## Summary
Test/diagnostics-only change that pins Dynamic Type and chat text-size environment inputs in the settled-Markdown dormancy harnesses, adds DEBUG-only gate-reopen/fresh-mount diagnostics, and adds window-trait/shared-preference churn regressions. The earlier critical findings (erased-`AnyView` churn root, shadowed `chatTextSize` pin, ledger over-count) are properly addressed on this head.

## Findings

### [major] `gateReopenLock` does not actually make `TranscriptPerf.storage` safe
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:136-172 (see also :393-404, :509-525)
**Issue:** The new lock only serializes `noteGateReopen`'s append against the report getter. Every other access to the same global value type (`note(_:)` incrementing fields at :27-49, `read`/`write` at :509-525, and `reset()` replacing the whole struct at :393-404) is unsynchronized. `noteGateReopen` is invoked from a nonisolated `static func ==` (`ChatView.swift:1069`, `:1355`), while `note` is also reachable from non-`@MainActor` code (e.g. `ChatScrollState.swift:118`). Mutating one stored property of the `static var storage` struct still requires exclusive access to the whole variable, so the lock does **not** remove the data race it is documented to close, and `reset()` can concurrently replace the struct the getter is reading. This is also a hard Swift 6 strict-concurrency blocker.
**Suggestion:** Isolate the instrumentation on `@MainActor` (or protect all `storage` access, including `note`/`read`/`reset`, with one lock), and preferably move recording out of `Equatable.==` into an explicitly main-actor call site. If neither is possible, document precisely that the lock only orders gate-report writes and does not make `storage` safe.

### [minor] "Bounded ring"/"Most recent" is actually a first-32 log
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:150, 163, 479-481
**Issue:** `guard storage.recentGateReopenReports.count < 32 else { return }` keeps the *first* 32 reopens and silently drops all later ones, yet the accessor doc says "Most recent gate-reopen reports (bounded ring…)" and the stored-property comment repeats "Bounded ring". In a long DEBUG run the decisive in-window reopen can be discarded, and a reader is told they are seeing the newest events.
**Suggestion:** Implement a drop-oldest ring (`removeFirst()` before append once full) or rename to `firstGateReopenReports` and correct both comments.

### [minor] Fresh-mount ledger retains full Markdown source strings
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:77-81, 462-464
**Issue:** `everSeenMarkdownSources` stores the raw `context` source string for up to 4096 distinct rows. The fixtures' sources are small, but in a DEBUG device build large documents (the code paths explicitly support sub-100 KB messages) would be retained wholesale, duplicating the existing render-cache retention. The doc calls the ledger "bounded" only in entry count, not bytes.
**Suggestion:** Store a hash/fingerprint (e.g. `source.hashValue` or a stable digest) as the ledger key instead of the full string, or note the byte-boundedness assumption.

### [minor] Test-only seam compiled into Release changes production `ChatView` surface
**File:** Conduit/Views/ChatView.swift:185
**Issue:** `var chatTextSizeOverride: ChatTextSize? = nil` is internal, mutable, and present in Release; production correctness depends only on convention (nil). Any in-module code could set it and silently bypass the user's persisted preference, and it becomes part of the view's value identity.
**Suggestion:** Gate it behind `#if DEBUG`, make it `private(set)` with a dedicated initializer, or add an explicit "production must never set this" contract at the declaration (the existing comment is close; make the invariant enforceable).

### [minor] `DormancyHarnessEnvironment` claim is narrower than the PR body implies
**File:** ConduitTests/LongContextScalingFixtureTests.swift:120-131, ConduitTests/ChatViewFollowCorrectionTests.swift:84-91
**Issue:** Two other hosted `ChatView` performance fixtures still mount via `AnyView(ChatView())` with no `chatTextSizeOverride` and no environment pin (`isHidden = false`). They read the same shared `@AppStorage` preference this PR exists to neutralize, so under the same hosted churn class they remain exposed. `DormancyHarnessEnvironment`'s "ONLY sanctioned way to mount a dormancy-fixture hierarchy" is only accurate if these are deliberately out of scope.
**Suggestion:** Route them through the pinned/concrete-root mount, or explicitly scope the doc claim to the dormancy suites.

### [nit] Isolation suite does not honor the new "clean per-test classification" doc
**File:** ConduitTests/SettledMessageIsolationTests.swift:20-23 (vs. Conduit/Services/TranscriptPerformanceInstrumentation.swift:411-415)
**Issue:** `setUp` calls `TranscriptPerf.reset()`, which deliberately preserves `everSeenMarkdownSources`/`settledMarkdownFreshMountBody`, so the new doc promise ("each test starts with a clean mount classification") is not delivered in this suite. It is unasserted today, so this is documentation drift rather than a functional bug.
**Suggestion:** Call `resetRenderLedgerForTesting()` in this suite's `setUp`, or scope the doc claim to the suites that call it.

### [nit] DEBUG gate compares `message` twice
**File:** Conduit/Views/ChatView.swift:1058-1069, 1337-1355
**Issue:** The `#if DEBUG` branch computes `equal` and then recomputes `lhs.message != rhs.message` to build the field list, so the largest value in the view (Markdown content) is compared twice whenever a gate opens. DEBUG-only, but the instrumentation runs inside hot render paths exactly where these fixtures measure work.
**Suggestion:** Capture the per-field booleans once and derive both `equal` and `fields` from them.

## Verdict
COMMENT

No production behavior change in Release; the prior critical/major issues are resolved. Remaining items are diagnostic-honesty and concurrency hygiene. The incomplete `gateReopenLock` is the one worth fixing before relying on these diagnostics under Swift 6, but it does not gate the release-validation goal.

<a href="https://opencode.ai/s/Kc0xm8Vd">New%20session%20-%202026-09-21T22%3A07%3A03.100Z</a>
[opencode session](https://opencode.ai/s/Kc0xm8Vd)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35660870670)

kaishi00:
## Summary
Test/diagnostics-only PR that pins Dynamic Type and chat text-size in the settled-Markdown dormancy harnesses (new `DormancyHarnessEnvironment` + `ChatView.chatTextSizeOverride` seam), adds DEBUG gate-reopen/fresh-mount diagnostics, and reshapes the single-row fixture plus new churn regressions to make hosted determinism reproducible.

## Findings

### [major] The renamed single-row test's assertion is vacuous in the shape it runs
**File:** ConduitTests/SettledMessageIsolationTests.swift:294-304
**Issue:** The test now asserts `TranscriptPerf.recentGateReopenReports.isEmpty`. But the PR's own measurement (and docstring at lines 275-286) states that in this standalone-row shape `SettledAssistantMessageContent.==` is *never consulted*, so no open report can ever be produced. The assertion therefore passes if the `Equatable` gate is deleted outright — the PR itself reports that mutation probe `gate` ("make `==` never prune") was not caught. The anti-vacuity guard at :261 only proves the row's body chain re-ran; it does not prove the gate was consulted. The test name and "strict" framing imply protection the test does not provide.
**Suggestion:** Either add a counter asserting `==` was actually consulted in this shape (and fail the test if zero), or rename/relax the assertion and note in the docstring that this vehicle cannot detect a non-pruning gate. Don't describe it as the strict gate-input regression.

### [major] `gateReopenLock` does not make `TranscriptPerf.storage` thread-safe
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:136-172 (see also :27-49, :393-403, :509-525)
**Issue:** The lock only wraps `noteGateReopen`'s append and the report getter. Every other access to the same global `private static var storage` (`note`, `read`, `write`, `reset`, `resetRenderLedgerForTesting`) is unsynchronized. `noteGateReopen` is called from nonisolated `static func ==` (ChatView.swift:1069, :1355), and mutating a single stored property on a `static var` struct still requires exclusive access to the whole variable. So if the lock's premise is true (SwiftUI may diff off the main actor), `reset()` can swap the struct a concurrent append/getter is using and the other `note` calls race as well; if the premise is false, the lock is unnecessary. Either way the documented guarantee is incomplete, and this will not compile under Swift 6 strict concurrency.
**Suggestion:** Isolate `TranscriptPerf` to `@MainActor` (all call sites are already main-actor), or protect *all* `storage` access with one lock. At minimum, correct the doc to state the lock only orders gate-report writes and does not make `storage` safe.

### [minor] `recentGateReopenReports` is documented as "most recent"/"bounded ring" but is a first-N log
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:150, 163, 479-481
**Issue:** `guard storage.recentGateReopenReports.count < 32 else { return }` keeps the *first* 32 reports and drops all later ones. The getter doc says "Most recent gate-reopen reports (bounded ring…)" and the storage comment repeats "Bounded ring". In a long DEBUG run the decisive in-window reopen can be discarded while the reader is told they are seeing the newest events.
**Suggestion:** Implement a true drop-oldest ring (`removeFirst()` before append once full) or rename to `firstGateReopenReports` and fix both comments.

### [minor] `settledMarkdownKnownSources` retains full source strings, unbounded, and its doc claims a false bound
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:82, 396-397, 465-468
**Issue:** `reset()` unions `windowSources` into `settledMarkdownKnownSources` process-lifetime, and it is never bounded or evicted, holding the full Markdown source string for every distinct message ever rendered in a DEBUG session. The comment says it is "bounded by the same order as the render cache," but `MarkdownRenderCache` is an `NSCache` (self-evicting, MarkdownText.swift:259-261) — not a bound this set shares.
**Suggestion:** Store a hash/fingerprint instead of the full source, bound the set, or correct the comment to state it is DEBUG-only and grows with distinct sources.

### [minor] Docstrings still assert mutation sensitivity the PR body says did not reproduce
**File:** ConduitTests/DormancyHarnessEnvironment.swift:22-26; ConduitTests/TranscriptPerformanceFixtureTests.swift:387-389
**Issue:** `DormancyHarnessEnvironment` claims the `chatTextSizeOverride` mutation "fails deterministically," and the fixture docstring claims stripping the override makes interior rows fail. The PR body's own caveat says that probe "did not reproduce." Source comments contradicting the measured caveat will mislead future maintainers.
**Suggestion:** Weaken both claims to match the measured behavior, or add a probe that actually discriminates.

### [minor] Test-only seam is compiled into Release
**File:** Conduit/Views/ChatView.swift:185
**Issue:** `var chatTextSizeOverride: ChatTextSize? = nil` is internal, mutable, and present in Release; production correctness depends only on convention. Any in-module code could set it and silently override the user's persisted preference, and it participates in the view's value identity.
**Suggestion:** Gate it behind `#if DEBUG`, make it `private(set)` with a dedicated initializer, or document the invariant enforcement at the declaration.

### [minor] "ONLY sanctioned way to mount" claim excludes two live ChatView fixtures
**File:** ConduitTests/DormancyHarnessEnvironment.swift:36-37 (see ConduitTests/LongContextScalingFixtureTests.swift:122, ConduitTests/ChatViewFollowCorrectionTests.swift:86)
**Issue:** Two hosted performance fixtures still mount `AnyView(ChatView())` with no pin and no override, so they read the same shared `@AppStorage` preference this PR exists to neutralize. The "ONLY sanctioned way" doc is inaccurate unless those are deliberately out of scope.
**Suggestion:** Route them through `DormancyHarnessEnvironment`/the concrete root, or scope the doc claim to the dormancy suites.

### [nit] Isolation suite never clears the process-lifetime mount ledger
**File:** ConduitTests/SettledMessageIsolationTests.swift:28-31 (vs. Conduit/Services/TranscriptPerformanceInstrumentation.swift:411-415)
**Issue:** `setUp` calls `TranscriptPerf.reset()`, which deliberately preserves `everSeenMarkdownSources`/`settledMarkdownFreshMountBody`, while `resetRenderLedgerForTesting()`'s doc promises "each test starts with a clean mount classification." This suite never calls it, so the documented per-test clean classification is not delivered here.
**Suggestion:** Call `resetRenderLedgerForTesting()` in this suite's `setUp`, or scope the doc claim to the suites that call it.

### [nit] Regression mutates process-global `UserDefaults.standard`
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:409-416
**Issue:** The test flips `ChatTypography.preferenceKey` in `UserDefaults.standard` (restored via `defer`), but the key is also read elsewhere (e.g. `AuxiliaryViews.swift`). Under parallel test execution this is a cross-suite race; a hard crash skips the restore.
**Suggestion:** Note the serial-execution assumption, or use an isolated defaults instance plus the override seam for the assertion.

### [nit] Dead `initialSTVUpdates`
**File:** ConduitTests/SettledMessageIsolationTests.swift:226, 305
**Issue:** `initialSTVUpdates` is captured and then discarded via `_ = initialSTVUpdates` after the zero assertion was removed.
**Suggestion:** Delete the now-unused local (and the capture).

## Verdict
REQUEST_CHANGES

The `pin`-only mutation sensitivity is honestly reported, but the reshaped single-row test no longer detects a broken gate at all (so the gate-pruning invariant has no discriminating regression), and the new lock's stated concurrency guarantee is materially incomplete. The remaining items are documentation-accuracy and hygiene fixes.

<a href="https://opencode.ai/s/QzxIMT6f">New%20session%20-%202026-09-23T00%3A56%3A29.238Z</a>
[opencode session](https://opencode.ai/s/QzxIMT6f)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35804108887)

kaishi00:
## Summary
Test/diagnostics-only PR that pins Dynamic Type and chat text-size in the settled-Markdown dormancy harnesses, adds DEBUG gate-reopen/fresh-mount diagnostics and a `ChatView.chatTextSizeOverride` seam, and rewrites the single-row fixture (deleting the unassertable hosted half) plus a newly added churn regression.

## Findings

### [major] `gateReopenLock` does not make `TranscriptPerf.storage` thread-safe, contrary to its stated guarantee
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:137-178 (see also :27-50, :64-102, :410-438); Conduit/Views/ChatView.swift:1344
**Issue:** The lock serializes only `noteGateReopen`'s append and the `recentGateReopenReports` getter. Every other access to the same `private static var storage` is unsynchronized: `note(_:)`, `note(_:context:)`, `read`/`write`, `reset()` (which replaces the whole struct), and the new `TranscriptPerf.note(.settledContentGateComparison)` called from `static func ==` at ChatView.swift:1344. The comment at :132-136 justifies the lock because "SwiftUI may evaluate a nonisolated Equatable conformance off the main actor". If that premise is true, then `reset()` can swap `storage` while a concurrent `==` mutates it, and the unlocked consultation counter reinstates exactly the race the adjacent `noteGateReopen` lock was added to close. If the premise is false the lock is unnecessary. Either way the documented guarantee is not delivered, and none of this compiles under Swift 6 strict concurrency.
**Suggestion:** Isolate `TranscriptPerf` to `@MainActor` (all call sites are main-actor), or guard all `storage` access (`note`, `read`, `write`, `reset`, `resetRenderLedgerForTesting`) with the same lock. At minimum, correct the comment to say the lock only orders gate-report writes and does not make `storage` safe.

### [major] The dormancy regression does not discriminate the gate-pruning invariant it is meant to guard
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:274-280, 329-335; Conduit/Services/TranscriptPerformanceInstrumentation.swift:453-463
**Issue:** The PR's own caveat reports that the `gate` mutation (make `SettledAssistantMessageContent.==` never prune) was not caught by these fixtures. `interiorAtRestRerenders` discards any source in `prefix(edgeMargin) + suffix(edgeMargin)` with `edgeMargin = 8`; in a bottom-anchored 250-row `LazyVStack` the mounted subset is only a handful of rows at one end, so all re-renders can fall inside that margin and the assertion passes regardless of whether the gate prunes. The newly added consultation counter (`TranscriptPerf.settledContentGateComparisons`, :209-211) exists precisely to detect the bypass shape but no fixture reads it. Net effect: the "interior rows stay flat" assertion may be vacuous against the regression class this PR exists to make trustworthy.
**Suggestion:** Assert `TranscriptPerf.settledContentGateComparisons > 0` in the transcript fixtures (or a dedicated fixture) so a bypass is detected, and/or reduce `edgeMargin` relative to the actually-mounted row count, or add a fixture that deliberately mounts interior rows. If a mutation-sensitive gate regression is not achievable in this shape, state that explicitly in the fixture docstrings.

### [minor] `settledMarkdownWindowSources` retains full source strings without bound; the "bounded" comment does not apply
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:88-92, 482-488
**Issue:** `TranscriptPerf.reset()` is never called in the app target (only in tests), so in a DEBUG device build `settledMarkdownWindowSources` accumulates one full Markdown source string per distinct message rendered, process-lifetime and unevicted. The comment on `settledMarkdownKnownSources` (:482-485) claims it is "bounded by the same order as the render cache", but `MarkdownRenderCache` is an `NSCache` (MarkdownText.swift:260-263) that self-evicts; this set does not. With messages up to the documented 100 KB threshold this is real retained memory.
**Suggestion:** Store a hash/fingerprint as the ledger key instead of the full source, bound the set, or correct the comment to state it is DEBUG-only and grows with distinct sources.

### [minor] Test-only seam is compiled into Release and participates in `ChatView`'s value identity
**File:** Conduit/Views/ChatView.swift:185
**Issue:** `var chatTextSizeOverride: ChatTextSize? = nil` is an internal, mutable stored property present in Release. Production correctness depends only on convention (nil), any in-module code could set it to bypass the user's persisted preference, and it becomes part of the view's stored state. The comment documents the intent but nothing enforces it.
**Suggestion:** Wrap it in `#if DEBUG`, or make it `private(set)` with a dedicated initializer, or otherwise make the "production must never set this" invariant enforceable.

### [minor] Docstrings assert mutation sensitivity the PR body reports did not reproduce
**File:** ConduitTests/DormancyHarnessEnvironment.swift:21-26; ConduitTests/TranscriptPerformanceFixtureTests.swift:368-372
**Issue:** `DormancyHarnessEnvironment` says stripping `ChatView.chatTextSizeOverride` makes `testSharedChatTextSizePreferenceChurnDoesNotReopenSettledGate` "fail deterministically", and the fixture docstring repeats that stripping the override fails the assertions below. The PR body's "Honest caveats" state that probe did not reproduce. Source comments contradicting the measured caveat will mislead maintainers about the coverage.
**Suggestion:** Weaken these claims to match the measured behavior, or add a probe that actually discriminates.

### [minor] Churn regression mutates process-global `UserDefaults.standard`
**File:** ConduitTests/TranscriptPerformanceFixtureTests.swift:406-426
**Issue:** The test writes `ChatTypography.preferenceKey` in `UserDefaults.standard` (restored via `defer`). That key is read elsewhere and by any concurrently executing suite, so under parallel test execution this is a cross-suite race; a hard crash also skips the restore.
**Suggestion:** Note the serial-execution assumption explicitly, or seed the value on an isolated defaults instance and assert through the override seam.

### [minor] `LongContextScalingFixtureTests.mountChat` claims the harness contract but keeps the shapes the contract warns against
**File:** ConduitTests/LongContextScalingFixtureTests.swift:125-142
**Issue:** This mount was just brought under `DormancyHarnessEnvironment`, yet it still wraps in `AnyView` and uses `window.isHidden = false`. The PR's own `mountRow`/`mountConcreteRow` comments (SettledMessageIsolationTests.swift:115-120, 166-168) state that a bare `isHidden` flip can leave the key/attach transition in flight into the measurement window — the exact churn class this PR neutralizes. The concrete-root discipline is likewise omitted.
**Suggestion:** Use the concrete root and `makeKeyAndVisible()` here (or reuse the shared mount helper) so all the fixtures this PR touches share one deterministic mount path.

### [nit] `interiorAtRestRerenders` default `edgeMargin` can swallow every mounted row
**File:** Conduit/Services/TranscriptPerformanceInstrumentation.swift:453-463
**Issue:** With `edgeMargin = 8` and a LazyVStack that mounts fewer than eight rows at one end of a 250-row transcript, the interior filter removes the entire mounted set, making any interior assertion vacuous. This predates the PR but is directly load-bearing for its new claims.
**Suggestion:** Derive the margin from the mounted row count, document the required mounted-row coverage, or assert that at least one source under test is classified interior.

### [nit] `DormancyHarnessEnvironment` "ONLY sanctioned way" claim is broader than the repo
**File:** ConduitTests/DormancyHarnessEnvironment.swift:36-37; ConduitTests/ChatViewFollowCorrectionTests.swift:84-87
**Issue:** `ChatViewFollowCorrectionTests` still mounts `AnyView(ChatView())` with no pin/override. It does not read the dormancy counters, so this is arguably out of scope, but the doc claim is inaccurate as written.
**Suggestion:** Scope the claim to the dormancy suites, or route that fixture through the pinned mount.

## Verdict
REQUEST_CHANGES

No Release behavior change and the previously blocking single-row vacuity was correctly resolved by deleting the unassertable test. But the new lock's concurrency guarantee is materially incomplete, and the fixtures that now carry the dormancy claim are not shown to discriminate the gate-pruning regression the PR targets.

<a href="https://opencode.ai/s/hF9ojojQ">New%20session%20-%202026-09-23T01%3A20%3A53.552Z</a>
[opencode session](https://opencode.ai/s/hF9ojojQ)&nbsp;&nbsp;|&nbsp;&nbsp;[github run](/kaishi00/hermes-conduit/actions/runs/35805833034)

kaishi00:
## Merge record — dormancy work accepted with a recorded environmental caveat

**Head:** `921f2256eef0dcb92a2d210966e1721cb061f6fc` (unchanged since review round 2; hosted CI run [35805835535](https://github.com/kaishi00/hermes-conduit/actions/runs/35805835535) fully green on this exact commit, including the required `CI Gate`).

### Local Mac exhaustive gate on this exact SHA — two invocations, zero genuine assertions in either

| run dir | `tested_sha` | unit | UI | repeats | `assertion_failures` | persistent infra |
|---|---|---|---|---|---|---|
| `921f2256eef0-20260923T015727Z` | `921f225…` | 2778 exec / 0 fail | 29 / 0 | 42 / 0 | `[]` | 4 |
| `921f2256eef0-20260923T081729Z` (`--allow-another-run`) | `921f225…` | 2778 / 0 | 29 / 0 | 34 / 0 | `[]` | 7 |

Both runs report exactly one problem: `persistent infrastructure failure(s)/hang(s) present (N infra, 0 timeout)`. Every persistent event is a repeat-lane iteration whose lane never launched the app under test:

```
Simulator device failed to launch com.milim.relay … denied by service delegate
(SBMainWorkspace) for reason: Busy ("Application failed preflight checks")
FBSOpenApplicationServiceErrorDomain Code=1
```

surfaced by XCTest as the synthetic `System Failures` class. For contrast, the `f582a2e…` runs in the same runs directory show what a genuine failure looks like here: a populated `assertion_failures` entry with `status: test-failures`. That array is **empty** in both `921f225` runs.

The second run was an operator-authorized infrastructure retry that began with the documented safe recovery — stale-process check, `simctl shutdown all`, CoreSimulatorService restart, then erase + fresh boot of the dedicated `Conduit CI Gate` device (6D08B063-B890-4D18-893B-D1E89E119919). The launch refusals still recurred and got worse (4 → 7 persistent), so this is a host/CoreSimulator condition rather than a property of this SHA.

**Caveat accepted for merge (operator decision):** two exact-SHA exhaustive runs establish complete unit/UI execution wherever a launch succeeded and zero genuine assertion failures; the residual failure is simulator launch infrastructure on this Mac. A third full gate run was explicitly not authorized and none was launched.

No further local Mac gate should be expected on this host until that CoreSimulator launch pathology is addressed separately.

### Review round 3 (CodeRabbit, 2026-09-23T01:00Z) — three findings, recorded, none blocking, none fixed here

Fixing any of them would move the head off the SHA the two exhaustive runs certified, which the merge governance for this change forbids. They are added to the accepted follow-ups above:

- 🟠 **`reset()` is not taken under `gateReopenLock`** while `noteGateReopen`/`recentGateReopenReports` are. Real, but DEBUG-only diagnostics, and the same family as the recorded "unlocked counter increments" follow-up: the race can corrupt a DEBUG report ring or a window timestamp, never a shipped behaviour or a lane verdict.
- 🟡 **Consume `settledContentGateComparisons` as an explicit equality-comparison anti-vacuity counter** in the trait-churn fixture. The counter was added by this PR for exactly that purpose and currently has no consumer — same family as the recorded "counter has no consumer at this head" follow-up. It is the right next step for the user-gate coverage gap too.
- 🟡 **Wait for the recreated hierarchy to go quiet before capturing the `rowsReran` baseline** in the streaming fixture. Hardening of the anti-vacuity guard (delayed descendant work could satisfy it), not of the dormancy assertion itself.

